PR #1, opened 2019-10-30
Find element by path

ngfreiter:
This fleshes out the implementation of the `findElementByPath` function by accounting for finding paths that use bracket syntax, or may refer to a choice element that does not explicitly exist. There are multiple private helper functions that are called in this function that are also added to  the `StructureDefinition` class. This PR also adds tests for `findElementByPath`. Finally, there is a minor change in `ElementDefinition` which was made to be consistent with the handling of choice elements in `findElementByPath` and (we believe) in FHIR.

### DIFF
--- a/src/fhirtypes/StructureDefinition.ts
+++ b/src/fhirtypes/StructureDefinition.ts
@@ -299,6 +299,9 @@ export class StructureDefinition {
     });
     if (matchingXElement) {
       // If we find a matching [x] element, we need to slice it to create the child element
+      // NOTE: The spec is somewhat incosistent on handling choice slicing, we decided on this
+      // approach per consistency with 4.0.1 observation-vitalsigns profiles and per this post
+      // https://blog.fire.ly/2019/09/13/type-slicing-in-fhir-r4/.
       matchingXElement.sliceIt('type', '$this', false, 'open');
       // Get the sliceName for the new element
       const sliceName = fhirPath.slice(fhirPath.lastIndexOf('.') + 1);
@@ -315,6 +318,10 @@ export class StructureDefinition {
    * @returns {ElementDefinition} - The sliceElement if found, else undefined
    */
   private findMatchingSlice(pathPart: PathPart, elements: ElementDefinition[]): ElementDefinition {
+    // NOTE: This function will assume the 'brackets' field contains information about slices. Even
+    // if you search for foo[sliceName][refName], this will try to find a re-slice
+    // sliceName/refName. To find the matching element for foo[sliceName][refName], you must
+    // use the findMatchingRef function. Be aware of this ambiguity in the bracket path syntax.
     return elements.find(e => e.sliceName === pathPart.brackets.join('/'));
   }
 
--- a/src/fhirtypes/StructureDefinition.ts
+++ b/src/fhirtypes/StructureDefinition.ts
@@ -342,7 +342,15 @@ export class StructureDefinition {
           return (
             t.code === 'Reference' &&
             t.targetProfile &&
-            t.targetProfile.find(tp => tp.endsWith(`/${pathPart.brackets[0]}`))
+            t.targetProfile.find(tp => {
+              const refName = pathPart.brackets.slice(-1)[0];
+              // Slice to get last part of url
+              // http://hl7.org/fhir/us/core/StructureDefinition/profile|3.0.0 -> profile|3.0.0
+              let tpRefName = tp.split('/').slice(-1)[0];
+              // Slice to get rid of version, profile|3.0.0 -> profile
+              tpRefName = tpRefName.split('|')[0];
+              return tpRefName === refName;
+            })
           );
         }
       }
--- a/src/fhirtypes/StructureDefinition.ts
+++ b/src/fhirtypes/StructureDefinition.ts
@@ -335,7 +335,7 @@ export class StructureDefinition {
           return (
             t.code === 'Reference' &&
             t.targetProfile &&
-            t.targetProfile.find(tp => tp.endsWith(pathPart.brackets[0]))
+            t.targetProfile.find(tp => tp.endsWith(`/${pathPart.brackets[0]}`))
           );
         }
       }
--- a/src/fhirtypes/StructureDefinition.ts
+++ b/src/fhirtypes/StructureDefinition.ts
@@ -2,7 +2,7 @@ import capitalize from 'lodash/capitalize';
 import cloneDeep from 'lodash/cloneDeep';
 import { CodeableConcept } from './CodeableConcept';
 import { Coding } from './Coding';
-import { ElementDefinition } from './ElementDefinition';
+import { ElementDefinition, ElementDefinitionType, ResolveFn } from './ElementDefinition';
 
 /**
  * A class representing a FHIR R4 StructureDefinition.  For the most part, each allowable property in a StructureDefinition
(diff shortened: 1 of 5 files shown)


src/fhirtypes/StructureDefinition.ts
```diff
@@ -200,6 +253,95 @@ export class StructureDefinition {
...
+      // If we have foo[a][b][c], and c is the ref, we need to find an element with sliceName = a/b
+      if (
+        pathPart.brackets.length === 1 ||
+        e.sliceName === pathPart.brackets.slice(0, -1).join('/')
+      ) {
+        for (const t of e.type) {
+          return (
+            t.code === 'Reference' &&
+            t.targetProfile &&
+            t.targetProfile.find(tp => tp.endsWith(pathPart.brackets[0]))
```
cmoesel: In theory, this could catch more than it should.  For example:
```
foo[Group]
```
should find a reference to a `Group` -- but would also match on a reference to a `RequestGroup` (which is actually a very different thing).

So... I think you need something more like:
```typescript
t.targetProfile.find(tp => tp === pathPart.brackets[0] || tp.endsWith(`/${pathPart.brackets[0]}`))
```

src/fhirtypes/StructureDefinition.ts
```diff
@@ -200,6 +253,95 @@ export class StructureDefinition {
...
+  }
+
+  /**
+   * Looks for a slice within the set of elements that matches the fhirPath
+   * @param {PathPart} pathPart - The path to match sliceName against
+   * @param {ElementDefinition[]} elements - The set of elements to search through
+   * @returns {ElementDefinition} - The sliceElement if found, else undefined
+   */
+  private findMatchingSlice(pathPart: PathPart, elements: ElementDefinition[]): ElementDefinition {
+    return elements.find(e => e.sliceName === pathPart.brackets.join('/'));
```
cmoesel: We talked before about a possibility of `foo[sliceName][referenceName]` -- will this erroneously try to find a slice called `sliceName/referenceName`?

test/fhirtypes/StructureDefinition.test.ts
```diff
@@ -97,11 +89,130 @@ describe('StructureDefinition', () => {
     });
   });
 
+  // Should get top level for '' element
```
cmoesel: This comment seems out of place.

test/fhirtypes/StructureDefinition.test.ts
```diff
@@ -97,11 +89,130 @@ describe('StructureDefinition', () => {
...
   describe('#findElementByPath', () => {
-    it('should find an element by path', () => {
-      const valueX = observation.findElementByPath('value[x]');
+    let jsonRespRate: any;
+    let respRate: StructureDefinition;
+    beforeAll(() => {
+      jsonRespRate = defs.findResource('resprate');
```
cmoesel: Isn't `resprate` the weird one that we said was wrong because it doesn't use slicing for the choice type?  That seems a little bit of an odd choice to use in a test unless you're purposefully testing to see if you can handle incorrect inputs.

test/fhirtypes/StructureDefinition.test.ts
```diff
@@ -97,11 +89,130 @@ describe('StructureDefinition', () => {
...
+      expect(refRange.id).toBe('Observation.referenceRange');
+    });
+
+    it('should find a child element by a path that exists', () => {
+      const refRangeLow = respRate.findElementByPath('referenceRange.low', getResolver(defs));
+      expect(refRangeLow).toBeDefined();
+      expect(refRangeLow.id).toBe('Observation.referenceRange.low');
+    });
+
+    it('should find the base element by a null path', () => {
```
cmoesel: I'd suggest `empty` path rather than `null` path -- since `''` is not `null`.

test/fhirtypes/StructureDefinition.test.ts
```diff
@@ -97,11 +89,130 @@ describe('StructureDefinition', () => {
...
+
+    // Choices
+    it('should make explicit a non-existent choice element by path', () => {
+      const originalLength = observation.elements.length;
+      const valueX = observation.findElementByPath('value[x]');
+      expect(valueX.slicing).toBeUndefined();
+      const valueQuantity = observation.findElementByPath('valueQuantity', getResolver(defs));
+      expect(valueQuantity).toBeDefined();
+      expect(valueQuantity.id).toBe('Observation.value[x]:valueQuantity');
+      expect(valueX.slicing).toBeDefined();
```
cmoesel: Can you check that it adds the correct discriminator values?

test/fhirtypes/StructureDefinition.test.ts
```diff
@@ -97,11 +89,130 @@ describe('StructureDefinition', () => {
...
+    });
+
+    // Choices
+    it('should make explicit a non-existent choice element by path', () => {
+      const originalLength = observation.elements.length;
+      const valueX = observation.findElementByPath('value[x]');
+      expect(valueX.slicing).toBeUndefined();
+      const valueQuantity = observation.findElementByPath('valueQuantity', getResolver(defs));
+      expect(valueQuantity).toBeDefined();
+      expect(valueQuantity.id).toBe('Observation.value[x]:valueQuantity');
```
cmoesel: Can you check that the `sliceName` and `path` are correct?

test/fhirtypes/StructureDefinition.test.ts
```diff
@@ -97,11 +89,130 @@ describe('StructureDefinition', () => {
...
+      expect(valueX.slicing).toBeDefined();
+      expect(observation.elements.length).toBe(originalLength + 1);
+    });
+
+    it('should make explicit a non-existent choice element by child path', () => {
+      const originalLength = observation.elements.length;
+      const valueX = observation.findElementByPath('value[x]');
+      expect(valueX.slicing).toBeUndefined();
+      const valueQuantity = observation.findElementByPath(
```
cmoesel: Rename variable to `valueQuantitySystem` just to be clear.

test/fhirtypes/StructureDefinition.test.ts
```diff
@@ -97,11 +89,130 @@ describe('StructureDefinition', () => {
...
+        'valueQuantity.system',
+        getResolver(defs)
+      );
+      expect(valueQuantity).toBeDefined();
+      expect(valueQuantity.id).toBe('Observation.value[x]:valueQuantity.system');
+      expect(valueX.slicing).toBeDefined();
```
cmoesel: To be thorough, I guess we can/should ensure the discriminator values are correct here too.

src/fhirtypes/StructureDefinition.ts
ngfreiter: It will try to find `sliceName/referenceName`, but that should fail so this will return null, and then we will go on to the `findMatchingRef` function, which will then treat the last bracketed element as a reference, so `referenceName` here, and search for the element with a slice `sliceName` that has a reference to `referenceName`. If there does exist a re-slice with the name `sliceName/referenceName`, and a slice `sliceName` with a reference to `referenceName`, we wouldn't have a way to distinguish between those two, but I don't think there really is a way. Plus that seems very improbable.

test/fhirtypes/StructureDefinition.test.ts
ngfreiter: So `resprate` itself isn't weird, all of the vital sign profiles in this bundle we have use the same convention. Which is that instead of slicing `value[x]`, they directly create elements like `value[x].value` to represent `value[x]:valueQuantity.value`. Which seems to me to just be incorrect. I think once some of the weird behavior we were observing on the FHIR spec site is done, we may want to update this. But, I am not using `resprate` to test choice types. I only used `resprate` to test slicing, and to test elements that have children. The handling of those cases is correct.

src/fhirtypes/StructureDefinition.ts
cmoesel: Ah.  OK.  I think it's cleaner as separate functions like this, but it's a little unfortunate that this function kind of depends on another function to be called after it in order to really get the right answer. If we re-use this function anywhere else (which is probably doubtful?), that's something we'll need to keep in mind.  Maybe we can at least document that behavior in the comment?

test/fhirtypes/StructureDefinition.test.ts
cmoesel: Actually, the body-weight profile (which is also a vital sign) seems to do it right:
http://hl7.org/fhir/R4/bodyweight.profile.json.html

src/fhirtypes/StructureDefinition.ts
ngfreiter: Yeah I can document the ambiguity involved.

test/fhirtypes/StructureDefinition.test.ts
cmoesel: Ah.  Perhaps.  I thought I remembered looking at it before 4.0.1 and body-weight was right then too, but who knows.

We should update the FHIR definitions w/ the new 4.0.1 ones -- but I want to wait until we get the "all clear" to do so, as it seems they're not quite done yet.  Just this morning Grahame made a change to the 4.0.1 package.json file.

src/fhirtypes/StructureDefinition.ts
cmoesel: I see you went with just the `endsWith` part.  Since this is a `canonical` I think that is OK, but in past versions of FHIR, it was possible for this just to be resource name, like `"Condition"`.

Also, note that a `canonical` _can_ have the version number appended to the end (delimited by `|`).  We're not supporting that here.  It's probably ok for now, but we may have to come back to it.  And we'll need to decide if we just ignore the version in that case or if we need to do something else special.

ngfreiter: My mistake, I can add the other check for just a resource name. And then I will add a comment for the version number situation.

cmoesel: Actually, I'm afraid this might come back and bite us at exactly the wrong moment.  How about we fix it so it will work even when there is a version at the end -- and for now we ignore the version.  So... for example, a path with `[us-core-practitioner]` should match on a type with `"profile": "http://hl7.org/fhir/us/core/StructureDefinition/us-core-practitioner|3.0.0"`.

ngfreiter: Ok I will happily match whether the version number is there or not, but ignore the version number.

cmoesel: Oooh... _happily_!  I'm glad to have made your morning!

ngfreiter: This should be addressed here: https://github.com/standardhealth/sushi/pull/1/commits/b9dc27051fe4e9a6fb3f28f031fb3dce8a6946a0.

PR #2, opened 2019-11-01
Adding FixedValueRule class

ngfreiter:
This adds a class for fixed value rules in the mold of the current existing `Rule` classes.

### DIFF
--- a/src/fshtypes/Quantity.ts
+++ b/src/fshtypes/Quantity.ts
@@ -1,0 +1,5 @@
+import { Code } from './Code';
+
+export class Quantity {
+  constructor(public value: number, public unit: Code) {}
+}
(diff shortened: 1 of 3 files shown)


kjmahalingam:
I'm on board with this, and agree with Chris's assessment!

src/fshtypes/Quantity.ts
```diff
@@ -0,0 +1,5 @@
+import { Code } from './Code';
+
+export class Quantity {
+  constructor(public value: number, public unit: Code) {}
```
cmoesel: I failed to mention this (and perhaps even implied otherwise), but `unit` is not actually required -- so that should be `public unit?: Code`.

cmoesel: It was such a trivial change, I just went ahead and made it.

PR #3, opened 2019-11-04
Initial FSHImporter

cmoesel:
- ANTLR4 grammar for profiles and extensions
- Infrastructure to build javascript targets from grammar
- FSHImporter using visitor pattern
  - profiles and extensions only
  - no support for slicing or invariants
- Updates to FSH types as necessary
- Added index.ts to simplify imports


package.json
```diff
@@ -5,8 +5,10 @@
   "scripts": {
     "build": "tsc",
     "build:watch": "tsc -w",
+    "build:grammar": "./antlr/gradlew -p ./antlr generateGrammarSource",
```
ngfreiter: This line doesn't work for me on windows, but if we change it to `bash ./antlr/gradlew -p ./antlr generateGrammarSource` it does work. I had @kjmahalingam test as well and it still worked for him with this new command, so we may want to change this.

src/import/FSHImporter.ts
```diff
@@ -0,0 +1,393 @@
...
+    return 'required';
+  }
+
+  visitFixedValueRule(ctx: pc.FixedValueRuleContext): FixedValueRule {
+    const fixedValueRule = new FixedValueRule(this.visitPath(ctx.path()));
+    fixedValueRule.fixedValue = this.visitValue(ctx.value());
+    return fixedValueRule;
+  }
+
+  visitValue(ctx: pc.ValueContext): boolean | number | string | Code | Quantity | Ratio {
```
ngfreiter: Could it possible be a good idea to define a `FixedValueRuleType` so we don't have to re-write this union in two different places. @kjmahalingam brought this up to me when I was making the FixedValueRule class but it got lost in the shuffle.

test/import/FSHImporter.Profile.test.ts
```diff
@@ -0,0 +1,528 @@
...
+        expect(profile.id).toBe('ObservationProfile');
+      });
+
+      it('should parse profile with additional metadata properties', () => {
+        const input = `
+        Profile: ObservationProfile
+        Parent: Observation
+        Id: observation-profile
+        Title: "An Observation Profile"
+        Description: "A profile on Observation"
```
ngfreiter: Do we want to test multi-line descriptions as well? I think that'll be a common use case, so we probably want a test to validate it.

test/import/FSHImporter.Profile.test.ts
```diff
@@ -0,0 +1,528 @@
...
+        const profile = result.profiles.get('ObservationProfile');
+        expect(profile.rules).toHaveLength(1);
+        assertFixedValueRule(
+          profile.rules[0],
+          'valueQuantity',
+          new Quantity(1.5, new Code('mm', 'http://unitsofmeasure.org'))
+        );
+      });
+
+      it('should parse fixed value Ratio rule', () => {
```
ngfreiter: In the spirit of lots of coverage, we could add a test for when a ratio is just a number, and not a `Quantity`.

src/fshtypes/rules/Rule.ts
```diff
@@ -1,5 +1,5 @@
 // TODO: Is this really an Interface
```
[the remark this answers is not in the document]
kjmahalingam: In looking at this more, I think I'm solidifying my opinion on this unless you have an opinion to the contrary.
Interfaces seem like they're more commonly used as a "contract" to define rules that other classes must follow. They are beneficial in the sense that you can have a class implement multiple interfaces to have multiple contracts to follow.
In our specific case, it seems like it would be more beneficial to have Rule be an abstract class from which the other rule types can inherit. That way we are binding the rule class to a specific set of requirements, but don't have to explicitly define them within each rule class.

src/import/FSHImporter.ts
kjmahalingam: I think this is a good idea. I was wondering if it would be useful to make such a type in that previous PR in case we ever needed to use the same type set elsewhere at some point; and it looks like we did!

antlr/src/main/antlr/FSH.g4
```diff
@@ -0,0 +1,134 @@
...
+
+// RULES
+cardRule:           STAR path CARD flag*;
+flagRule:           STAR (path | paths) flag+;
+valueSetRule:       STAR path KW_FROM SEQUENCE strength?;
+fixedValueRule:     STAR path EQUAL value;
+containsRule:       STAR path KW_CONTAINS item (KW_AND item)*;
+onlyRule:           STAR path KW_ONLY SEQUENCE (KW_OR SEQUENCE)*;
+obeysRule:          STAR path? KW_OBEYS SEQUENCE (KW_AND SEQUENCE)*;
+carrotValueRule:    STAR path? carrotPath EQUAL value;
```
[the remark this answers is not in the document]
cmoesel: Ha -- you are exactly right.  And it's not nit-picky; it drives me crazy when there are misspellings in code; especially on class, function, or property names.  I must have been hungry when I wrote that code.  I'll fix it!

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1,18 +1,19 @@
...
+import { Code } from '../fshtypes';
+import {
+  BindingStrengthError,
+  CodedTypeNotFoundError,
+  CodeAlreadyFixedError,
+  InvalidCardinalityError,
+  SlicingDefinitionError,
+  SlicingNotDefinedError,
+  WideningCardinalityError
+} from '../errors';
```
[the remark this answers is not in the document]
cmoesel: Yes, but it only works if you create an `index.ts` file that exports all the other things in that folder.  You don't get it for free.

src/fshtypes/Extension.ts
```diff
@@ -3,10 +3,13 @@ import { Rule } from './rules/Rule';
 export class Extension {
   id: string;
   parent?: string;
+  title?: string;
+  description?: string;
   rules: Rule[];
 
   constructor(public name: string) {
     this.id = name; // init same as name
+    this.parent = 'Extension'; // init to 'Extension'
```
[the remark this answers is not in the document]
cmoesel: There is no obvious _default_ parent for a profile -- it could be any one of the 130 or so FHIR resource types.  But I'd say that 99.9% of extensions are based directly on `Extension` (and 100% must eventually be based on `Extension`).

I'll add a comment to make this more clear.

src/fshtypes/Extension.ts
```diff
@@ -3,10 +3,13 @@ import { Rule } from './rules/Rule';
 export class Extension {
   id: string;
   parent?: string;
```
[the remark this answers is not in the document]
cmoesel: Yeah -- I didn't initially set `parent` to `Extension` in the constructor, but after I did, I forgot to update the property.  I'll fix it.

src/fshtypes/rules/Rule.ts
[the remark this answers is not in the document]
cmoesel: I think I'm inclined to agree at this point.  I think I initially did it this way because there's not really much going on there (no actual _logic_ that needs to be inherited -- just properties).  But... I foresee a future in which rules may have some additional features added -- and it would be a pain to have to add them in every separate rule class!  So... good call.  I'll make the change.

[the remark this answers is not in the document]
cmoesel: I'm also going to evaluate if we really need the `ruleType` property.  I'm sure I had a reason for adding it before, but I'm not so sure what that reason is anymore.  If I confirm we don't need it, then I'm going to remove it.

package.json
cmoesel: OK.  I can make the change.  I wondered about that -- but assumed it was probably ok because I thought I had done something similar in the past.  Oh well.

src/import/FSHImporter.ts
cmoesel: Yeah, as I was writing that I thought it seemed kind of repetitive.  Thanks for the encouragement to go ahead and `type` it.

test/import/FSHImporter.Profile.test.ts
cmoesel: I do test it in one spot (the `FixedValueRule` tests, I think) -- but I can test it here too.

test/import/FSHImporter.Profile.test.ts
cmoesel: Yeah, totally should.  I think it was like 1:30am when I was writing these tests.  Subconsciously, I was probably just trying to get to bed as soon as I could!

src/fshtypes/rules/Rule.ts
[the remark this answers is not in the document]
cmoesel: OK -- it looks like rule types are not needed, so I have removed them (will be in next PR).

cmoesel:
@kjmahalingam and @ngfreiter  -- I believe I've addressed your concerns.  @ngfreiter -- as for the difference in `FSH.interp` and `FSHLexer.interp`, I'm not sure that that's about.  We may be able to apply a git configuration to handle that, but I'd have to do some research.  I think *maybe* we can tell git that the `.interp` extension is a text file and to ignore line endings, but I forget how.

PR #4, opened 2019-11-04
Add various FHIR types

mint-thompson:
New types are: UsageContext, Range, Identifier, Period,
ContactDetail, ContactPoint, Narrative, Extension, Meta,
and Resource. Each type's definition includes a reference
to FHIR documentation.

The value[x] field of UsageContext is not fully defined, as some of the
possible types for the value are not yet defined.

A UsageContext's value may be a Range.
Information about the Range FHIR type can be found at:
http://hl7.org/fhir/R4/datatypes.html#Range
The validateRange method will throw an error if given a Range that
violates a constraint.

This pull request addresses https://standardhealthrecord.atlassian.net/browse/CIMPL-96

Test cases are included as part of the pull request.


src/fhirtypes/ContactDetail.ts
```diff
@@ -0,0 +1,24 @@
+import { Period } from './Identifier';
+
+/**
+ * Represents the FHIR R4 metadata type ContactDetail.
+ *
+ * @see {@link http://hl7.org/fhir/R4/metadatatypes.html#ContactDetail}
+ */
+export type ContactDetail = {
```
cmoesel: When my PR #3 is merged, it will introduce a src/fhirtypes/index.ts file. This and the other new FHIR types introduced in this PR should be added to that file.

src/fhirtypes/UsageContext.ts
```diff
@@ -0,0 +1,22 @@
...
+ * valueRange: Range
+ * valueReference: Reference<PlanDefinition, ResearchStudy, InsurancePlan, HealthcareService, Group, Location, Organization>
+ *
+ * @see {@link http://hl7.org/fhir/R4/metadatatypes.html#UsageContext}
+ */
+export type UsageContext = {
+  code: Coding;
+  valueCodeableConcept?: CodeableConcept;
+  valueQuantity?: Quantity;
+  valueRange?: Range;
```
cmoesel: Seeing as the only one you're missing is `valueReference`, I think it makes sense to add it.

test/fhirtypes/Range.test.ts
```diff
@@ -0,0 +1,112 @@
+import cloneDeep from 'lodash/cloneDeep';
+import { Range, validateRange } from '../../src/fhirtypes/Range';
+import { Quantity } from '../../src/fhirtypes/Quantity';
+import { InvalidRangeValueError } from '../../src/errors/InvalidRangeValueError';
+import { UnitMismatchError } from '../../src/errors/UnitMismatchError';
+import { CodeAndSystemMismatchError } from '../../src/errors/CodeAndSystemMismatchError';
```
cmoesel: We can clean up these imports a bit once we add these types to their respective `index.ts` files.

test/fhirtypes/Range.test.ts
```diff
@@ -0,0 +1,112 @@
...
+      };
+      const onlyHigh: Range = {
+        high: small
+      };
+      const neitherBound: Range = {};
+      expect(validateRange(onlyLow)).toBe(true);
+      expect(validateRange(onlyHigh)).toBe(true);
+      expect(validateRange(neitherBound)).toBe(true);
+    });
+    it('should require low bound value to be less than or equal to high bound value', () => {
```
cmoesel: Can you put a blank line between `it` tests?  For me, that helps me to read them easier.

test/fhirtypes/Range.test.ts
```diff
@@ -0,0 +1,112 @@
...
+        validateRange(differentSystem);
+      }).toThrow(CodeAndSystemMismatchError);
+      expect(() => {
+        validateRange(missingCode);
+      }).toThrow(CodeAndSystemMismatchError);
+      expect(() => {
+        validateRange(missingSystem);
+      }).toThrow(CodeAndSystemMismatchError);
+    });
+  });
```
cmoesel: Probably a stylistic thing, but... you seem to test multiple things in a single `it` -- whereas I might have broken them up some more.  I also like to describe the expected behavior a bit more specifically.  As opposed to saying it `should require...` -- which is ambiguous about what happens if the requirement is broken, I like saying it `should throw...` For example, this last test I might have broken into tests like this:
* `it('should throw when the low and high have different codes', () =>{`
* `it('should throw when the low and high have different systems', () =>{`
* `it('should throw when low or high is missing a code', () =>{`
* `it('should throw when low or high is missing a system', () =>{`

This way when a test fails you can more easily pinpoint exactly what failed.  It also provides a nice set of requirements describing what should happen in what cases.

src/fhirtypes/Identifier.ts
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Enforces the constraint on Period instances that the start
+ * cannot come after the end.
+ * @param period - the Period to validate.
+ * @returns {boolean} - true when the Period is valid.
+ * @throws {InvalidPeriodError} when the start comes before the end.
+ */
+
+export function validatePeriod(period: Period): boolean {
```
cmoesel: I think it makes more sense to have these return `void` -- because as it is now, the only possible value it can ever return is `true`, so there's no reason for a caller to ever check the value.  By actually saying it returns a `boolean`, the caller may mistakenly assume that they should expect `true/false` to indicate if it validates (rather than a thrown exception).  If we return `void`, it's a bit more obvious that an error will throw (since the function would be pointless otherwise).

cmoesel:
Also, I wonder if we might make sense to bundle these types together in few files rather than a file per type. This might be easier to manage -- especially if we aligned it to the spec and called the files `primitiveTypes.ts`, `dataTypes.ts`, `metaDataTypes.ts`, and `specialTypes.ts`.  Then we'd also pull out those few extra types we define in `StructureDefinition` into those files as well.

Thoughts?

Screenshot from FHIR spec:

image

kjmahalingam:
@mint-thompson This should now be rebased, since PR #3 has been officially merged into `master`.

mint-thompson:
@cmoesel I'm completely on board with your suggestion to group these type declarations based on the FHIR spec. Also, thanks for pointing out the auto-timezone-conversion that javascript does with dates. I'll change those back to strings and update the validate logic.

src/fhirtypes/metaDataTypes.ts
```diff
@@ -19,4 +25,5 @@ export type UsageContext = {
   valueCodeableConcept?: CodeableConcept;
   valueQuantity?: Quantity;
   valueRange?: Range;
+  valueReference?: any;
```
cmoesel: I'd suggest you go ahead and create the `Reference` type and put it in the `dataType.ts` file:
http://hl7.org/fhir/R4/references.html#Reference

src/fhirtypes/primitiveTypes.ts
```diff
@@ -0,0 +1,11 @@
+import { InvalidDateTimeError } from '../errors/InvalidDateTimeError';
+
+export type FHIRdateTime = string;
```
cmoesel: Is there a reason you called this `FHIRdateTime`?  Is it because some of the other primitive names would clash with reserved words in JavaScript, so you're establishing the pattern now?

I think that's a valid reason, but I would have expected the D to be capitalized (e.g., `FHIRDateTime`).  We already have an example of this with our `FHIRDefinitions` class.

src/fhirtypes/primitiveTypes.ts
```diff
@@ -0,0 +1,11 @@
...
+
+const dateTimeRegex = /^([0-9]([0-9]([0-9][1-9]|[1-9]0)|[1-9]00)|[1-9]000)(-(0[1-9]|1[0-2])(-(0[1-9]|[1-2][0-9]|3[0-1])(T([01][0-9]|2[0-3]):[0-5][0-9]:([0-5][0-9]|60)(\.[0-9]+)?(Z|(\+|-)((0[0-9]|1[0-3]):[0-5][0-9]|14:00)))?)?)?$/;
+
+export function validateFHIRdateTime(dateTime: FHIRdateTime): void {
```
cmoesel: Again, I'd expect the D to be capitalized (e.g., `validateFHIRDateTime`).

test/fhirtypes/FHIRdateTime.test.ts
```diff
@@ -0,0 +1,51 @@
...
+  let zTime: FHIRdateTime;
+  let noTimeZone: FHIRdateTime;
+
+  beforeAll(() => {
+    yearOnly = '1964';
+    yearMonth = '1975-05';
+    dateOnly = '2006-03-29';
+    hasTimeZone = '2010-02-07T13:28:17+03:00';
+    zTime = '2017-01-01T00:00:00.000Z';
+    noTimeZone = '2010-02-07T13:28:17';
```
cmoesel: I think these tests would be cleaner if you just put the literal strings right into the test cases.  E.g.:
```ts
    it('should allow just a year', () => {
      validateFHIRdateTime('1964');
    });
```

Creating objects in the higher-level scope is really only useful if they're used in more than one test.

test/fhirtypes/Period.test.ts
```diff
@@ -1,44 +1,43 @@
...
-  let later: Date;
+  let dateOnly: string;
+  let hasTimeZone: string;
 
   beforeAll(() => {
-    earlier = new Date();
-    later = new Date();
-    later.setFullYear(later.getFullYear() + 1);
+    dateOnly = '2006-03-29';
+    hasTimeZone = '2010-02-07T13:28:17+03:00';
```
cmoesel: Even here, even though these are re-used, they're so trivial that I wonder if it's better just to put the string literals right in the tests?  As it is now, when I'm reading the tests, I need to remember what those variables represent and which one comes before the other (chronologically).  That's all very obvious, however, if the test just uses string literals.

The other approach, which is closer to what you had before, is to give the variables more meaningful names so it's clear what they are and how they relate to each other -- for example `earlyDate` and `laterDateTime`.

test/fhirtypes/Range.test.ts
```diff
@@ -1,6 +1,5 @@
 import cloneDeep from 'lodash/cloneDeep';
-import { Range, validateRange } from '../../src/fhirtypes/Range';
-import { Quantity } from '../../src/fhirtypes/Quantity';
+import { Range, validateRange, Quantity } from '../../src/fhirtypes/dataTypes';
 import { InvalidRangeValueError } from '../../src/errors/InvalidRangeValueError';
 import { UnitMismatchError } from '../../src/errors/UnitMismatchError';
 import { CodeAndSystemMismatchError } from '../../src/errors/CodeAndSystemMismatchError';
```
cmoesel: The above three imports can be consolidated to a single line.

src/fhirtypes/dataTypes.ts
```diff
@@ -0,0 +1,126 @@
...
+ * which are enforced in the following validatePeriod method.
+ *
+ * @see {@link http://hl7.org/fhir/R4/datatypes.html#Period}
+ */
+export type Period = {
+  start?: FHIRdateTime;
+  end?: FHIRdateTime;
+};
+
+// http://hl7.org/fhir/R4/datatypes.html#dateTime
```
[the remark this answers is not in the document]
mint-thompson: nope! forgot to take it out when i moved the FHIR dateTime primitive to its own file

mint-thompson:
@cmoesel I don't think these findings are nit-picky at all. It's important to establish and adhere to usage patterns, doubly so in a project's early stages. Since I'm still very new to using typescript, these kinds of reminders are helpful.

src/fhirtypes/StructureDefinition.ts
```diff
@@ -376,6 +379,11 @@ type PathPart = {
   brackets?: string[];
 };
 
+export type Narrative = {
```
cmoesel: I think we should probably pull this into one of the data types files.  FHIR lists this in a weird section on the data types page, so maybe we should just put it into the `dataTypes.ts`?

src/fhirtypes/dataTypes.ts
```diff
@@ -0,0 +1,125 @@
+import {
+  InvalidRangeValueError,
+  UnitMismatchError,
+  CodeAndSystemMismatchError,
+  InvalidPeriodError
+} from '../errors';
+import { FHIRDateTime, validateFHIRDateTime } from './primitiveTypes';
+
+export type CodeableConcept = {
```
cmoesel: I really like how you documented (in comments) the other types and included the links.  I (obviously) did not do that in the types I initially created, but could you add that documentation here so they're all documented consistently?

mint-thompson:
I think with this, it's finally ready. There will be more work to do with types later, but the types currently present should be correctly organized and documented.

PR #7, opened 2019-11-06
Exporting profiles with basic metadata

ngfreiter:
This PR adds exporting of profiles and extensions with basic metadata (id, description, name, title, url). To do that, the basic structure flow for exporting a profile or extension was set up. One thing to call out is that I pulled the exporting functions into the `StructureDefinitionExporter` because I realized I was copying all the same code for `ExtensionExporter` that I did for `ProfileExporter`. Don't hesitate to be critical of any decisions I made here, because any bad setup here could come back to bite us later on.


src/export/ExtensionExporter.ts
```diff
@@ -1,15 +1,27 @@
...
+   * Exports Extensions to StructureDefinitions
+   * @param {FSHTank} tank - The FSH tank we are exporting
+   * @returns {StructureDefinition[]}
+   */
   export(tank: FSHTank): StructureDefinition[] {
-    // TODO: Make meaningful export of StructureDefinitions from contents of tank
-    return [];
+    const structDefs: StructureDefinition[] = [];
+    for (const doc of tank.docs) {
+      for (const [, extension] of doc.extensions) {
```
cmoesel: Honest question... do you think that it might be more straightforward to iterate over `doc.extensions.values()`?
```ts
for (const extension of doc.extensions.values()) {
```

src/export/ExtensionExporter.ts
```diff
@@ -1,15 +1,27 @@
...
+        // TODO: catch errors and log them once logging is in place
```
cmoesel: So it's easier to integrate the logger later, I'd suggest you go ahead and wrap this in a try/catch now and put a `console.error(...)` in the catch.

Also, just so we're on the same page, I think the only error we should toss up to this level is when the parent is not found.  I think most other errors should be caught inside the `exportStructDef` function so that we don't bail on the whole structdef just because of one error...

src/export/ProfileExporter.ts
```diff
@@ -1,15 +1,27 @@
...
-    return [];
+    const structDefs: StructureDefinition[] = [];
+    for (const doc of tank.docs) {
+      for (const [, profile] of doc.profiles) {
+        // TODO: catch errors and log them once logging is in place
+        const structDef = this.exportStructDef(profile, tank);
+        structDefs.push(structDef);
+      }
+    }
+    return structDefs;
```
cmoesel: Same comments on this function as on the one in the `ExtensionExporter`.

src/export/StructureDefinitionExporter.ts
```diff
@@ -1,6 +1,51 @@
...
+  }
+
+  /**
+   * Exports Profile or Extension to StructureDefinition
+   * @param {Profile | Extension} fshDefinition - The Profile or Extension we are exporting
+   * @param {FSHTank} tank - The FSH tank we are exporting
+   * @returns {StructureDefinition}
+   */
+  exportStructDef(fshDefinition: Profile | Extension, tank: FSHTank): StructureDefinition {
+    const parentName = fshDefinition.parent ? fshDefinition.parent : 'Resource';
```
cmoesel: I think you can do:
```ts
const parentName = fshDefinition.parent || 'Resource';
```

src/export/StructureDefinitionExporter.ts
```diff
@@ -1,6 +1,51 @@
...
+    const jsonParent =
+      fshDefinition instanceof Profile
+        ? this.FHIRDefs.findResource(parentName)
+        : this.FHIRDefs.findType(parentName);
```
cmoesel: Do you see any danger in just using `this.FHIRDefs.find(parentName)`?

test/export/ExtensionsExporter.test.ts
```diff
@@ -0,0 +1,66 @@
...
+    expect(exported[0].type).toBe('Extension');
+  });
+
+  it('should throw ParentNotDefinedError when parent extension is not found', () => {
+    const extension = new Extension('Foo');
+    extension.parent = 'Bar';
+    doc.extensions.set(extension.name, extension);
+    expect(() => {
+      exporter.export(input);
+    }).toThrow('Parent Bar not found for Foo');
```
cmoesel: I don't think this is quite right.  While we would want the function that exports a single profile to throw, we wouldn't want it to bubble up all the way to `export(tank: FSHTank)`.  This would effectively stop the entire process on a single error.

test/export/ProfileExporter.test.ts
```diff
@@ -0,0 +1,65 @@
...
+    expect(exported[0].type).toBe('Resource');
+  });
+
+  it('should throw ParentNotDefinedError when parent resource is not found', () => {
+    const profile = new Profile('Foo');
+    profile.parent = 'Bar';
+    doc.profiles.set(profile.name, profile);
+    expect(() => {
+      exporter.export(input);
+    }).toThrow('Parent Bar not found for Foo');
```
cmoesel: Same comment re: where we allow the error to bubble up.

test/export/ExtensionsExporter.test.ts
```diff
@@ -0,0 +1,66 @@
+import { ExtensionExporter } from '../../src/export';
+import { FSHTank, FSHDocument } from '../../src/import';
+import { FHIRDefinitions, load } from '../../src/fhirdefs';
+import { Extension } from '../../src/fshtypes';
+
+describe('ExtensionsExporter', () => {
```
kjmahalingam: For consistency, rename the file and this line to `ExtensionExporter` rather than `ExtensionsExporter`.

src/export/ExtensionExporter.ts
ngfreiter: Aw no one liked my cheeky syntax

src/export/StructureDefinitionExporter.ts
ngfreiter: Oh this cheeky syntax makes up for the other lost cheeky syntax.

src/export/StructureDefinitionExporter.ts
ngfreiter: No I do not.

test/export/ExtensionsExporter.test.ts
ngfreiter: That makes sense, I changed this and refactored the tests so that they are more unit-ish.

ngfreiter:
Ok I believe I've addressed the comments, and have now rebased, so it should be good to go.

PR #8, opened 2019-11-07
Adding exporting of cardinality rules

ngfreiter:
This adds exporting of cardinality rules onto a Structure Definition. Note that the code to `constrainCardinality` already existed, and had its own tests for validating a cardinality. This just uses that function in the `StructureDefinitionExporter`, and tests that the function is being used correctly.

### DIFF
--- a/src/export/StructureDefinitionExporter.ts
+++ b/src/export/StructureDefinitionExporter.ts
@@ -3,6 +3,7 @@ import { StructureDefinition } from '../fhirtypes';
 import { Profile, Extension } from '../fshtypes';
 import { FSHTank } from '../import';
 import { ParentNotDefinedError } from '../errors/ParentNotDefinedError';
+import { CardRule } from '../fshtypes/rules';
 
 /**
  * The StructureDefinitionExporter is a parent class for ProfileExporter and ExtensionExporter.
@@ -31,6 +32,30 @@ export class StructureDefinitionExporter {
   }
 
   /**
+   * Sets the rules for the StructureDefinition
+   * @param {StructureDefinition} structDef - The StructureDefinition to set rules on
+   * @param {Profile | Extension} fshDefinition - The Profile or Extension we are exporting
+   */
+  private setRules(structDef: StructureDefinition, fshDefinition: Profile | Extension) {
+    for (const rule of fshDefinition.rules) {
+      const element = structDef.findElementByPath(rule.path);
+      if (element) {
+        try {
+          if (rule instanceof CardRule) {
+            element.constrainCardinality(rule.min, rule.max);
+          }
+        } catch (e) {
+          console.error(e.stack);
+        }
+      } else {
+        console.error(
+          `No element found at path ${rule.path} for ${fshDefinition.name}, skipping rule`
+        );
+      }
+    }
+  }
+
+  /**
    * Exports Profile or Extension to StructureDefinition
    * @param {Profile | Extension} fshDefinition - The Profile or Extension we are exporting
    * @param {FSHTank} tank - The FSH tank we are exporting
@@ -46,6 +71,7 @@ export class StructureDefinitionExporter {
       throw new ParentNotDefinedError(fshDefinition.name, parentName);
     }
     this.setMetadata(structDef, fshDefinition, tank);
+    this.setRules(structDef, fshDefinition);
     // Set the rules
     return structDef;
   }
--- a/src/export/StructureDefinitionExporter.ts
+++ b/src/export/StructureDefinitionExporter.ts
@@ -23,7 +23,7 @@ export class StructureDefinitionExporter {
     structDef: StructureDefinition,
     fshDefinition: Profile | Extension,
     tank: FSHTank
-  ) {
+  ): void {
     structDef.name = fshDefinition.name;
     structDef.id = fshDefinition.id;
     structDef.url = `${tank.packageJSON.canonical}/StructureDefinition/${structDef.id}`;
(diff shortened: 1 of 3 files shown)


test/export/StructureDefinitionExporter.test.ts
```diff
@@ -95,4 +96,63 @@ describe('StructureDefinitionExporter', () => {
...
+    baseProfile.parent = 'Observation';
+    const changedProfile = new Profile('Foo');
+    changedProfile.parent = 'Observation';
+
+    const rule = new CardRule('subject');
+    rule.min = 1;
+    rule.max = '1';
+    changedProfile.rules.push(rule);
+
+    const baseStructDef = exporter.exportStructDef(baseProfile, input);
```
cmoesel: While this approach to getting the base cardinality is theoretically sound -- I think it might be better to get the base structure definition from the `FHIRDefinitions` and then load it via `StructureDefinition.fromJSON(...)`.  That way you really are dealing w/ the base resource and not a profile on it.

Maybe we should use this as an opportunity to implement `StructureDefinition.getBaseStructureDefinition()` -- which might be useful in other cases too.  Then you'd do:

```ts
const changedStructDef = exporter.exportStructDef(changedProfile, input);
const baseStructDef = changedStructDef.getBaseStructureDefinition();
```

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -95,4 +96,63 @@ describe('StructureDefinitionExporter', () => {
       exporter.exportStructDef(extension, input);
     }).toThrow('Parent Bar not found for Foo');
   });
+
+  // Rules
+  it('should emit an error and continue when the path is not found', () => {
+    // TODO: This should check for emitting an error once we have logging
+    const profile = new Profile('Foo');
+    const rule = new CardRule('fakePath');
+    rule.min = 0;
```
kjmahalingam: Something that I didn't view as odd until just seeing this here in writing...is there a reason that the `CardRule` has different types for the `max` and `min`? I know this is technically unrelated to your work, but if there's not a reason (or a good enough counter-argument), perhaps there's room to edit here.

My only guess is if we did this so that `max` could be `*`, but then my counterargument would be to represent that as `null` or `undefined` like we did in CIMPL.

cmoesel: It reflects how FHIR models min/max (which is `integer`/`string`, respectively).  Although we used `null` in CIMPL to represent max, in hindsight, I think that was confusing.  It's kind of odd for `null` or `undefined` to actually have specific semantic meaning (other than "no data").

kjmahalingam: Well, if FHIR itself does it that way, I won't dispute the point!

kjmahalingam:
I'm fine with this, pending changes that Chris requested based on the implementation of some tests.

test/export/StructureDefinitionExporter.test.ts
ngfreiter: Yeah that makes sense. I think I could actually also use the `ElementDefinition` functions that tracks `diff`s. Could it be cleaner to use those here?

test/export/StructureDefinitionExporter.test.ts
cmoesel: We could make the max type `number | string`, but I think that would make things more difficult because you'd always have to check which type it is before doing anything with it.

test/export/StructureDefinitionExporter.test.ts
cmoesel: I don't know...  If you use `diff`, I'd suggest you still check the regular `elements` _and_ the `diff`.  But really, I think we should have separate unit tests to ensure diff is working correctly anyway.  So I'd lean toward not using `diff` here.

If you don't want to pull and check the base definition here, I think you could also check the element's `base.min` and `base.max` which are also supposed to represent the original cardinality.  In past versions of FHIR though, I feel like `base` was not always present.

test/export/StructureDefinitionExporter.test.ts
kjmahalingam: I agree. Let's keep it as is.

test/export/StructureDefinitionExporter.test.ts
ngfreiter: Upon further investigation, I don't think `diff` is used yet when we make a new `ElementDefinition`, so I will go the `getBaseStructureDefinition` route.

src/export/StructureDefinitionExporter.ts
```diff
@@ -30,6 +31,30 @@ export class StructureDefinitionExporter {
     if (fshDefinition.description) structDef.description = fshDefinition.description;
   }
 
+  /**
+   * Sets the rules for the StructureDefinition
+   * @param {StructureDefinition} structDef - The StructureDefinition to set rules on
+   * @param {Profile | Extension} fshDefinition - The Profile or Extension we are exporting
+   */
+  private setRules(structDef: StructureDefinition, fshDefinition: Profile | Extension) {
```
cmoesel: I may be guilty of forgetting to do this myself, but I think it's good practice to always declare the intended return type of the function (which, in this case, would be `void`).  While ts can infer the return type, declaring it is more intentional and ensures that your function body adheres to that intent.

src/fhirtypes/StructureDefinition.ts
```diff
@@ -56,6 +56,11 @@ export class StructureDefinition {
    */
   elements: ElementDefinition[];
 
+  /**
+   * A base clone of the Structure Definition from before any rules were applied
+   */
+  _baseStructureDefinition: StructureDefinition;
```
cmoesel: Since this is intended to be _private_ we should enforce that via the `private` accessor.

ngfreiter:
You overestimated me by about 6 minutes and 5.3 seconds.

PR #12, opened 2019-11-12
Apply ValueSetRules when exporting a structure definition

mint-thompson:
Main functionality for applying ValueSetRule instances is contained in the bindToVS function.

As a side note: tests for cases of using invalid ValueSetRule definitions are not included
in the StructureDefinitionExporter test file, as these scenarios are already covered by tests
in the ElementDefintion.bindToVS test file.

This pull request addresses https://standardhealthrecord.atlassian.net/browse/CIMPL-115


mint-thompson:
@cmoesel I've added a pair of tests: one for applying a ValueSetRule to a path that doesn't support getting a binding, and one for applying a ValueSetRule of lower binding strength. I didn't make a lot of specific tests for these, since that level of detail is covered in the ElementDefinition.bindToVS tests. Please re-review when you have a chance.

PR #13, opened 2019-11-13
ElementDefinition.constrainType

cmoesel:
This PR adds the implementation of `ElementDefinition.constrainType`.  See the inline documentation and tests for information about what it supports and how to use it.

It also updates the grammar and importer to better support specifying references on the right-hand-side of the `only` rule.


src/fhirtypes/ElementDefinition.ts
```diff
@@ -238,6 +240,290 @@ export class ElementDefinition {
...
+   * - attempting to add a profile that doesn't match any of the existing types
+   * - attempting to add a base reference that wasn't already a reference
+   * - attempting to add a reference to a profile that doesn't match any of the existing references
+   * - specifying a target that does not match any of the existing type choices
+   * - specifying types or a target whose definition cannot be found
+   * @see {@link http://hl7.org/fhir/R4/elementdefinition-definitions.html#ElementDefinition.type}
+   * @param {{ type: string; isReference?: boolean }[]} types - the set of constrained types,
+   *   identified by id/type/url strings and an optional reference flag (defaults false)
+   * @param {ResolveFn} resolve - a function that can resolve a type to a StructureDefinition instance
+   * @param {string} [target] - a specific target type to constrain.  If supplied, will attempt to
```
kjmahalingam: Just a nitpick, but is there a reason that `target` is in brackets here? Is that some notation I don't understand, or is it a mistake?

src/fhirtypes/ElementDefinition.ts
```diff
@@ -238,6 +240,290 @@ export class ElementDefinition {
...
+
+  /**
+   * Given an input type (the constraint) and a set of target types (the things to potentially
+   * constrain), find the match and return information about it.
+   * @param {{ type: string; isReference?: boolean }} type - the constrained types, identified by
+   *   id/type/url string and an optional reference flag (defaults false)
+   * @param {ElementDefinitionType[]} targetTypes - the element types that the constrained type
+   *   can be potentially applied to
+   * @param {ResolveFn} resolve - a function that can resolve a type to a StructureDefinition instance
+   * @param {string} [target] - a specific target type to constrain.  If supplied, will attempt to
```
kjmahalingam: Same question as before.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -238,6 +240,290 @@ export class ElementDefinition {
...
+   * @throws {InvalidTypeError} when the type doesn't match any of the targetTypes
+   */
+  private findTypeMatch(
+    type: { type: string; isReference?: boolean },
+    targetTypes: ElementDefinitionType[],
+    resolve: ResolveFn
+  ): ElementTypeMatchInfo {
+    let matchedType: ElementDefinitionType;
+
+    // Get the lineage (type hierarchy) so we can walk up it when attempting to match
```
kjmahalingam: Lineage is a very nice term for this. I like it.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -238,6 +240,290 @@ export class ElementDefinition {
...
+
+    return results;
+  }
+
+  /**
+   * Given a new ElementTypeDefinition (based on the existing one), will apply the matching
+   * profiles and targetProfiles as appropriate.  If a targetType was specified, will filter out
+   * the other profiles or targetProfiles.
+   * @param {ElementDefinitionType} newType - the new type to apply the profiles/targetProfiles to
+   * @param {ElementDefinitionType} [targetType] - the (potentially null) target type for the
```
kjmahalingam: I'm beginning to guess that this is a convention I don't understand. Perhaps being the convention for optional arguments?

src/fhirtypes/ElementDefinition.ts
```diff
@@ -498,8 +784,13 @@ export class ElementDefinition {
...
    */
   unfold(resolve: ResolveFn = () => undefined): ElementDefinition[] {
-    if (this.type.length === 1) {
-      const def = resolve(this.type[0]);
+    if (
+      this.type.length === 1 &&
+      (this.type[0].profile == null || this.type[0].profile.length <= 1)
+    ) {
+      // If it has a profile, use that, otherwise use the code
+      const type = this.type[0].profile?.[0] ?? this.type[0].code;
```
kjmahalingam: This is the most beautiful line of code I've ever seen.

src/fhirtypes/ElementDefinition.ts
cmoesel: In JSDoc, you can use brackets to indicate that an argument is optional or can be null.

src/fhirtypes/ElementDefinition.ts
cmoesel: In JSDoc, you can use brackets to indicate that an argument is optional or can be null.

src/fhirtypes/ElementDefinition.ts
cmoesel: Thanks.  I felt a bit like I was a historian studying royalty, but it was fun nonetheless.

src/fhirtypes/ElementDefinition.ts
cmoesel: 🥇

src/fhirtypes/ElementDefinition.ts
cmoesel: There are a lot of `?`s, but it's definitely better than:
```js
let type;
if (this.type[0].profile && this.type[0].profile.length > 0) {
  type = this.type[0].profile[0];
} else {
  type = this.type[0].code;
}
```

PR #14, opened 2019-11-13
Apply FlagRules when exporting a structure definition

mint-thompson:
Not all of the requirements regarding how flags can be changed are implemented, but errors will be thrown when trying to unset the `mustSupport` or `isModifier` flags.

None of the FHIR resources have `mustSupport` set, so a custom resource is used for a test involving `mustSupport`.

This pull request addresses https://standardhealthrecord.atlassian.net/browse/CIMPL-113


src/fhirtypes/ElementDefinition.ts
```diff
@@ -524,6 +525,34 @@ export class ElementDefinition {
...
+   * @see {@link http://hl7.org/fhir/R4/elementdefinition-definitions.html#ElementDefinition.isSummary}
+   * @see {@link http://hl7.org/fhir/R4/elementdefinition-definitions.html#ElementDefinition.isModifier}
+   * @param mustSupport - whether to make this element a Must Support element
+   * @param summary - whether to include this element when querying for a summary
+   * @param modifier - whether this element acts as a modifier on the resource
+   * @throws {DisableFlagError} when attempting to disable a flag that cannot be disabled
+   */
+  applyFlags(mustSupport: boolean, summary: boolean, modifier: boolean): void {
+    const disabledFlags = [];
+    if (this.mustSupport && !mustSupport) {
```
cmoesel: If `mustSupport` is `null`, I would think that _should_ mean to leave it as-is; but I think this logic would treat it as if it's trying to change the value of `mustSupport` and issue an error.  I think we should check explicitly for `false` here (e.g., `mustSupport === false`).

src/fhirtypes/ElementDefinition.ts
```diff
@@ -524,6 +525,34 @@ export class ElementDefinition {
...
+      disabledFlags.push('Must Support');
+    }
+    if (this.isModifier && !modifier) {
```
cmoesel: Same as note above.  I think we should interpret a `null` value as "don't change it".  It's not always the case that the caller will know the value they want for _every_ flag.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -524,6 +525,34 @@ export class ElementDefinition {
...
+      disabledFlags.push('Is Modifier');
+    }
+    if (disabledFlags.length) {
+      throw new DisableFlagError(disabledFlags);
+    }
+    this.mustSupport = mustSupport;
+    this.isSummary = summary;
+    this.isModifier = modifier;
```
cmoesel: In the same theme, I think we should be setting each of these only if the incoming argument is not `null` or `undefined`.

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -153,4 +155,82 @@ describe('StructureDefinitionExporter', () => {
...
+      fs.readFileSync(
+        path.join(__dirname, '../fhirdefs/testdefs/patient-photo-must-support.json'),
+        'utf-8'
+      )
+    );
+    const mustSupportDefs: FHIRDefinitions = new FHIRDefinitions('4.0.1');
+    mustSupportDefs.add(jsonMustSupport);
+    const mustSupportExporter: StructureDefinitionExporter = new StructureDefinitionExporter(
+      mustSupportDefs
+    );
```
cmoesel: The above is clever, but actually unnecessary.  We have access to any of the profiles defined as part of core FHIR ([listed here](http://hl7.org/fhir/R4/profilelist.html)).  So, you could skip all this and just create a profile that declares its parent as `http://hl7.org/fhir/StructureDefinition/vitalsigns` (which has lots of MS, as [shown here](http://hl7.org/fhir/R4/vitalsigns.html)).

src/fhirtypes/ElementDefinition.ts
mint-thompson: I agree with and will implement your suggestions about only making changes or throwing errors when the arguments are actual `true` or `false` values.

test/export/StructureDefinitionExporter.test.ts
mint-thompson: That's

that's good. That's much better. I'll do that. I am extremely embarrassed by this.

cmoesel: Ha.  Nothing to be embarrassed by at all.  I don't think I _told_ anyone we supported those other FHIR profiles until now.  So really, I shouldn't have expected you to have known!

PR #20, opened 2019-11-14
Fixes needed to support npm publish

cmoesel:
In order to support npm publish we need to ensure that we do publish the right things (compiled js, json, and *.d.ts definition files) and we don't publish the wrong things (original ts files, map files, antlr jar, etc).  We also need to use a name that isn't already in use at npm (sushi *is* already in use).

To support this, the following changes were made:
- package.json: changed "name" to "fsh-sushi"
- package.json: added a "prepare" script to run "npm build" before packaging (to ensure compiled js is up to date)
- package.json: added a "prepublishOnly" script to run "npm run check" before publishing (to ensure we don't publish code w/ errors)
- package.json: added "files" to whitelist what we want published
- package.json: added "types" to point to the main *.d.ts file (currently empty, but that's ok)
- tsconfig.json: set "declaration" to true to generate .d.ts files in dist
- importText.ts: fixed bad relative paths in import statements

To test this, run:
```
npm pack
```

This will show you the list of things going into the package (which should _mainly_ be js, json, and d.ts file in dist -- as well as the FSH.g4 file).  It will also produce a `fsh-sushi-0.1.0.tgz` file.

To test the package itself, run:
```
npm install -g fsh-sushi
```

After this, you should be able to run sushi on the commandline:
```
sushi --help
```
or
```
sushi ./path/to/fshtank
```

When you're done, if you wish, you can uninstall it:
```
npm remove -g fsh-sushi
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,5 +1,5 @@
 {
-  "name": "sushi",
+  "name": "fsh-sushi",
   "version": "0.1.0",
   "lockfileVersion": 1,
   "requires": true,
--- a/package.json
+++ b/package.json
@@ -1,11 +1,11 @@
 {
-  "name": "sushi",
+  "name": "fsh-sushi",
   "version": "0.1.0",
   "description": "Sushi Unshortens Short Hand Inputs (FSH Compiler)",
   "scripts": {
     "build": "rm -rf dist && tsc && cp -r src/fhirdefs/fhir-4.0.1 dist/fhirdefs/fhir-4.0.1",
     "build:watch": "tsc -w",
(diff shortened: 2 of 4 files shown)


mint-thompson:
Everything seems to be working fine, although for the installation step from the package, I had to run `npm install -g fsh-sushi-0.1.0.tgz` rather than what was given in your instructions.

cmoesel:
Oh shoot!  You're totally right.  I was in a hurry to get the PR out and mis-wrote that.  Thanks for catching it!

PR #24, opened 2019-11-15
Implementation of fixValue

ngfreiter:
This adds `ElementDefinition.fixValue` which can be used to fix values on the primitive datatypes listed here https://www.hl7.org/fhir/datatypes.html#primitive, Quantity, Ratio, and code datatypes.


src/fshtypes/rules/FixedValueRule.ts
```diff
@@ -1,7 +1,7 @@
 import { Rule } from './Rule';
-import { Code, Quantity, Ratio } from '../index';
+import { Code, FshQuantity, FshRatio } from '../index';
```
kjmahalingam: I understand why we wanted to add the `Fsh` prefix to these classes, so the names don't conflict with the FHIR types. That said, is there a reason we don't also do that with `Code`? Is there no FHIR type for `Code`? I wonder if we should name it `FshCode` just for consistency unless there's a reason not to do that.

test/fhirtypes/ElementDefinition.fixFshQuantity.test.ts
```diff
@@ -0,0 +1,71 @@
...
+    it('should throw NoSingleTypeError when element has multiple types', () => {
+      const valueX = observation.elements.find(e => e.id === 'Observation.value[x]');
+      expect(() => {
+        valueX.fixFshQuantity(new FshQuantity(1.23, new Code('mm', 'http://unitsofmeasure.org')));
+      }).toThrow(
+        'Cannot fix Quantity value on this element since this element does not have a single type'
+      );
+    });
+
+    it('should throw QuantityAlreadFixed when the value is fixed to a different value', () => {
```
kjmahalingam: Typo, should be `QuantityAlreadyFixed`

src/fhirtypes/ElementDefinition.ts
```diff
@@ -602,6 +615,192 @@ export class ElementDefinition {
...
+    currentElementValue: boolean | string | number,
+    elementType: string
+  ): boolean {
+    if (currentElementValue != null && currentElementValue !== value) {
+      throw new PrimitiveValueAlreadyFixedError(value, elementType, currentElementValue);
+    }
+    return true;
+  }
+
+  private matchesRegex(value: string, regex: RegExp) {
```
kjmahalingam: I don't know that this function should exist. It seems like it only handles the one operation within it that we could just use instead.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -602,6 +615,192 @@ export class ElementDefinition {
...
+    } else if (value instanceof Code) {
+      this.fixFshCode(value);
+    } else if (value instanceof FshQuantity) {
+      this.fixFshQuantity(value);
+    } else if (value instanceof FshRatio) {
+      this.fixFshRatio(value);
+    }
+  }
+
+  private checkIfFixed(
```
kjmahalingam: This is misleadingly named. I would expect it to return `true` if it *is* fixed. It seems like it does the opposite.

src/fshtypes/rules/FixedValueRule.ts
cmoesel: I wouldn't be opposed to adding `Fsh` prefixes to other FSH types that might conflict.  The reason `Code` doesn't generally conflict is that in FHIR the `code` type is really just a string -- so we didn't really need to build a type around it.

src/fhirtypes/ElementDefinition.ts
cmoesel: Agreed.  I'm thinking maybe this function once did something else too?  But given its simple implementation, I'd just use the standard `/myregex/.test(value)` where it's needed.  If @ngfreiter did this based on conversation w/ me, then I apologize -- I probably did not express myself well.

src/fhirtypes/ElementDefinition.ts
cmoesel: Perhaps rename it to `checkIfCanBeFixed` or `checkIfFixable`

kjmahalingam:
I'm going to push a commit to address the issues that I mentioned.

cmoesel:
I won't have much time to review, so don't count on me -- but.. is this still "Preliminary"?  If not, I suggest you change the title.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -602,6 +623,301 @@ export class ElementDefinition {
...
+   */
+  private checkIfFixable(
+    value: boolean | string | number,
+    currentElementValue: boolean | string | number,
+    elementType: string
+  ): boolean {
+    if (currentElementValue != null && currentElementValue !== value) {
+      throw new ValueAlreadyFixedError(value, elementType, currentElementValue);
+    }
+    return true;
```
cmoesel: Since this can only return one value (`true`) and throws in all cases, we should consider making it a `void` function.  By stating that it returns a `boolean`, someone calling the function might be led to believe it returns `true`/`false` rather than `true/throw` -- and might structure their code around the misconception.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -602,6 +623,301 @@ export class ElementDefinition {
...
+  fixString(value: string): void {
+    if (!this.hasSingleType()) {
+      throw new NoSingleTypeError(typeof value);
+    }
+    const type = this.type[0].code;
+    if (type === 'string' && this.checkIfFixable(value, this.fixedString, type)) {
+      this.fixedString = value;
+    } else if (
+      type === 'uri' &&
+      /^\S*$/.test(value) &&
```
cmoesel: Assuming you got these regular expressions from the FHIR spec, I think it would be good to include a link to the spec where the regular expressions are specified.  You probably don't need a link for every regex -- but putting a link at the top of the function might be good.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -602,6 +623,301 @@ export class ElementDefinition {
...
+            value.toString(),
+            'Quantity',
+            foundFshQuantity.toString()
+          );
+        }
+        // if they do match, there is nothing to do, so return
+        return;
+      }
+      if (value.unit) {
+        this.fixFshCode(value.unit);
```
cmoesel: Oh nice.  I forgot that I implemented `fixedCode` for `Quantity`.  That said, for a more casual reader of the code, this line might be confusing.  I'd consider adding a comment indicating that the `fixFshCode` has special support for fixing `Quantity` units as codes.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -602,6 +623,301 @@ export class ElementDefinition {
...
+        }
+        // If they do match, there is nothing more to do, so return
+        return;
+      }
+      // There is no existing patternRatio, so create it
+      this.patternRatio = {};
+      this.patternRatio.numerator = {};
+      this.patternRatio.denominator = {};
+      this.patternRatio.numerator.value = value.numerator.value;
+      this.patternRatio.denominator.value = value.denominator.value;
```
cmoesel: Nitpicky and probably a matter of preference, but couldn't the above 5 statements have been:
```ts
this.patternRatio = {
  numerator: { value: value.numerator.value }
  denominator: { value: value.denominator.value }
}
```

src/fhirtypes/ElementDefinition.ts
```diff
@@ -614,27 +930,30 @@ export class ElementDefinition {
...
    * @throws {CodeAlreadyFixedError} where the code is already fixed to a different code
    */
-  fixFshCode(code: Code): void {
+  fixFshCode(code: FshCode): void {
     // This is the element to fix it to
-    const types = this.type || [];
-    if (types.some(t => t.code === 'CodeableConcept')) {
+    if (!this.hasSingleType()) {
+      throw new NoSingleTypeError('Code');
+    }
```
cmoesel: Thanks for fixing this!

src/fhirtypes/ElementDefinition.ts
cmoesel: I still feel like it's better as a `void` function, but I'm willing to be wrong if the rest of the team disagrees with me.  The new approach (`true|undefined`) does address my main concern.

ngfreiter: Ah sorry I meant to respond to this explaining my choice there. The issue with making the function `void` is that we then can't use the `checkIfFixable` function in an `if` statement, because we can't test a function that returns `void` for true or false. So I could make it `void`, but then I would have to move the `checkIfFixable` function out of the condition and into the `if` block itself. Which you could argue is a more sensible way of doing things, but I personally liked having all the checking logic in the condition of the `if`.

PR #26, opened 2019-11-18
Add FixedValueRule and tests

jafeltra:
Added the FixedValueRule to StructureDefinitionExporter and two tests. The tests only test that a code is correct fixed to a value and that an incorrect fixed value is not set. I believe that all the tests for the actual `fixValue` function cover all the other cases for calling this function and all its other error cases, but if anyone feels I missed something, I'm happy to add additional tests.

### DIFF
--- a/test/export/StructureDefinitionExporter.test.ts
+++ b/test/export/StructureDefinitionExporter.test.ts
@@ -424,8 +424,8 @@ describe('StructureDefinitionExporter', () => {
     expect(constrainedValue.type).toHaveLength(11);
   });
 
-  // FixValue rule
-  it('should apply a correct FixValueRule', () => {
+  // Fixed Value Rule
+  it('should apply a correct FixedValueRule', () => {
     const profile = new Profile('Foo');
     profile.parent = 'Observation';
 
@@ -446,7 +446,7 @@ describe('StructureDefinitionExporter', () => {
     });
   });
 
-  it('should not apply an incorrect FixValueRule', () => {
+  it('should not apply an incorrect FixedValueRule', () => {
     // TODO: this should check for emitting an error once logging is set up
     const profile = new Profile('Foo');
     profile.parent = 'Observation';
--- a/test/export/StructureDefinitionExporter.test.ts
+++ b/test/export/StructureDefinitionExporter.test.ts
@@ -1,8 +1,14 @@
 import { StructureDefinitionExporter } from '../../src/export';
 import { FSHTank, FSHDocument } from '../../src/import';
 import { FHIRDefinitions, load } from '../../src/fhirdefs';
-import { Profile, Extension } from '../../src/fshtypes';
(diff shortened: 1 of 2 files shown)


test/export/StructureDefinitionExporter.test.ts
```diff
@@ -1,8 +1,14 @@
 import { StructureDefinitionExporter } from '../../src/export';
```
kjmahalingam: For this file in general, in my usual line of inconsequential typo feedback, there are a few instances of `FixValue` or `FixValueRule` that should be changed to `FixedValue` or `FixedValueRule`, respectively.

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -418,6 +424,47 @@ describe('StructureDefinitionExporter', () => {
     expect(constrainedValue.type).toHaveLength(11);
   });
 
+  // FixValue rule
```
kjmahalingam: Typo

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -418,6 +424,47 @@ describe('StructureDefinitionExporter', () => {
...
+  it('should apply a correct FixValueRule', () => {
```
kjmahalingam: Typo

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -418,6 +424,47 @@ describe('StructureDefinitionExporter', () => {
...
+    const baseCode = baseStructDef.findElement('Observation.code');
+    const fixedCode = sd.findElement('Observation.code');
+
+    expect(baseCode.patternCodeableConcept).toBeUndefined();
+    expect(fixedCode.patternCodeableConcept).toEqual({
+      coding: [{ code: 'foo', system: 'http://foo.com' }]
+    });
+  });
+
+  it('should not apply an incorrect FixValueRule', () => {
```
kjmahalingam: Typo

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -418,6 +424,47 @@ describe('StructureDefinitionExporter', () => {
...
+    // TODO: this should check for emitting an error once logging is set up
```
kjmahalingam: I might be misunderstanding, but isn't logging already set up? Why can't we perform this test already?

jafeltra: We have the logging set up, but we haven't set up the ability to check what errors are emitted directly in a test yet. That was going to be an additional feature to be added after the initial logging task, if I'm recalling correctly.

kjmahalingam: Ah yes, I understand now. Cool, then this point sounds good to me.

test/export/StructureDefinitionExporter.test.ts
jafeltra: Typos should be addressed now! Good catch!

PR #27, opened 2019-11-18
Fixing a bug with ordering of elements

ngfreiter:
Currently, when adding an element, we find an element that starts with the same id, and add the element after that. So an `identifier.value` element should be added after a `identifier` element. However, if there is an `id` element, then the `identifier.value` element would be added after that, since `identifiier.value` starts with `id`. This fixes that issue by checking for a `.` or `:` since these demarcate the complete path elements.


ngfreiter:
Sorry for the force-push, I am forgetful of my rebasing today.

PR #29, opened 2019-11-18
Fix issues w/ metadata (derivation, baseDefinition, context) and bump version

jafeltra:
Fixes the bug that @kjmahalingam noticed where the `derivation` needs to be set to `constraint`. Also bumps the version number in preparation for a bug fix release and updates a dependency with a security vulnerability.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "fsh-sushi",
-  "version": "0.2.0",
+  "version": "0.2.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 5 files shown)


cmoesel:
Thanks for jumping on this @jafeltra!  In future versions of FSH, however, we will likely support logical models -- which are `StructureDefinition`s whose `derivation` is _not_ `"constraint"`.  For that reason, I think it's better to set `derivation` in code that is specific to profiles and extensions (rather than hard-coded into `StructureDefinition.ts`).

It also would be awesome to have a couple of _simple_ tests.  This might even be just adding another `expect(...)` in some of the existing tests.

kjmahalingam:
I also noticed that, sadly, when we run the FHIR validator on the `StructureDefinition-fsh-birthsex.json` extension that we output, we get this error:

`  Error @ StructureDefinition (line 1, col2) : sdf-5: If the structure defines an extension then the structure must have context information [(type != 'Extension') or (derivation = 'specialization') or (context.exists())]`

@cmoesel Do you have a preferred solution to this issue?

cmoesel:
Ugh.  The correct solution is _probably_ to have a special top-level keyword for `Context` since it is needed in every extension -- or to have authors specify it via the caret syntax `^context[0].type = "element"` and `^context[0].expression = "Patient"`-- but I don't want to try to introduce either of those at this point.  For now, we probably should hard code it (for extensions only) to:

```json
"context" : [
    {
      "type" : "element",
      "expression" : "*"
    }
  ]
```

cmoesel:
@jafeltra and @kjmahalingam -- I've just pushed commits to address three issues:
(1) Set `derivation` to `constraint` for `Profile`/`Extension` only
(2) Set `baseDefinition` to the parent `url` (oops)
(3) Set a default extension `context` when necessary

I'm also going to rename the PR to more closely reflect what it is now.

PR #31, opened 2019-11-21
Instance may have title, only has FixedValueRules

mint-thompson:
Just a small update to the Instance class to prepare for the import/export work.


cmoesel:
Travis seems stuck... We'll give it a bit more to try to work itself out -- but if it remains stuck, I'll verify locally and then merge.

PR #32, opened 2019-11-21
Import contains rule

ngfreiter:
This adds a `ContainsRule` class that stores information for constraints that use the `contains` keyword https://github.com/HL7/fhir-shorthand/wiki/2.4-Constraints. It also adds logic to the importer that will use the `ContainsRule` class to store information when the `contains` keyword is used on a `Profile` or an `Extension`, and it adds tests of this functionality.

This also update `.gitignore` so that any directory starting with `out` is ignored. I was running the cli a few times and this became an issue for me.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,7 +12,7 @@ pids
 logs
 results
 tmp
-out
+/out*
 
 # Build
 public/css/main.css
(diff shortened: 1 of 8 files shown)


ngfreiter:
I went with `item` and also used the existing functionality for resolving an Alias when setting a value to set the `item` on the `ContainsRule`.

PR #33, opened 2019-11-25
Import instances when parsing FSH

mint-thompson:
This change involves a small update to the grammar, but it should not be necessary to recompile it manually in order to test it. New tests are in `FSHImporter.Instance.test.ts`. Note that the tests do not include every possible datatype for a `FixedValueRule`, as the tests in `FSHImporter.Profile.test.ts` already provide coverage of that code.

The current implementation is based on the documentation available at https://github.com/HL7/fhir-shorthand/wiki/3.7-Instances.

This pull request addresses https://standardhealthrecord.atlassian.net/browse/CIMPL-147


antlr/src/main/antlr/FSH.g4
```diff
@@ -1,12 +1,13 @@
...
 
 doc:                entity*;
-entity:             alias | profile | extension | invariant;
+entity:             alias | profile | extension | invariant | instance;
 
 alias:              KW_ALIAS SEQUENCE EQUAL SEQUENCE;
 
 profile:            KW_PROFILE SEQUENCE sdMetadata+ sdRule*;
 extension:          KW_EXTENSION SEQUENCE sdMetadata* sdRule*;
+instance:           KW_INSTANCE SEQUENCE KW_INSTANCEOF SEQUENCE title? fixedValueRule*;
```
cmoesel: This requires the exact order:
```
Instance: xxx
InstanceOf: yyy
Title: zzz
```
In our other types (profile and extension), the metadata can come in any order (via the `sdMetadata*`). For user-friendliness, I think we should do the same here.  If we make an `instanceMetadata` parser rule, this also gives us a good place to put further metadata if we come up with any.  The downside is that this means that we can't mark `InstanceOf` as _required_ in the grammar -- so we'll have to ensure it is specified when we parse it instead.

src/import/FSHImporter.ts
```diff
@@ -125,6 +137,18 @@ export class FSHImporter extends FSHVisitor {
     });
   }
 
+  visitInstance(ctx: pc.InstanceContext) {
+    const instance = new Instance(ctx.SEQUENCE()[0].getText())
+      .withLocation(this.extractStartStop(ctx))
+      .withFile(this.file);
+    instance.instanceOf = ctx.SEQUENCE()[1].getText();
```
cmoesel: It's possible that the parent might be specified via a previously defined alias, so I think we actually want to use:
```ts
instance.instanceOf = this.aliasAwareValue(ctx.SEQUENCE()[1].getText());
```

mint-thompson: Ah, that's correct! I'll make that change.

antlr/src/main/antlr/FSH.g4
mint-thompson: I have mixed feelings about this one. As you point out, it's more user-friendly to not enforce a specific order to fields. The documentation implies, but does not outright state, that these should always be the first two keywords when defining an instance: 
> Instance specifications are introduced using the `Instance` and `InstanceOf` keywords.

And, all examples have `Instance` as the first and `InstanceOf` as the second. Enforcing the order like this is not something done by the grammar for other entities. And, thinking ahead to a potential future where there are multiple required metadata fields, it would be unusual to enforce an order. Therefore, I'll go ahead and make it an optional field from the grammar's perspective, and enforce its presence in `FSHImporter`.

kjmahalingam: I'm comfortable with the ultimate decision here to make it an optional field in the grammar and enforce it in the importer. I have a similar feeling to @mint-thompson about how, in practice, it seems like it will always be in this exact order. But I will similarly concede that the benefits of flexibility outweigh the negatives.

src/import/FSHImporter.ts
```diff
@@ -120,16 +127,32 @@ export class FSHImporter extends FSHVisitor {
...
+          declaredParent = true;
         } else if (pair.key === SdMetadataKey.Title) {
+          if (def.title) {
+            throw new DuplicateMetadataError('Title', def.title);
+          }
           def.title = pair.value;
         } else if (pair.key === SdMetadataKey.Description) {
+          if (def.description) {
+            throw new DuplicateMetadataError('Description', def.description);
+          }
```
cmoesel: So... I apologize (a lot)... I wasn't very clear when I said I think an error should be generated.  Probably because I wasn't clearly thinking it through myself.  But... in general, I think we shouldn't halt execution unless we have truly encountered an _irrecoverable_ error.  If at all possible, we should generally prefer to log an error and keep going.  So... I think the above could actually be modified to support that and be a bit cleaner.

There are probably a few variations on the approach (involving still throwing errors but catching them sooner, or not throwing at all but just logging -- as well as variations on if we look for duplicates in a standalone function or do right in the loop).  One approach, sticking more closely to what you already have might look something like this:
```ts
const seenPairs: Map<InMetadataKey, string> = new Map();
metaCtx
  .map(sdMeta => this.visitSdMetadata(sdMeta))
  .forEach(pair => {
    if (seenPairs.has(pair.key)) {
      logger.error(`Metadata field ${pair.key} already declared with value ${pair.value}.`);
      return;
    }
    seenPairs.set(pair.key, pair.value);
    if (pair.key === SdMetadataKey.Id) {
      def.id = pair.value;
    } // etc.
  });
```

For the above to work, you'd also need to change the enum to be a [string-enum](https://www.typescriptlang.org/docs/handbook/enums.html#string-enums).  (Also, warning: I wrote that code write into the code review box, not my IDE -- so I apologize for any small syntax errors!)

test/import/FSHImporter.Extension.test.ts
```diff
@@ -54,6 +55,66 @@ describe('FSHImporter', () => {
...
+        Parent: ParentExtension
+        Id: some-extension
+        Title: "Some Extension"
+        Description: "An extension on something"
+        Parent: ParentExtension
+        `;
+
+        expect(() => {
+          importText(input);
+        }).toThrow(DuplicateMetadataError);
```
cmoesel: Unfortunately, if we don't throw errors any more, then this test becomes less interesting.  Instead of checking for a thrown error, it should check to ensure that the *first* value of the duplicated key is the one kept.

Once we've implemented the bit about being able to validate that an _error_ was logged, we can revisit the test and add in that logic.

src/import/FSHImporter.ts
```diff
@@ -126,6 +153,44 @@ export class FSHImporter extends FSHVisitor {
...
+        }
+        seenPairs.set(pair.key, pair.value);
+        if (pair.key === InstanceMetadataKey.InstanceOf) {
+          instance.instanceOf = pair.value;
+        } else if (pair.key === InstanceMetadataKey.Title) {
+          instance.title = pair.value;
+        }
+      });
+    if (!instance.instanceOf) {
+      throw new RequiredMetadataError('InstanceOf', 'Instance');
```
cmoesel: If I'm following the code path correctly, I think this error will bubble all the way to the top and ultimately halt execution of the whole import process.  I think we want to catch this error in `visitInstance` and then log it.

mint-thompson: That makes sense. As for updating the test associated with this (`FSHImporter.Instance.Test.ts:44`), would it be fine to just check that the result has no instances?

cmoesel: Yes, I think so -- although I suppose you _could_ also use your fancy jest spy to see if an error was logged.  But we do actually have that functionality (check for logged errors) as a separate story in our backlog.  So it could wait for that instead.  Up to you.  But either way, at least check that no instance is imported.

PR #34, opened 2019-11-26
Set Instance Property

ngfreiter:
This addresses CIMPL-98, https://standardhealthrecord.atlassian.net/browse/CIMPL-98, by implementing `StructureDefinition.setInstancePropertyByFshPath`. This will be used with the `^` syntax to directly set attributes on a `StructureDefinition` when making a profile, as described here https://github.com/HL7/fhir-shorthand/wiki/2.2-Paths#addressing-structure-definition-attributes-directly.

The overall approach is to get a path, and then check if that path is valid using the FHIR StructureDefinition of the StructureDefinition resource. This is done by the `validateValueAtPath` function. This function will also add implied 0 indices to the path, meaning that if we have a path `foo.bar`, and `foo` is an array, that path is treated as `foo[0].bar`. If the path is valid, we set the value at that path on our `StructureDefinition` class.

### DIFF
--- a/src/fhirtypes/StructureDefinition.ts
+++ b/src/fhirtypes/StructureDefinition.ts
@@ -230,11 +230,14 @@ export class StructureDefinition {
   ): void {
     // The StructureDefinition of the StructureDefinition resource is stored on our StructureDefinition class
     const structureDefinitionStructureDefinition = this.getStructureDefinitionStructureDefinition();
-    const fixedValue = structureDefinitionStructureDefinition.canFixValue(path, value, resolve);
+    const { fixedValue, pathParts } = structureDefinitionStructureDefinition.validateValueAtPath(
+      path,
+      value,
+      resolve
+    );
     if (fixedValue != null) {
       // If we can fix the value on the StructureDefinition StructureDefinition, then we can set the
       // instance property here
-      const pathParts = this.parseFSHPath(path);
       // eslint-disable-next-line
       let current: any = this;
       for (const [i, pathPart] of pathParts.entries()) {
@@ -248,21 +251,21 @@ export class StructureDefinition {
           for (let i = current[key].length; i <= index; i++) {
             current[key].push({});
           }
-          current = current[key][index];
-        } else if (current[key] == null && i < pathParts.length - 1) {
-          // If we aren't yet at the end of the path, create a new empty object
-          current[key] = {};
-          current = current[key];
+          // If it isn't the last element, move on, if it is, set the value
+          if (i < pathParts.length - 1) {
+            current = current[key][index];
+          } else {
+            current[key][index] = fixedValue;
+          }
+        } else {
+          // If it isn't the last element, move on, if it is, set the value
+          if (i < pathParts.length - 1) {
+            if (current[key] == null) current[key] = {};
+            current = current[key];
+          } else {
+            current[key] = fixedValue;
+          }
         }
-      }
-      const pathEnd = pathParts.slice(-1)[0];
-      const index = this.getArrayIndex(pathEnd);
-      if (index != null) {
-        // If we are ending on an array, an empty {} should exist to copy values into
-        Object.assign(current, fixedValue);
-      } else {
-        // If we are not ending on an array, we set the key-value pair
-        current[pathParts.slice(-1)[0].base] = fixedValue;
       }
     }
   }
@@ -360,7 +363,11 @@ export class StructureDefinition {
    * @throws {CannotResolvePathError} when the path cannot be resolved to an element
    * @returns {any} - The object or value to fix
    */
-  canFixValue(path: string, value: FixedValueType, resolve: ResolveFn = () => undefined): any {
+  validateValueAtPath(
+    path: string,
+    value: FixedValueType,
+    resolve: ResolveFn = () => undefined
+  ): { fixedValue: any; pathParts: PathPart[] } {
     const pathParts = this.parseFSHPath(path);
     let currentPath = '';
     let currentElement: ElementDefinition;
@@ -385,6 +392,10 @@ export class StructureDefinition {
         // We throw an error if the currentElement doesn't exist, has been zeroed out,
         // or is being incorrectly accessed as an array
         throw new CannotResolvePathError(path);
+      } else if (currentElement.max === '*' && arrayIndex == null) {
+        // Modify the path to have 0 indices
+        if (!pathPart.brackets) pathPart.brackets = [];
+        pathPart.brackets.push('0');
       }
     }
     const clone = currentElement.clone();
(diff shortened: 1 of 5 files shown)


src/fhirtypes/StructureDefinition.ts
```diff
@@ -316,6 +437,21 @@ export class StructureDefinition {
...
+  /**
+   * Tests to see if the last bracket in a PathPart is a positive int, and if so returns it
+   * @param {PathPart} pathPart - The part of the path to test
+   * @returns {number} The index if it exists and is positive, otherwise undefined
+   *
+   */
+  private getArrayIndex(pathPart: PathPart): number {
+    const lastBracket = pathPart.brackets?.slice(-1)[0];
+    let arrayIndex: number;
+    if (/[0]|[-+]?[1-9][0-9]*/.test(lastBracket)) {
```
cmoesel: I think this regex may be more confusing than it needs to be...  and isn't quite correct, in that it would match on anything with a number in it (for example, it would match on `'abc25xyz'`).  I don't see why we shouldn't allow leading zeros (e.g., `01`), even if they are a little weird.  So fixing it up to not allow those other non-digit characters, and to allow leading zeroes (mainly to simplify the regex), we would have something like:
```
if (/^[-+]?\d+$/.test(lastBracket)) {
```

test/fhirtypes/StructureDefinition.test.ts
```diff
@@ -259,6 +260,255 @@ describe('StructureDefinition', () => {
...
+
+    // Simple values
+    it('should set an instance property which has a value', () => {
+      observation.setInstancePropertyByPath('version', '1.2.3', getResolver(defs));
+      expect(observation.version).toBe('1.2.3');
+    });
+
+    it('should set an instance property which must be created', () => {
+      observation.setInstancePropertyByPath('contextInvariant', 'foo', getResolver(defs));
+      expect(observation.contextInvariant[0]).toBe('foo');
```
cmoesel: This is actually testing a few things here -- arrays, arrays with implied index 0, and a property which must be created (which is the purpose of the test).  If there is another property that must be created and is *not* an array, that might be better since it is a more specific "unit".  If not, then I'd suggest adding `[0]` to the path so you're not testing implicit indexes here.

src/fhirtypes/StructureDefinition.ts
```diff
@@ -60,11 +62,13 @@ export class StructureDefinition {
    * A base clone of the Structure Definition from before any rules were applied
    */
   private _baseStructureDefinition: StructureDefinition;
+  private _structureDefinitionStructureDefinition: StructureDefinition;
```
cmoesel: I don't know about you, but I'd be comfortable with `sdStructureDefinition`.  It's shorter and less repetitive.

src/fhirtypes/StructureDefinition.ts
```diff
@@ -60,11 +62,13 @@ export class StructureDefinition {
...
 
   /**
    * Constructs a StructureDefinition with a root element.
+   * @param {StructureDefinition} structureDefinitionStructureDefinition - The StructureDefinition of StructureDefinition
    */
-  constructor() {
+  constructor(structureDefinitionStructureDefinition: StructureDefinition = null) {
```
cmoesel: I don't love having to pass in the StructureDefinition definition into the constructor.  It just feels real weird to me.

I think we require the `resolve` function as an argument to any functions that need the structdef anyway, and `resolve` can look up the StructureDefinition in the FHIR defs, so... I'd suggest taking this out of the constructor -- and using `resolve` at the appropriate time/place instead.

src/fhirtypes/StructureDefinition.ts
```diff
@@ -83,6 +88,13 @@ export class StructureDefinition {
     return this._baseStructureDefinition;
   }
 
+  /**
+   * Get the Structure Definition for Structure Definition
+   */
+  getStructureDefinitionStructureDefinition() {
```
cmoesel: Should this maybe be `private`?  Is there a good reason to expose it?

cmoesel: Also, if you made this take a `resolve` function as an argument, then it could essentially do something like:
```
if (this._structureDefinitionStructureDefinition == null) {
  this._structureDefinitionStructureDefinition = resolve('StructureDefinition');
}
return this._structureDefinitionStructureDefinition;
```
This allows dynamic resolution of the SD when needed, but caches it after that.

src/fhirtypes/StructureDefinition.ts
```diff
@@ -263,8 +328,11 @@ export class StructureDefinition {
    * @param {any} json - the FHIR 3.0.1 JSON representation of a StructureDefinition to construct
    * @returns {StructureDefinition} a new StructureDefinition instance representing the passed in JSON
    */
-  static fromJSON(json: LooseStructDefJSON): StructureDefinition {
-    const sd = new StructureDefinition();
+  static fromJSON(
+    json: LooseStructDefJSON,
+    structureDefinitionStructureDefinition: StructureDefinition = null
```
cmoesel: If we take the SD out of the constructor, then we don't need it here anymore either.

src/fhirtypes/StructureDefinition.ts
```diff
@@ -205,6 +217,59 @@ export class StructureDefinition {
...
+   * This function sets an instance property of an SD if possible
+   * @param {string} path - The path to the ElementDefinition to fix
+   * @param {FixedValueType} value - The value to fix
+   * @param {ResolveFn} resolve - A function that can resolve a type to a StructureDefinition instance
+   */
+  setInstancePropertyByPath(
+    path: string,
+    value: FixedValueType,
+    resolve: ResolveFn = () => undefined
+  ): void {
```
cmoesel: So... I'm realizing now that this approach will work well for many parts of the SD, but it will _not_ work for `snapshot` or `differential` (or any of their child paths) since we abstract that to `elements`.  For now, I'm going to say that's OK -- and in fact, we should embrace it; throw an error if the path starts with `snapshot` or `differential`.  To set instances on those, they should be using the FSH path to identify the element and then caret path after that.

src/fhirtypes/StructureDefinition.ts
ngfreiter: Oh nice this is much cleaner.

src/fhirtypes/StructureDefinition.ts
```diff
@@ -205,6 +217,59 @@ export class StructureDefinition {
...
+      for (const [i, pathPart] of pathParts.entries()) {
+        const key = pathPart.base;
+        // If this part of the path indexes into an array, the index will be the last bracket
+        const index = this.getArrayIndex(pathPart);
+        if (index != null) {
+          // If the array doesn't exist, create it
+          if (current[key] == null) current[key] = [];
+          // If the index doesn't exist in the array, add it and lesser indices
+          for (let i = current[key].length; i <= index; i++) {
+            current[key].push({});
```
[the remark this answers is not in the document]
kjmahalingam: @ngfreiter I see that you've addressed all feedback except I don't see an 'outdated' mark on this one. Did I miss something here?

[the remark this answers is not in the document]
kjmahalingam: Never mind, it was right above it all along!

PR #35, opened 2019-11-26
Add support for FSHy Parents

kjmahalingam:
This PR adds support for FSHy Parents. This means that if the Parent of a Profile or Extension is another Profile or Extension, respectively, then the exporter process will export in the correct order so that Profiles and Extensions can be based on one another.

In order to test this, just run `npm run check` on this branch. I added what I believe to be comprehensive tests to cover various cases. If you think I'm missing any, let me know!

### DIFF
--- a/src/export/ProfileExporter.ts
+++ b/src/export/ProfileExporter.ts
@@ -15,17 +15,16 @@ export class ProfileExporter extends StructureDefinitionExporter {
    * @returns {StructureDefinition[]}
    */
   export(tank: FSHTank): StructureDefinition[] {
-    const structDefs: StructureDefinition[] = [];
     for (const doc of tank.docs) {
       for (const profile of doc.profiles.values()) {
         try {
           const structDef = this.exportStructDef(profile, tank);
-          structDefs.push(structDef);
+          this.structDefs.push(structDef);
         } catch (e) {
           logger.error(e.message);
         }
       }
     }
-    return structDefs;
+    return this.structDefs;
   }
 }
--- a/src/export/ProfileExporter.ts
+++ b/src/export/ProfileExporter.ts
@@ -5,20 +5,19 @@ import { FHIRDefinitions } from '../fhirdefs';
 import { logger } from '../utils/FSHLogger';
 
 export class ProfileExporter extends StructureDefinitionExporter {
-  constructor(FHIRDefs: FHIRDefinitions) {
-    super(FHIRDefs);
+  constructor(FHIRDefs: FHIRDefinitions, tank: FSHTank) {
+    super(FHIRDefs, tank);
   }
 
   /**
    * Exports Profiles to StructureDefinitions
-   * @param {FSHTank} tank - The FSH tank we are exporting
    * @returns {StructureDefinition[]}
    */
-  export(tank: FSHTank): StructureDefinition[] {
-    for (const doc of tank.docs) {
+  export(): StructureDefinition[] {
+    for (const doc of this.tank.docs) {
       for (const profile of doc.profiles.values()) {
         try {
-          this.exportStructDef(profile, tank);
+          this.exportStructDef(profile);
         } catch (e) {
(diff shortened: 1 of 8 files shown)


kjmahalingam:
Comments that have been addressed from a code review:

- The tests have been updated to check the metadata of parents, in order to verify that the parent information is actually being processed correctly.
- The exporters have been refactored so that they take in a tank as an argument to their constructor, so it doesn't need to be passed into the export function. This way, there will be one exporter per tank when we have the capability to process multiple.
-  Some utility functions have been added to the FSHTank class so that we have access to them elsewhere. They have also been used in places that existed where they would have been useful.
- Moved extraneous resolution functionality in to the resolver function rather than the exporter function. This way, the resolver can also be used to resolve parents that don't yet exist by first exporting them.

src/export/ProfileExporter.ts
```diff
@@ -5,27 +5,22 @@ import { FHIRDefinitions } from '../fhirdefs';
...
-        } catch (e) {
-          logger.error(e.message);
-        }
+  export(): StructureDefinition[] {
+    for (const profile of this.tank.getAllProfiles()) {
+      try {
+        this.exportStructDef(profile);
+      } catch (e) {
+        logger.error(e.message);
       }
```
ngfreiter: For both `ProfileExporter` and `ExtensionExporter`, I think looping in this way can be inefficient. Let's say we have Foo with parent Bar with parent Baz. If Foo is exported first, then Bar and Baz (by my understanding) will be exported during the export of Foo. But then on the next iteration of the loop won't we re-export Bar, getting the Baz SD, copying it, and applying all the rules again? Then the next iteration I think we will re-export Baz, getting the Resource SD, copying it, and applying all the rules. I have stepped through the code and I'm pretty confident this is what is happening. I guess it isn't explicitly going to cause incorrect behavior, but I think it could be much more efficient if when we are exporting a SD, we check if that SD has already been exported, either here or at the beginning of the `exportStructDef` function.

test/export/ExtensionExporter.test.ts
```diff
@@ -46,8 +46,75 @@ describe('ExtensionExporter', () => {
     const extensionBar = new Extension('Bar');
     doc.extensions.set(extensionFoo.name, extensionFoo);
     doc.extensions.set(extensionBar.name, extensionBar);
-    const exported = exporter.export(input);
+    const exported = exporter.export();
     expect(exported.length).toBe(1);
     expect(exported[0].name).toBe('Bar');
   });
+
+  it('should export extensions with FSHy parents', () => {
```
ngfreiter: I am willing to be wrong on this because I'd rather have more tests than less, but couldn't you make a case that the new tests here for `ExtensionsExporter` and `ProfileExporter` are testing the `exportStructDef` function in the same way? I think it might make more sense to have maybe one test in each of these files that ensures basic FSHy paths works for both Extensions and Profiles, but to then have one set of more in depth tests in the `StructureDefinitionExporter.test.ts` file.

src/export/ProfileExporter.ts
kjmahalingam: Good call. I moved the check for whether something was exported to the beginning of `exportStructDef` rather than the end.

test/export/ExtensionExporter.test.ts
kjmahalingam: I'm not sure I agree in this specific case. It seems to me that it would be better to test individually in each exporter because technically parents should only work from profile to profile or extension to extension. Adding tests to the base `StructureDefinitionExporter` would require either picking one or doing both. I suppose that's also fine, but I think I personally prefer it this way. Especially given the fact that the rest of the `ProfileExporter` and `ExtensionExporter` tests are duplicative in this way.

ngfreiter: That seems fair to me, I'd rather more tests than too few anyway.

PR #36, opened 2019-11-29
Include FSH source information in log statements

mint-thompson:
This comes with several test cases that check the source information, but not the specific text associated with the error being thrown.

I ended up only using the `WithSource` interface on `ParentNotDefinedError`, since that's the only error that gets caught and logged in a location where the source information would not normally be accessible.

Also, as it turns out, jest's `SpyInstance` can do some pretty handy stuff.

This pull request addresses https://standardhealthrecord.atlassian.net/browse/CIMPL-141.

### DIFF
--- a/src/export/ProfileExporter.ts
+++ b/src/export/ProfileExporter.ts
@@ -22,7 +22,7 @@ export class ProfileExporter extends StructureDefinitionExporter {
           const structDef = this.exportStructDef(profile, tank);
           structDefs.push(structDef);
         } catch (e) {
-          logger.error(e.message);
+          logger.error(e.message, e.sourceInfo);
         }
       }
     }
--- a/src/errors/ParentNotDefinedError.ts
+++ b/src/errors/ParentNotDefinedError.ts
@@ -1,8 +1,10 @@
 import { Annotated } from './Annotated';
+import { WithSource } from './WithSource';
+import { SourceInfo } from '../fshtypes';
 
-export class ParentNotDefinedError extends Error implements Annotated {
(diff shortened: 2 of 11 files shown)


kjmahalingam:
Looks great and works great, @mint-thompson ! You're right, the `SpyInstance` is a very cool feature. I like it.

Could you just rebase to resolve some conflicts here, and then once you do that, we should be good to go? Additionally, could you also just check the new tests for `ProfileExporter` and `ExtensionExporter` to see if it makes sense to add error checks to those?

After that, we are good to go!

src/export/ProfileExporter.ts
```diff
@@ -22,7 +22,7 @@ export class ProfileExporter extends StructureDefinitionExporter {
           const structDef = this.exportStructDef(profile, tank);
           structDefs.push(structDef);
         } catch (e) {
-          logger.error(e.message);
+          logger.error(e.message, e.sourceInfo);
```
kjmahalingam: Another quick comment...is there a reason that we don't want source info in the extensions? It seems that we explicitly log them for profiles, but there's no support added for extensions. There's probably a reason I'm just not aware of, but I want to clarify that.

mint-thompson: Nope, just forgot to add it in. I'll be sure to include it in the next set of updates after resolving the merge conflicts.

kjmahalingam: Ah, okay. Sounds good. Then I guess as you do that, just make sure to update the corresponding tests for extensions and their exporter as well. Thanks!

PR #38, opened 2019-12-02
Export contains rule

ngfreiter:
This adds exporting of a `ContainsRule` constraint when there already existing a `slicing` on the `ElementDefinition`. To do this, we just need to call the `addSlice` function. This addresses CIMPL-139: https://standardhealthrecord.atlassian.net/browse/CIMPL-139.

### DIFF
--- a/src/errors/InvalidMaxOfSliceError.ts
+++ b/src/errors/InvalidMaxOfSliceError.ts
@@ -1,0 +1,10 @@
+import { Annotated } from './Annotated';
+
+export class InvalidMaxOfSliceError extends Error implements Annotated {
+  specReferences = ['http://www.hl7.org/fhiR/profiling.html#slice-cardinality'];
+  constructor(public sliceMax: string, public slicedElementMax: string) {
+    super(
+      `No individual slice may have max >= max of sliced element, but max of slice ${sliceMax} is > max of sliced element ${slicedElementMax}.`
+    );
+  }
+}
--- a/src/errors/InvalidMaxOfSliceError.ts
+++ b/src/errors/InvalidMaxOfSliceError.ts
@@ -2,9 +2,9 @@ import { Annotated } from './Annotated';
 
 export class InvalidMaxOfSliceError extends Error implements Annotated {
   specReferences = ['http://www.hl7.org/fhiR/profiling.html#slice-cardinality'];
-  constructor(public sliceMax: string, public slicedElementMax: string) {
+  constructor(public sliceMax: string, public sliceName: string, public slicedElementMax: string) {
     super(
-      `No individual slice may have max >= max of sliced element, but max of slice ${sliceMax} is > max of sliced element ${slicedElementMax}.`
+      `No individual slice may have max >= max of sliced element, but max ${sliceMax} of slice ${sliceName} > max of sliced element ${slicedElementMax}.`
     );
   }
 }
--- a/src/errors/InvalidMaxOfSliceError.ts
+++ b/src/errors/InvalidMaxOfSliceError.ts
@@ -1,10 +1,10 @@
 import { Annotated } from './Annotated';
 
 export class InvalidMaxOfSliceError extends Error implements Annotated {
-  specReferences = ['http://www.hl7.org/fhiR/profiling.html#slice-cardinality'];
+  specReferences = ['http://www.hl7.org/fhir/profiling.html#slice-cardinality'];
   constructor(public sliceMax: string, public sliceName: string, public slicedElementMax: string) {
     super(
-      `No individual slice may have max >= max of sliced element, but max ${sliceMax} of slice ${sliceName} > max of sliced element ${slicedElementMax}.`
+      `No individual slice may have max > max of sliced element, but max of slice ${sliceName} (${sliceMax}) > max of sliced element (${slicedElementMax}).`
     );
   }
 }
--- a/src/errors/InvalidSumOfSliceMinsError.ts
+++ b/src/errors/InvalidSumOfSliceMinsError.ts
@@ -1,0 +1,10 @@
+import { Annotated } from './Annotated';
+
+export class InvalidSumOfSliceMinsError extends Error implements Annotated {
+  specReferences = ['http://www.hl7.org/fhiR/profiling.html#slice-cardinality'];
+  constructor(public sumMins: number, public max: string) {
+    super(
+      `The sum of mins of slices must be <= max of sliced element, but sum of mins ${sumMins} is > max ${max}.`
+    );
+  }
+}
--- a/src/errors/InvalidSumOfSliceMinsError.ts
+++ b/src/errors/InvalidSumOfSliceMinsError.ts
@@ -1,10 +1,10 @@
 import { Annotated } from './Annotated';
 
 export class InvalidSumOfSliceMinsError extends Error implements Annotated {
-  specReferences = ['http://www.hl7.org/fhiR/profiling.html#slice-cardinality'];
+  specReferences = ['http://www.hl7.org/fhir/profiling.html#slice-cardinality'];
   constructor(public sumMins: number, public max: string, public slicedElementId: string) {
     super(
-      `The sum of mins of slices must be <= max of sliced element, but sum of mins ${sumMins} > max ${max} of ${slicedElementId}.`
+      `The sum of mins of slices must be <= max of sliced element, but sum of mins (${sumMins}) > max (${max}) of ${slicedElementId}.`
     );
   }
 }
--- a/src/errors/InvalidSumOfSliceMinsError.ts
+++ b/src/errors/InvalidSumOfSliceMinsError.ts
@@ -2,9 +2,9 @@ import { Annotated } from './Annotated';
 
 export class InvalidSumOfSliceMinsError extends Error implements Annotated {
   specReferences = ['http://www.hl7.org/fhiR/profiling.html#slice-cardinality'];
-  constructor(public sumMins: number, public max: string) {
+  constructor(public sumMins: number, public max: string, public slicedElementId: string) {
     super(
-      `The sum of mins of slices must be <= max of sliced element, but sum of mins ${sumMins} is > max ${max}.`
+      `The sum of mins of slices must be <= max of sliced element, but sum of mins ${sumMins} > max ${max} of ${slicedElementId}.`
     );
   }
 }
--- a/src/fhirtypes/ElementDefinition.ts
+++ b/src/fhirtypes/ElementDefinition.ts
@@ -16,7 +16,9 @@ import {
   SlicingDefinitionError,
   SlicingNotDefinedError,
   TypeNotFoundError,
-  WideningCardinalityError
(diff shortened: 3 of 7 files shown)


mint-thompson:
Can a profile have more than one `ContainsRule` on the same path? Either way, it could be a useful test to add, either to correctly apply both rules or to throw an error when the second `ContainsRule` is encountered.

mint-thompson:
Semi-related to previous comment: does the existing cardinality constraint functionality guarantee that the sum of the slices' minimum cardinalities do not exceed the element's maximum cardinality? This might be out of scope for this PR.

cmoesel:
> Can a profile have more than one `ContainsRule` on the same path?

I *think* so.  You wouldn't see this probably in a normal profile, but maybe when we support mix-ins, it might become relevant -- because you might have one contains rule on the path from the mix-in, and then another contains rule on the path from the profile.  So, I *think* it should operate as if they were all a part of the same rule.

cmoesel:
> does the existing cardinality constraint functionality guarantee that the sum of the slices' minimum cardinalities do not exceed the element's maximum cardinality?

That's a great questions/observation.  I think we *should* be comparing the sum of the mins to the overall max to ensure we're not going out of bounds.

A similar question would be if we should *change* the overall min based on slice cardinalities...  In other words, if we have three slices each with min `1`, should we update the overall cardinality to be min `3` (if it is currently < 3)?  I *think* so?  What do you all think?

ngfreiter:
I can add a test that tests out exporting multiple contains rules on the same path.

I think checking the minimums of the slices against the max of the parent element makes sense, and if we are going to do that, I don't see why we shouldn't also update the min cardinality of the parent based on the min cardinalities of the children.

mint-thompson:
> A similar question would be if we should _change_ the overall min based on slice cardinalities... In other words, if we have three slices each with min `1`, should we update the overall cardinality to be min `3` (if it is currently < 3)? I _think_ so? What do you all think?

I agree that the overall min should be updated based on the cardinalities of its slices.

ngfreiter:
Ok in the latest commit I added logic to `ElementDefinition.constrainCardinality` for handling the special rules around slice cardinalities. If an element is sliced, then when its cardinality changes we make sure it's max is not getting set lower than the sum of the minimums of the slices, or lower than the maximum of any individual slices. If an element is a slice, then when its cardinality changes, we make sure that increasing its min is not pushing the overall sum of mins to be greater than the max of the base sliced element.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -269,6 +271,40 @@ export class ElementDefinition {
...
 
+    // Sliced elements and slices have special card rules described here:
+    // http://www.hl7.org/fhiR/profiling.html#slice-cardinality
+    // If element is slice definition
+    if (this.slicing) {
+      const slices = this.structDef.elements.filter(e => e.id !== this.id && e.path === this.path);
+      const sumOfMins = slices.reduce((prev, curr) => (prev += curr.min), 0);
+      // Check that new max >= sum of mins of children
+      if (!isUnbounded && sumOfMins > maxInt) {
+        throw new InvalidSumOfSliceMinsError(sumOfMins, max);
```
mint-thompson: For this error and the other cases where `InvalidSumOfSliceMinsError` and `InvalidMaxOfSliceError` are thrown in this method, I think the message should include the name of the element being sliced. Also, if a specific slice's cardinality caused the error to be thrown, that element's name should also be included in the message.

ngfreiter: Oh yeah good point. I think since the `InvalidMaxOfSliceError` is only when we are constraining cardinality on the element being sliced, we will have information on the sliced element from the `sourceInfo` correct? But I do think that error should list the name of the slice that has too large a maximum. And then since the `InvalidSumOfSliceMinsError` can happen either when we are constraining an element being sliced, or constraining a slice, I agree we should put the name of the element being sliced in the error since we may not have `sourceInfo` on it.

src/errors/InvalidMaxOfSliceError.ts
```diff
@@ -0,0 +1,10 @@
+import { Annotated } from './Annotated';
+
+export class InvalidMaxOfSliceError extends Error implements Annotated {
+  specReferences = ['http://www.hl7.org/fhiR/profiling.html#slice-cardinality'];
+  constructor(public sliceMax: string, public sliceName: string, public slicedElementMax: string) {
+    super(
+      `No individual slice may have max >= max of sliced element, but max ${sliceMax} of slice ${sliceName} > max of sliced element ${slicedElementMax}.`
```
cmoesel: I think you mean `>`, _not_ `>=`.  It's perfectly fine for the `max`es to be `=`.

Also, the wording is a little inconsistent at the end of the sentence.  It should either be:
```
but max ${sliceMax} of slice ${sliceName} > max ${slicedElementMax} of sliced element.
```
or
```
but max of slice ${sliceName} (${sliceMax}) > max of sliced element (${slicedElementMax}).
```

I _think_ I prefer the latter, but it's not a strong opinion.

src/errors/InvalidMaxOfSliceError.ts
```diff
@@ -0,0 +1,10 @@
+import { Annotated } from './Annotated';
+
+export class InvalidMaxOfSliceError extends Error implements Annotated {
+  specReferences = ['http://www.hl7.org/fhiR/profiling.html#slice-cardinality'];
```
cmoesel: `fhiR`?  Interesting...

src/errors/InvalidSumOfSliceMinsError.ts
```diff
@@ -0,0 +1,10 @@
+import { Annotated } from './Annotated';
+
+export class InvalidSumOfSliceMinsError extends Error implements Annotated {
+  specReferences = ['http://www.hl7.org/fhiR/profiling.html#slice-cardinality'];
```
cmoesel: More `fhiR`...

src/errors/InvalidSumOfSliceMinsError.ts
```diff
@@ -0,0 +1,10 @@
...
+  constructor(public sumMins: number, public max: string, public slicedElementId: string) {
+    super(
+      `The sum of mins of slices must be <= max of sliced element, but sum of mins ${sumMins} > max ${max} of ${slicedElementId}.`
```
cmoesel: I feel like maybe the counts should be in parentheses?

src/fhirtypes/ElementDefinition.ts
```diff
@@ -269,6 +271,40 @@ export class ElementDefinition {
       throw new WideningCardinalityError(this.min, this.max, min, max);
     }
 
+    // Sliced elements and slices have special card rules described here:
+    // http://www.hl7.org/fhiR/profiling.html#slice-cardinality
+    // If element is slice definition
+    if (this.slicing) {
+      const slices = this.structDef.elements.filter(e => e.id !== this.id && e.path === this.path);
+      const sumOfMins = slices.reduce((prev, curr) => (prev += curr.min), 0);
```
cmoesel: Ah.  Ain't nothin' quite like an opportunity to use the `reduce` function...

src/fhirtypes/ElementDefinition.ts
```diff
@@ -269,6 +271,40 @@ export class ElementDefinition {
...
+      }
+      // Check that new max >= every individual child max
+      const overMaxChild = slices.find(child => child.max === '*' || parseInt(child.max) > maxInt);
+      if (!isUnbounded && overMaxChild) {
+        throw new InvalidMaxOfSliceError(overMaxChild.max, overMaxChild.sliceName, max);
+      }
+    }
+
+    // If element is a slice
+    const slicedElement = this.structDef.elements.find(e => e.path === this.path && e.slicing);
```
cmoesel: I'd suggest surrounding this whole set of logic with `if (this.sliceName) { ... }`.  It will avoid always iterating through every element when `this` isn't a slice anyway.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -269,6 +271,40 @@ export class ElementDefinition {
...
+    if (slicedElement) {
+      const slices = this.structDef.elements.filter(
+        e => e.id !== this.id && e.path === this.path && !e.slicing
+      );
+      const sumOfMins = min + slices.reduce((prev, curr) => (prev += curr.min), 0);
+      // Check that slicedElement max >= new sum of mins
+      if (slicedElement.max !== '*' && sumOfMins > parseInt(slicedElement.max)) {
+        throw new InvalidSumOfSliceMinsError(sumOfMins, slicedElement.max, slicedElement.id);
+      }
+      // If new sum of mins > slicedElement min, increase slicedElement min
```
cmoesel: I think you mean `<`?

src/fhirtypes/ElementDefinition.ts
```diff
@@ -269,6 +271,40 @@ export class ElementDefinition {
...
+    const slicedElement = this.structDef.elements.find(e => e.path === this.path && e.slicing);
+    if (slicedElement) {
+      const slices = this.structDef.elements.filter(
+        e => e.id !== this.id && e.path === this.path && !e.slicing
+      );
+      const sumOfMins = min + slices.reduce((prev, curr) => (prev += curr.min), 0);
+      // Check that slicedElement max >= new sum of mins
+      if (slicedElement.max !== '*' && sumOfMins > parseInt(slicedElement.max)) {
+        throw new InvalidSumOfSliceMinsError(sumOfMins, slicedElement.max, slicedElement.id);
+      }
```
cmoesel: I think the above 8 lines of code is extremely similar to lines 278 - 283 above.  Can these be consolidated into a single function?

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1266,11 +1302,9 @@ export class ElementDefinition {
     delete slice.slicing;
     slice.id = `${this.id}:${name}`;
     slice.sliceName = name;
-    // When a choice is sliced, we do not inherit min cardinality, but rather make it 0
+    // When a we slice, we do not inherit min cardinality, but rather make it 0
```
cmoesel: There's a straggling `a` that got left behind in the edit.  I also think it might be good to also briefly note that this allows multiple slices to be defined without violating the sliced element's min and/or that we expect the card will be explicitly constrained after the slice is created.

src/fhirtypes/ElementDefinition.ts
ngfreiter: I'm not sure. This is when the element is a slice, and it's min is getting increased. If the new sum of all the mins is greater than the old min of the sliced element, then we increase the min of the sliced element accordingly. So I _think_ this is the right phrasing.

cmoesel: Oh.  Yeah.  I think my brain got confused for a minute.  You're totally right.  Sorry.

ngfreiter:
Ok the latest commit should address all of your comments @cmoesel. I also rebased since some stuff had been merged into master.

PR #39, opened 2019-12-02
Record source info in ContainsRule

ngfreiter:
I noticed that the importing of `ContainsRule` was not tracking the source information as all the other rules were. This was my mistake from when I added the import of `ContainsRule`, so this fixes that by adding tracking of source info to the importing of `ContainsRule`.

### DIFF
--- a/src/import/FSHImporter.ts
+++ b/src/import/FSHImporter.ts
@@ -412,7 +412,7 @@ export class FSHImporter extends FSHVisitor {
       containsRule.items.push(item);
 
       const cardRule = new CardRule(`${containsRule.path}[${item}]`)
-        .withLocation(this.extractStartStop(ctx))
+        .withLocation(this.extractStartStop(i))
         .withFile(this.file);
       const card = this.parseCard(i.CARD().getText());
       cardRule.min = card.min;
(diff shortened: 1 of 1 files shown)


src/import/FSHImporter.ts
```diff
@@ -402,21 +402,27 @@ export class FSHImporter extends FSHVisitor {
...
+      .withFile(this.file);
 
     rules.push(containsRule);
     ctx.item().forEach(i => {
       const item = this.aliasAwareValue(i.SEQUENCE().getText());
       containsRule.items.push(item);
 
-      const cardRule = new CardRule(`${containsRule.path}[${item}]`);
+      const cardRule = new CardRule(`${containsRule.path}[${item}]`)
+        .withLocation(this.extractStartStop(ctx))
```
cmoesel: I think you should be extracting the start/stop from `i` (the specific `ctx.item()`) or `i.CARD()` (the actual cardinality in the item).  Extracting it from the overall `ContainsRuleContext` is too broad.

src/import/FSHImporter.ts
```diff
@@ -402,21 +402,27 @@ export class FSHImporter extends FSHVisitor {
...
+        .withFile(this.file);
       const card = this.parseCard(i.CARD().getText());
       cardRule.min = card.min;
       cardRule.max = card.max;
       rules.push(cardRule);
 
       if (i.flag() && i.flag().length > 0) {
-        const flagRule = new FlagRule(`${containsRule.path}[${item}]`);
+        const flagRule = new FlagRule(`${containsRule.path}[${item}]`)
+          .withLocation(this.extractStartStop(ctx))
```
cmoesel: I think you should be extracting the start/stop from `i` (the specific `ctx.item()`) or `i.flag()` (the actual cardinality in the item).  Extracting it from the overall `ContainsRuleContext` is too broad.

src/import/FSHImporter.ts
ngfreiter: Good point, in the last commit I changed it to be specific to the specific `ctx.item()`.

PR #41, opened 2019-12-04
ElementDefinition setInstancePropertyByPath

ngfreiter:
This adds a `setInstancePropertyByPath` function to the `ElementDefinition` class. To write this function I was just copying code from the `setInstancePropertyByPath` function on the `StructureDefinition` class, so I refactored that common code from both of these functions into the `common.ts` file. So functionally this is almost exactly the same as how setting an instance property works on the `StructureDefinition`.


src/fhirtypes/ElementDefinition.ts
```diff
@@ -143,6 +145,17 @@ export class ElementDefinition {
     return this.path.slice(this.structDef.type.length + 1);
   }
 
+  /**
+   * Get the StructureDefinition for ElementDefinition
+   * @param {ResolveFn} resolve - A function that can resolve a type to a StructureDefinition instance
+   */
+  getOwnStructureDefinition(resolve: ResolveFn = () => undefined) {
```
cmoesel: Can you add a return type annotation so it's clear to users (instead of inferred)?  This also ensures if you ever change the function, you don't unintentionally change the return type.

src/fhirtypes/StructureDefinition.ts
```diff
@@ -89,7 +90,7 @@ export class StructureDefinition {
    * Get the Structure Definition for Structure Definition
    * @param {ResolveFn} resolve - A function that can resolve a type to a StructureDefinition instance
    */
-  private getSdStructureDefinition(resolve: ResolveFn = () => undefined) {
+  getOwnStructureDefinition(resolve: ResolveFn = () => undefined) {
```
cmoesel: Can you add a return type annotation so it's clear to users (instead of inferred)?  This also ensures if you ever change the function, you don't unintentionally change the return type.

src/fhirtypes/common.ts
```diff
@@ -0,0 +1,75 @@
...
+function getArrayIndex(pathPart: PathPart): number {
+  const lastBracket = pathPart.brackets?.slice(-1)[0];
+  let arrayIndex: number;
+  if (/^[-+]?\d+$/.test(lastBracket)) {
+    arrayIndex = parseInt(lastBracket);
+  }
+  return arrayIndex >= 0 ? arrayIndex : null;
+}
+
+export { getArrayIndex, setPropertyOnInstance };
```
cmoesel: I think in most other files, we've been using the `export` keyword as a prefix before the function or class declaration (rather than exporting at the end of the file).  We should probably stick with that convention for consistency sake.

src/fhirtypes/common.ts
```diff
@@ -0,0 +1,75 @@
...
+        // If the array doesn't exist, create it
+        if (current[key] == null) current[key] = [];
+        // If the index doesn't exist in the array, add it and lesser indices
+        let j = current[key].length;
+        for (; j < index; j++) {
+          current[key].push(undefined);
+        }
+        if (j === index) {
+          current[key].push({});
+        }
```
mint-thompson: I don't think it is necessary to use this loop here. If you assign a value to an array index beyond its current length, the elements between the current end and the new element will automatically be set to `undefined`. It should be fine to do something like: `if(index >= current[key].length) { current[key][index] = {}; }`

ngfreiter: Oh cool! I did not know that. That's so much neater.

cmoesel: Woah.  I did not know that either!  Thanks for the tip, @mint-thompson!

PR #42, opened 2019-12-04
Import CaretValueRule

ngfreiter:
This adds the `CaretValueRule` class, and the importing of the `CaretValueRule` class. This rule is used to read in rules like `* ^description = "foo"`, or `* status ^short = "foo"`. The path before the `^` is optional. For description of this syntax, see https://github.com/HL7/fhir-shorthand/wiki/2.2-Paths#addressing-structure-definition-attributes-directly.

Note that in `FSHImporter.Profile.test.ts`, the only relevant changes are the `CaretValueRule` tests. The other changes come from moving some other tests inside a block of code that I believe they should have been inside.

### DIFF
--- a/src/import/FSHImporter.ts
+++ b/src/import/FSHImporter.ts
@@ -18,7 +18,8 @@ import {
   FixedValueRule,
   FixedValueType,
   OnlyRule,
-  ContainsRule
+  ContainsRule,
+  CaretValueRule
 } from '../fshtypes/rules';
 import { ParserRuleContext } from 'antlr4';
 import { logger } from '../utils/FSHLogger';
@@ -257,6 +258,8 @@ export class FSHImporter extends FSHVisitor {
       return [this.visitOnlyRule(ctx.onlyRule())];
     } else if (ctx.containsRule()) {
       return this.visitContainsRule(ctx.containsRule());
+    } else if (ctx.caretValueRule()) {
+      return [this.visitCaretValueRule(ctx.caretValueRule())];
     }
     logger.warn(`Unsupported rule: ${ctx.getText()}`, {
       file: this.file,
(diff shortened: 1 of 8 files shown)


ngfreiter:
> Are caret value rules applicable to extensions? If so, please add at least one test involving caret rules to FSHImporter.Extension.test.ts.

Yes they are! Good catch.

test/import/FSHImporter.Extension.test.ts
```diff
@@ -161,5 +162,20 @@ describe('FSHImporter', () => {
         assertOnlyRule(extension.rules[0], 'value[x]', { type: 'Quantity' });
       });
     });
+
+    // Since Extensions use the same rule parsing code as Profiles, only do a minimal tests
```
mint-thompson: Should be "minimal test" (singular), since there's only one `CaretValueRule` test. Actually, I see that this line appears repeatedly in `FSHImporter.Extension.test.ts`, and it would probably be fine to just have it once before all the tests that overlap with the Profile tests.

src/import/FSHImporter.ts
```diff
@@ -514,6 +521,18 @@ export class FSHImporter extends FSHVisitor {
     return rules;
   }
 
+  visitCaretValueRule(ctx: pc.CaretValueRuleContext): CaretValueRule {
+    const path = ctx.path() ? this.visitPath(ctx.path()) : null;
```
cmoesel: Do you think it's weird to have a `null` path?  I'm wondering if it would be better to use empty string `''` to signify the path at the root?

test/import/FSHImporter.Extension.test.ts
ngfreiter: Yeah fair point it's repetitive. I made it one comment at the top that is more grammatically sound.

src/import/FSHImporter.ts
ngfreiter: Keeping the path a string if we can seems like a good reason to do `''`, and I can't think of any good reason to make it `null`, so yeah empty string works for me.

PR #43, opened 2019-12-05
Unfolded elements should not cause diffs (plus other refactorings)

cmoesel:
This PR contains three separate (but somewhat related) changes:

1. It modifies the behavior of the unfold function such that it captures the original elements right after unfolding.  This makes it so that a diff is only produced if something changes from the base definition.

2. It moves the invocation of captureElements into the SD and ED `fromJSON` functions, since it will almost always be wanted when deserializing from a JSON document.  An argument override can supress this behavior.

3. It removes a clone of the base SD that was not used in production code but served only to help in unit tests.

### DIFF
--- a/src/export/StructureDefinitionExporter.ts
+++ b/src/export/StructureDefinitionExporter.ts
@@ -137,8 +137,6 @@ export class StructureDefinitionExporter {
       throw new ParentNotDefinedError(fshDefinition.name, parentName, fshDefinition.sourceInfo);
     }
 
-    // Capture the orginal elements so that any further changes are reflected in the differential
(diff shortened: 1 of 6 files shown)


ngfreiter:
Ah! I had this open on my computer and then left for a while and when I came back the page had not refreshed and I approved before seeing your comment @mint-thompson. Very very sorry for merging that before @cmoesel had a chance to respond to that comment.

cmoesel:
Ha.  Looks like @ngfreiter just earned himself the privilege (and duty) of writing that unit test for me.  Thanks Nick!

PR #44, opened 2019-12-05
Add export of caretValueRule

ngfreiter:
This adds exporting of `caretValueRule`. This exporting can operate on an individual `ElementDefinition`, or a `StructureDefinition` itself. Also adds tests of this exporting on the `StructureDefinitionExporter`.

### DIFF
--- a/src/export/StructureDefinitionExporter.ts
+++ b/src/export/StructureDefinitionExporter.ts
@@ -9,7 +9,8 @@ import {
   FlagRule,
   OnlyRule,
   ValueSetRule,
-  ContainsRule
+  ContainsRule,
+  CaretValueRule
 } from '../fshtypes/rules';
 import { logger } from '../utils/FSHLogger';
 import cloneDeep from 'lodash/cloneDeep';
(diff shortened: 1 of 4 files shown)


src/export/StructureDefinitionExporter.ts
```diff
@@ -80,6 +81,20 @@ export class StructureDefinitionExporter {
...
           } else if (rule instanceof ContainsRule) {
             rule.items.forEach(item => element.addSlice(item));
+          } else if (rule instanceof CaretValueRule) {
+            if (rule.path !== '') {
+              element.setInstancePropertyByPath(
+                rule.caretPath,
+                rule.value,
+                this.resolve.bind(this)
+              );
+            } else {
```
cmoesel: Do we provide a way to access the root element?  I expect that when `rule.path === ''`, then the `element` received in line 68 is probably the root element, but we toss that value and interpret it as setting something on the structure definition.  That's how we've been thinking about it all along (so you've implemented it right), but I'm thinking we need a way to set something on the root as well.  Have we discussed this before?  Any ideas?  Maybe if path is `$root` (or `$rootElement`)?  Actually the [FHIR FHIRPath implementation](http://hl7.org/fhir/R4/fhirpath.html) uses `%` for special variables, so if we wanted to conform to that... then `%rootElement`?

ngfreiter:
The latest commit updates the tests to work with some changes made in this PR https://github.com/standardhealth/sushi/pull/43. It also adds tests of the `fromJSON` methods when `ccaptureOriginalElements` is set to false. This should have been added on this PR https://github.com/standardhealth/sushi/pull/43 but someone didn't see a comment and merged that PR. That someone is me, and I would once more like to apologize for my hastiness.

src/export/StructureDefinitionExporter.ts
cmoesel: I spoke w/ @ngfreiter about this yesterday and we agreed that we *should* support it somehow, but that designing and implementing that support should be separate from this PR.

cmoesel:
Oh @ngfreiter ... and we had such high hopes for you!  Now... all down the drain.  Thanks for trying to redeem yourself though.

PR #47, opened 2019-12-06
Add ValueSet class

mint-thompson:
The FSH ValueSet class is used to represent value sets. Value sets
may contain specific codes. Value sets may also specify that they
include or exclude codes that are descendents of a specified code.

This pull request addresses https://standardhealthrecord.atlassian.net/browse/CIMPL-149.


src/fshtypes/ValueSet.ts
```diff
@@ -0,0 +1,19 @@
+import { FshEntity } from './FshEntity';
+import { FshCode } from './FshCode';
+import { ValueSetTerm } from './ValueSetTerm';
+
+export class ValueSet extends FshEntity {
+  id: string;
+  description?: string;
+  codes: ValueSetTerm[];
+  includesDescendants: FshCode[];
+  excludesDescendants: FshCode[];
```
ngfreiter: I'm not sure if these are the only two ways that we can apply some sort of filter to a ValueSet. For example, I think something like this `* include concept is-a RACE#2133-7` could be done, and I'm not sure how that would be handled. There are a set of operators http://hl7.org/fhir/valueset-filter-operator.html, and I'm not sure if we have decided what is and isn't allowed. If you did already discuss this and decided this is what we need for now, sorry!

mint-thompson: You are correct that there are additional operators that could be applied. I was basing this implementation off https://github.com/HL7/fhir-shorthand/wiki/3.3-Value-Sets, which only lists `include descendant-of` and `exclude descendant-of` and mentions that it borrows a subset of the available operators.

cmoesel: I haven't had a chance to review yet myself, but I saw these comments come by.  The intent is to support _all_ of the filter operators.  That being the case, I don't know if it's better to explicitly put each one in the grammar or if it might be better just to have a generic slot for the filter in the grammar that is then validated against the supported list in the importer.  If/when we eventually add support for DSTU2, it has a different set (well, a subset) of filter operators.

cmoesel: It's fine to have them explicitly in the grammar, but a few things to note about that:
* although you find the error sooner, the error message may be harder for the user to understand (since it will be an ANTLR message complaining about not being able to find a matching parserule)
* if I recall correctly, ANTLR tries to do as exact a match as possible on lexer rules when creating the token stream; so any time you define something as a specific lexer token, that means that this is the token that will be used _wherever_ if finds that word.  For example, since `exists` is one of the filter ops, if you create a lexer rule `KW_EXISTS` for it, then if there is a FHIR property called exists and someone uses it in a definition:
  ```
  * exists = true
  ```
  then the lexer will make that a `KW_EXISTS` token even though the parse rule for that part of the grammar might be expecting a `SEQUENCE` token.  For this set of keywords, it's probably not a problem, but it's always something to think about when deciding whether to make a new lexer token or not.

As for where the filter ops can be used, yes -- they can be in the `include` or `exclude` rules.

cmoesel: > ANTLR tries to do as exact a match as possible on lexer rules
I can't remember if that's right or if it's that ANTLR uses the *first* lexer rule that matches (going top-down).  Either way, the result would actually be the same because in order to get `KW_EXISTS` at all, it would have to be higher than `SEQUENCE`.

mint-thompson:
I've changed the strategy for the `ValueSet` class a bit. It now has lists of what to include and exclude. This is similar to the specification found at http://hl7.org/fhir/valueset-definitions.html#ValueSet.compose.

In the inclusion/exclusion lists, I was planning to represent remote value sets by setting their id to their url. Does that sound acceptable, or should I have a different way of representing an included/excluded remote value set?

src/fshtypes/ValueSet.ts
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * For more information about the composition of a ValueSet,
+ * @see {@link http://hl7.org/fhir/valueset-definitions.html#ValueSet.compose}
+ */
+export class ValueSet extends FshEntity {
+  id: string;
+  url?: string;
+  description?: string;
+  inclusions: (ValueSetTerm | ValueSetFilter | ValueSet)[];
+  exclusions: (ValueSetTerm | ValueSetFilter | ValueSet)[];
```
cmoesel: I wonder if we should model this more closely to how FHIR defines it?  Would it be simpler that way?
```ts
// ...
inclusions: ValueSetRule[]
exclusions: ValueSetRule[]
// ...

class ValueSetRule extends FshEntiry {
  public system?: string;
  public rule?: ValueSetConcept | ValueSetFilter | ValueSetReference;
  // with...
}

type ValueSetReference: string;

Comment (cmoesel) [reply src/fshtypes/ValueSet.ts]: Then you would not need to put the `system` into the other classes.

There is a question of whether `rule` should be an array or not -- because technically, in the FHIR spec, you can put all the rules that share a code system into the same `include` item.  But probably, to be consistent w/ the FSH grammar, it makes more sense as a rule per line.

File: src/fshtypes/ValueSetFilter.ts [hunk 3]
```diff
@@ -0,0 +1,53 @@
...
+  withSystem(system: string) {
+    this.system = system;
+    return this;
+  }
+
+  withProperty(property: string) {
+    this.property = property;
+    return this;
+  }
+}
```
cmoesel: Here too, staying closer to the FHIR spec might simplify things.  Why not:
```ts
class ValueSetFilter extends FshEntity {
  public property: string;
  public op: '=' | 'is-a' | 'descendent-of' | 'is-not-a' | 'regex' // | etc.
  public value: string;
}
```

cmoesel: Then we don't need all those subclasses -- and when we process it, it's probably easier to setup a `switch` statement on `op` than having to check `instanceof` against each subclass.

cmoesel: Also... as it is currently written, where would the `value` go?

src/fshtypes/ValueSetTerm.ts
```diff
@@ -0,0 +1,7 @@
+import { FshCode } from './FshCode';
+
+export class ValueSetTerm extends FshCode {
+  constructor(code: string, system: string, display?: string, public definition?: string) {
```
cmoesel: If we pop the `system` up a level, we don't need it here.  Also... I don't think there is anywhere that `definition` comes into play?  It might if we allow defining _new_ codes in a `ValueSet`, but I think we're actually leaning toward not allowing that (you can only use _existing_ codes from an _existing_ value set).

mint-thompson:
I've made another attempt at redesigning the ValueSet class. This still has the limitation of only having one rule per component (the FHIR specification allows more complex definitions), but as @cmoesel mentioned, this limitation will be a better match for the FSH grammar.

PR #48, opened 2019-12-06
Update location format in log messages

cmoesel:
Update location format in log messages to:
```
error: My Message
  File: /absolute/path/to/MyFile.fsh
  Line: 12 - 14
```
Based on team discussion.

### DIFF
--- a/test/export/StructureDefinitionExporter.test.ts
+++ b/test/export/StructureDefinitionExporter.test.ts
@@ -170,7 +170,7 @@ describe('StructureDefinitionExporter', () => {
     expect(structDef).toBeDefined();
     expect(structDef.type).toBe('Resource');
     expect(mockWriter.mock.calls[mockWriter.mock.calls.length - 1][0].message).toMatch(
-      /File: Foo\.fsh.*Line: 3 - 4/s
+      /File: Foo\.fsh.*Line: 3 - 4\D/s
     );
   });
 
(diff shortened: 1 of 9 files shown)


test/export/StructureDefinitionExporter.test.ts
```diff
@@ -170,7 +170,7 @@ describe('StructureDefinitionExporter', () => {
     expect(structDef).toBeDefined();
     expect(structDef.type).toBe('Resource');
     expect(mockWriter.mock.calls[mockWriter.mock.calls.length - 1][0].message).toMatch(
-      /File: Foo\.fsh.*Line 3\D.*Column 8\D.*Line 4\D.*Column 22\D/s
+      /File: Foo\.fsh.*Line: 3 - 4/s
     );
```
mint-thompson: In this match expression and a few others, the `\D` at the end of the expression is removed. But some of the others still have the `\D` at the end. Did the presence/absence of the `\D` cause test case failures in any of the cases? In the old format, the `\D` at the end would match the closing parentheses at the end of the message, but now, the end of the message should be the start line (for single-line sources) or the end line (for multi-line sources).

cmoesel: Oh - that was unintentional.  I must have selected too much text when deleting.  I can fix that.

PR #50, opened 2019-12-09
Slice Extension in ContainsRule

ngfreiter:
Now when an extension element is sliced by a `ContainsRule`, if no slicing exists on the element, a slicing is added to the `ElementDefinition`. After the extension slice is added, if the extension is already defined and can be resolved, the canonical url of the extension is added to the `type.profile` on the extension `ElementDefinition`.

This addresses https://standardhealthrecord.atlassian.net/browse/CIMPL-168.

### DIFF
--- a/src/export/StructureDefinitionExporter.ts
+++ b/src/export/StructureDefinitionExporter.ts
@@ -80,7 +80,20 @@ export class StructureDefinitionExporter {
           } else if (rule instanceof ValueSetRule) {
             element.bindToVS(rule.valueSet, rule.strength as ElementDefinitionBindingStrength);
           } else if (rule instanceof ContainsRule) {
-            rule.items.forEach(item => element.addSlice(item));
+            const isExtension = element.type?.length === 1 && element.type[0].code === 'Extension';
+            if (isExtension && !element.slicing) {
+              element.sliceIt('value', 'url');
+            }
+            rule.items.forEach(item => {
+              const slice = element.addSlice(item);
+              if (isExtension) {
+                const extension = this.resolve(item);
+                if (extension) {
+                  if (!slice.type[0].profile) slice.type[0].profile = [];
+                  slice.type[0].profile.push(extension.url);
+                }
+              }
+            });
           } else if (rule instanceof CaretValueRule) {
             if (rule.path !== '') {
               element.setInstancePropertyByPath(
--- a/src/export/StructureDefinitionExporter.ts
+++ b/src/export/StructureDefinitionExporter.ts
@@ -2,7 +2,7 @@ import { FHIRDefinitions } from '../fhirdefs';
 import { StructureDefinition, ElementDefinitionBindingStrength } from '../fhirtypes';
 import { Profile, Extension } from '../fshtypes';
 import { FSHTank } from '../import';
-import { ParentNotDefinedError } from '../errors/ParentNotDefinedError';
+import { ParentNotDefinedError, InvalidExtensionSliceError } from '../errors';
 import {
   CardRule,
   FixedValueRule,
@@ -89,7 +89,9 @@ export class StructureDefinitionExporter {
               if (isExtension) {
                 const extension = this.resolve(item);
                 if (extension) {
-                  if (!slice.type[0].profile) slice.type[0].profile = [];
+                  if (!slice.type[0].profile) {
+                    slice.type[0].profile = [];
+                  }
                   slice.type[0].profile.push(extension.url);
                 }
               }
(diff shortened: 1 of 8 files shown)


ngfreiter:
Good point, I will add that.

Also I believe some changes need to be made with how we handle urls being used to reference extensions, and how lookup of extensions works in general. It seems that for extensions defined by FHIR, we lookup by `id` or `url`, but not `name`, but for extensions defined within FSH, we look things up by `name`, but not `id` or `url`. Since this inconsistency makes it difficult to complete this task, I will work on resolving that inconsistency as a part of this task.

ngfreiter:
The latest commit adds handling of Aliases to exporting. The functions for finding profiles and extensions will now try to find the relevant profile or extension by name or id. If a profile or extension is still not found, it will try to find a name or id by first resolving an alias.

src/export/StructureDefinitionExporter.ts
```diff
@@ -122,11 +135,10 @@ export class StructureDefinitionExporter {
       let structDef = cloneDeep(this.structDefs.find(sd => sd.name === type));
```
cmoesel: This is only looking at `name`.  It seems that it should also look at `id` and `url` to be consistent w/ other resolve behavior?

src/export/StructureDefinitionExporter.ts
```diff
@@ -80,7 +80,20 @@ export class StructureDefinitionExporter {
...
-            rule.items.forEach(item => element.addSlice(item));
+            const isExtension = element.type?.length === 1 && element.type[0].code === 'Extension';
+            if (isExtension && !element.slicing) {
+              element.sliceIt('value', 'url');
+            }
+            rule.items.forEach(item => {
+              const slice = element.addSlice(item);
+              if (isExtension) {
+                const extension = this.resolve(item);
+                if (extension) {
```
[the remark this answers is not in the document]
cmoesel: Nevermind.  Now I see this can happen if you're defining a new extension inline.  See my comments in the tests about some caveats though.

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -609,6 +609,120 @@ describe('StructureDefinitionExporter', () => {
...
+    });
+  });
+
+  it('should apply a ContainsRule of an undefined extension on an extension element', () => {
+    const profile = new Profile('Foo');
+    profile.parent = 'Observation';
+
+    const rule = new ContainsRule('extension');
+    rule.items = ['foo'];
+    profile.rules.push(rule);
```
cmoesel: Technically, this is an invalid SD since the `slicing` says to slice on `url`, but the slice does not set `extension[foo] ^type[0].profile` or fix `extension[foo].url`.  It seems like maybe we should actually detect this situation and report an error?  Otherwise it seems easy for the author to fat-finger what is meant to be an existing extension name/id, but not realize it since we happily accept it and treat it like a new (undefined) extension.

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -609,6 +609,120 @@ describe('StructureDefinitionExporter', () => {
...
+    expect(bar.type[0]).toEqual({
+      code: 'Extension',
+      profile: ['http://example.com/StructureDefinition/Bar']
+    });
+    expect(baz).toBeDefined();
+    expect(baz.type[0]).toEqual({
+      code: 'Extension',
+      profile: ['http://example.com/StructureDefinition/BazId']
+    });
+  });
```
cmoesel: Can you also add a test where the alias points to a URL for an existing FHIR extension?  That's the way I would expect alias to usually be used in this context.

test/export/StructureDefinitionExporter.test.ts
ngfreiter: Hmm. This is a good point, but I am not sure how to go about it. Since we don't know when that rule that sets the url will be. It'll be sometime after the contains rule, but not necessarily immediately after.

test/export/StructureDefinitionExporter.test.ts
ngfreiter: Ah good point! I will add that.

src/import/FSHTank.ts
```diff
@@ -22,20 +22,43 @@ export class FSHTank {
...
+   * @param {string} key - The name or id of the profile we're looking for
    * @returns {Profile | undefined}
    */
-  public findProfileByName(name: string): Profile | undefined {
-    return this.getAllProfiles().find(profile => profile.name === name);
+  public findProfile(key: string): Profile | undefined {
+    let profile = this.getAllProfiles().find(p => p.name === key || p.id === key);
+    if (!profile) {
+      const aliased = this.findAlias(key);
+      if (aliased) profile = this.getAllProfiles().find(p => p.name === key || p.id === key);
```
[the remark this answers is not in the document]
ngfreiter: That's also usually my personal preference. What came over me...

ngfreiter:
The latest commit should address the comments. I added a function that post processes the sd and makes sure that extension slices are fixing a url. It's very special casey, but this is a special case.

src/import/FSHTank.ts
```diff
@@ -22,20 +22,33 @@ export class FSHTank {
...
+  public findExtension(key: string): Extension | undefined {
+    return this.getAllExtensions().find(p => p.name === key || p.id === key);
+  }
+
+  /**
+   * Finds the alias in the tank, if it exists
+   * @param {string} name - The name of the alias we're looking for
+   * @returns {string | undefined}
+   */
+  public resolveAlias(name: string): string | undefined {
```
cmoesel: The logic of this function returns back the alias if it is not found, but the doc and the return type seems to indicate otherwise (implying it will return undefined if it is not found).  I feel like it may be important to know whether or not the alias actually resolved, which is difficult as implemented -- so my preference would be that it returns `undefined` when the alias cannot be resolved.  This would, of course, mean updating any code that relies on the current behavior.

src/import/FSHTank.ts
```diff
@@ -22,20 +22,33 @@ export class FSHTank {
...
-   * Finds the profile in the tank by name, if it exists
-   * @param {string} name - The name of the profile we're looking for
+   * Finds the profile in the tank by name, id, or alias, if it exists
+   * @param {string} key - The name or id of the profile we're looking for
    * @returns {Profile | undefined}
    */
-  public findProfileByName(name: string): Profile | undefined {
-    return this.getAllProfiles().find(profile => profile.name === name);
+  public findProfile(key: string): Profile | undefined {
+    return this.getAllProfiles().find(p => p.name === key || p.id === key);
```
cmoesel: For consistency, should this also check for a match on URL?  I know URL isn't a field on the object, but since you have the config, you could easily construct what the URL _will_ be and check against that.

src/import/FSHTank.ts
```diff
@@ -22,20 +22,33 @@ export class FSHTank {
...
-   * @param {string} name - The name of the extension we're looking for
-   * @returns {Extension | undefined}
+   * Finds the extension in the tank by name, id, or alias, if it exists
+   * @param {string} key - The name or id of the extension we're looking for
+   * @returns {[Extension, string]}
    */
-  public findExtensionByName(name: string): Extension | undefined {
-    return this.getAllExtensions().find(extension => extension.name === name);
+  public findExtension(key: string): Extension | undefined {
+    return this.getAllExtensions().find(p => p.name === key || p.id === key);
```
cmoesel: Same as above re: `URL`.

ngfreiter:
Ok @cmoesel I fixed those latest comments and rebased onto the new master.

cmoesel:
@mint-thompson -- if you're happy w/ Nick's resolution of your comments, can you add your approval?

mint-thompson:
Ah, sorry for the delay! Thank you for adding those test cases.

PR #51, opened 2019-12-11
Log number of errors and warnings

kjmahalingam:
Adds a global `stats` object to the `FSHLogger` which (right now) just records the number of info, warn, and error messages. Also reports the number of errors and warnings at the end of a CLI run.

To test, either just run against something invalid so it reports errors and warnings, or add errors and warning scattered around the code to test the stats.


src/app.ts
```diff
@@ -80,3 +80,5 @@ fs.writeFileSync(
 logger.info(
   `Exported ${outPackage.profiles.length} profile(s) and ${outPackage.extensions.length} extension(s).`
 );
+
+logger.info(`There was a total of ${stats.numError} error(s) and ${stats.numWarn} warning(s).`);
```
cmoesel: Can we reformat how this gets displayed?  I'm thinking something a little cleaner and easier to visually parse might be nice.  For example, we could replace the two logging statements above so that they produce something more like this:

```
Profiles:    12
Extensions:  5
Errors:      0
Warnings:    0
```

Or something fancier:
```
+============================+
|          RESULTS           |
|----------------------------|
| Profiles:   12             |
| Extensions:  5             |
+============================+
    Errors: 0  Warnings: 0
```

src/utils/FSHLogger.ts
```diff
@@ -17,7 +17,22 @@ const withLocation = format(info => {
   return info;
 });
 
+const incrementCounts = format(info => {
+  if (info.level === 'info') stats.numInfo++;
+  if (info.level === 'warn') stats.numWarn++;
+  if (info.level === 'error') stats.numError++;
```
cmoesel: I think this would be bit cleaner as a `switch` statement on `info.level` -- but if you prefer `if`, then I'd use `else if` to be more efficient.

kjmahalingam:
I super strangely never got notified that you reviewed this. I'll fix these now.

src/utils/FSHLogger.ts
kjmahalingam: Sure, I can change it to a `switch`.

src/app.ts
kjmahalingam: The fancier version seems a bit much, to me (I can change it if you want though), but I will update to report results more cleanly!

kjmahalingam:
All comments here should be addressed!

test/utils/FSHLogger.test.ts
```diff
@@ -0,0 +1,50 @@
+import { logger, stats } from '../../src/utils/FSHLogger';
+
+describe('FSHLogger', () => {
+  beforeEach(() => {
+    stats.reset();
+    expect(stats.numInfo).toBe(0);
+    expect(stats.numWarn).toBe(0);
+    expect(stats.numError).toBe(0);
```
cmoesel: Super minor, but I'd avoid putting assertions into the `beforeEach`.  Instead, have `beforeEach` do what it needs to do (`stats.reset()`) and create a new test that validates that `reset()` works correctly.

kjmahalingam:
I'm a little confused about what just happened with the commits here after I rebased...it should be up to date now and there should be no issue if we just squash and merge, but I don't understand why it seems to have duplicated some commits.

PR #53, opened 2019-12-12
Add LoggerSpy class and use in tests

kjmahalingam:
Adds a wrapper class to the Jest Spy functionality, to clean up the tests a bit. This doesn't change any functionality at all. If you run the tests and they pass, we're good to go. Just let me know if there's any feedback you have on the implementation itself.

### DIFF
--- a/test/export/ExtensionExporter.test.ts
+++ b/test/export/ExtensionExporter.test.ts
@@ -3,17 +3,17 @@ import { FSHTank, FSHDocument } from '../../src/import';
 import { FHIRDefinitions, load } from '../../src/fhirdefs';
 import { Extension } from '../../src/fshtypes';
 import { logger } from '../../src/utils/FSHLogger';
+import { LoggerSpy } from '../testhelpers/loggerSpy';
 
 describe('ExtensionExporter', () => {
   let defs: FHIRDefinitions;
   let doc: FSHDocument;
   let input: FSHTank;
   let exporter: ExtensionExporter;
-  let mockWriter: jest.SpyInstance<boolean, [any, string, ((error: Error) => void)?]>;
+  const loggerSpy = new LoggerSpy(logger);
 
   beforeAll(() => {
     defs = load('4.0.1');
-    mockWriter = jest.spyOn(logger.transports[0], 'write');
   });
 
   beforeEach(() => {
@@ -59,9 +59,7 @@ describe('ExtensionExporter', () => {
     extension.parent = 'DoesNotExist';
     doc.extensions.set(extension.name, extension);
     exporter.export();
-    expect(mockWriter.mock.calls[mockWriter.mock.calls.length - 1][0].message).toMatch(
-      /File: Wrong\.fsh.*Line: 14 - 24\D/s
-    );
+    expect(loggerSpy.getLastMessage()).toMatch(/File: Wrong\.fsh.*Line: 14 - 24\D/s);
   });
 
   it('should export extensions with FSHy parents', () => {
--- a/test/export/ExtensionExporter.test.ts
+++ b/test/export/ExtensionExporter.test.ts
@@ -2,15 +2,13 @@ import { ExtensionExporter } from '../../src/export';
 import { FSHTank, FSHDocument } from '../../src/import';
 import { FHIRDefinitions, load } from '../../src/fhirdefs';
 import { Extension } from '../../src/fshtypes';
-import { logger } from '../../src/utils/FSHLogger';
-import { LoggerSpy } from '../testhelpers/loggerSpy';
+import { loggerSpy } from '../testhelpers/loggerSpy';
 
 describe('ExtensionExporter', () => {
   let defs: FHIRDefinitions;
   let doc: FSHDocument;
   let input: FSHTank;
   let exporter: ExtensionExporter;
-  const loggerSpy = new LoggerSpy(logger);
 
   beforeAll(() => {
     defs = load('4.0.1');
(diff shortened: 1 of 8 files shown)


test/import/FSHImporter.Profile.test.ts
```diff
@@ -155,12 +152,8 @@ describe('FSHImporter', () => {
...
 
         importText(input, 'Dupe.fsh');
-        expect(mockWriter.mock.calls[mockWriter.mock.calls.length - 2][0].message).toMatch(
-          /File: Dupe\.fsh.*Line: 7\D/s
-        );
-        expect(mockWriter.mock.calls[mockWriter.mock.calls.length - 1][0].message).toMatch(
-          /File: Dupe\.fsh.*Line: 8\D/s
-        );
+        expect(loggerSpy.getMessageAtIndex(1, true)).toMatch(/File: Dupe\.fsh.*Line: 7\D/s);
+        expect(loggerSpy.getLastMessage()).toMatch(/File: Dupe\.fsh.*Line: 8\D/s);
```
mint-thompson: I was wondering how you'd handle this scenario. I like the reverse-index approach you've used here.

test/testhelpers/loggerSpy.ts
```diff
@@ -0,0 +1,22 @@
+import { Logger } from 'winston';
+
+export class LoggerSpy {
+  private mockWriter: jest.SpyInstance<boolean, [any, string, ((error: Error) => void)?]>;
+
+  constructor(logger: Logger) {
+    this.mockWriter = jest.spyOn(logger.transports[0], 'write');
+  }
```
mint-thompson: I'm wondering if it would be useful to have an optional parameter that defaults to 0 for indicating which transport you want to spy on, just in case we need to spy on a different transport later. But, that can probably wait until the situation actually arises.

test/export/ExtensionExporter.test.ts
```diff
@@ -3,17 +3,17 @@ import { FSHTank, FSHDocument } from '../../src/import';
...
 import { logger } from '../../src/utils/FSHLogger';
+import { LoggerSpy } from '../testhelpers/loggerSpy';
 
 describe('ExtensionExporter', () => {
   let defs: FHIRDefinitions;
   let doc: FSHDocument;
   let input: FSHTank;
   let exporter: ExtensionExporter;
-  let mockWriter: jest.SpyInstance<boolean, [any, string, ((error: Error) => void)?]>;
+  const loggerSpy = new LoggerSpy(logger);
```
cmoesel: Since `logger` itself is a global, is it necessary to pass it in to a constructor for `LoggerSpy`?  If not, then you could have `testhelpers/loggerSpy` just export an _instance_ of `loggerSpy` that is already initialized with the global logger.  Then you wouldn't need this extra line of code here.

test/testhelpers/loggerSpy.ts
kjmahalingam: That's definitely a good option for the future, but I agree that for now we probably don't need it.

test/import/FSHImporter.Profile.test.ts
kjmahalingam: I asked people around me what they thought until someone mentioned it and it felt like the right approach!

test/export/ExtensionExporter.test.ts
kjmahalingam: Sure, I suppose I could do that instead!

kjmahalingam: This should be done and ready to go.

test/testhelpers/loggerSpy.ts
```diff
@@ -0,0 +1,20 @@
+import { logger } from '../../src/utils/FSHLogger';
+
+class LoggerSpy {
+  private mockWriter = jest.spyOn(logger.transports[0], 'write');
+
+  getMessageAtIndex(index: number, reverse = false): string {
+    const i = reverse ? this.mockWriter.mock.calls.length - index - 1 : index;
```
cmoesel: Good thinking to support a way to do this from the end of the message array.

In many JS functions that take an index (for example, [slice](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice)), they use negative numbers to indicate offset from the end (e.g., reverse).  I'd suggest we follow suit, as it will be familiar to other JS developers.

So instead of:
```
expect(loggerSpy.getMessageAtIndex(1, true)).toMatch(/File: Dupe\.fsh.*Line: 7\D/s);
```
you have:
```
expect(loggerSpy.getMessageAtIndex(-1)).toMatch(/File: Dupe\.fsh.*Line: 7\D/s);
```

kjmahalingam: What would I then do if I wanted the last element? For example, the case where right now we would do:

```
loggerSpy.getMessageAtIndex(0, true);
```

kjmahalingam: Perhaps that could be in index of `-1` for the last element, and `-2` could be the second to last element?

kjmahalingam: I decided to go with that approach. @cmoesel what do you think of the newest push?

cmoesel: I'll look at the approach, but yeah -- `-1` should be the last element and `-2` is the second-to-last, etc.

cmoesel: Yep -- looks good to me!

PR #54, opened 2019-12-12
Add ValueSet to grammar

mint-thompson:
This pull request is only for updating the grammar and related files. The actual import processing comes later.

A value set is composed from terms, references, and filters. Terms and references are both fairly simple, grammatically speaking, but the grammar for filters is a little more complex. Because so many different forms of filter statements are allowable, the grammar must be flexible to allow those forms. The comment in `FSH.g4` helps explain some of this. The documentation at http://hl7.org/fhir/valueset-definitions.html#ValueSet.compose.include.filter explains the basic components of the filter: the system, property, operator, and value. FSH does not require that all of these components are explicitly declared. The following forms of filter definitions are all grammatically and semantically correct:
- `ABC display regex /important/i` : this is the typical form, with System, Property, Operator, and Value.
- `system = ABC` : this is a special form allowed when the Property is `system`, as there is no need to list the same information for both System and Value.
- `ABC descendant-of #765` : this form omits the Property, as it is implicitly looking for codes with the specified relationship to the code made by combining the given System and Value.
- `descendant-of ABC#765` : this form omits the System and Property, as the System is contained within the Value.

The supported flags for regular expressions are taken from https://javascript.info/regexp-introduction#regular-expressions: `i` for case-insensitive, `m` for multiline, `s` to allow dot to match newlines, and `u` for full unicode support. The FHIR specification for regular expressions at https://www.hl7.org/fhir/extension-regex.html doesn't mention flags, but these four flags seemed applicable to the usage of value set filter definitions.

The only metadata field I know of for value sets is the description, but I structured the `vsMetadata` similarly to the metadata rules for other entities to make it easier to add more fields later.

As this pull request updates the grammar, but not the importer, and we do not have any tests that are directly on the grammar, no new tests are included. Of course, all existing tests must still succeed.

As a bonus, the extraneous `STRING?` in the definition of the `CODE` lexer rule is removed, since the display string for a code is included in the `code` parser rule.

This pull request addresses https://standardhealthrecord.atlassian.net/browse/CIMPL-178.


antlr/src/main/antlr/FSH.g4
```diff
@@ -120,6 +143,9 @@ REFERENCE:          'Reference' WS* '(' WS* SEQUENCE WS* ('|' WS* SEQUENCE WS*)*
                  //  ^  NON-WHITESPACE
 CARET_SEQUENCE:     '^' ~[ \t\r\n\f]+;
 
+                 // '/' EXPRESSION '/' FLAGS
+REGEX:              '/' ('\\/' | ~[/\r\n])+ '/' REGEX_FLAGS;
```
cmoesel: Did you determine that there's a place to put those flags in the FHIR VS?  It seemed to me like there is not, but I wasn't sure if you researched it further.  If there's not, then I don't think we should support it in the grammar.

cmoesel:
@markkramerus and I discussed this further and came up w/ this syntax.  I think it still has all the same components -- just arranged in a different way:
* All codes from VS: `[exclude] codes from valueset XXX`
* All codes from CS: `[exclude] codes from system XXX`
* Code from CS with filter: `[exclude] codes from system XXX where {filter}`
* Coding (e.g., specific concept): `[exclude] XXX`

where `[exclude]` is optional.  The word "include" is not part of the grammar -- as it's the default behavior.

antlr/src/main/antlr/FSH.g4
mint-thompson: Since the value field in a filter is just a string, filters could only be included if the stored string was something like `/expression/flags` and not just `expression`. But, that seems like the sort of fact that would be included in the documentation if it were true, so I guess it expects the value to just be `expression`, which unfortunately means no support for flags.

mint-thompson:
Given that including things is probably much more common than excluding things, I think leaving the explicit `include` keyword out is reasonable. Is using a filter when including from another value set allowed, or is a filter only allowed when including from a code system?

cmoesel:
Yeah, that's the basic conclusion I came to as well. :-/

antlr/src/main/antlr/FSH.g4
cmoesel: Yeah, that's the basic conclusion I came to as well. :-/

cmoesel:
@mint-thompson -- that's a good question and it makes me wonder if we still don't have the syntax quite right.  You can see a more elaborate description of how the include/exclude stuff works here: http://hl7.org/fhir/R4/valueset.html#compositions

cmoesel:
Based on that, there is some intersectionality that we don't support since we chose to split out the different approaches (concept/filter/valuset) into separate rules.  For example, you *should* be able to indicate a VS and a filter (and result is intersection of the two), or a VS and concepts (and again, result is intersection again).  In fact, technically I think you could set system and VS to get just the codes from the VS in that system....  So... the question is, do we try to account for that intersectionality, or do we say you have to drop into `^` to get that?  At first blush, I'm thinking we should try to account for it.

cmoesel:
Oh -- also -- you can select multiple `valueSet` references to get the intersection of them all.

mint-thompson:
@cmoesel I took a look at the documentation for composition rules, and my read is that if you specify a value set, but do not specify a system, then you don't get access to filters (and importantly, a single include/exclude statement can have any number of value sets, but at most one system).

As for what FSH ought to support: I'm completely in favor of having it support the full power of the FHIR value set definition. Anything less than that is asking to be hoist by our own petard. This, of course, means that the grammar must be enhanced accordingly. I'm sketching out some potential grammars for those enhancements now.

cmoesel:
@mint-thompson -- Ah... you are right!  It looks like `system` is needed in order to understand the meaning of `property` in the `filter` -- and similarly, would be needed for `concept` as well (since `concept` doesn't have a property for system inside itself).  And there is even a constraint that says you must have system in order to use concept or filter:
image
Good analysis!

mint-thompson:
Based on helpful insights from @cmoesel and @markkramerus , the grammar is once again updated. The full power of included and excluded components should now be at your disposal, grammatically speaking. The specification is at http://hl7.org/fhir/valueset-definitions.html#ValueSet.compose.include, but the short version is, all of the following are valid components grammatically and semantically:
```
* codes from system ABC
* codes from valueset DEF, GHI
* ABC#123 ABC#234
* #123 #234 from system ABC
* #123 #234 from system ABC valueset DEF, GHI
* codes from system ABC where zip = hog and zap = egg
* codes from valueset DEF, GHI system ABC where zip = hog and zap = egg
```
Note that the following components are valid grammatically but not semantically:
```
* ABC#123 XYZ#765
* #123 #765
```
The first violates the `0..1` cardinality of system, and the second violates invariant rule `vsd-2`.

cmoesel:
This is coming along very nicely and is proving to be quite powerful!  I have a few preferences / suggestions, but maybe you shouldn't change anything until @markkramerus also pipes in, as he may have different opinions.

---

```
// Current
* ABC#123 ABC#234

// Proposed
* ABC#123
* ABC#234
// or
#123, #234 from ABC
```
Given that this invites syntactic error (when two codes from different systems are used), and that it's an error that probably isn't obvious to authors, I'd suggest we should support only _one_ code at a time when this format is used.  If an author wants multiples on one line they can always use the `from {system}` syntax.

---

```
// Current
* #123 #234 from system ABC

// Proposed
* #123, #234 from system ABC
```
In other places where we allow multiples, we use commas to separate items.  I feel like we should do that here too -- especially as codes can be long (if you include the display too).

---

```
// Current
* #123 #234 from system ABC valueset DEF, GHI

// Proposed
* #123 #234 from system ABC and valueset DEF, GHI
```

It's a minor syntactic sugar change, but I feel like it really improved readability.

---

```
// Current
* codes from valueset DEF, GHI system ABC where zip = hog and zap = egg

// Proposed
* codes from system ABC and valueset DEF, GHI where zip = hog and zap = egg
```
Similar to above, I think the `and` adds a lot.  I also changed order to be system then VS.  Do you think we should allow interchangeable order on that?  I can't decide.  Interchangeable order does complicate the grammar a little.

mint-thompson:
I've made concept lists into comma-separated lists, and the `and` token is now required when specifying both a valueset and system in a `from` statement. Note that this may result in undesired behavior if a concept has `, #` somewhere in its value. So, I hope nobody wants to do that.

One other thing that I discussed with @cmoesel that is not yet implemented is allowing for whitespace in codes. This is permitted based on FHIR specifications (http://hl7.org/fhir/datatypes.html#code), but I think that this will require some deeper consideration. My initial attempt was easily defeated, as `* #123 from system ABC` was parsed as a single concept with a very curious name and no `from` statement.

mint-thompson:
Spaces should be allowed in concepts now, but the strings have to be quoted. So, the following value set component is grammatically and semantically correct: `* #"some spaces", #"some more spaces" "and a description" from system ABC`. Thank you all for your patience as I worked on this one.

antlr/src/main/antlr/FSH.g4
```diff
@@ -102,8 +124,11 @@ NUMBER:             [+\-]? [0-9]+('.' [0-9]+)?;
                  //   '  UCUM UNIT   '
 UNIT:               '\'' (~[\\'])* '\'';
 
-                 // SYSTEM     #  SYSTEM   "DISPLAY"
-CODE:               SEQUENCE? '#' SEQUENCE STRING?;
+                 // SYSTEM     #  SYSTEM
+CODE:               SEQUENCE? '#' (SEQUENCE | CONCEPT_STRING);
```
cmoesel: Since you changed the existing `CODE` rule here, can you update the importer to accommodate the change?  Basically, anywhere that parses a rule that uses the `CODE` token, it will need to check if the resulting code value (after `#`) is wrapped in `"`.  If so, it will want to:
* remove surrounding `"`
* unescape any `\"`
* unescape any `\\`

antlr/src/main/antlr/FSH.g4
```diff
@@ -91,7 +113,7 @@ COMMA:              ',';
 // PATTERNS
 
                  //  "    CHARS    "
-STRING:             '"' (~[\\"])* '"';
+STRING:             '"' (~[\\"] | '\\"' | '\\\\')* '"';
```
cmoesel: Since you changed the existing `STRING` rule here, can you update the importer to accommodate the change?  Basically, anywhere that parses a rule that uses the `STRING` token, I think it will need to:
* unescape any `\"`
* unescape any `\\`

antlr/src/main/antlr/FSH.g4
```diff
@@ -120,6 +145,12 @@ REFERENCE:          'Reference' WS* '(' WS* SEQUENCE WS* ('|' WS* SEQUENCE WS*)*
                  //  ^  NON-WHITESPACE
 CARET_SEQUENCE:     '^' ~[ \t\r\n\f]+;
 
+                 // '/' EXPRESSION '/'
+REGEX:              '/' ('\\/' | ~[/\r\n])+ '/';
+
+
+COMMA_DELIMITED_CODES: (CODE (WS+ STRING)? WS* COMMA WS+)+ CODE (WS+ STRING)?;
```
cmoesel: I see why you needed this, and have no better solution, but it kind of sucks that we're going to have to parse out all these parts in the importer. :-/

mint-thompson:
The importer has been updated to unquote and unescape all `STRING` tokens (made easier by the `extractString` function), as well as quoted strings found as the concept when handling a `code`. A test has been added to demonstrate this fact.

cmoesel:
So Travis seems to be stuck and I somehow lost my admin privileges, so I can't even force this PR through.  I'll try to get those admin privileges back tomorrow and merge this thing!

cmoesel:
It seems that the Travis integration is broken since moving to github.com/fhir.  I've checked out the branch and run the tests locally to confirm that they work.

PR #55, opened 2019-12-13
Add tests for OnlyRule with FSHy target

kjmahalingam:
Originally, this task was about allowing OnlyRules to have a FSHy target. Turns out, this is already supported given work we did to support FSHy parents for Profiles and Extensions! This PR now just adds tests to prove that it already works.

### DIFF
--- a/test/export/StructureDefinitionExporter.test.ts
+++ b/test/export/StructureDefinitionExporter.test.ts
@@ -496,6 +496,129 @@ describe('StructureDefinitionExporter', () => {
     );
   });
 
+  it('should apply a correct OnlyRule on a non-reference FSHy choice', () => {
+    const profile = new Profile('Foo');
+    profile.parent = 'Observation';
+
+    const targetProfile = new Profile('MySpecialQuantity');
+    targetProfile.parent = 'Quantity';
+    doc.profiles.set(targetProfile.name, targetProfile);
+
+    const rule = new OnlyRule('value[x]');
+    rule.types = [{ type: 'MySpecialQuantity' }];
+    profile.rules.push(rule);
+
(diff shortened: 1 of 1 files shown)


test/export/StructureDefinitionExporter.test.ts
```diff
@@ -496,6 +496,129 @@ describe('StructureDefinitionExporter', () => {
...
+    const sd = exporter.structDefs[1];
+    const baseStructDef = resolve('Observation');
+
+    const baseValue = baseStructDef.findElement('Observation.value[x]');
+    const constrainedValue = sd.findElement('Observation.value[x]');
+
+    expect(baseValue.type).toHaveLength(11);
+    expect(baseValue.type[0]).toEqual({ code: 'Quantity' });
+    expect(baseValue.type[1]).toEqual({ code: 'CodeableConcept' });
+    expect(baseValue.type[2]).toEqual({ code: 'string' });
```
cmoesel: These three need to be updated now that @jafeltra's PR with the new `ElementDefinitionType` class is merged.  For example, line 519 should be
```ts
expect(baseValue.type[0]).toEqual(new ElementDefinitionType('Quantity'));
```

cmoesel: I won't specifically call out the others, but anywhere in these new tests where you're comparing against the `type`, you'll need to use an instance of the `ElementDefinitionType` class instead.

kjmahalingam:
I fixed the tests. As far as circular dependencies go, it turns out they don't work. It might be tricky to implement those right now. I am happy to work on it if it's a priority, but could move to something else otherwise.

I'll leave it up to @cmoesel. If that's not a priority, then we should be good to go on this.

PR #56, opened 2019-12-16
Extension value[x] bug

ngfreiter:
This fixes a bug that @markkramerus pointed out to me. When an element had to be unfoled and had to have a `value[x]` element resolved all in one step, the `findElementByPath` function would fail. The new test tests this case.

To see the code that led us to find this issue, clone the https://github.com/standardhealth/fsh-mcode repo and run sushi on it. The bug results from this line `* extension[EvidenceType].valueCodeableConcept from CancerDiseaseStatusEvidenceTypeVS (required)` in CancerDiseaseStatus.fsh. When sushi is on master, you should see 12 errors, the first of which says "No element found at path extension[EvidenceType].valueCodeableConcept". You will also notice in the output that the `CancerDiseaseStatusEvidenceTypeVS` does not get bound to `Observation.extension:EvidenceType.value[x]:valueCodeableConcept`. When sushi is on this branch, you should see 11 errors, and in the output you will see that `CancerDiseaseStatusEvidenceTypeVS` is bound to `Observation.extension:EvidenceType.value[x]:valueCodeableConcept`, which is the desired outcome of the problematic line.

### DIFF
--- a/src/fhirtypes/StructureDefinition.ts
+++ b/src/fhirtypes/StructureDefinition.ts
@@ -156,8 +156,8 @@ export class StructureDefinition {
       newMatchingElements = matchingElements.filter(e => e.path.startsWith(fhirPathString));
 
       if (newMatchingElements.length === 0 && matchingElements.length === 1) {
-        // If we did not find an [x] element, and there was previously only one match,
-        // We want to unfold that match and dig deeper into it
+        // If there was previously only one match,
+        // we want to unfold that match and dig deeper into it
         unfoldedElements = matchingElements[0].unfold(resolve);
         if (unfoldedElements.length > 0) {
           // Only get the children that match our path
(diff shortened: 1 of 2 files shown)


src/fhirtypes/StructureDefinition.ts
```diff
@@ -148,35 +148,37 @@ export class StructureDefinition {
     let fhirPathString = this.type;
     let matchingElements = this.elements;
     let newMatchingElements: ElementDefinition[] = [];
+    let unfoldedElements: ElementDefinition[] = [];
```
cmoesel: Does `unfoldedElements` really need to be in this scope (outside of the `for` loop)?  It seems like this might cause some unintentional behavior because (unlike `newMatchingElements`) it is not guaranteed to be re-assigned on each iteration of the `for` loop -- so it seems it is theoretically possible it will be assigned during one iteration of the loop and used in another.  Even if that's not possible or likely, putting it in the smaller scope still seems like the right thing to do.

src/fhirtypes/StructureDefinition.ts
```diff
@@ -148,35 +148,37 @@ export class StructureDefinition {
...
+        ]);
         if (matchingSlice) {
           newMatchingElements.push(matchingSlice, ...matchingSlice.children());
           fhirPathString = matchingSlice.path;
         }
       }
 
-      // TODO: If path is A.B.C, and we unfold B, but C is invalid, the unfolded
-      // elements are still on the structDef. We may want to change this to remove the elements
-      // upon error
```
cmoesel: It seems we dropped this TODO comment.  Was that intentional?  It seems to me that it probably still applies?

PR #58, opened 2019-12-19
IG Builder

cmoesel:
This introduces support for exporting a folder structure (and files) that can be used with the template-based IG Publisher.  You can test it locally by following these steps:

### First, run it against an existing unmodified FSH Tank.

```
$ ts-node src/app.ts path/to/tank
```

Note that SUSHI outputs to `build` by default now (instead of `out`).  Also note that the output in `build` is no different than the current functionality.  This is intended.

### Then add an ig-data folder to the FSH Tank.

If you add an `ig-folder` sub-folder to your tank, this tells SUSHI to generate an IG.  If it is empty, it will just use all the defaults based on your package.json.  We don't expect it to normally be empty though.

```
$ mkdir path/to/tank/ig-data
$ ts-node src/app.ts path/to/tank
```

Note that `build` now contains a different file structure.  CD into it, download and run the publisher:

```
$ cd build
$ ./_updatePublisher.sh
$ ./_genonce.sh
```

_(NOTE: These scripts -- including their names -- are actually provided by HL7.  I didn't make them up myself.  So don't complain to me about the weird naming of `_genonce.sh`)_

That should succeed (fingers crossed), after which, you can open `build/output/index.html` and check it out!

### Customize your IG

Try customizing your IG by adding the file `path/to/tank/ig-data/ig.ini` with the following contents:
```ini
[IG]
template = hl7.fhir.template
```

You can also add a `path/to/tank/ig-data/input/pagecontent/index.md` page to customize the main page of the ID.

Then generate it again and run `./_genonce.sh` again.  And... look at the different (different template and new index content).

### Files for Testing

For your testing pleasure, I've uploaded a few zip files.  One is a FSH Tank for an uncustomized IG, another is for a customized IG, and the last is for an IG with invalid customization (which should cause errors).  Enjoy!

[simple-ig.zip](https://github.com/FHIR/sushi/files/3982533/simple-ig.zip)
[customized-ig.zip](https://github.com/FHIR/sushi/files/3982534/customized-ig.zip)
[invalid-data-ig.zip](https://github.com/FHIR/sushi/files/3982535/invalid-data-ig.zip)


test/ig/IGExporter.test.ts
```diff
@@ -0,0 +1,390 @@
...
+    });
+
+    afterAll(() => {
+      temp.cleanupSync();
+    });
+
+    it('should should log errors for invalid user-specified values in ig.ini', () => {
+      // Check for log messages indicating invalid input
+      expect(loggerSpy.getMessageAtIndex(-5)).toMatch(
+        /igi\.ini: sushi does not currently support overriding ig value\..*File: .*\/invalid-data-ig\/ig-data\/ig.ini/s
```
ngfreiter: This fails for me on Windows because it looks like you are checking for the "/" in the file path.

test/ig/IGExporter.test.ts
```diff
@@ -0,0 +1,390 @@
...
+      expect(content.IG.excludexml).toEqual('Yes');
+      expect(content.IG.excludejson).toEqual('Yes');
+      expect(content.IG.excludettl).toEqual('Yes');
+      expect(content.IG.excludeMaps).toEqual('Yes');
+    });
+
+    it('should log an error if the user attempted to add more pages', () => {
+      // Check for log messages indicating invalid input
+      expect(loggerSpy.getMessageAtIndex(-6)).toMatch(
+        /SUSHI does not yet support custom pagecontent other than index\.md\..*File: .*\/invalid-data-ig\/ig-data\/input\/pagecontent/s
```
ngfreiter: Also fails on Windows.

test/ig/IGExporter.test.ts
```diff
@@ -0,0 +1,390 @@
...
+      );
+    });
+
+    it('should log an error if supplied package-list.json does not match package.json', () => {
+      // Check for log messages indicating invalid input
+      expect(loggerSpy.getMessageAtIndex(-2)).toMatch(
+        /package-list\.json: package-id value \(wrong-package-id\) does not match name declared in package\.json \(sushi-test\)\.  Ignoring custom package-list\.json\..*File: .*\/invalid-data-ig\/ig-data\/package-list.json/s
```
ngfreiter: Also fails on Windows.

test/ig/IGExporter.test.ts
cmoesel: Oops.

cmoesel:
@ngfreiter -- Good catch.  I've updated the regular expressions in the tests to support `/` or `\` as the file separator.  Please try again!

cmoesel:
@ngfreiter -- our new automated CI found a bug with case sensitivity of filenames in our tests -- so I had to push another commit.  By doing so, it dismissed your review.  Can you approve again?

PR #60, opened 2019-12-19
Add CodeSystem class

mint-thompson:
The CodeSystem FSH type represents a set of related concepts.

The only sort of interesting thing here is the `addConcept` helper method, since any `FshCode` added to the system ought to have its `system` property set correctly.

This pull request addresses https://standardhealthrecord.atlassian.net/browse/CIMPL-179


src/fshtypes/CodeSystem.ts
```diff
@@ -0,0 +1,23 @@
...
+  concepts: FshCode[];
+
+  constructor(public name: string) {
+    super();
+    this.id = name;
+    this.concepts = [];
+  }
+
+  addConcept(code: string, display?: string) {
+    this.concepts.push(new FshCode(code, this.id, display));
```
cmoesel: So... this is interesting, because really, the system in `FshCode` *ought* to be a URL -- but since we haven't yet generated the URL (it's based on the config `canonical`), we don't have a URL to put into the `FshCode`.

While I totally understand the preference for re-using `FshCode`, there is an added problem since CodeSystem concepts can have both a `display` _and_ a `definition` -- and `FshCode` has nowhere to put `definition`.

So... I wonder if it's better to create a new special type (or class) representing a CodeSystem concept, that has `code`, `display`, and `definition`, but does _not_ have `system`?

mint-thompson: A new class seems like the best approach, since as you pointed out, this really isn't quite the same as a `FshCode`. I've updated the branch accordingly.

PR #61, opened 2019-12-19
CodeSystem grammar

mint-thompson:
The grammar definition now includes CodeSystems. A CodeSystem has
a few metadata fields, and contains a list of codes. Existing rules
provide most of the definition of this new entity.

This one is a lot simpler than the ValueSet grammar, and addresses issue https://standardhealthrecord.atlassian.net/browse/CIMPL-180.


antlr/src/main/antlr/FSH.g4
```diff
@@ -18,6 +18,8 @@ invariantMetadata:  description | expression | xpath | severity;
 
 valueSet:           KW_VALUESET SEQUENCE vsMetadata* vsComponent+;
 vsMetadata:         id | title | description;
+codeSystem:         KW_CODESYSTEM SEQUENCE csMetadata* code*;
```
cmoesel: So this is the one where we *do* have that extra `definition` field for each concept, so the `code` parser rule isn't quite the right fit on its own.  We also need to add in that optional `"`-delimited `definition`.

cmoesel: And in the importer we'll have to make sure that the `code` doesn't have a system before the `#` -- because in the CodeSystem context, all concepts should start w/ `#` (but that's hard to enforce in a clean way in the grammar since other contexts optionally allow the system).

mint-thompson: Thank you for reminding me about `definition`! I've added an extra optional `STRING` token to the new `concept` rule in order to capture it along with the `code`.

PR #65, opened 2019-12-23
Load packages

ngfreiter:
This adds package loading to SUSHI. As a result, the definitions of FHIR for 4.0.1 have been removed, and now those will be loaded in from the `~/.fhir` cache of the user. The tests mock out the functionality, but if you want to see it in action, you can create a folder with a few FSH files and a package.json file. There is an existing example here https://github.com/HL7/fhir-shorthand/tree/master/dev-days-2019/working-samples that could be used. Then in the package.json file, you can add dependencies as shown below:
```
"dependencies": {
    "hl7.fhir.r4.core": "4.0.1",
    "hl7.fhir.us.core": "3.1.0"
 }
```
If you go into your own FHIR cache at `~/.fhir/packages` and delete the folders corresponding to these versions, and then run SUSHI, you should see that the packages are downloaded and extracted from https://simplifier.net/packages back in to where they were deleted. If you then run SUSHI again, you should notice that since the packages are already in the local cache, they are not changed, and they are just loaded from the cache.

There are two special version tags that change the behavior. The `dev` tag can be used in place of a version number, and if this is used, the package will only be loaded from the local cache, it will not be downloaded. The `current` tag can also be used, and if this is used the package will be downloaded from http://build.fhir.org/ig/HL7/, where the most current versions of packages are hosted.


mint-thompson:
Worked as described for me. I'm on a windows machine, so I'd like someone on a mac to also confirm that this works for them.

kjmahalingam:
Works perfectly on my Mac as well, good work!

src/fhirdefs/load.ts
```diff
@@ -1,29 +1,113 @@
...
+    let packageUrl: string;
+    if (version === 'dev') {
+      // Dev packages must be present in local FHIR cache
+      throw new DevPackageLoadError(fullPackageName);
+    } else if (version === 'current') {
+      // Current packages need to be loaded using build.fhir.org
+      const baseUrl = 'http://build.fhir.org/ig';
+      const res = await rp.get({ uri: `${baseUrl}/qas.json`, encoding: null });
+      const decoded = String.fromCharCode(...new Uint16Array(res));
+      const qaJSON = JSON.parse(decoded);
```
mint-thompson: Since the response body is expected to be JSON, it should be possible make the request with the option `json: true`. This will automatically parse the response body as JSON, so you won't need to do that manually. Additionally, it would be useful to include some error handling in case the package list is unavailable or broken.

src/fhirdefs/load.ts
```diff
@@ -1,29 +1,113 @@
...
+      })[0];
+      // Current packages are stored at build.fhir.org
+      packageUrl = `${baseUrl}/${newestPackage.repo}/package.tgz`;
+    } else {
+      // All non-current packages are stored at packages.fhir.org
+      packageUrl = `http://packages.fhir.org/${packageName}/${version}`;
+    }
+    // Create a temporary file and write the package to there
+    temp.track();
+    const tempFile = temp.openSync(); // TODO CHANGE THIS TO USE OTHER TEMP LIB
```
mint-thompson: Is this using the correct temp library now? If so, you can remove the TODO.

src/fhirdefs/load.ts
ngfreiter: Ah true, that is neater, good catch. I'll also add some error handling.

src/fhirdefs/load.ts
ngfreiter: Yes it's using `temp` not `tmp` now, I'll remove that.

src/fhirdefs/load.ts
```diff
@@ -1,29 +1,118 @@
...
       }
+    } else {
+      // All non-current packages are stored at packages.fhir.org
+      packageUrl = `http://packages.fhir.org/${packageName}/${version}`;
     }
+    // Create a temporary file and write the package to there
+    temp.track();
+    const tempFile = temp.openSync();
+    const targetDirectory = path.join(cachePath, fullPackageName);
+    const res = await rp.get({ uri: packageUrl, encoding: null });
```
mint-thompson: There's an additional error case that I think should be handled here: if a dependency specifies a version that doesn't exist, the promise here will be rejected, which currently causes an `UnhandledPromiseRejectionWarning` to appear in the output. When this `get` fails, it would be useful to show a message based on the HTTP response code: 404 probably means that the user specified a package and version combination that doesn't exist, 5XX probably just means the server is down.

src/app.ts
```diff
@@ -7,79 +7,98 @@ import { importText, FSHDocument, FSHTank } from './import';
...
+  // Load external dependencies
+  const defs = new FHIRDefinitions();
+  const dependencyDefs: Promise<FHIRDefinitions>[] = [];
+  for (const dep of Object.keys(config?.dependencies ?? {})) {
+    try {
+      dependencyDefs.push(loadDependency(dep, config.dependencies[dep], defs));
+    } catch (e) {
+      logger.error(e.message);
+      logger.error(`Failed to load ${dep}#${config.dependencies[dep]}`);
+    }
```
mint-thompson: A normal try-catch block doesn't work here because `loadDependency` is an async function. When a `throw` statement runs in an async function, the function returns a rejected promise. So, you'll need to handle that with something like:
`loadDependency(...).catch(e => { /* current catch block goes here */ });`

ngfreiter:
Ok @mint-thompson the latest commit should address those concerns you raised.

PR #66, opened 2019-12-23
Initial Instance Export

jafeltra:
This PR adds initial support for exporting instances.

Supported features include:
- Fixing values on instances
- Fixing values on extensions by accessing the extension directly through an index (not by using a slice name)
- Fixing values on an attribute that has been constrained from a `1..*` cardinality to a `1..1` cardinality will correctly set the value within an array
- Instances are available within the Table of Contents in an IG at the same level as other profiles and extensions

Unsupported features that will come in later tasks:
- Implying fixed values that are fixed in a profile or extension
- Fixing values using slices
- Adding extensions to primitives


src/ig/IGExporter.ts
```diff
@@ -155,6 +155,18 @@ export class IGExporter {
         exampleBoolean: false
       });
     });
+    this.pkg.instances.forEach(instance => {
+      const instanceFileName = `${instance.resourceType}-${instance.id ?? instance.instanceName}`;
+      const instancePath = path.join(igPath, 'input', 'instances', `${instanceFileName}.json`);
+      outputJSONSync(instancePath, instance, { spaces: 2 });
```
mint-thompson: Just to double-check: this is supposed to include the `instanceName` field, which would get removed if only the resources were being generated in `app.ts` line 81.

src/ig/IGExporter.ts
```diff
@@ -155,6 +155,18 @@ export class IGExporter {
         exampleBoolean: false
       });
     });
+    this.pkg.instances.forEach(instance => {
+      const instanceFileName = `${instance.resourceType}-${instance.id ?? instance.instanceName}`;
```
ngfreiter: I think you could make a case here that the `instance.instanceName` should always be used to name the file. If you check out something like USCore, https://www.hl7.org/fhir/medicationadministration0311.json.html, you will see that the file is not necessarily named after the id of the instance. Also I don't think there would necessarily be any rule in FHIR against having two instances with the same id, which could cause an issue here (although I think the example of USCore is a stronger reason to always use instanceName).

src/app.ts
```diff
@@ -76,10 +76,20 @@ if (fs.existsSync(igDataPath)) {
...
     );
   }
+  for (const instance of outPackage.instances) {
+    const fileName = `${instance.resourceType}-${instance.id ?? instance.instanceName}`;
+    delete instance.instanceName; // Only needed for the file name - not a FHIR property
+    fs.writeFileSync(
+      path.join(program.out, `${fileName}.json`),
+      JSON.stringify(instance, null, 2),
+      'utf8'
+    );
```
ngfreiter: As @mint-thompson mentioned above, could it maybe be cleaner to make `InstanceDefinition` a class with a `toJSON` method that handled removing `instanceName` and a `getFileName` method? Then I think you could just make line `72` above `for (const sd of [...outPackage.profiles, ...outPackage.extensions, ...outPackage.instances])`, and it would work as is.

src/export/InstanceExporter.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    // All other values of the instance will be fixedValues explicitly on the FSH Instance
+    instanceDef = this.setFixedValues(
+      fshDefinition,
+      instanceDef,
+      instanceOfStructureDefinition,
+      structDefExporter
+    );
+
+    // Add any set values that were set on the StructDef but not explicitly on the FSH Instance
```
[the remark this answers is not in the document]
ngfreiter: In general, those to-do tests look helpful we should definitely be using those. But I do have a task this sprint for implementing this feature, so I'd be ok with it if Julia wants to skip that for now. Of course, if she wants to write some tests for me I wouldn't stop her 😉 .

src/fhirtypes/common.ts
```diff
@@ -19,6 +20,14 @@ export function setPropertyOnInstance(
     value,
     resolve
   );
+  setFixedValueOfPath(instance, pathParts, fixedValue);
+}
+
+export function setFixedValueOfPath(
```
ngfreiter: Could you just add a comment header to this function? Also I know I stole the name that this function should have, but maybe we can come up with better names for this function and `setPropertyOnInstance`. I feel like `setFixedValueOfPath` should truly be named `setPropertyOnInstance`. Maybe then `setPropertyOnInstance` could become `setPropertyOnDefinitionInstance`? I guess still kinda confusing but maybe a little more clear.

test/export/InstanceExporter.test.ts
```diff
@@ -0,0 +1,94 @@
...
+    const instance = new Instance('Bar');
+    instance.instanceOf = 'Patient';
+    const fixedValRule = new FixedValueRule('gender');
+    const fixedFshCode = new FshCode('foo', 'http://foo.com');
+    fixedValRule.fixedValue = fixedFshCode;
+    instance.rules.push(fixedValRule);
+    doc.instances.set(instance.name, instance);
+    const exported = exporter.export();
+    expect(exported.length).toBe(1);
+    // @ts-ignore
```
ngfreiter: I think if you add a line to `InstanceDefinition` like this: `[key: string]: any`, that indicates that any key value pair is allowed on `InstanceDefinition`, so the `ts-ignore` wouldn't be needed. In my mind adding this line would make sense, since an Instance can have any key value pair defined by FHIR. Of which there are so many that trying to list them all out would be a bad idea. But I'm not confident enough in my TypeScript best practices to know if `[key: string]: any` is a bad idea. @mint-thompson, do you have any thoughts?

src/fhirtypes/StructureDefinition.ts
```diff
@@ -347,6 +356,12 @@ export class StructureDefinition {
         pathPart.brackets?.forEach(p => (currentPath += `[${p}]`));
       }
       currentElement = this.findElementByPath(currentPath, resolve);
+
+      // If the element has a base.max that is great than 1, but the element has been constrained, still set properties in an array
+      const valueShouldBeWithinArray =
+        currentElement?.base?.max !== '0' &&
+        currentElement?.base?.max !== '1' &&
+        (currentElement?.max === '0' || currentElement?.max === '1');
```
cmoesel: I'm confused about this last line.  So if the `currentElement.base.max` is `*` _and_ the `currentElement.max` is also `*`, we're setting `valueShouldBeWithinArray` to `false`?  Why?

The logic above also effectively says that if there is no `currentElement.base` defined, but `currentElement.max` is `*`, then we also say `valueShouldBeWithinArray` is `false`.  I don't think that's right.  If there is no `base`, then the `currentElement.max` should be used to determine if it's an array (e.g., if `currentElement.max` is not `0` or `1`).  Actually the opposite of how it works above if `currentElement.base` is not defined.

cmoesel: I see it gets handled ok down below, but since this variable is called `valueShouldBeWithinArray`, it's a bit confusing that it's not always correct.

test/export/InstanceExporter.test.ts
cmoesel: I think `[key: string]: any` makes sense in this very specific context.

src/export/InstanceExporter.ts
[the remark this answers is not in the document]
jafeltra: Ah, those todo tests look really cool! I've never used them before. I am thinking it might make sense to hold off on them in this particular case since Nick is working on the task to add this functionality in this sprint, but if either of you feel strongly, I can add them now.

src/ig/IGExporter.ts
jafeltra: That's a great catch! The `instanceName` field should be removed here as well. I'll fix that.

src/export/InstanceExporter.ts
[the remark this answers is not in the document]
mint-thompson: If it's going to be addressed in this sprint, then I feel like it's not super necessary to add the todo tests.

src/ig/IGExporter.ts
jafeltra: Those are two good points. I think when I originally discussed the naming pattern with @cmoesel, we were looking at the USCore patient example [here](https://www.hl7.org/fhir/us/core/Patient-example.json.html) and followed that. Given that doesn't seem consistent, I don't think I have a strong opinion either way. Are the `instanceName`s given in the FSH file supposed to be unique? Because if so, that seems like a fair enough reason to just use that.

src/app.ts
jafeltra: I like this! I think at one point I had it implemented as a class, but I think I switched it back because it was so simple. I do like that it would simplify the output both here and in the IGExport and would resolve the other issue Mint pointed out where `instanceName` needs to be removed there. I'll update!

src/ig/IGExporter.ts
jafeltra: I talked more with @ngfreiter about this, and I think we are going to leave the naming as it is for now. Using the specified `id` if it exists seems to mirror what USCore and mCODE are using as file names for various instances.

src/fhirtypes/common.ts
jafeltra: I agree the naming is confusing. Maybe the existing `setPropertyOnInstance` could be renamed to `setPropertyOnStructureDefinitionInstance`? @ngfreiter do you think that's too wordy? Then the `setFixedValueOfPath` could actually be `setPropertyOnInstance`.

ngfreiter: The existing `setPropertyOnInstance` gets used on both `StructureDefinition`s and `ElementDefinition`s. So the most accurate name would probably be `setPropertyOnStructureDefinitionOrElementDefinitionInstance`. But now that is pretty dang wordy, which is why the best I could come up with was `setPropertyOnDefinitionInstance`, but that isn't as clear.

jafeltra: Hm I honestly don't mind `setPropertyOnDefinitionInstance` and I definitely prefer it over my suggestion now that you pointed that out. So I'm good with using that, as long as you don't find it too unclear.

src/errors/InstanceOfNotDefinedError.ts
```diff
@@ -0,0 +1,14 @@
+import { Annotated } from './Annotated';
+import { WithSource } from './WithSource';
+import { SourceInfo } from '../fshtypes';
+
+export class InstanceOfNotDefinedError extends Error implements Annotated, WithSource {
+  specReferences = [''];
```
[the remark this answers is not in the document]
jafeltra: There isn't a reason. I don't know of any spec reference, so I'll remove `Annotated`. However, I could definitely be missing a spec reference that exists for this, so if anyone knows of it, I'll add it back in.

src/fhirtypes/StructureDefinition.ts
```diff
@@ -145,7 +145,7 @@ export class StructureDefinition {
     // If the path already exists, get it and return the match
     // If !path just return the base parent element
     const fullPath = path ? `${this.type}.${path}` : this.type;
-    const match = this.elements.find(e => e.path === fullPath);
+    const match = this.elements.find(e => e.path === fullPath && !e.id.includes(':'));
```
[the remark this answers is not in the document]
jafeltra: The test "should allow setting values directly on extensions by accessing indexes" in `StructureDefinition.test.ts` ([here](https://github.com/FHIR/sushi/pull/66/files#diff-11a71c2cc494f9aa25b4ea6946cc28caR578))  is what caused me to add this extra logic. The issue I was having was that I wanted to set a value on a later slice that hadn't yet been unfolded and it was matching on the previous slices that already had a fixed value.

src/fhirtypes/StructureDefinition.ts
jafeltra: I think maybe to clarify what this variable is actually doing, I will rename the variable to be more accurate. I'm thinking something along the lines of `baseMaxGreaterThanMax` or `constrainedElementBaseIsArray`. Neither of those exactly capture what the variable is checking, but I'm struggling to come up with a short name that fully captures the logic. @ngfreiter or @mint-thompson any ideas?

ngfreiter: Maybe `nonArrayElementBasedOnArray`. A mouthful but it makes sense I think.

jafeltra: Oh yes! I like that!

jafeltra:
I think I've addressed all of the comments. If I missed any comments or there are any further suggestions, just let me know!

@mint-thompson hopefully making InstanceDefinition into a class will help with some of the confusion you had when first following along the export. If you feel like there is anything else I can do to make it less confusing, let me know!

ngfreiter:
The changes addressed my concerns, so I'm happy with it. To my eye it also seemed like you hit on all of @cmoesel 's comments, but @mint-thompson when you check out the changes you requested, if you could also verify that the comments from Chris have been addressed, that would make me feel more safe and secure. Or if we don't merge until after the holiday Chris can verify Chris's changes.

jafeltra:
I just rebased after the two updates to master and updated one test file to reflect the new way we load packages.

PR #67, opened 2019-12-30
Add identifying integer to rules in FSH grammar

kjmahalingam:
This adds support for an identifying integer for rules, as per: https://standardhealthrecord.atlassian.net/browse/CIMPL-194

Running the tests should prove it works just fine. Additionally, if you want to test it manually, you can edit one of the files in the FHIR Dev Days 2019 Working Samples in order to have a rule or a few rules who have an integer directly after their `*`. For example, you could change something like this:
`* method = SCT#5880005 "Clinical examination"`
to this:
`*201 method = SCT#5880005 "Clinical examination"`
That should have zero effect on the toolchain whatsoever.


kjmahalingam:
Yeah, eventually it might be worth including in the logs since I can imagine that would be useful too, but that's probably out of scope for now.

Good call on the test! I'll add one now.

kjmahalingam:
@mint-thompson The last commit should include a new test for this. I decided to add it to the root `FSHImporter.test.ts` because it seemed redundant to include a metadata-related test like this in all other locations. Let me know if you think it would be better suited to somewhere else.

PR #68, opened 2019-12-30
Import ValueSet

mint-thompson:
A ValueSet is defined as a set of included and excluded components. There are two types of components: concept components, which specify one or more concepts, and filter components, which use filter expressions to specify concepts. http://hl7.org/fhir/valueset-definitions.html#ValueSet.compose.include has the relevant details about what can get included in a ValueSet. http://hl7.org/fhir/valueset-filter-operator.html describes the set of filter operators that can be used in a filter expression.

SUSHI is not responsible for expanding ValueSet definitions. As a result, filter expressions do not get evaluated to the greatest possible degree. For example, SUSHI does not determine the elements of a comma-separated list used as a filter value.

Because this feature has a lot of associated grammar definitions, please also consider in your review the usability of the grammar for writing ValueSet specifications.

This pull request addresses https://standardhealthrecord.atlassian.net/browse/CIMPL-150.


test/import/FSHImporter.ValueSet.test.ts
```diff
@@ -0,0 +1,552 @@
...
+                .withFile('Felines.fsh')
+            }
+          ]
+        );
+      });
+
+      it('should parse a value set that uses filter operator descendent-of, but spelled correctly', () => {
+        const input = `
+        ValueSet: AllFelinesVS
+        * codes from valueset ZooVS where code descendant-of ZOO#cat
```
kjmahalingam: I think I'm a bit confused about this test. Is this supposed to indicate that `descendent-of` and `descendant-of` are both valid? If that's the case, could you just rename the test to clarify that it's an alternate spelling? If that's not the case, could you explain what is going on here?

test/import/FSHImporter.ValueSet.test.ts
```diff
@@ -0,0 +1,552 @@
...
+        expect(valueSet.components.length).toBe(1);
+        assertValueSetFilterComponent(valueSet.components[0], 'ZOO', undefined, []);
+        expect(loggerSpy.getLastMessage()).toMatch(/"generalizes".*code/);
+        expect(loggerSpy.getLastMessage()).toMatch(/File: Mustelids\.fsh.*Line: 3\D/s);
+      });
+
+      it('should parse a value set that uses filter operator exists', () => {
+        const input = `
+        ValueSet: ZooVS
+        * codes from system ZOO where display exists true
```
kjmahalingam: This is a minor comment, so if this is difficult to implement, I would say really don't worry about it for now. That said, I find this wording a little bit odd. Maybe it would make sense to do something like:
`where display exists` or `where display does not exist` instead of `where display exists true` or `where display exists false`. 
But trust me, I (and @jafeltra) have done plenty of work that has to deal with the complexities of English grammar before, so the added code to make something like that work is honestly maybe not worthwhile.

src/fshtypes/ValueSet.ts
```diff
@@ -7,15 +7,13 @@ import { ValueSetComponent } from './ValueSetComponent';
  */
 export class ValueSet extends FshEntity {
   id: string;
-  url?: string;
+  title?: string;
   description?: string;
-  inclusions: ValueSetComponent[];
-  exclusions: ValueSetComponent[];
+  components: ValueSetComponent[];
```
kjmahalingam: I'm fine with this decision, but just for understanding, was this change made just because there are now far more components that exist in addition to just inclusions and exclusions?

test/import/FSHImporter.ValueSet.test.ts
mint-thompson: Yes, both spellings are valid. I'll clarify that in the test name.

test/import/FSHImporter.ValueSet.test.ts
mint-thompson: I like the idea that `where display exists` should be allowed as an equivalent to `where display exists true`, and it should introduce minimal complications, so I'll try to add it. Having the special grammar for `does not exist` as an equivalent to `where display exists false` would be nice, but would more or less need to be treated as a special case in the grammar, since it doesn't match the current rule `vsFilterOperator: EQUAL | SEQUENCE`.

src/fshtypes/ValueSet.ts
mint-thompson: This change was made for ease of implementation, but I could honestly go either way on it, since the alternative isn't really much more complicated. It might be easier to deal with during export to have them separated in advance.

kjmahalingam: This is fine to me, thanks for explaining!

test/import/FSHImporter.ValueSet.test.ts
kjmahalingam: Okay, that sounds good! Don't worry about `does not exist`, then. But if it doesn't take much time to do `where display exists` then that would be nice.

test/import/FSHImporter.ValueSet.test.ts
kjmahalingam: Perfect, thanks!

src/import/FSHImporter.ts
```diff
@@ -550,6 +630,215 @@ export class FSHImporter extends FSHVisitor {
...
+      } else {
+        logger.error(
+          `Concept ${singleCode.code} must include system as "SYSTEM#CONCEPT" or "#CONCEPT from system SYSTEM"`,
+          {
+            file: this.file,
+            location: this.extractStartStop(ctx)
+          }
+        );
+      }
+    } else if (ctx.COMMA_DELIMITED_CODES()) {
```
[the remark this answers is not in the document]
mint-thompson: Good catch! adding this one

src/import/FSHImporter.ts
```diff
@@ -205,6 +229,51 @@ export class FSHImporter extends FSHVisitor {
...
+  ) {
+    const seenPairs: Map<VsMetadataKey, string> = new Map();
+    metaCtx
+      .map(vsMetadata => ({
+        ...this.visitVsMetadata(vsMetadata),
+        context: vsMetadata
+      }))
+      .forEach(pair => {
+        if (seenPairs.has(pair.key)) {
+          logger.error(
```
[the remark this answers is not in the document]
mint-thompson: Ah, I forgot to add this because the duplicate metadata logic is the same as for other entities. But, they're different functions, so I'll add tests. Maybe I should also try to figure out how to not have that logic duplicated, too, given that the duplicate metadata rules applies to all entity types.

src/errors/ValueSetFilterOperatorError.ts
```diff
@@ -0,0 +1,5 @@
+export class ValueSetFilterOperatorError extends Error {
+  constructor(public value: string) {
+    super(`Operator "${value}" not available in value set filter.`);
```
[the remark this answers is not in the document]
mint-thompson: Thanks for reminding me about the `Annotated` interface, I had forgotten about it and was wondering "huh, should I list things that are allowed when I show the error?" Yes, I should, and this is how.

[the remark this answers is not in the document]
mint-thompson: I took another look at spec's definition for `property`, and it might actually be correct to allow any string (and not throw an error when it isn't one of the four specified). All it says on http://hl7.org/fhir/valueset-definitions.html#ValueSet.compose.include.filter.property is that it needs a code which "identifies a property or a filter defined in the code system." The four enum values are the four properties of a `code`, but since system-specific filters can be defined, "a filter defined in the code system" could basically be anything. I'll keep the current behavior for now, but every time I reread that part of the spec, it seems more vague.

mint-thompson:
New test cases added, including a test for using the `exists` operator and omitting the value.

src/errors/ValueSetFilterOperatorError.ts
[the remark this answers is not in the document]
ngfreiter: I think you are right that we may want to be more permissive here. For example, for this ValueSet from mCODE here: http://standardhealthrecord.org/guides/mcode-r4/ValueSet-onco-core-CancerBodyLocationVS.json.html, wouldn't we want to say `* codes from SCT where concept is-a 123037004`? But I think that wouldn't be allowed. If that is correct, I would vote that we be more permissive here. Do you think it is too permissive to just allow any single word property here? You're right, the spec is a little frustratingly vague though.

ngfreiter:
I had just one response to a comment above that I want to clear up, and I also think it may be a good idea to add a test for a case where the `and` is used:
```
* codes from system SCT where 
    code is-a SCT#399983006 and
    code is-not-a SCT#399983006
```
I call out this case because I noticed previously that when running SUSHI on fsh-mcode, https://github.com/standardhealth/fsh-mcode, this case did cause an error. I think you've made changes to fix that error, but better safe than sorry.

src/import/FSHImporter.ts
```diff
@@ -550,6 +631,217 @@ export class FSHImporter extends FSHVisitor {
...
+   * which spells operator "descendant-of" as "descendent-of".
+   * @see {@link http://hl7.org/fhir/valueset-filter-operator.html}
+   */
+  visitVsFilterDefinition(ctx: pc.VsFilterDefinitionContext): ValueSetFilter {
+    const property = ctx
+      .SEQUENCE()
+      .getText()
+      .toLocaleLowerCase() as VsProperty;
+    if (Object.values(VsProperty).indexOf(property) < 0) {
+      throw new ValueSetFilterPropertyError(ctx.SEQUENCE().getText());
```
cmoesel: I know this has already been merged, but I agree with the second thoughts that @mint-thompson expressed earlier -- we should be permissive here and allow any property (not just the four).  If FHIR intended for only four possible values, it would have bound `property` with a `required` value set containing only those four values.  Since it did not bind `property` at all, we should assume that means that anything that is a valid `code` type is allowed.

PR #69, opened 2019-12-31
Allow importer to resolve aliases from any input file

kjmahalingam:
Addresses https://standardhealthrecord.atlassian.net/browse/CIMPL-190

This adds support for the importer to resolve an alias from any input file, in any other input file. This was done by updating the `importText.ts` file so that the function takes in all input files, instead of the previous approach of taking in only one at a time. There is still a single importer per input file, though.

An optional `allAliases` argument has been provided to the importers' `visitDoc` method. If it is provided (which it is for our toolchain), then each individual importer will resolve aliases from this `allAliases` map. Otherwise, it will operate based on the aliases within the importer's sole document, as before.

Test cases have been added to test this functionality. Tests have also been updated to reflect the structural change to `importText`.

There's an odd side effect to this. I will explain that in an inline comment in the code.

### DIFF
--- a/src/import/FSHImporter.ts
+++ b/src/import/FSHImporter.ts
@@ -1,6 +1,10 @@
 import * as pc from './parserContexts';
 import { FSHDocument } from './FSHDocument';
+import { RawFSH } from './RawFSH';
+import { FSHErrorListener } from './FSHErrorListener';
 import { FSHVisitor } from './generated/FSHVisitor';
+import { FSHLexer } from './generated/FSHLexer';
+import { FSHParser } from './generated/FSHParser';
 import {
   Profile,
   Extension,
@@ -29,7 +33,7 @@ import {
   ContainsRule,
   CaretValueRule
 } from '../fshtypes/rules';
-import { ParserRuleContext } from 'antlr4';
+import { ParserRuleContext, InputStream, CommonTokenStream } from 'antlr4';
 import { logger } from '../utils/FSHLogger';
 import { TerminalNode } from 'antlr4/tree/Tree';
 import {
@@ -39,6 +43,7 @@ import {
   ValueSetFilterValueTypeError,
   ValueSetFilterMissingValueError
 } from '../errors';
+import flatMap from 'lodash/flatMap';
 
 enum SdMetadataKey {
   Id = 'Id',
@@ -76,13 +81,41 @@ enum Flag {
  * we must call the explicit visitX functions.
  */
 export class FSHImporter extends FSHVisitor {
-  private used = false;
-  private readonly doc: FSHDocument;
+  private currentFile: string;
+  private currentDoc: FSHDocument;
   private allAliases: Map<string, string>;
 
-  constructor(public readonly file: string = '') {
+  constructor() {
     super();
-    this.doc = new FSHDocument(file);
+  }
+
+  import(rawFSHes: RawFSH[]): FSHDocument[] {
+    const docs: FSHDocument[] = [];
+    const contexts: pc.DocContext[] = [];
+    rawFSHes.forEach(rawFSH => {
+      docs.push(new FSHDocument(rawFSH.path));
+      contexts.push(this.parseDoc(rawFSH.content, rawFSH.path));
+    });
+
+    // Import all aliases first
+    this.allAliases = new Map(
+      flatMap(
+        contexts.map((context, index) => {
+          this.currentDoc = docs[index];
+          this.currentFile = this.currentDoc.file ?? '';
+          return this.getAliases(context);
+        }),
+        aliases => Array.from(aliases)
+      )
+    );
+
+    contexts.forEach((context, index) => {
+      this.currentDoc = docs[index];
+      this.currentFile = this.currentDoc.file ?? '';
+      this.visitDoc(context);
+    });
+
+    return docs;
   }
 
   getAliases(ctx: pc.DocContext): Map<string, string> {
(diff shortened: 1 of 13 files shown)


test/import/FSHImporter.Alias.test.ts
```diff
@@ -67,9 +69,25 @@ describe('FSHImporter', () => {
...
+      const input = `
+      Profile: ObservationProfile
+      Parent: Observation
+      * code from LOINC
+      `;
+      const input2 = `
+      Alias: LOINC = http://loinc.org
+      `;
+
+      const results = importText([new FileInfo(input), new FileInfo(input2)]);
```
kjmahalingam: The odd side effect mentioned at the top level can be seen here.

We can see from the passing test that `results[0]` is able to correctly resolve the alias as we need it to.

Weirdly, though, the alias is located in `results[1].aliases` and is nowhere to be found in `results[0].aliases`. This is technically not a problem, and in fact I think is objectively correctly; however, the weird result is that you technically would have no knowledge in `results[0]` that there was any alias used at all.

Do people think that's fine?

jafeltra: I agree that this feels a little weird, but I think that you're right that it is objectively correct, so I think I'm okay with this as it is.

src/import/FileInfo.ts
```diff
@@ -0,0 +1,3 @@
+export class FileInfo {
```
cmoesel: So... thinking about what this represents -- and the fact that it might not actually come from a file (for example, in our unit tests, but also if we support a web-based IDE) -- the name `FileInfo` didn't seem _quite_ right.  But I was having trouble coming up with something better.

Then it hit me.

...

`RawFSH`.  How do you feel about renaming this to `RawFSH`?

test/import/FSHImporter.Alias.test.ts
cmoesel: I agree -- it is correct as implemented.

test/import/FSHImporter.Alias.test.ts
```diff
@@ -22,7 +22,8 @@ describe('FSHImporter', () => {
       Alias: UCUM = http://unitsofmeasure.org
       `;
 
-      const result = importText(input);
+      const results = importText([new FileInfo(input)]);
+      const result = results[0];
```
cmoesel: Seeing as this is a common pattern in the importer tests, you could simplify it by creating a helper function something like:
```ts
export function importSingleText(content string, path? string): FSHDocument {
  return importText([new FileInfo(content, path)])[0];
}
```
Then in each test, you can go back to the simpler statement:
```ts
const result = importSingleText(input);
```

kjmahalingam:
I pushed up a fix that addresses the inline comments. I'll work on the other larger refactor to make one `FSHImporter` later today or tomorrow.

src/import/importText.ts
```diff
@@ -5,20 +5,20 @@ import { DocContext } from './parserContexts';
...
 import flatMap from 'lodash/flatMap';
 
 /**
  * Parses various text strings into individual FSHDocuments.
- * @param {FileInfo[]} filesInfo - the list of FileInfo to parse into FSHDocuments
+ * @param {RawFSH[]} filesInfo - the list of RawFSH to parse into FSHDocuments
  * @returns {FSHDocument[]} - the FSH documents representing each parsed text
  */
-export function importText(filesInfo: FileInfo[]): FSHDocument[] {
+export function importText(filesInfo: RawFSH[]): FSHDocument[] {
```
cmoesel: Cool! Only other suggestion I'd make is to change the argument name now (`rawFSHes`?).

kjmahalingam:
@cmoesel , your comments should all be addressed. I also added an `importer.import()` function that takes in all `RawFSHes` in order to run its operations individually on each.

I had to remove the operation (and test) to track whether the `importer.visitDoc()` method has already been used, because that is now irrelevant. I also updated the importer to keep track of its current document and current file, since the importer is no longer associated with a specific document and file at any given time.

src/import/FSHImporter.ts
```diff
@@ -76,41 +81,60 @@ enum Flag {
...
+        }),
+        aliases => Array.from(aliases)
+      )
+    );
+
+    contexts.forEach((context, index) => {
+      this.currentDoc = docs[index];
+      this.currentFile = this.currentDoc.file ?? '';
+      this.visitDoc(context);
+    });
```
cmoesel: I doubt it will be a problem, but I would feel better if we reset `this.currentDoc` and `this.currentFile` to `null` here -- since we're done with them.

I'm probably being paranoid, but I've been bitten by something like this before -- where a value stayed in memory from a prior iteration and was used out of context.  Again, it might not even be possible here -- but I'll feel better if we just guard against it. ;-)

src/import/importText.ts
```diff
@@ -1,40 +1,14 @@
...
-  const parser = new FSHParser(tokens);
-  // @ts-ignore
-  parser.removeErrorListeners();
-  // @ts-ignore
-  parser.addErrorListener(listener);
-  // @ts-ignore
-  parser.buildParseTrees = true;
-  // @ts-ignore
-  return parser.doc() as DocContext;
+  return importer.import(rawFSHes);
```
cmoesel: Nice.  I like how you moved this to the importer.  Makes sense.

src/import/FSHImporter.ts
kjmahalingam: That's a good idea. I thought about it and initially decided not to, since they are `private`. That said, I can now imagine some situations where someone might try to overwrite it erroneously, since they can't be set as `readonly`. I'll fix it now!

src/import/importText.ts
kjmahalingam: Yeah, it was more of a necessity, now that it actually needs to be used within the importer. But it's a nice side effect!

kjmahalingam:
@cmoesel , your newest comment should be addressed; plus a bonus by resetting after each alias import as well.

PR #70, opened 2020-01-03
Fix value on Instance from Structure Definition

ngfreiter:
This addresses https://standardhealthrecord.atlassian.net/browse/CIMPL-189: Automatically set fixed values on instance from profile.

The desired behavior is that fixed values are set on an instance for one of two reasons:
*  If a top level (aka the element is not nested within some other element) element in the SD that we are making an instance of has a fixed value.
* If the direct parent of a nested element is created by the user within the instance, and the nested element has a fixed value.

For example, let's say we made a profile of Patient, https://www.hl7.org/fhir/patient.html, and `active` was fixed to `true` on this profile. Then when making an instance of Patient, `active` is auto set to true since it is top level. If on the profile of patient we set `communication.preferred` to `true`, then `communication.preferred` is only auto set to `true` on the instance if the user writes a rule on the instance that creates the `communication` object. So this would be any rule that fixes a value on `communication` or a child of `communication`.

### DIFF
--- a/src/export/InstanceExporter.ts
+++ b/src/export/InstanceExporter.ts
@@ -1,5 +1,10 @@
 import { FSHTank } from '../import/FSHTank';
-import { StructureDefinition, InstanceDefinition, ResolveFn } from '../fhirtypes';
+import {
+  StructureDefinition,
+  InstanceDefinition,
+  ResolveFn,
+  ElementDefinition
+} from '../fhirtypes';
 import { Instance } from '../fshtypes';
 import { FHIRDefinitions } from '../fhirdefs';
 import { logger } from '../utils/FSHLogger';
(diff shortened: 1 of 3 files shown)


src/export/InstanceExporter.ts
```diff
@@ -27,11 +33,63 @@ export class InstanceExporter {
...
+   * @param {InstanceDefinition} instanceDef - The InstanceDefinition to fix values on
+   */
+  private setFixedValuesForDirectChildren(
+    element: ElementDefinition,
+    existingPath: PathPart[],
+    instanceDef: InstanceDefinition
+  ) {
+    const directChildren = element
+      .children()
+      .filter(c => c.path.split('.').length === element.path.split('.').length + 1);
```
cmoesel: We may need something like this in the future.  I think it might make sense to move this into the `StructureDefinition` class.  One approach would be to add a new optional `excludeGrandchildren` (or `directOnly`) argument to the `children` function.

ngfreiter: This seemed fair so I added it

ngfreiter:
Ok, so the latest commit adds some tests to verify that we are checking that values aren't being re-fixed to a value different than that on the Structure Definition. However, we did realize that there is an issue with how values are being fixed. If an element has a parent element that is fixed via a pattern, the user can still set the child elements to a value different than that required by the pattern. This is really an issue with the `fixValue` function, which is used as a part of this PR, but we decided this is a separate issue and it will not be addressed here. I added a test that is currently skipped that should pass when we fix this issue. I'll also add this issue to our backlog.

ngfreiter:
Here is the new issue https://standardhealthrecord.atlassian.net/browse/CIMPL-200 mentioned above.

PR #73, opened 2020-01-06
Set child elements on primitives

ngfreiter:
This addresses https://standardhealthrecord.atlassian.net/browse/CIMPL-187, Support for setting IDs and extensions on primitives in instances.

Primitives in FHIR can have child elements, and when a child element is set on a primitive, this is differentiated from setting a value on the primitive by prefixing the name of the primitive with an `_`. For further documentation of how this works, see https://www.hl7.org/fhir/json.html#primitive.

### DIFF
--- a/src/fhirtypes/StructureDefinition.ts
+++ b/src/fhirtypes/StructureDefinition.ts
@@ -383,6 +383,9 @@ export class StructureDefinition {
         if (!pathPart.brackets) pathPart.brackets = [];
         pathPart.brackets.push('0');
       }
+      if (currentElement.type?.length === 1 && PRIMITIVES.includes(currentElement.type[0].code)) {
+        pathPart.primitive = true;
+      }
     }
     const clone = currentElement.clone();
     // fixValue will throw if it fails
@@ -545,6 +548,7 @@ export type StructureDefinitionContext = {
 export type PathPart = {
   base: string;
   brackets?: string[];
+  primitive?: boolean;
 };
 
 /**
@@ -595,4 +599,26 @@ const PROPS = [
   'type',
   'baseDefinition',
   'derivation'
+];
+
+const PRIMITIVES = [
+  'boolean',
+  'integer',
+  'string',
+  'decimal',
+  'uri',
+  'url',
+  'canonical',
+  'base64Binary',
+  'instant',
+  'date',
+  'dateTime',
+  'time',
+  'code',
+  'oid',
+  'id',
+  'markdown',
+  'unsignedInt',
+  'positiveInt',
+  'uuid'
 ];
--- a/src/fhirtypes/StructureDefinition.ts
+++ b/src/fhirtypes/StructureDefinition.ts
@@ -383,7 +383,12 @@ export class StructureDefinition {
         if (!pathPart.brackets) pathPart.brackets = [];
         pathPart.brackets.push('0');
       }
-      if (currentElement.type?.length === 1 && PRIMITIVES.includes(currentElement.type[0].code)) {
+      // Primitive and only primitives have a lower case first letter
+      if (
+        currentElement.type?.length === 1 &&
+        currentElement.type[0].code.charAt(0) ===
+          currentElement.type[0].code.charAt(0).toLowerCase()
+      ) {
         pathPart.primitive = true;
       }
     }
@@ -599,26 +604,4 @@ const PROPS = [
   'type',
   'baseDefinition',
   'derivation'
-];
-
-const PRIMITIVES = [
-  'boolean',
-  'integer',
-  'string',
-  'decimal',
-  'uri',
-  'url',
-  'canonical',
-  'base64Binary',
-  'instant',
-  'date',
-  'dateTime',
-  'time',
-  'code',
-  'oid',
-  'id',
-  'markdown',
-  'unsignedInt',
-  'positiveInt',
-  'uuid'
 ];
--- a/src/fhirtypes/common.ts
+++ b/src/fhirtypes/common.ts
@@ -43,12 +43,12 @@ export function setPropertyOnInstance(
         if (current[key] == null) current[key] = [];
         // If the index doesn't exist in the array, add it and lesser indices
         // Empty elements should be null, not undefined, according to https://www.hl7.org/fhir/json.html#primitive
-        let j = current[key].length;
-        for (; j < index; j++) {
-          current[key].push(null);
-        }
-        if (j === index) {
-          current[key].push({});
+        for (let j = 0; j <= index; j++) {
+          if (j < current[key].length && j === index && !current[key][index]) {
+            current[key][index] = {};
+          } else if (j >= current[key].length) {
+            current[key].push(j === index ? {} : null);
+          }
         }
         // If it isn't the last element, move on, if it is, set the value
         if (i < pathParts.length - 1) {
--- a/src/fhirtypes/common.ts
+++ b/src/fhirtypes/common.ts
@@ -44,7 +44,7 @@ export function setPropertyOnInstance(
         // If the index doesn't exist in the array, add it and lesser indices
         // Empty elements should be null, not undefined, according to https://www.hl7.org/fhir/json.html#primitive
         for (let j = 0; j <= index; j++) {
-          if (j < current[key].length && j === index && !current[key][index]) {
+          if (j < current[key].length && j === index && current[key][index] == null) {
             current[key][index] = {};
           } else if (j >= current[key].length) {
             current[key].push(j === index ? {} : null);
--- a/src/fhirtypes/common.ts
+++ b/src/fhirtypes/common.ts
@@ -32,15 +32,23 @@ export function setPropertyOnInstance(
     // instance property here
     let current: any = instance;
     for (const [i, pathPart] of pathParts.entries()) {
-      const key = pathPart.base;
+      // When a primitive has child elements, a _ is appended to the name of the primitive
+      // According to https://www.hl7.org/fhir/json.html#primitive
+      const key =
+        pathPart.primitive && i < pathParts.length - 1 ? `_${pathPart.base}` : pathPart.base;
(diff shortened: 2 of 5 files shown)


src/fhirtypes/common.ts
```diff
@@ -32,15 +32,23 @@ export function setPropertyOnInstance(
...
         // If the index doesn't exist in the array, add it and lesser indices
-        if (index >= current[key].length) {
-          current[key][index] = {};
+        // Empty elements should be null, not undefined, according to https://www.hl7.org/fhir/json.html#primitive
+        let j = current[key].length;
+        for (; j < index; j++) {
+          current[key].push(null);
+        }
+        if (j === index) {
+          current[key].push({});
```
ngfreiter: Note that in a previous PR, this loop was changed to
```
if (index >= current[key].length) {
          current[key][index] = {};
}
```
to be more compact. However, I had to revert back to using this loop, because the more compact syntax fills the array with `undefined`, while according to this documentation, https://www.hl7.org/fhir/json.html#primitive, the array needs to be filled with `null`.

test/export/InstanceExporter.test.ts
```diff
@@ -247,5 +247,28 @@ describe('InstanceExporter', () => {
...
+      fixedValRule.fixedValue = 'foo';
+      instance.rules.push(fixedValRule);
+      doc.instances.set(instance.name, instance);
+      const exported = exporter.exportInstance(instance);
+      expect(exported.address.length).toBe(1);
+      expect(exported.address[0]._line.length).toBe(2);
+      expect(exported.address[0]._line[0]).toBeNull();
+      expect(exported.address[0]._line[1].extension.length).toBe(1);
+      expect(exported.address[0]._line[1].extension[0].url).toBe('foo');
+    });
```
mint-thompson: I found an issue when taking a look at this PR. It might be more appropriately addressed in a separate task, as it isn't introduced by your changes. It has to do with setting values on array elements when the indices don't arrive in ascending order. If the following block as line 264, a `TypeError` occurs in `Object.setPropertyOnInstance (src/fhirtypes/common.ts:43:20)` when attempting to apply the rule on `address[0].line[1].extension[0].url` because `line[1]` exists, but is null.
```
      const badStuff = new FixedValueRule('address[0].line[2].extension[0].url');
      badStuff.fixedValue = 'bogus';
      instance.rules.push(badStuff);
```

test/fhirtypes/ElementDefinition.setInstancePropertyByPath.test.ts
```diff
@@ -110,5 +110,35 @@ describe('ElementDefinition', () => {
...
+      status.setInstancePropertyByPath('alias[1]', 'foo', resolve);
+      status.setInstancePropertyByPath('alias[1].id', 'bar', resolve);
+      expect(status.alias.length).toBe(2);
+      expect(status.alias[0]).toBeNull();
+      expect(status.alias[1]).toBe('foo');
+      // @ts-ignore
+      expect(status._alias[0]).toBeNull();
+      // @ts-ignore
+      expect(status._alias[1].id).toBe('bar');
+    });
```
mint-thompson: A similar `TypeError` can be created here when calling `setInstanceByPropertyPath` directly.

test/fhirtypes/StructureDefinition.test.ts
```diff
@@ -406,6 +406,36 @@ describe('StructureDefinition', () => {
...
+      observation.setInstancePropertyByPath('contextInvariant[1]', 'foo', resolve);
+      observation.setInstancePropertyByPath('contextInvariant[1].id', 'bar', resolve);
+      expect(observation.contextInvariant.length).toBe(2);
+      expect(observation.contextInvariant[0]).toBeNull();
+      expect(observation.contextInvariant[1]).toBe('foo');
+      // @ts-ignore
+      expect(observation._contextInvariant[0]).toBeNull();
+      // @ts-ignore
+      expect(observation._contextInvariant[1].id).toBe('bar');
+    });
```
mint-thompson: I didn't modify this test case, but since it also uses `setPropertyByPath`, it should be breakable in a similar way.

src/fhirtypes/StructureDefinition.ts
```diff
@@ -596,3 +600,25 @@ const PROPS = [
   'baseDefinition',
   'derivation'
 ];
+
+const PRIMITIVES = [
```
mint-thompson: If this list comes from FHIR documentation, it would be useful to link to that documentation here.

src/fhirtypes/common.ts
mint-thompson: Alas, my clever trick! But any solution that involves appending a pre-filled `null` array of the appropriate size will probably be harder to understand, so this is completely fine with me.

src/fhirtypes/StructureDefinition.ts
cmoesel: If we only use this list to check if something is a primitive, then I *think* we can just test to see if the first letter is lowercase.  I'm pretty sure that is a reliable convention for checking on primitiveness in FHIR.

If we do actually need a complete list of primitive names, then I think maybe we should consider dynamically building this list from the FHIR definitions instead.  That should be possible and it would ensure that we reflect any new/different primitives between FHIR versions.

src/fhirtypes/common.ts
```diff
@@ -32,15 +32,23 @@ export function setPropertyOnInstance(
     // instance property here
     let current: any = instance;
     for (const [i, pathPart] of pathParts.entries()) {
-      const key = pathPart.base;
+      // When a primitive has child elements, a _ is appended to the name of the primitive
+      // According to https://www.hl7.org/fhir/json.html#primitive
+      const key =
+        pathPart.primitive && i < pathParts.length - 1 ? `_${pathPart.base}` : pathPart.base;
```
cmoesel: Nice approach.

test/export/InstanceExporter.test.ts
cmoesel: Good find, @mint-thompson.  @ngfreiter -- let's make a unit test for this situation as well.

test/fhirtypes/ElementDefinition.setInstancePropertyByPath.test.ts
```diff
@@ -110,5 +110,35 @@ describe('ElementDefinition', () => {
...
       expect(status.code[1]).toEqual({ code: 'bar', system: 'http://example.com' });
     });
+
+    // Children of primitives
+    it('should set a child of a primitive instance property which has a value', () => {
+      status.setInstancePropertyByPath('short', 'foo', resolve);
+      status.setInstancePropertyByPath('short.id', 'bar', resolve);
+      expect(status.short).toBe('foo');
+      // @ts-ignore
+      expect(status._short.id).toBe('bar');
```
cmoesel: Do you think we should create new `_` properties in `StructureDefinition`/`ElementDefinition` for all primitive properties?  Then we could avoid the `@ts-ignore`?

ngfreiter: I think for now, I am going to vote no. I feel like the number of `_` elements that we would add could kinda muddy up what is right now a pretty clear representation of what can be in an ED or SD. If we find that we are referencing these `_` elements a lot in other parts of the code, and are using a lot of `@ts-ignore`, I'd be for it. But I'd prefer to push it off until we feel like we have to, unless you feel really strongly.

cmoesel: OK. I'm fine with that.  Just thought I'd ask!

ngfreiter:
The latest commit should address both of your comments. I did decide to fix the bug with array elements being added out of order, and added some tests to verify it is working. It also does look like we can check primitiveness just by looking at the first element of the type. Thanks FHIR!

src/fhirtypes/common.ts
```diff
@@ -43,12 +43,12 @@ export function setPropertyOnInstance(
...
         // If the index doesn't exist in the array, add it and lesser indices
         // Empty elements should be null, not undefined, according to https://www.hl7.org/fhir/json.html#primitive
-        let j = current[key].length;
-        for (; j < index; j++) {
-          current[key].push(null);
-        }
-        if (j === index) {
-          current[key].push({});
+        for (let j = 0; j <= index; j++) {
+          if (j < current[key].length && j === index && !current[key][index]) {
```
cmoesel: I'm trying to work out if `!current[key][index]` is a safe way to do this check or if you need to use `current[key][index] != null`...  The current form would be a problem if it's possible for `current[key][index]` to have the value `false` or `0`.  Is that possible?

Since `j < current[key].length`, then I think we know we're not at the _end_ of the path.  So we can assume more comes after -- which means `false` or `0` wouldn't be valid.  Even in the case of children on primitives, if the real primitive value was set to `false`, then `current[key][index]` still wouldn't be pointing to it, because it would be pointing to the key with the `_` prefix, right?

I think that's right -- just want to confirm with you.

cmoesel: It looks like we don't have any tests that set a real primitive value _and_ primitive children at the same time.  So, I think it would be good to actually add a test for this scenario anyway.  Basically it should be testing for this:
```
active = false
active.id = "foo"
```

ngfreiter: I think we actually could have a case where `current[key][index]` is `false`. The `j < current[key].length` is testing if we are at the end of the current array, not the end of the path. So if we had an array of boolean primitives that looked like `arr = [false, true, true]`, and we tried to set `arr[0] = true`, we would  mistakenly think `arr[0]` is not set, and set it to `{}`. Now, later on in the code, we would actually overwrite that `{}` and change it to `true`. But even though it works it isn't really "right", so I am going to change it to just `current[key][index] == null`.

ngfreiter: And I'll add that test.

cmoesel: Ah, OK.  I misread the code.  Good call.  Thanks!

PR #74, opened 2020-01-07
Unify Profile and Extension Exporters

jafeltra:
This addresses CIMPL-192: https://standardhealthrecord.atlassian.net/browse/CIMPL-192

This PR eliminates the separate profile and extension exporters as all functionality was the same between the two exporters. Additionally, having the single StructureDefinitionExporter helped to solve the issue where profiles and extensions were exported in the opposite exporter and were therefore exported multiple times. In the exporter now, we have separate arrays for profile definitions and extension definitions. When the export is finished for a single definition, we push it onto the appropriate array.

A few tests had to be changed since I removed the general `this.structDefs` array for the two separate arrays. I also added a test to check that we no longer export duplicate definitions.


src/export/StructureDefinitionExporter.ts
```diff
@@ -158,17 +158,12 @@ export class StructureDefinitionExporter {
       // Maybe it's a FSH-defined definition and not a FHIR one
     } else {
       let structDef = cloneDeep(
-        this.structDefs.find(sd => sd.name === type || sd.id === type || sd.url === type)
+        [...this.profileDefs, ...this.extensionDefs].find(
```
kjmahalingam: This isn't a breaking issue or anything, but I wonder if you think it would be reasonable to still have `this.structDefs` be a getter that returns `[...this.profileDefs, ... this.extensionDefs]`. It seems to be used in a couple places, so it might be useful to have.

src/export/StructureDefinitionExporter.ts
```diff
@@ -202,6 +197,34 @@ export class StructureDefinitionExporter {
...
+      this.profileDefs.push(structDef);
+    }
+  }
+
+  /**
+   * Exports Profiles and Extensions to StructureDefinitions
+   * @returns {AllStructureDefinitions}
+   */
+  export(): AllStructureDefinitions {
+    [...this.tank.getAllProfiles(), ...this.tank.getAllExtensions()].forEach(sd => {
```
kjmahalingam: Similar to comment on line 161. Maybe adding another function to `FSHTank` called `getAllStructureDefinitions()`.

src/export/StructureDefinitionExporter.ts
```diff
@@ -202,6 +197,34 @@ export class StructureDefinitionExporter {
...
+    });
+    return {
+      profileDefs: this.profileDefs,
+      extensionDefs: this.extensionDefs
+    };
   }
 }
+
+// Type to hold both profile and extension structure definitions
+export type AllStructureDefinitions = {
```
kjmahalingam: Nice, I like this type being added.

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -1012,4 +1012,24 @@ describe('StructureDefinitionExporter', () => {
       min: 1
     });
   });
+
+  // No duplicate structure definitions exported
+  it('should not export duplicate structure definitions', () => {
```
kjmahalingam: This is most likely me just misunderstanding something, but how exactly is this testing for duplicate structure definitions? I'm guessing that the `ContainsRule` is supposed to do something behind the scenes that adds a second copy of the profile or extension, but I don't know what that would be exactly.

src/export/StructureDefinitionExporter.ts
jafeltra: Yeah, I was kind of torn on this as well and I didn't want to maintain the three arrays, but I like the idea of having a getter or a function that does this concatenation.

src/export/StructureDefinitionExporter.ts
jafeltra: I like this idea as well, similar to the previous one.

test/export/StructureDefinitionExporter.test.ts
jafeltra: This trying to test that there are _not_ duplicate structure definitions. `ContainsRule` makes it so the profile is using the extension, I believe, which is needed to test that only one copy of the profile and extension are exported. The test sets up a profile and extension that would look like the following: 
```
Profile: Foo
Parent: Patient
* extension contains Bar 0..*

Extension: Bar
Id: bar
```

If you were to create those on master and not use the deduplication that Chris added in as a work around, the final package would include the extension twice, since it would get exported once during the profile export and once on its own. You can't exactly copy and paste this test onto master, but a very similar version of it would fail because the length of the `profileDefs` would be `2`, since it would contain the profile and the extension. The test is trying to check that now the length is `1` since the extension should only get added to the appropriate `extensionDefs` array.

If you feel like there's a clearer way to write this test, let me know, because I want to to be clear what it is actually testing. I can also add some of this explanation as a comment if it was actually helpful.

kjmahalingam: Ah, yes, this makes sense. Thank you for clarifying!

jafeltra:
@kjmahalingam the changes should be made to the two places you pointed out!

PR #75, opened 2020-01-07
Adding gitattributes

ngfreiter:
This adds a `.gitattributes` file which manages the line endings. This shouldn't really have an impact on Mac users, but for @mint-thompson and I, what this should do is ensure that `.sh` files never get converted to CRLF line endings on check out, and always stay as LF line endings. All other files will be automatically handled by `git`, which should be the same behavior as having the `core.autocrlf` configuration param set to `true`. This means that for Windows users other files will be converted to CRLF on checkout, and converted back to LF on check in.

Note that this does mean that we could still `npm publish` files with CRLF endings, since the `.js` and `.ts` files will still have CRLF endings locally on Windows. This shouldn't be a problem though, but if we would like to be ultra safe, we could make it so all files stay LF all the time. This would be possible since most text editors now can switch between the two. But, this current solution is more in line with how we have been doing things, so I went with the less drastic shift.


ngfreiter:
If @mint-thompson has some time, it would probably be helpful just to make sure what I did doesn't accidentally mess with their settings. I am not sure exactly what git caches locally, so to test this I cleanly cloned this branch, and verified that the `.sh` files have LF endings.

mint-thompson:
I checked this branch on a fresh clone of the repo, and the `*.sh` files unfortunately have the windows-style CRLF line endings.

cmoesel:
I wonder, though, if that's because .gitattributes in not in the master branch?  I don't think it will *change* line endings on existing files.  I think it only does that when creating the file on disk.  So... in your case, when you cloned it, Git _created_ the `.sh` files on disk from `master` (no gitattributes) -- hence CRLF.  When you switched branches, since those files already existed, it didn't change the line endings.

It's just a theory.  Can you try this?
* switch to Nick's branch
* delete the *.sh files
* `git checkout -- .` to get them back
* then check the line endings

mint-thompson:
Ah! That did the trick, so I'd say it's good to go on my windows setup.

PR #76, opened 2020-01-07
Set executable flags on IG bash scripts

cmoesel:
If the sushi package was built on a Windows machine, the bash scripts might not retain the proper permissions.  To ensure correct behavior, fix the permissions on the output bash scripts after they've been copied.

@ngfreiter -- can you use this branch to build a new npm package and send it to me so I can ensure it works as intended when the package has been built on a Windows system?


cmoesel:
Thanks for sending the package, @ngfreiter.  I've confirmed it works as designed.  If you get a chance, please review the actual code and approve or request changes as necessary.

PR #77, opened 2020-01-08
Export value set

mint-thompson:
This pull request adds the FHIR value set to the export capabilities. Many of the fields in the ValueSet class are currently unused, but are included so that the class matches the specifications at http://hl7.org/fhir/R4/valueset.html. The cardinality requirement of the `include` property is checked during the export process.

Additionally, there are some changes in the import process: aliased values are now handled in `from` clauses, and the presence of of a system is checked when composing from a list of concepts.

This pull request addresses https://standardhealthrecord.atlassian.net/browse/CIMPL-152.


src/export/ValueSetExporter.ts
```diff
@@ -0,0 +1,104 @@
...
+  ValueSetConceptComponent,
+  ValueSetFilterComponent,
+  FshCode
+} from '../fshtypes';
+import { logger } from '../utils/FSHLogger';
+import { ValueSetComposeError } from '../errors';
+
+export class ValueSetExporter {
+  public readonly valueSets: ValueSet[] = [];
+  constructor(public readonly FHIRDefs: FHIRDefinitions, public readonly tank: FSHTank) {}
```
ngfreiter: Does the `ValueSetExporter` actually need the `FHIRDefs`? I don't see if they get used here. Doesn't hurt much to leave them though if you can see them being needed in the future.

src/export/ValueSetExporter.ts
```diff
@@ -0,0 +1,104 @@
...
+        if (component.inclusion) {
+          valueSet.compose.include.push(composeElement);
+        } else {
+          valueSet.compose.exclude.push(composeElement);
+        }
+      });
+    }
+  }
+
+  private filterValueToString(value: string | RegExp | boolean | FshCode): string {
```
ngfreiter: I think it could be worthwhile to make a type for the `string | RegExp | boolean | FshCode` like we did for `FixedValueType`, since I think I've seen it pop up in 3 or 4 places.

src/fhirtypes/ValueSet.ts
```diff
@@ -0,0 +1,72 @@
...
+  useContext: UsageContext[];
+  jurisdiction: CodeableConcept[];
+  immutable: boolean;
+  purpose: string;
+  copyright: string;
+  compose: ValueSetLogicalDefinition;
+}
+
+export type ValueSetLogicalDefinition = {
+  lockedDate?: string;
```
cmoesel: So I'm noticing that we're a little inconsistent in using `?` to indicate optionality.  We don't use it in any of the properties directly on `ValueSet`, but we are using it here.  I think the same can be said for how we did `StructureDefinition`.

It doesn't need to be in this PR, but I think we should consider adding the optionality flags to the higher-level properties too (here and in other fhir classes we've already written).

src/fhirtypes/ValueSet.ts
```diff
@@ -0,0 +1,72 @@
...
+  inactive?: boolean;
+  include: ValueSetCompose[];
```
cmoesel: Following the convention we've used elsewhere, this should be:
```
include: ValueSetComposeInclude
```

src/fhirtypes/ValueSet.ts
```diff
@@ -0,0 +1,72 @@
...
+  exclude?: ValueSetCompose[];
```
cmoesel: Ah...  I now see the method to your madness...  Still, I like the idea of sticking to existing conventions as much as possible, so even though it may seem strange, I'd suggest this should also use the `ValueSetComposeInclude` type.  Or if that really doesn't sit well with you, we could use the convention as inspiration and call that type `ValueSetComposeIncludeOrExclude`.

src/fhirtypes/ValueSet.ts
```diff
@@ -0,0 +1,72 @@
...
+};
+
+export type ValueSetCompose = {
```
cmoesel: As noted above, rename to `ValueSetComposeInclude` -- or if you really don't like that, `ValueSetComposeIncludeOrExclude`.

src/fhirtypes/ValueSet.ts
```diff
@@ -0,0 +1,72 @@
...
+  system?: string;
+  version?: string;
+  valueSet?: string[];
+  concept?: ValueSetConcept[];
```
cmoesel: Oh man, you're going to hate me.  `concept?: ValueSetComposeIncludeConcept[]`.

??

src/fhirtypes/ValueSet.ts
```diff
@@ -0,0 +1,72 @@
...
+  filter?: ValueSetLogicalFilter[];
```
cmoesel: And `filter?: ValueSetComposeIncludeFilter[]`

src/fhirtypes/ValueSet.ts
```diff
@@ -0,0 +1,72 @@
...
+};
+
+export type ValueSetConcept = {
+  code: string;
+  display?: string;
+  designation?: ValueSetConceptDesignation[];
```
cmoesel: Oh this just keeps getting more fun, doesn't it: `designation?: ValueSetComposeIncludeConceptDesignation[]`.

src/import/FSHImporter.ts
```diff
@@ -753,46 +760,57 @@ export class FSHImporter extends FSHVisitor {
...
-        .forEach(filterDefinition => {
-          try {
-            filters.push(this.visitVsFilterDefinition(filterDefinition));
-          } catch (e) {
-            logger.error(e, {
-              location: this.extractStartStop(filterDefinition),
-              file: this.currentFile
-            });
-          }
+      if (!from.system) {
```
cmoesel: Above, you do:
```ts
if (from.system) {
  // process it...
} else {
  // log an error...
}
```
 Here you do the reverse order:
```ts
if (!from.system) {
  // log an error...
} else {
  // process it...
}
```

It's not wrong, but it feels inconsistent since they both are doing the same thing at a high level.

cmoesel: In addition, even though each error is essentially saying the same thing (that you must state the system), the two messages are worded very differently.

src/import/FSHImporter.ts
```diff
@@ -753,46 +760,57 @@ export class FSHImporter extends FSHVisitor {
...
   }
 
   visitVsComponentFrom(ctx: pc.VsComponentFromContext): ValueSetComponentFrom {
     const from: ValueSetComponentFrom = {};
     if (ctx.vsFromSystem()) {
-      from.system = ctx
-        .vsFromSystem()
-        .SEQUENCE()
-        .getText();
+      from.system = this.aliasAwareValue(
```
cmoesel: Good catch!

test/export/ValueSetExporter.test.ts
```diff
@@ -0,0 +1,229 @@
...
+    exporter = new ValueSetExporter(defs, input);
+  });
+
+  it('should output empty results with empty input', () => {
+    const exported = exporter.export();
+    expect(exported).toEqual([]);
+  });
+
+  it('should export a single value set', () => {
+    const valueSet = new FshValueSet('BreakfastVS');
```
cmoesel: Mmmmmm... Breakfast!

test/export/ValueSetExporter.test.ts
```diff
@@ -0,0 +1,229 @@
...
+    let exported = exporter.export();
+    expect(exported.length).toBe(1);
+    exported = exporter.export();
+    expect(exported.length).toBe(1);
+  });
+
+  it('should export a value set that includes a component from a system', () => {
+    const valueSet = new FshValueSet('DinnerVS');
+    const component = new ValueSetConceptComponent(true);
+    component.from = { system: 'FOOD' };
```
cmoesel: Not a real big deal, but technically the string `FOOD` is not really a valid system identifier.  This test would be more realistic if it was something like: `http://food.org/food`.

test/export/ValueSetExporter.test.ts
```diff
@@ -0,0 +1,229 @@
...
+    const exported = exporter.export();
+    expect(exported.length).toBe(1);
+    expect(exported[0].compose.include.length).toBe(1);
+    expect(exported[0].compose.include[0].system).toBe('FOOD');
+  });
+
+  it('should export a value set that includes a component from a value set', () => {
+    const valueSet = new FshValueSet('DinnerVS');
+    const component = new ValueSetConceptComponent(true);
+    component.from = { valueSets: ['HotFoodVS', 'ColdFoodVS'] };
```
cmoesel: Same as above (and I won't mention it for anything else) -- technically a string like `HotFoodVS` is not a valid VS identifier.  It would be a more realistic test if it were a URL.

test/export/ValueSetExporter.test.ts
```diff
@@ -0,0 +1,229 @@
...
+    doc.valueSets.set(valueSet.name, valueSet);
+    const exported = exporter.export();
+    expect(exported.length).toBe(1);
+    expect(exported[0].compose.include.length).toBe(1);
+    expect(exported[0].compose.include[0].system).toBe('FOOD');
+    expect(exported[0].compose.include[0].concept.length).toBe(2);
+    expect(exported[0].compose.include[0].concept[0].code).toBe('Pizza');
+    expect(exported[0].compose.include[0].concept[0].display).toBe('Delicious pizza to share.');
+    expect(exported[0].compose.include[0].concept[1].code).toBe('Salad');
+    expect(exported[0].compose.include[0].concept[1].display).toBe('Plenty of fresh vegetables.');
```
cmoesel: If it's simpler for you, it's totally valid to do an object-based equality assertion.  I.e., instead of above, do:
```ts
expect(exported[0]).toEqual({
  compose: {
    include: [{
      system: 'FOOD',
      concept: [
        { code: 'Pizza', display: 'Delicious pizza to share.' },
        { code: 'Salad', display: 'Plenty of fresh vegetables.' }
      ]
    }]
});
```

This actually has the added benefit in that if the result contains any unexpected keys (like `"foo": "bar"`), the test will fail.  That said, I'm not saying you have to do this -- just saying you _can_ do it, and it might be simpler than a bunch of long expect statements.  Up to you.

src/fhirtypes/ValueSet.ts
```diff
@@ -0,0 +1,72 @@
...
+  language?: string;
+  use?: Coding;
+  value: string;
+};
+
+export type ValueSetLogicalFilter = {
+  property: string;
+  op: string;
+  value: string;
+};
```
cmoesel: To actually write this out as a file, we might need a `toJSON()` method defined -- unless the default `toJSON` functionality works (and it might if all of your property names and cardinalities exactly match what FHIR expects).

src/errors/ValueSetComposeError.ts
```diff
@@ -0,0 +1,8 @@
+import { Annotated } from './Annotated';
+
+export class ValueSetComposeError extends Error implements Annotated {
+  specReferences = ['http://hl7.org/fhir/R4/valueset-definitions.html#ValueSet.compose.include'];
+  constructor(valueSetName: string) {
+    super(`Non-empty ValueSet ${valueSetName} must be composed from at least one inclusion.`);
```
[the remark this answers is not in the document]
mint-thompson: Your error message definitely reads a lot better than mine. I would like to still include "non-empty" or something like it in the message, because a ValueSet is allowed to be empty (`compose` has cardinality 0..1). Maybe something like: `ValueSet ${valueSetName} must contain at least one inclusion rule or be empty.`?

cmoesel:
I realized one other thing.  [ValueSet]() requires `status` (it is `1..1`).  We don't set it, but since `^` isn't supported yet, there is no way for the author to set it either.  So for now, I suggest we default `status` to `active`.

src/errors/ValueSetComposeError.ts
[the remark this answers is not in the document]
cmoesel: Oh, right.  I get it now.  So inclusion is only needed if it is not empty... So... what does that really mean?  I'm trying to understand how a FSH user will get in this situation...  Is it basically if they define an exclusion rule but not an inclusion rule?

At any rate, yes, I like your new message.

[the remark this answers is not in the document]
mint-thompson: Yes, that's the case that gets the user this error: if there is an exclusion rule, but no inclusion rules. This is the test case that demonstrates this: [log a message when a value set has a logical definition without inclusions](https://github.com/FHIR/sushi/pull/77/files#diff-b8de52dfe4a6614e7f67eca387f5f652R217).

src/export/ValueSetExporter.ts
mint-thompson: You're absolutely right. Those definitions never get used, so there's no reason to have them in here.

mint-thompson:
Thank you for your helpful and powerful review comments. The latest three commits should resolve all of the issues that have been presented thus far.

cmoesel:
One higher-level issue is when FSH value sets are defined like this:
```
ValueSet: 				YesNoUnknownVS
Description:			"A value set containing yes, no, and unknown."
* SCT#373066001			"Yes"
* SCT#373067005			"No"
* SCT#261665006			"Unknown"
```

The exporter exports it like this:
```json
"compose": {
    "include": [
      {
        "system": "http://snomed.info/sct",
        "concept": [
          {
            "code": "373066001",
            "display": "Yes"
          }
        ]
      },
      {
        "system": "http://snomed.info/sct",
        "concept": [
          {
            "code": "373067005",
            "display": "No"
          }
        ]
      },
      {
        "system": "http://snomed.info/sct",
        "concept": [
          {
            "code": "261665006",
            "display": "Unknown"
          }
        ]
      }
    ]
  }
```
Since these are all from the same system, this should be:
```
"compose": {
    "include": [
      {
        "system": "http://snomed.info/sct",
        "concept": [
          {
            "code": "373066001",
            "display": "Yes"
          },
          {
            "code": "373067005",
            "display": "No"
          },
          {
            "code": "261665006",
            "display": "Unknown"
          }
        ]
      }
    ]
  }
```
This affects the rendering of the IG, so we do want to make this change.  I suspect we can choose if we want to address it in the exporter (by checking to see if there is already an `includes` entry w/ the system and codes (without filters)) -- or during import by collecting the codes by system and grouping them together in the FshValueSet.  I'd lean toward the latter, but I think either should be fine.

cmoesel:
The second high-level issue is that since we are now exporting value sets, the profiles and extensions that bind to these value sets need to specify the value set they bind to by its URL instead of its name.  The FSH statement can (and should) still bind by name -- but the `StructureDefinition` exporter needs to turn it into a URL.

Using one of the mcode definitions as an example...  We have FSH:
```
Profile: CancerStageParent
Parent: Observation
// bunch of other stuff
* method from CancerStagingSystemVS (extensible)
```
And `CancerStagingSystemVS` is also defined in the same tank:
```
ValueSet:   CancerStagingSystemVS
Title: "Cancer Staging System Value Set"
Description:    "System used for staging. Because SNOMED CT does not currently have a code representing AJCC Version 8, specify the exact text 'AJCC Version 8' in the text sub-field of the code structure, and omit the code."
* SCT#444256004 "American Joint Commission on Cancer, Cancer Staging Manual, 6th edition neoplasm staging system (tumor staging)"
* SCT#443830009 "American Joint Commission on Cancer, Cancer Staging Manual, 7th edition neoplasm staging system (tumor staging)"
* SCT#258235000 "International Union Against Cancer (tumor staging)"
```
Currently, the exported `CancerStageParent` profile has this:
```json
      {
        "id": "Observation.method",
        "path": "Observation.method",
        // bunch of stuff
        "binding": {
          "strength": "extensible",
          "valueSet": "CancerStagingSystemVS"
        }
        // more stuff
      }
```
But that `binding` should actually be:
```
        "binding": {
          "strength": "extensible",
          "valueSet": "http://hl7.org/fhir/us/mcode/ValueSet/CancerStagingSystemVS"
        }
```
So... it seems like the StructureDefinition exporter will need to know how to resolve `CancerStagingSystemVS` to its `url` and bind to _that_.

src/ig/IGExporter.ts
```diff
@@ -183,6 +183,15 @@ export class IGExporter {
...
     });
+    this.pkg.valueSets.forEach(valueSet => {
+      const valueSetPath = path.join(igPath, 'input', 'resources', valueSet.getFileName());
+      outputJSONSync(valueSetPath, valueSet.toJSON(), { spaces: 2 });
+      this.ig.definition.resource.push({
+        reference: { reference: `ValueSet/${valueSet.id}` },
+        name: valueSet.title ?? valueSet.name,
+        description: valueSet.description
+      });
+    });
```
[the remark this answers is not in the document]
mint-thompson: Should instances also be sorted before being added to `this.ig.definition.resource`, or do only profiles, extensions, and valuesets need to be sorted?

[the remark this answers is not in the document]
cmoesel: Ah, yeah, I guess we should sort instances too?  TBH -- I'm not sure how instances are surfaced in the IG, but yeah, let's sort them.

mint-thompson:
The latest findings are addressed: value sets are sorted before being added to the IG, value set exports include the `resourceType` field, and compatible concept components get merged together.

src/ig/IGExporter.ts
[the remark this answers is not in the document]
jafeltra: Instances are added to the same folder as the other resources. I had thought about other places to potentially put them in the IG given the available folders, but I wasn't completely sure where they belonged.

mint-thompson:
@cmoesel All of your changes look fine to me. Thank you!

PR #79, opened 2020-01-09
Remove infinite loops in cases with circular dependencies

kjmahalingam:
This task removes infinite loops in the event of circular dependencies occurring. It does so by allowing a structure definition to be read even when it is incomplete, only containing its metadata. This technically leaves the possibility for a structure definition to be accessed when it is incomplete and needs more than just its metadata, but this chance is incredibly slim.

In order to test this, run the tests as usual (since I've added a couple of circular tests). In order to test an actual implementation, first run `master` of this repository against `master of `fsh-mcode`. You should either notice errors that the maximum call stack size was exceeded, or notice that the execution hangs and doesn't complete (depending on your OS, I believe). Then, run this branch of this repository against `master` of `fsh-mcode` and notice that the errors are resolved (other irrelevant errors will still remain, though) and the execution completes.

When reviewing this, please note that I added a debug statement for when elements are unfolded. That is, I believe, the only current instance of our code where the `resolve` function is used and requires more than just the metadata. If you know of other instances where this happens, please let me know, and I'll add a debug statement in those locations, as well.

### DIFF
--- a/src/export/StructureDefinitionExporter.ts
+++ b/src/export/StructureDefinitionExporter.ts
@@ -139,6 +139,8 @@ export class StructureDefinitionExporter implements Fishable {
         throw new InvalidExtensionSliceError(ext.sliceName);
       }
     });
+
+    structDef.complete = true;
   }
 
   fishForFHIR(item: string, ...types: Type[]) {
--- a/src/export/StructureDefinitionExporter.ts
+++ b/src/export/StructureDefinitionExporter.ts
@@ -189,7 +189,9 @@ export class StructureDefinitionExporter implements Fishable {
     const structDef = StructureDefinition.fromJSON(json);
     if (structDef.inProgress) {
       logger.warn(
-        `The definition of ${fshDefinition.name} may end up incomplete because its parent ${parentName} is still in progress.`
+        `The definition of ${fshDefinition.name} may be incomplete because there is a circular ` +
+          `dependency with its parent ${parentName} causing the parent to be used before the ` +
+          'parent has been fully processed.'
       );
     }
 
--- a/src/export/StructureDefinitionExporter.ts
+++ b/src/export/StructureDefinitionExporter.ts
@@ -139,8 +139,6 @@ export class StructureDefinitionExporter implements Fishable {
         throw new InvalidExtensionSliceError(ext.sliceName);
       }
     });
-
-    structDef.complete = true;
   }
 
   fishForFHIR(item: string, ...types: Type[]) {
@@ -189,8 +187,12 @@ export class StructureDefinitionExporter implements Fishable {
     }
 
     const structDef = StructureDefinition.fromJSON(json);
+    structDef.inProgress = true;
+
     this.setMetadata(structDef, fshDefinition);
 
+    // These are being pushed now in order to allow for
+    // incomplete definitions to be used to resolve circular reference issues.
     if (structDef.type === 'Extension') {
       this.pkg.extensions.push(structDef);
     } else {
@@ -199,6 +201,8 @@ export class StructureDefinitionExporter implements Fishable {
 
     this.setRules(structDef, fshDefinition);
     this.validateStructureDefinition(structDef);
+
+    structDef.inProgress = false;
 
     return structDef;
   }
--- a/src/export/StructureDefinitionExporter.ts
+++ b/src/export/StructureDefinitionExporter.ts
@@ -188,14 +188,15 @@ export class StructureDefinitionExporter implements Fishable {
 
     const structDef = StructureDefinition.fromJSON(json);
     this.setMetadata(structDef, fshDefinition);
-    this.setRules(structDef, fshDefinition);
-    this.validateStructureDefinition(structDef);
 
     if (structDef.type === 'Extension') {
       this.pkg.extensions.push(structDef);
     } else {
       this.pkg.profiles.push(structDef);
     }
+
+    this.setRules(structDef, fshDefinition);
+    this.validateStructureDefinition(structDef);
 
     return structDef;
   }
--- a/src/export/StructureDefinitionExporter.ts
+++ b/src/export/StructureDefinitionExporter.ts
@@ -187,6 +187,12 @@ export class StructureDefinitionExporter implements Fishable {
     }
 
     const structDef = StructureDefinition.fromJSON(json);
+    if (structDef.inProgress) {
+      logger.warn(
+        `The definition of ${fshDefinition.name} may end up incomplete because its parent ${parentName} is still in progress.`
+      );
+    }
+
     structDef.inProgress = true;
 
     this.setMetadata(structDef, fshDefinition);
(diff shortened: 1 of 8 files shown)


src/fhirtypes/ElementDefinition.ts
```diff
@@ -1326,6 +1327,9 @@ export class ElementDefinition {
       const type = this.type[0].profile?.[0] ?? this.type[0].code;
       const def = resolve(type);
       if (def) {
+        if (!def.complete) {
+          logger.debug(`${def.name} is incomplete and may be missing required fields.`);
```
ngfreiter: I think now that since we are using `debug`, we want a command line option to set the log level to show `debug` messages. As it stands now, I don't think there is a way to see `debug` messages without manually changing app.ts to have `logger.level = 'debug'`.

mint-thompson: Will any existing test cases execute this logging statement? If not, could you add a test case that does?

cmoesel: I'd suggest a slightly more specific log message:
```
logger.debug(`Warning: Circular relationship detected between ${this.name} and ${def.name}.  As a result, it is possible that the definition of ${this.name} may be based on incomplete components of ${def.name}.`); 
```

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -618,6 +618,110 @@ describe('StructureDefinitionExporter', () => {
...
+        'http://example.com/StructureDefinition/Bar',
+        'http://hl7.org/fhir/StructureDefinition/QuestionnaireResponse',
+        'http://hl7.org/fhir/StructureDefinition/MolecularSequence'
+      )
+    );
+
+    expect(constrainedHasMemberBar.type).toHaveLength(1);
+    expect(constrainedHasMemberBar.type[0]).toEqual(
+      new ElementDefinitionType('Reference').withTargetProfiles(
+        'http://hl7.org/fhir/StructureDefinition/Observation',
```
cmoesel: Uh oh.  I think this might be wrong...  If `Bar` has parent `Foo`, and `Foo` says `* hasMember[Observation] only Reference(Bar)` -- then shouldn't `Bar` inherit that constraint?  Meaning that the above URL should *actually* be `http://example.com/StructureDefinition/Bar`?

src/export/StructureDefinitionExporter.ts
```diff
@@ -196,14 +198,15 @@ export class StructureDefinitionExporter {
     }
```
cmoesel: GitHub won't let me comment on the line I really want to comment on, but a few lines above, there is:
```ts
const structDef = this.resolve(parentName);
```

The resulting `structDef` is what this structdef is built on (it's a cloned version of the parent).  So... if that `structDef` is incomplete, then that definitely could be (and almost certainly _would_ be) a problem.  I think we need a log statement there -- and it probably should be warn level because it's almost certainly a problem.

In fact, there is one unit test that I questioned the results on.  I wonder if it's not working as expected because of this situation.  In fact, I think maybe that *is* the case...  I'll update my other comment w/ more details.

test/export/StructureDefinitionExporter.test.ts
cmoesel: So I think I know why this happens.  Here's the breakdown:

1. First it exports `Foo`, so it starts processing `Foo`...
2. It gets to `* hasMember[Observation] only Bar` so it tries to resolve `Bar`
3. `Bar` hasn't been exported yet, so it starts processing `Bar`
4. The first thing it does when processing `Bar` is resolve its parent: `Foo`
5. It gets back the _incomplete_ `Foo`, which doesn't yet have the `hasMember` constraint applied
6. Since it uses this incomplete `Foo` as its starting point it misses the `hasMember` constraint

cmoesel: So what do we do?  Well...

`Foo` really only needed the metadata of `Bar` -- because it just needed its `url`.  The problem is that even though that is all we needed, we generated the whole `Bar` then and there and stored its results.  If we had only processed `Bar`'s metadata and didn't generate all of `Bar` then and there, we would have been fine.

So... we might be back to adding an argument to `resolve` that says `metadataOnly` -- in which case `resolve` would only process the metadata and not push `Bar` at that point...  Does that make sense?

src/fhirtypes/ElementDefinition.ts
kjmahalingam: Good catch! 
I don't think any existing test cases will log this. I will add this test case if still relevant after the work that @cmoesel is working on is complete.

src/export/StructureDefinitionExporter.ts
cmoesel: @kjmahalingam -- I still think we need a _warning_ here if the parent is in progress.

cmoesel:
Rebased on master and fixed related issues.  Also added new tests to ensure that debug messages are actually getting logged.  This required a few changes to loggerSpy and a new test helper function.

The last two commits (with me as author) are the relevant changes.  @kjmahalingam -- can you review those?

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -21,6 +21,8 @@ import {
 import { loggerSpy, TestFisher } from '../testhelpers';
 import { ElementDefinitionType } from '../../src/fhirtypes';
 import path from 'path';
+import { logger } from '../../src/utils';
```
ngfreiter: As github actions has pointed out 'logger' is defined but used. Could you remove that.

test/testhelpers/withDebugLogging.ts
```diff
@@ -0,0 +1,12 @@
+import { logger } from '../../src/utils';
+
+export function withDebugLogging(fn: () => any) {
```
kjmahalingam: I like this inclusion. Definitely helps out for potential future debug statements as well.

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -759,6 +769,75 @@ describe('StructureDefinitionExporter', () => {
...
+    extension.rules = [extOnlyRule];
+    doc.extensions.set(extension.name, extension);
+
+    withDebugLogging(() => exporter.export());
+
+    const lastLog = loggerSpy.getLastLog();
+    expect(lastLog.level).toMatch(/debug/);
+    expect(lastLog.message).toMatch(/Warning: Circular .* BarQuantity and FooQuantity/);
+
+    expect(loggerSpy.getLastMessage()).toMatch(/Warning: Circular .* BarQuantity and FooQuantity/);
```
kjmahalingam: I would have never thought to just abbreviate a test message in a regex like that. Now I know.

test/export/StructureDefinitionExporter.test.ts
cmoesel: Hey... so I have a follow-on PR to this for the IG dependencies and I actually already found that issue and fixed it in a commit on my other branch.  Can you trust me to fix it in my next PR, which will be ready 2 minutes after we merge this?

kjmahalingam: Alright @cmoesel , if you say so. :) 💯

PR #88, opened 2020-01-13
Add slicing on instances

ngfreiter:
This addresses https://standardhealthrecord.atlassian.net/browse/CIMPL-188, Support for slicing on instances.

This PR makes it so that when using the `element[sliceName]` syntax, we will select the correct slice out of the overall array that is being sliced. For example, if we were creating an instance of this SD: https://www.hl7.org/fhir/us/core/StructureDefinition-us-core-patient.html, we would expect that `extension[race].extension[ombCategory][1]` will select the second `ombCategory` in the extension array, not the second overall value in the extension array, since the extension array can also contain `detailed` or `text` slices. There is no (easy) reliable way to tell slices apart, so this was accomplished by adding a temporary `_sliceName` field on the instance, that is then stripped out before the instance is exported.

I believe I've added thorough tests, but slices and instances can be tricky, so I'd appreciate if reviewers take a good look at the tests and think about any cases I may have missed.


src/fhirtypes/common.ts
```diff
@@ -41,22 +41,44 @@ export function setPropertyOnInstance(
...
+          const sliceIndices: number[] = [];
+          // Find the indices where slices are placed
+          current[key].forEach((el: any, i: number) => {
+            if (el?._sliceName === sliceName) {
+              sliceIndices.push(i);
+            }
+          });
+          // Convert the index in terms of the slice to the corresponding index in the overall array
+          if (index >= sliceIndices.length) {
+            index = index - sliceIndices.length + current[key].length;
```
cmoesel: I'm having a little trouble following... so this may or may not be a good question, but let's assume we have:
```
foo: [
  { _sliceName: 'a', value: 'a0' },
  { _sliceName: 'a', value: 'a1' },
  { _sliceName: 'b', value: 'b0' },
  { value: 'sliceless0' }
]
```
Based on the logic above, would `foo[a][2].value = "a2"` end up overwriting the first element of slice `b`?  Don't you need to insert an element in the array between the last slice a and the first slice b?

Again, might not be a valid question.  Just want to be sure.

src/fhirtypes/common.ts
```diff
@@ -41,22 +41,44 @@ export function setPropertyOnInstance(
...
+            // _sliceName is used to later differentiate which slice an element represents
+            const arrayFiller = sliceName ? { _sliceName: sliceName } : null;
+            current[key].push(j === index ? {} : arrayFiller);
           }
         }
         // If it isn't the last element, move on, if it is, set the value
         if (i < pathParts.length - 1) {
           current = current[key][index];
+          if (sliceName) {
+            current._sliceName = sliceName;
```
cmoesel: How does this approach work if the thing we're slicing is a primitive?  Does it?

cmoesel: Actually, slicing primitives is real weird.  I'm not sure I care if it works or not.

test/export/InstanceExporter.test.ts
```diff
@@ -333,5 +341,100 @@ describe('InstanceExporter', () => {
...
+      const barRule = new FixedValueRule('extension[type][0].valueCoding.system');
+      barRule.fixedValue = 'bar';
+      patientProfInstance.rules.push(barRule);
+      const exported = exporter.exportInstance(patientProfInstance);
+      expect(exported.extension).toEqual([
+        { url: 'type', valueCoding: { system: 'bar' } },
+        { url: 'type', valueCoding: { system: 'foo' } },
+        { url: 'level', valueCoding: { system: 'baz' } }
+      ]);
+    });
```
cmoesel: This is good, but I think we need a test for the thing I asked about above... essentially adding a new element to the end of a slice in the middle of the array.  Something like:

```ts
    it('should fix mixed sliced elements in an array out of order', () => {
      const barRule = new FixedValueRule('extension[type][0].valueCoding.system');
      barRule.fixedValue = 'bar';
      patientProfInstance.rules.push(barRule);
      const bazRule = new FixedValueRule('extension[level].valueCoding.system');
      bazRule.fixedValue = 'baz';
      patientProfInstance.rules.push(bazRule);
      const fooRule = new FixedValueRule('extension[type][1].valueCoding.system');
      fooRule.fixedValue = 'foo';
      patientProfInstance.rules.push(fooRule);
      const exported = exporter.exportInstance(patientProfInstance);
      expect(exported.extension).toEqual([
        { url: 'type', valueCoding: { system: 'bar' } },
        { url: 'type', valueCoding: { system: 'foo' } },
        { url: 'level', valueCoding: { system: 'baz' } }
      ]);
    });
```

ngfreiter: I think that the test starting on line `404, "should fix mixed sliced elements in an array" tests this situation if I'm not missing something. As the test shows, the `level` slice does not get overwritten, but the second `type` slice gets added to the end of the array, not in the middle:
```
expect(exported.extension).toEqual([
        { url: 'type', valueCoding: { system: 'foo' } },
        { url: 'level', valueCoding: { system: 'baz' } },
        { url: 'type', valueCoding: { system: 'bar' } }
      ]);
```
Is that an issue? Are slices of the same type supposed to be adjacent?

cmoesel: Huh. Yeah, line 404 looks a lot like the test I just suggested.  Obviously I must have been very sorry while doing this review.  I'm sorry...  TBH, I'm not sure if slice parts need to be grouped together.  We might want to look at the slicing doc to be sure.

One thing we're not paying attention to here is if the discriminator says slices are ordered.  If they're ordered then we need to force the author to declare the slice values in order or we need to re-order them for the author.

The other thing is open/closed.  If slicing is closed then the author should not be allowed to add any elements that don't fit in a slice.  That one might be a bit harder to enforce though.  It would only work if the author used the slice naming syntax...  Should we force that for closed slicings?

ngfreiter: I don't think the slicing doc says anything explicitly, so I guess we just want to go with whatever behavior is expected. I guess it probably would make sense to have slice parts grouped together, since the definition of slicing does say "sub-list". To me that implies all the elements of the list are together, and at the very least it seems to be a best practice used in all the examples I see.

I will look at adding checking of the ordering, I'd prefer to throw an error at the author, not to re-order, since that could lead to confusion in how the FSH correlates to the SD being created.

We would have to use the explicit slice naming syntax because otherwise we would have to try to check which slice something fits right? And that would involve using the discriminator and checking fixed values and overall be pretty difficult, correct? And when you say slice naming syntax you mean the `Slice` keyword right? Assuming the answer to all those questions is yes or mostly yes, then yeah I think it makes sense to enforce that for closed slicings. But I don't know if we want to do that yet, since the `Slice` keyword doesn't exist yet, we would effectively be shutting down and closed slicings in the meantime. So until the `Slice` keyword would it make sense to just allow anything for closed?

ngfreiter:
Ok @cmoesel, I added some failing tests for the cases we discussed above. The test for slicing a primitive is notably empty, because I could not for the life of me find an example in FHIR of a sliced primitive. That doesn't mean it's impossible to test, but it will probably be a bit of a pain, and since this is a feature we aren't even totally confident we need, it didn't seem like a good use of time. There are three issues on JIRA I made to track these three features:
https://standardhealthrecord.atlassian.net/browse/CIMPL-211
https://standardhealthrecord.atlassian.net/browse/CIMPL-212
https://standardhealthrecord.atlassian.net/browse/CIMPL-213

There is also one additional change, which is a bug I noticed in `findElementByPath` while writing the skipped tests. We actually had an issue where if element A's path was contained in the path of element B, we would be unable to unfold element A. So in the situation I saw, A was `result`, and B was `resultInterpreter`, and this caused an issue. I fixed this bug and added a test showing why it was necessary.

PR #89, opened 2020-01-13
Unfold children of slices directly from SD

ngfreiter:
This addresses https://standardhealthrecord.atlassian.net/browse/CIMPL-206, unable to find inline children of slices.

When an element is sliced, and that element has children that are on the SD, we should unfold those children onto slices of the element. Previously we were unfolding the element from the FHIR definition of the type of the element. Now we will first check if the sliced element has children on the SD, and if it does, we unfold with those children. If the sliced element does not have children, then we revert to the old behavior where we unfold from the FHIR definition of the type.

If you want to test this further, checkout master on fsh-mcode https://github.com/standardhealth/fsh-mcode, but go into CancerGenomics.fsh and uncomment lines 45 through 80, which deal with setting slices on `component`. When you run SUSHI on master against fsh-mcode, you should see a bunch of errors coming from these lines. When you run SUSHI on this branch against fsh-mcode, these errors should go away, and you should see these lines accurately reflected in the output StructureDefinition-CancerGeneticVariant.json file.

### DIFF
--- a/src/fhirtypes/ElementDefinition.ts
+++ b/src/fhirtypes/ElementDefinition.ts
@@ -387,15 +387,13 @@ export class ElementDefinition {
     }
 
     // If element is a slice
-    if (this.sliceName) {
-      const slicedElement = this.structDef.elements.find(e => e.path === this.path && e.slicing);
-      if (slicedElement) {
-        // Check that slicedElement max >= new sum of mins
-        const sumOfMins = slicedElement.checkSumOfSliceMins(slicedElement.max, min - this.min);
-        // If new sum of mins > slicedElement min, increase slicedElement min
-        if (sumOfMins > slicedElement.min) {
-          slicedElement.min = sumOfMins;
-        }
+    const slicedElement = this.slicedElement();
+    if (slicedElement) {
+      // Check that slicedElement max >= new sum of mins
+      const sumOfMins = slicedElement.checkSumOfSliceMins(slicedElement.max, min - this.min);
+      // If new sum of mins > slicedElement min, increase slicedElement min
+      if (sumOfMins > slicedElement.min) {
+        slicedElement.min = sumOfMins;
       }
     }
 
@@ -411,9 +409,9 @@ export class ElementDefinition {
    * @throws {InvalidSumOfSliceMinsError} when the sum of mins of the slices exceeds max of sliced element
    */
   private checkSumOfSliceMins(newSlicedElementMax: string, sliceMinIncrease = 0) {
-    const slices = this.structDef.elements.filter(
-      e => e.id !== this.id && e.path === this.path && !e.slicing
-    );
+    const slices = this.parent()
+      .children()
+      .filter(e => e.id !== this.id && e.path === this.path && !e.slicing);
     const sumOfMins = sliceMinIncrease + slices.reduce((prev, curr) => (prev += curr.min), 0);
     if (newSlicedElementMax !== '*' && sumOfMins > parseInt(newSlicedElementMax)) {
       throw new InvalidSumOfSliceMinsError(sumOfMins, newSlicedElementMax, this.id);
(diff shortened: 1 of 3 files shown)


src/fhirtypes/ElementDefinition.ts
```diff
@@ -411,9 +409,9 @@ export class ElementDefinition {
    * @throws {InvalidSumOfSliceMinsError} when the sum of mins of the slices exceeds max of sliced element
    */
   private checkSumOfSliceMins(newSlicedElementMax: string, sliceMinIncrease = 0) {
-    const slices = this.structDef.elements.filter(
-      e => e.id !== this.id && e.path === this.path && !e.slicing
-    );
+    const slices = this.parent()
+      .children()
+      .filter(e => e.id !== this.id && e.path === this.path && !e.slicing);
```
cmoesel: I think this is just an optimization, right? Functionally the same?

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1364,18 +1372,32 @@ export class ElementDefinition {
...
+        newElements = def?.elements.slice(1).map(e => {
           const eClone = e.clone();
           eClone.id = eClone.id.replace(def.type, `${this.id}`);
           eClone.structDef = this.structDef;
           // Capture the original so that diffs only show what changed *after* unfolding
           eClone.captureOriginal();
           return eClone;
         });
+      }
+      if (newElements.length > 0) {
```
cmoesel: This changes the signature so now it might return `undefined`.  Was there any danger in the previous approach (which would return an empty list instead)?

test/fhirtypes/ElementDefinition.test.ts
```diff
@@ -319,6 +319,53 @@ describe('ElementDefinition', () => {
...
     });
 
+    it('should add children from the Structure Definition when they exist', () => {
+      const numOriginalElements = observation.elements.length;
+      const component = observation.elements.find(e => e.path === 'Observation.component');
+      component.slicing = {
+        ordered: false,
+        rules: 'open',
+        discriminator: [{ type: 'value', path: 'code' }]
+      };
```
cmoesel: BTW, `ElementDefinition` has a `sliceIt` function.  So you could have done:
```ts
component.sliceIt('value', 'code');
```

src/fhirtypes/ElementDefinition.ts
ngfreiter: Sorry I should have mentioned this actually. This is not functionally the same, there was in fact a bug that I noticed when testing on fsh-mcode. Let's say we had the following four lines.
```
* component[GeneStudied].value[x] only CodeableConcept
* component[GeneStudied].valueCodeableConcept 1..1
* component[VariationCode].value[x] only CodeableConcept
* component[VariationCode].valueCodeableConcept 1..1
```
In the last line, when we try to set the cardinality, we need to check we aren't setting the min too high, so we need to get all slices of `component[VariationCode].value[x]`. But since slices aren't reflected in the path, the previous method would get all elements with path `component.value[x]`, which includes the `component[GeneStudied].value[x]:valueCodeableConcept` slice, which we don't want. Using the new method, we only look at children of `component[VariationCode]`, so we only get the `component[VariationCode].value[x]:valueCodeableConcept` slice that we want.

src/fhirtypes/ElementDefinition.ts
ngfreiter: There is still a `return []` at the bottom of the function, it gets cut off by this view in GitHub but if you view the full file it is there.

test/fhirtypes/ElementDefinition.test.ts
ngfreiter: I wasn't sure if it was better to try to avoid depending on other functions if I could, but I can change that if you want the compactness.

cmoesel: Oh, I see.  Personally, I think it would be fine since `sliceIt` isn't terribly complicated and is well-tested elsewhere (and you're already using `slice`) -- but I understand your point, so I'm fine with it as-is.

src/fhirtypes/ElementDefinition.ts
cmoesel: Oh!  OK.  Sorry I missed that!

src/fhirtypes/ElementDefinition.ts
cmoesel: Got it.  I forgot the parent would just be the slice and not the whole array.  It makes sense to me now!

cmoesel:
@ngfreiter -- unless you're planning any further changes, I'm fine with merging this now.

ngfreiter:
Nope I wasn't thinking of adding anything else, merge away!

PR #91, opened 2020-01-14
Code System Exporter

jafeltra:
This PR adds support for exporting Code Systems (https://standardhealthrecord.atlassian.net/browse/CIMPL-198).

It supports exporting FshCodeSystems to a FHIR Code System. The CodeSystem class, similar to the ValueSet class, has many fields that are unused by FSH, but are included to match the specification at http://hl7.org/fhir/R4/codesystem.html.

### DIFF
--- a/src/export/FHIRExporter.ts
+++ b/src/export/FHIRExporter.ts
@@ -36,7 +36,7 @@ export class FHIRExporter {
     const { profileDefs, extensionDefs } = this.structureDefinitionExporter.export();
     const instanceDefs = this.instanceExporter.export();
     const valueSets = this.valueSetExporter.export();
-    const codeSystems = this.codeSystemExporter.export([]);
+    const codeSystems = this.codeSystemExporter.export();
 
     return new Package(
       profileDefs,
--- a/src/export/FHIRExporter.ts
+++ b/src/export/FHIRExporter.ts
@@ -36,7 +36,7 @@ export class FHIRExporter {
     const { profileDefs, extensionDefs } = this.structureDefinitionExporter.export();
     const instanceDefs = this.instanceExporter.export();
     const valueSets = this.valueSetExporter.export();
-    const codeSystems = this.codeSystemExporter.export();
+    const codeSystems = this.codeSystemExporter.export([]);
 
     return new Package(
       profileDefs,
--- a/src/export/FHIRExporter.ts
+++ b/src/export/FHIRExporter.ts
@@ -1,9 +1,12 @@
 import { FSHTank } from '../import/FSHTank';
 import { Package } from './Package';
-import { StructureDefinitionExporter } from './StructureDefinitionExporter';
+import {
+  CodeSystemExporter,
+  InstanceExporter,
+  StructureDefinitionExporter,
+  ValueSetExporter
+} from '.';
 import { FHIRDefinitions } from '../fhirdefs';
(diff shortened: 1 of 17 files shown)


src/export/FHIRExporter.ts
```diff
@@ -1,9 +1,12 @@
 import { FSHTank } from '../import/FSHTank';
 import { Package } from './Package';
-import { StructureDefinitionExporter } from './StructureDefinitionExporter';
+import {
+  CodeSystemExporter,
+  InstanceExporter,
+  StructureDefinitionExporter,
+  ValueSetExporter
+} from '.';
```
cmoesel: Good call.

src/fhirtypes/CodeSystem.ts
```diff
@@ -0,0 +1,106 @@
...
+  date?: string;
+  publisher?: string;
+  contact?: ContactDetail[];
+  description?: string;
+  useContext?: UsageContext[];
+  jurisdiction?: CodeableConcept[];
+  purpose?: string;
+  copyright?: string;
+  caseSensitive?: boolean;
+  valueSet?: ValueSet;
```
cmoesel: In the FHIR doc, this is `canonical(ValueSet)`, which, in fact, is just a fancy `url` (so should be a `string` here).

src/fhirtypes/CodeSystem.ts
```diff
@@ -0,0 +1,106 @@
...
+  hierarchyMeaning?: string;
+  compositional?: boolean;
+  versionNeeded?: boolean;
+  content = 'complete';
+  supplements?: CodeSystem;
```
cmoesel: This is actually `canoncial(CodeSystem)`, so, like above, we should just count it as a `string`.

test/export/CodeSystemExporter.test.ts
```diff
@@ -0,0 +1,102 @@
...
+      id: 'MyCodeSystem',
+      status: 'active',
+      content: 'complete',
+      url: 'http://example.com/CodeSystem/MyCodeSystem'
+    });
+  });
+
+  it('should export a code system with additional metadata', () => {
+    const codeSystem = new FshCodeSystem('MyCodeSystem');
+    codeSystem.title = 'My Fancy Code System';
```
cmoesel: Can you also set `id` to something different here?  I'm sure it works, but it's a common use case so let's reflect it in the tests.

test/export/CodeSystemExporter.test.ts
```diff
@@ -0,0 +1,102 @@
...
+      url: 'http://example.com/CodeSystem/MyCodeSystem',
+      concept: [
+        {
+          code: 'myCode',
+          display: 'My code',
+          definition: 'This is the formal definition of my code'
+        }
+      ]
+    });
+  });
```
cmoesel: Can we either make the above two tests each export _two_ concepts or create a new test with multiple concepts?  Usually there will be more than one, and while I'm sure you got it right, it's good to ensure in a test.

src/fhirtypes/CodeSystem.ts
jafeltra: Thanks for catching this! (And the similar one below)

jafeltra:
The latest changes should address your comments @cmoesel. I also rebased off master again.

Issue #92, opened 2020-01-15
Instances not correctly written to ImplementationGuide JSON


cmoesel:
The references to instances in the generated ImplementationGuide.json is incorrect.  For example, I have a project with an instance of Patient.  When I run sushi to generate an IG, I find this in the generated ImplementationGuide JSON:

```json
        {
        "reference": {
          "reference": "Patient/undefined"
        },
        "name": "Patient-undefined",
        "exampleBoolean": true
      }
```

I've attached a sample project that displays this issue using the version of SUSHI on `master`.

[simple-ig.zip](https://github.com/FHIR/sushi/files/4065357/simple-ig.zip)

kjmahalingam:
Fixed in 0.6.0

PR #95, opened 2020-01-15
Fix undefined references to instances in ImplementationGuide.json

ngfreiter:
This addresses https://github.com/FHIR/sushi/issues/92.

The toJSON method on instances strips out `instanceName`, and we were then trying to use that later on, so we were getting `undefined`. This fixes that issue.

### DIFF
--- a/src/ig/IGExporter.ts
+++ b/src/ig/IGExporter.ts
@@ -173,13 +173,15 @@ export class IGExporter {
     });
     sortBy(this.pkg.instances, instance => instance.id ?? instance.instanceName).forEach(
       instance => {
-        const instancePath = path.join(igPath, 'input', 'resources', instance.getFileName());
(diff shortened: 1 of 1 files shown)


cmoesel:
Oh.  So that's it.  It turns out I already fixed this in my branch.  Just in a different way (and without realizing it was the culprit behind this bug): https://github.com/FHIR/sushi/compare/fishable#diff-a5c53134bbb898a8c42d991113110dad

cmoesel:
So do we want to still merge this or just wait for my PR?

ngfreiter:
Let's go ahead and wait for your PR then, I think it's cleaner that way.

cmoesel:
I tested and confirmed that my new PR (#97) address this, so I'm going to close this one without merging.

PR #96, opened 2020-01-15
Add meta.profile to Instance

ngfreiter:
This addresses https://standardhealthrecord.atlassian.net/browse/CIMPL-210.

An instance will now have a meta.profile field, which will contain the name of the profile that we are making an instance of.

### DIFF
--- a/src/export/InstanceExporter.ts
+++ b/src/export/InstanceExporter.ts
@@ -119,9 +119,11 @@ export class InstanceExporter {
     instanceDef.resourceType = instanceOfStructureDefinition.type; // ResourceType is determined by the StructureDefinition of the type
     instanceDef.instanceName = fshDefinition.id; // This is name of the instance in the FSH
 
-    // Add the profile we are making an instance of to meta.profile
+    // Add the SD we are making an instance of to meta.profile, as long as SD is not a base FHIR resource
     // If we end up adding more metadata, we should wrap this in a setMetadata function
-    instanceDef.meta = { profile: [instanceOfStructureDefinition.url] };
+    if (instanceOfStructureDefinition.derivation === 'constraint') {
+      instanceDef.meta = { profile: [instanceOfStructureDefinition.url] };
+    }
     // Set Fixed values based on the FSH rules and the Structure Definition
     instanceDef = this.setFixedValues(fshDefinition, instanceDef, instanceOfStructureDefinition);
 
--- a/src/export/InstanceExporter.ts
+++ b/src/export/InstanceExporter.ts
@@ -119,6 +119,9 @@ export class InstanceExporter {
     instanceDef.resourceType = instanceOfStructureDefinition.type; // ResourceType is determined by the StructureDefinition of the type
     instanceDef.instanceName = fshDefinition.id; // This is name of the instance in the FSH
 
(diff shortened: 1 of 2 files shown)


src/export/InstanceExporter.ts
```diff
@@ -119,6 +119,9 @@ export class InstanceExporter {
     instanceDef.resourceType = instanceOfStructureDefinition.type; // ResourceType is determined by the StructureDefinition of the type
     instanceDef.instanceName = fshDefinition.id; // This is name of the instance in the FSH
 
+    // Add the profile we are making an instance of to meta.profile
+    // If we end up adding more metadata, we should wrap this in a setMetadata function
+    instanceDef.meta = { profile: [instanceOfStructureDefinition.url] };
```
cmoesel: You should only do this if it is indeed an instance of a _profile_ (not a base _resource_).  In other words, if I do an instance of plain old `Patient`, then I should _not_ have the `meta.profile` set.

ngfreiter: This is my way of checking if something is a base FHIR resource:
```
if (
      instanceOfStructureDefinition.url !==
      `http://hl7.org/fhir/StructureDefinition/${instanceDef.resourceType}`
    ) 
```
But I want to double check before I push that there isn't some obvious `isABaseResource` flag that I am missing.

cmoesel: I think the more formally correct way is to check that `instanceOfStructureDefinition.derivation === 'constraint'`.

src/export/InstanceExporter.ts
```diff
@@ -119,6 +119,11 @@ export class InstanceExporter {
     instanceDef.resourceType = instanceOfStructureDefinition.type; // ResourceType is determined by the StructureDefinition of the type
     instanceDef.instanceName = fshDefinition.id; // This is name of the instance in the FSH
 
+    // Add the SD we are making an instance of to meta.profile, as long as SD is not a base FHIR resource
+    // If we end up adding more metadata, we should wrap this in a setMetadata function
+    if (instanceOfStructureDefinition.derivation === 'constraint') {
+      instanceDef.meta = { profile: [instanceOfStructureDefinition.url] };
```
cmoesel: So, ideally I would give you all the feedback at the same time.  But... I missed something last time.

It's totally possible that the author might set `meta.profile` themselves, either with this profile or with other profiles as well (it's actually possible to conform to multiple profiles at once).  In that case, we should not _overwrite_ `meta.profile`.

So I think we need to check if it exists:
* if it doesn't, set it as above
* if it does, check if it already contains the profile (in which case we do nothing) or if we need to push it.

ngfreiter: I'm not sure I understand where `meta.profile` would be set from. The `instanceDef` gets created on line 118, so don't we know that `meta.profile` isn't set?

I think it's possible someone could have rules on the instance that do set the `meta.profile`, but I think the `setFixedValues` function should handle that, and overwrite this setting of `meta.profile` if necessary.

cmoesel: Oh...  right.  For some reason I thought this was happening closer to the tail end.  Thanks for keeping me honest.

PR #97, opened 2020-01-16
Replace ResolveFn w/ a MasterFisher and Fishables

cmoesel:
**THIS PR SHOULD GET AT LEAST TWO REVIEWERS**

The primary purpose of this PR is to replace the former `ResolveFn` with a more flexible mechanism around a `Fishable` interface.  Things that are `Fishable` expose two functions: `fishForFHIR` and `fishForMetadata`.  The former will attempt to get the full FHIR definition (as JSON); the latter will get just core metadata (useful when you just need an id or url and do not want to fully export the logic yet).  This mechanism reduces circular dependency issues.  In addition, the new `Fishable` interface allows the caller to indicate what types to fish for -- allowing us to ensure that if we want a `ValueSet` we don't get back a `CodeSystem` with the same id!

There are a few things that now implement the `Fishable` interface, namely:
* `FSHTank` - allowing metadata to be retrieved from the tank (but no FHIR)
* `FHIRDefinitions` - allowing FHIR JSON to be retrieved from the FHIR cache
* `Package` - allowing artifacts that have been exported by SUSHI to be found
In addition, I've created a `MasterFisher` that uses all three of the above to catch the _best_ fish.  In the tests, we have a `TestFisher` that uses the test cache instead of the full FHIR cache -- and copies things over from the FHIR cache to the test cache when they're not found in the test cache (heavily re-using Nick's code there).

I've also made some changes to the exporters to better support this approach.  In short, every exporter is initialized with a `FSHTank` (to export) and a `Package` (to export into).  So, a single `Package` is made _before_ we create the exporters and then passed into the exporters.  Some of the exporters also take the `MasterFisher`, but the VS and CS ones don't yet (as they didn't use `resolve` either).  The `StructureDefinitionExporter` is a little special, in that it takes a `MasterFisher`, but then it wraps it with itself and implements `Fishable` on its own.  This is so that when someone requests the full FHIR JSON for a profile/extension that hasn't been generated yet, the SDExporter can determine that is the case and launch an export of the requested item so it can return the FHIR to the caller.

You can test this by running the unit tests, but you can also run against your favorite IG or `fsh-mcode` (which is your favorite IG anyway, right?).  If you compare the output of running `master` against `fsh-mcode` vs running this branch against `fsh-mcode`, you'll notice some differences.  Luckily, they are all for the good; it turns out this fixes bugs we didn't even know we had!

There's a lot here.  About half of it is tests.  If you have any questions or run into any issues, please let me know.


test/fhirdefs/FHIRDefinitions.test.ts
```diff
@@ -0,0 +1,592 @@
...
+    loadFromPath(
+      path.join(__dirname, '..', 'testhelpers', 'testdefs', 'package'),
+      'test#1.1.1',
+      defs
+    );
+  });
+
+  describe('#fishForFHIR()', () => {
+    it('should find base FHIR resources', () => {
+      const conditionByID = defs.fishForFHIR('Condition', Type.Resource);
```
ngfreiter: So this file will fail if you delete the definitions from the `testdefs/package` folder, but I guess there is probably really no way around that, since we are trying to test the fishing we don't want to use `TestFisher`. It's probably not a big deal overall, but it is something to be aware of.

test/fhirdefs/load.test.ts
```diff
@@ -1,113 +1,59 @@
...
       defs
     );
-    resolve = getResolver(defs);
-    resolve('Condition');
-    resolve('boolean');
-    resolve('Address');
-    resolve('patient-mothersMaidenName');
-    resolve('allergyintolerance-clinical');
+    // Run the dependency resources through TestFisher to force them into the testhelpers cache
+    const fisher = new TestFisher().withFHIR(defs);
```
ngfreiter: I guess in the `FHIRDefinitions.test.ts` file you could maybe do something like this to ensure that the definitions you are going to use will be in the cache? But I don't know I don't think it is the cleanest solution, and wouldn't want to hold this up over something that'll only be an issue if we someday switch to R5 or something.

src/import/FSHTank.ts
```diff
@@ -107,4 +71,96 @@ export class FSHTank {
...
+      return meta;
+    }
+    return;
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  fishForFHIR(item: string, ...types: Type[]): any | undefined {
+    // the FSHTank cannot return FHIR definitions, but we define this function
+    // in order to implement the Fishable interface
+  }
```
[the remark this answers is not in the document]
cmoesel: I think I prefer to go ahead and leave the parameters in.  I'm surprised that typescript still considers it as implementing `Fishable` when the function signature changed.  But let's just keep it as-is in case that's a bug in TypeScript. ;-)

src/export/Package.ts
```diff
@@ -1,13 +1,77 @@
...
+      const metadata: Metadata = {
+        id: result.id,
+        name: result instanceof InstanceDefinition ? result.instanceName : result.name,
+        url: result.url
+      };
+      if (result instanceof StructureDefinition) {
+        metadata.sdType = result.type;
+        metadata.parent = result.baseDefinition;
+      } else if (result instanceof InstanceDefinition) {
+        if (result.meta && result.meta.profile && result.meta.profile.length > 0) {
```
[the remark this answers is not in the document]
cmoesel: Oh yeah.  Sometimes old habits die hard.  Will do.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1387,15 +1396,24 @@ export class ElementDefinition {
...
-        const def = resolve(type);
-        newElements = def?.elements.slice(1).map(e => {
-          const eClone = e.clone();
-          eClone.id = eClone.id.replace(def.type, `${this.id}`);
-          eClone.structDef = this.structDef;
-          // Capture the original so that diffs only show what changed *after* unfolding
-          eClone.captureOriginal();
-          return eClone;
-        });
+        const json = fisher.fishForFHIR(
```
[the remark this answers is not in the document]
cmoesel: I think this is right, because based on the line above it, `type` could be referring to `this.type[0].code` or `this.type[0].profile`.  If it is `this.type[0].code` then it is most-likely a Type (e.g., `Quantity`) but _could_ be a Resource (there are a few rare cases of inline resources).  If it is `this.type[0].profile`, it could be a Profile (on a Type or Resource) or a specific Extension (which is more or less a profile on `Extension`).  So I think we need to support all four of those types.  I guess we just don't have tests that use inline resources or profiles in this spot.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1387,15 +1396,24 @@ export class ElementDefinition {
       if (newElements.length === 0) {
         // If it has a profile, use that, otherwise use the code
         const type = this.type[0].profile?.[0] ?? this.type[0].code;
```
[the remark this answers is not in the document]
cmoesel: I actually assumed it _was_ me.  And I think it is correct.  It's doing what the comment says -- if there is a `profile` defined, prefer that -- otherwise just take the `code`.  We always want to unfold the most specific representation, which in this case would be the profile if it is specified.

We have the `?.[0]` because it's possible that the bit before that is null, so it's null-safe -- and that's the way you do null-safe access to an indexer in TypeScript.

src/export/StructureDefinitionExporter.ts
```diff
@@ -186,51 +172,48 @@ export class StructureDefinitionExporter {
...
-    const structDef = this.resolve(parentName);
+    const json = this.fishForFHIR(parentName);
 
     // If we still don't have a resolution, then it's not defined
-    if (!structDef) {
+    if (!json) {
       throw new ParentNotDefinedError(fshDefinition.name, parentName, fshDefinition.sourceInfo);
     }
 
+    const structDef = StructureDefinition.fromJSON(json);
```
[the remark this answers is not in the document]
cmoesel: Yeah, so it's essentially doing what you say above (re: Profile A and Profile B).  There's a few reasons for that.

(1) I wanted a consistent approach to the `fishForFHIR` function.  The `fishForFHIR` function can get a result from the FHIRDefinitions (which is straight JSON) or from the Package (which is an instance of StructureDefinition).  For consistency, regardless of the source, the returned type should be the same -- so I had to decide between instances of `StructureDefinition`, `ValueSet`, and `CodeSystem`; or plain JSON for everything.  The problem with using instances of the specific classes is that `ValueSet` and `CodeSystem` don't have `fromJSON` functions yet, so when we get one of those from FHIRDefinitions we don't have a way to transform it into an instance of its respective class.  I could have implemented that, but it seemed low value.  It's also possible that the caller really only needs the JSON anyway, so half the time it might be useless.  So I decided just to have `fishForFHIR` always return JSON -- and here we are.

This actually also unintentionally fixed a bug that I didn't realize we had.  Before, if we started w/ the parent's `StructureDefinition`, we weren't recapturing the original, so the child ended up thinking that changes that the parent made were actually its own changes and putting them in its own diff.  You can see that if you compare an export of fsh-mcode from master to an export from this branch.

test/fhirdefs/FHIRDefinitions.test.ts
cmoesel: Oh right.  We could do like we did in the load test file and create a `TestFisher` in the setup and just manually fish for all the things we know we'll need.  I missed that here.

test/fhirdefs/load.test.ts
cmoesel: I should have read ahead.  Yeah, I noted this as a possible approach when replying to your last comment.

src/export/CodeSystemExporter.ts
```diff
@@ -28,24 +28,25 @@ export class CodeSystemExporter {
...
+  exportCodeSystem(fshDefinition: FshCodeSystem): CodeSystem {
+    if (this.pkg.codeSystems.some(cs => cs.name === fshDefinition.name)) {
       return;
     }
     const codeSystem = new CodeSystem();
     this.setMetadata(codeSystem, fshDefinition);
     this.setConcepts(codeSystem, fshDefinition);
-    this.codeSystems.push(codeSystem);
+    this.pkg.codeSystems.push(codeSystem);
+    return codeSystem;
```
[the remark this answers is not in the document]
cmoesel: Yeah.  I was torn on this one.  In the end, I decided they should be consistent.  And it was easier to add the return statement to these than to remove it from the SDExporter since there were already a bunch of tests that relied on that capability of the SDExporter.

TBH -- I'm still not 100% happy w/ how the exporters all work... but I think we're closer.  Still... they may get some more refactoring at some point.

src/export/Package.ts
```diff
@@ -1,13 +1,77 @@
...
+          metadata.instanceOf = result.meta.profile[0];
```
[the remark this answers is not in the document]
cmoesel: Well... I think any one of them would do... but maybe it would actually be more correct to specifically grab `instanceOf` from the `tank` definition instead.  That's more deterministic.  That said, I'm not even sure we're using this.  It may have been over-agressive of me to even make it a part of `Metadata`.   Hmmm...

src/export/StructureDefinitionExporter.ts
```diff
@@ -12,23 +11,23 @@ import {
...
 /**
  * The StructureDefinitionExporter is the class for exporting Profiles and Extensions.
  * The operations and structure of both exporters are very similar, so they currently share an exporter.
  */
-export class StructureDefinitionExporter {
-  public readonly profileDefs: StructureDefinition[] = [];
-  public readonly extensionDefs: StructureDefinition[] = [];
+export class StructureDefinitionExporter implements Fishable {
+  public readonly profileDefs: StructureDefinition[];
+  public readonly extensionDefs: StructureDefinition[];
```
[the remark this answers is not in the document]
cmoesel: No!  They aren't!  I kept them in at first while I was refactoring in order to keep a billion tests from failing -- but I forgot to remove them once the refactoring was done.  Thanks!

cmoesel:
Dunno why CI is failing, but maybe it will be happier on my next commit.  I'll also rebase.

src/export/CodeSystemExporter.ts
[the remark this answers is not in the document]
jafeltra: Okay! I just wanted to be sure. Yeah I agree it's definitely not worth changing all the tests for SDExporter, so I'm good with this change.

src/fhirtypes/StructureDefinition.ts
```diff
@@ -85,12 +86,14 @@ export class StructureDefinition {
...
+   * @param {Fishable} fisher - A fishable implementation for finding definitions and metadata
    * @returns {StructureDefinition} the StructureDefinition of StructureDefinition
    */
-  getOwnStructureDefinition(resolve: ResolveFn = () => undefined): StructureDefinition {
+  getOwnStructureDefinition(fisher: Fishable): StructureDefinition {
     if (this._sdStructureDefinition == null) {
-      this._sdStructureDefinition = resolve('StructureDefinition');
+      this._sdStructureDefinition = StructureDefinition.fromJSON(
+        fisher.fishForFHIR('StructureDefinition', Type.Resource)
+      );
```
mint-thompson: Similar to the case in `ElementDefinition`, if somehow this isn't found, a TypeError gets thrown.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -215,12 +216,14 @@ export class ElementDefinition {
...
+   * @param {Fishable} fisher - A fishable implementation for finding definitions and metadata
    * @returns {StructureDefinition} the StructureDefinition of ElementDefinition
    */
-  getOwnStructureDefinition(resolve: ResolveFn = () => undefined): StructureDefinition {
+  getOwnStructureDefinition(fisher: Fishable): StructureDefinition {
     if (this._edStructureDefinition == null) {
-      this._edStructureDefinition = resolve('ElementDefinition');
+      this._edStructureDefinition = StructureDefinition.fromJSON(
+        fisher.fishForFHIR('ElementDefinition', Type.Type)
+      );
```
mint-thompson: This might be a bit of a stretch, but if somehow there isn't anything named `ElementDefinition` available, `fromJSON` will receive `undefined` and throw a `TypeError`. While I recognize this is colossally unlikely to occur, is this the first place the processing would break when `ElementDefinition` doesn't exist?

src/utils/MasterFisher.ts
```diff
@@ -0,0 +1,104 @@
...
+  fishForFHIR(item: string, ...types: Type[]): any | undefined {
+    // Resolve the alias if necessary
+    item = this.tank?.resolveAlias(item) ?? item;
+
+    // First check for it in the package
+    let result = this.pkg?.fishForFHIR(item, ...types);
+    if (result == null) {
+      // If it is in the tank, return undefined. We don't want to return the external FHIR
+      // definition, even if it exists -- because it won't match what is in the tank.  This
+      // ensures consistency between the outputs of fishForFHIR and fishForMetadata.
```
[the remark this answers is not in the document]
cmoesel: Yes, that is exactly right!  The second bullet ensures we avoid the case where there is something in the tank with a name or id that matches something in FHIR.  For example, in mCODE, there is a `laterality` extension that mCODE defines.  As it happens, FHIR core also has a `laterality` extension.  If we fish for `laterality` but we haven't exported it yet, we _don't_ want to return FHIR core `laterality` because that's not the right `laterality`.

src/fhirtypes/ElementDefinition.ts
cmoesel: I'm not sure if this is the first place, but it would be a fatal error.  We absolutely need it.  Since it is part of the core FHIR spec, it's very unlikely it's not there.  The only case that might happen is if someone manually mucked with their `./~fhir` cache.  I think if they didn't have the FHIR core spec at all, we would fail way before this.  So... I'm not too concerned.

src/fhirtypes/StructureDefinition.ts
cmoesel: The chances of this happening are extremely small -- and it would be fatal in any case.  So... I'm not terribly concerned.  That said, I have considered if maybe we should make `fromJSON` return `undefined`/`null` when `undefined`/`null` is passed in.  What do you think?  It would still fail somewhere -- just not here.

cmoesel:
I just pushed three new commits.  One addresses all of the PR comments (that needed to be addressed).  The other two address limitations I forgot about -- they make it so that value set bindings can properly reference FSH-authored value sets and so that value set definitions can properly reference FSH-authored value sets and code systems.

cmoesel:
But I forgot to rebase.  I'll do that now and push again.

src/fhirtypes/StructureDefinition.ts
mint-thompson: If `fromJSON` returns `undefined`/`null`, the TypeError just gets moved up the stack to `setPropertyOnDefinitionInstance`. Since it's basically impossible for anything useful to be done when `StructureDefinition` is missing, it might be reasonable to have a specific message shown when it isn't found. I think that a case of missing `ElementDefinition` could be handled similarly. If I understand this situation correctly, the most probable causes would be that FHIR wasn't specified as a dependency, or that somehow it got removed from the cache? So a possible message could mention something like checking to make sure that your dependencies in `package.json` are correct.

cmoesel:
Ugh.  Lots of broken tests after rebasing.  Weird.  Will have to figure it out.

src/fhirtypes/StructureDefinition.ts
cmoesel: If we don't do so already, we really ought to log an error and fail fast earlier in the process if the user hasn't declared a dependency on FHIR core.  We could also do a spot check for `StructureDefinition` and `ElementDefinition` to ensure that the local FHIR installation isn't horribly wrong.  I'd suggest that as a separate task though.

cmoesel: I'll put it on our backlog.

cmoesel:
OK -- rebased and fixed.  For those of you who already reviewed, the last four commits are new since then.

src/utils/MasterFisher.ts
```diff
@@ -0,0 +1,104 @@
...
+    // Resolve the alias if necessary
+    item = this.tank?.resolveAlias(item) ?? item;
+
+    const fishables = [this.pkg, this.tank, this.fhir].filter(f => f != null);
+    for (const fishable of fishables) {
+      const result = fishable.fishForMetadata(item, ...types);
+      if (result) {
+        // If it came from the tank, we need to get the sdType because the tank doesn't know.
+        if (fishable instanceof FSHTank) {
+          result.sdType = this.findSdType(result, types, fishables);
```
ngfreiter: I tried deleting all of the cached definitions in `testdefs/package`, and there is still one test that fails. The test is 'should apply a correct OnlyRule on a non-reference FSHy choice' in StructureDefinitionExporter.test.ts. If you want to fix the test itself, you can add a `fisher.fishForStructureDefinition('Quantity');` line to it before exporting in that test.

But the test led me to here. In the case of the test, `MySpecialQuantity` is a child of `Quantity`, but `Quantity` is not present in the definitions, and we are constraining a `value[x]` element to only be `MySpecialQuantity`. Since `Quantity` is not defined, `findSdType` returns `undefined`, and `result` is returned with `result.sdType = undefined`. So I wonder if we should still return `result` if `fishable` is a `FSHTank` but we don't actually find an `sdType`.

cmoesel: Thanks for the thorough review, @ngfreiter!  In the case exercised by the test, if this were a real run of sushi, there would be much more troublesome errors than a null `sdType` -- as the export of `MySpecialQuantity` would fail completely (since its parent is missing).  In general, I think the only way this will happen is when there are FSHy profiles/extensions that have missing parents -- which will always present itself as a bigger problem.

That being the case, it's not possible that a null `sdType` would somehow go unnoticed -- there will always be bigger/badder errors alongside it; and once the author fixes those the `sdType` issue is also fixed.  So... I think I'm ok with it as-is, which is basically a "best effort" / "return as much as you can" approach.  It's actually consistent with our overall approach which is that even if we run into an error in processing a rule on a profile, we don't toss the whole profile -- we do our best to export as much of it as we can.

Don't let me fool you though.  It's not like I thought of this specific case and intentionally designed it that way -- so it's good that you brought it up and forced the conversation.  Let me know if you disagree with my thoughts on it.

ngfreiter: Yup that's totally fine by me! I can't imagine how it would be an issue without other errors finding it, so I'm good.

cmoesel: I just pushed up a commit to add the fishing statement.

Issue #104, opened 2020-01-18
Sushi 0.6.0 should put generated SDs and other artifacts in /build/input/resources by default


markkramerus:
When no `ig-data` directory is present, SUSHI 0.6.0 puts the generated artifacts directly into the /build directory. The IG publisher can't find them there.  The artifacts need to always go in  /build/input/resources directory. 

The workaround is to use the `-o /build/input/resources` option.

Note: This only happens when the `ig-data` directory does not exist, but that is a significant use case for when IG customizations are to be preserved.

cmoesel:
This seems to me like asking for SUSHI to change behavior to support what is essentially a workaround because of current (but hopefully short-lived) limitations in SUSHI's IG generation.

The current behavior is by design, as it's assumed that if you don't want to generate an IG, then you don't want to have to dig into the nested structure that only IGs require.

We can change it, but is this really the solution we want, or do we want SUSHI to play better with customized IGs without all the current hoops you still have to jump through when doing it the way that is proposed above?

cmoesel:
It's just that for people not interested in generating IGs, having output go to _build/**input**/resoures_ is really confusing!

cmoesel:
OK.  After discussion, we'll do this as suggested.

ngfreiter:
This was addressed by #109

cmoesel:
Re-opening until this is in an official release.

cmoesel:
Fixed in 0.6.1

PR #109, opened 2020-01-21
Place output in build/input/resources by default

ngfreiter:
This addresses https://standardhealthrecord.atlassian.net/browse/CIMPL-229, which is also #104 

The old default was to place output in a `build` folder, but we have decided to place output by default in `build/input/resources` to make it work better with the IG publisher.


ngfreiter:
Good point, I fixed that in the latest commit and rebased.

Issue #103, opened 2020-01-18
Instances (Examples) must have Ids


markkramerus:
IG publisher won't process an example instance that doesn't have an id. The id should be the same as the name of example, by default. FSH doesn't accept an Id: keywork on example instances. **Note:** This is an mCODE show-stopper.

cmoesel:
More discussion on this here: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Instance.20generation/near/186185641

As noted, the intent, at one point, was to have the name serve as the _id_.  That intent got lost somewhere along the way (sorry).  For now, you can workaround by using the normal instance approach to setting an instance field:
```
* id = "MyID"
```

But we will fix it so that the _name_ is re-used as the `id`.

markkramerus:
Implementing the workaround of adding IDs through rules.

cmoesel:
Fixed in 0.6.1

PR #111, opened 2020-01-21
Instance ids set by default

jafeltra:
This PR addresses Github issue #103.

It defaults the `id` to the name specified in the FSH definition (the id in the FSH definition is already being defaulted to the name). If the id is set by a rule in the FSH file, it is overwritten in the exported instance. There are two tests to test both these cases.


jafeltra:
Ah, that is the way I probably should have done this. I tried to do kind of the opposite by comparing it to an instances of InstanceDefinition, but I'll switch it now.

jafeltra:
Last commit should address your comment @cmoesel. The "whole exported instance" is pretty minimal, but it is comparing to that simple JSON, so if we add an extra field at the top level later on, those two tests should fail.

Issue #124, opened 2020-01-22
Conflicting Aliases are not reported


cmoesel:
If an author defines the same `Alias` with two different values, only one of them will be used -- but there is no error indicating what happened.  We should log an error when we encounter a duplicate Alias definition with a different value than the previous one.

This came up in fsh-mcode.  In `CodeSystemAliases.fsh`, it had:
```
Alias:   USCoreRace = http://hl7.org/fhir/us/core/ValueSet/omb-race-category
```
But in `USCoreAliases` and `Examples.fsh`, it had:
```
Alias:   USCoreRace = http://hl7.org/fhir/us/core/StructureDefinition/us-core-race
```

I noticed this when reviewing this line in `Examples.fsh`, where the same alias is used twice:  once to replace an Extension URL and once to replace a VS URL:
```
* extension[USCoreRace].extension[ombCategory].valueCoding = USCoreRace#2186-5 "Not Hispanic or Latino"
```

cmoesel:
Fixed in 0.6.1

PR #125, opened 2020-01-22
Report duplicate aliases w/ different values

cmoesel:
If a FSHTank contains duplicate aliases with different values, only keep the first one and report the others as errors.  Addresses #124.

### DIFF
--- a/src/import/FSHImporter.ts
+++ b/src/import/FSHImporter.ts
@@ -45,7 +45,6 @@ import {
   ValueSetFilterValueTypeError,
   ValueSetFilterMissingValueError
 } from '../errors';
-import flatMap from 'lodash/flatMap';
 import isEqual from 'lodash/isEqual';
 import sortBy from 'lodash/sortBy';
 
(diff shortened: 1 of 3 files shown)


src/import/FSHImporter.ts
```diff
@@ -101,28 +100,44 @@ export class FSHImporter extends FSHVisitor {
...
+    this.allAliases = new Map();
     const docs: FSHDocument[] = [];
     const contexts: pc.DocContext[] = [];
+
+    // Preprocess the FSH files
     rawFSHes.forEach(rawFSH => {
-      docs.push(new FSHDocument(rawFSH.path));
-      contexts.push(this.parseDoc(rawFSH.content, rawFSH.path));
+      // Create and store doc for main import process
+      const doc = new FSHDocument(rawFSH.path);
```
kjmahalingam: I like what you've done with the place.

test/import/FSHImporter.Alias.test.ts
```diff
@@ -31,6 +34,70 @@ describe('FSHImporter', () => {
...
+      expect(results[0].aliases.get('USCoreRace')).toBe(
+        'http://hl7.org/fhir/us/core/ValueSet/omb-race-category'
+      );
+      expect(results[1].aliases.size).toBe(0);
+      expect(loggerSpy.getLastMessage()).toMatch(
+        /Alias USCoreRace cannot be redefined to http:\/\/hl7.org\/fhir\/us\/core\/StructureDefinition\/us-core-race; it is already defined as http:\/\/hl7.org\/fhir\/us\/core\/ValueSet\/omb-race-category\..*File: Alias2\.fsh.*Line: 2\D/s
+      );
+    });
+
+    it('should not report an error when the same alias is defined twice w/ different values in different files', () => {
```
kjmahalingam: This name is confusing, since I think you might have a copy-paste error. Could you remove the 'w/ different values' section?

cmoesel: Ah.  Dang.  Good catch.

PR #126, opened 2020-01-22
Bump to 0.6.1 plus surprise feature ooh la la

cmoesel:
Now supports:
```
$ sushi -v
v0.6.1
```

EDIT: and bumps version to 0.6.1.  Almost forgot!

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -24,6 +24,7 @@ async function app() {
       path.join('.', 'build', 'input', 'resources')
     )
     .option('-d, --debug', 'output extra debugging information')
+    .version(getVersion(), '-v, --version', 'print SUSHI version')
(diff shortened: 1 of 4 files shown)


src/app.ts
```diff
@@ -24,6 +24,7 @@ async function app() {
       path.join('.', 'build', 'input', 'resources')
     )
     .option('-d, --debug', 'output extra debugging information')
+    .version(getVersion(), '-v, --version', 'print SUSHI version')
```
cmoesel: This is what made it take longer than expected.  I did not realize that you have to use a special `version` function -- I was trying to use the normal `option` function to do it.

jafeltra:
Ooh la la!

Issue #129, opened 2020-01-23
Unable to set the .text element


davidhay25:
These lines:

* ^text.status = #additional
* ^text.div = "<div xmlns=\"http://www.w3.org/1999/xhtml\">text</div>"

result in this error:

error: Cannot fix string value: <div xmlns="http://www.w3.org/1999/xhtml">text</div>. Value does not match element type: xhtml
  File: /Users/davidhay/IG/registry/shorthand/profile-HpiPractitioner.fsh
  Line: 8

cmoesel:
Thanks for reporting this.  We already were tracking this in our internal backlog but hadn't made a public bug for it.  Team, this is the same as _CIMPL-202: FSH: Support for fixing an xhtml element to a string._  Let's bump up priority on this one.

cmoesel:
Fixed in 0.6.2

PR #131, opened 2020-01-23
Fix strings to xhtml values

mint-thompson:
Strings can be fixed to xhtml values, such as Narrative.div.
Basic validation occurs to check if the value is well-formed xml.

This pull request addresses task https://standardhealthrecord.atlassian.net/browse/CIMPL-202 and issue https://github.com/FHIR/sushi/issues/129.


src/fhirtypes/ElementDefinition.ts
```diff
@@ -989,6 +991,12 @@ export class ElementDefinition {
       this.fixedMarkdown = value;
     } else if (type === 'uuid' && this.checkIfFixable(value, this.fixedUuid, type)) {
       this.fixedUuid = value;
+    } else if (
+      type == 'xhtml' &&
+      sax.parser(true).write(value).error == null &&
```
cmoesel: If I'm reading the source code for `sax` correctly, `write()` has several scenarios in which it throws.  I don't know if we'd ever encounter them, but should we try to handle thrown errors in case they occur?

mint-thompson: Ah, I didn't catch that when I glanced at the docs. With the new `checkXhtml` helper method, if it happens to throw an exception, the input value will be considered invalid.

PR #133, opened 2020-01-23
Add exampleCanonical field to instances that are examples

### LINKED ISSUE #127: Examples not showing up in tab on profile pages
Examples are not showing up on the pages of the profiles they are examples of.  This is because we're generating the ImplementationGuide resource descriptions of examples with `exampleBoolean` instead of `exampleCanonical`.  We need to update this to get the desired functionality.  See: http://hl7.org/fhir/R4/implementationguide-definitions.html#ImplementationGuide.definition.resource.example_x_

kjmahalingam:
Fixes #127 

This adds an exampleCanonical field to example instances, in order to allow examples to display in their tab for profiles.

To test, verify that the IG build with `master` doesn't have examples on their tab for a given profile, even if that example exists in the table of contents.

Then, build the IG with this branch, and verify that they show up.

### DIFF
--- a/src/fhirtypes/InstanceDefinition.ts
+++ b/src/fhirtypes/InstanceDefinition.ts
@@ -10,6 +10,7 @@ export class InstanceDefinition {
   resourceType: string;
   instanceName: string;
   id?: string;
+  meta?: any;
   [key: string]: any; // Allow any key value pair on InstanceDefinition due to the high number of potential properties that can be set on a FHIR instance
 
   /**
--- a/src/fhirtypes/InstanceDefinition.ts
+++ b/src/fhirtypes/InstanceDefinition.ts
@@ -1,4 +1,5 @@
 import cloneDeep = require('lodash/cloneDeep');
+import { Meta } from './specialTypes';
 
 /**
  * A class representing a FHIR Instance.
@@ -10,7 +11,7 @@ export class InstanceDefinition {
   resourceType: string;
   instanceName: string;
   id?: string;
-  meta?: any;
+  meta?: Meta;
   [key: string]: any; // Allow any key value pair on InstanceDefinition due to the high number of potential properties that can be set on a FHIR instance
 
   /**
--- a/src/ig/IGExporter.ts
+++ b/src/ig/IGExporter.ts
@@ -4,7 +4,11 @@ import ini from 'ini';
 import sortBy from 'lodash/sortBy';
 import { ensureDirSync, copySync, outputJSONSync, outputFileSync } from 'fs-extra';
 import { Package } from '../export';
-import { ContactDetail, ImplementationGuide, ImplementationGuideDefinitionResource } from '../fhirtypes';
+import {
+  ContactDetail,
+  ImplementationGuide,
+  ImplementationGuideDefinitionResource
+} from '../fhirtypes';
 import { logger, Type } from '../utils';
 import { FHIRDefinitions } from '../fhirdefs';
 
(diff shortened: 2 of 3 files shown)


jafeltra:
I know I asked this when @cmoesel had done something similar, but is it guarenteed that the example will be fore the `meta.profile[0]` element?

Also, I tried to build fsh-mcode using this branch and I got the following error:

```
onGenerate:
Generating Summary Outputs                                                       (26:34.0943)
Publishing Content Failed: Unable to find exampleFor resource http://hl7.org/fhir/us/core/StructureDefinition/us-core-practitioner for resource /Practitioner/mCODEPractitionerExample01 (26:36.0880)
                                                                                 (26:36.0880)
Use -? to get command line help                                                  (26:36.0880)
                                                                                 (26:36.0880)
Stack Dump (for debugging):                                                      (26:36.0880)
java.lang.Exception: Unable to find exampleFor resource http://hl7.org/fhir/us/core/StructureDefinition/us-core-practitioner for resource /Practitioner/mCODEPractitionerExample01
	at org.hl7.fhir.igtools.publisher.Publisher.generateSummaryOutputs(Publisher.java:4971)
	at org.hl7.fhir.igtools.publisher.Publisher.generate(Publisher.java:4262)
	at org.hl7.fhir.igtools.publisher.Publisher.createIg(Publisher.java:787)
	at org.hl7.fhir.igtools.publisher.Publisher.execute(Publisher.java:663)
	at org.hl7.fhir.igtools.publisher.Publisher.main(Publisher.java:6893)
```

I haven't tried to run it again, but I didn't get that error on SUSHI master.

cmoesel:
Ugh.  This is my fault.  @kjmahalingam did it exactly the way I suggested.  I wasn't aware that the publisher got mad if you set `exampleCanonical` to a URL for an external profile (in this case, US Core).  Dylan -- I think the new approach should be to try to search `meta?.profile?` for a URL that starts with the `canonical` from the project config (e.g., `package.json`).  If it finds one, use it as `exampleCanonical`; otherwise just set `exampleBoolean` to `true`.  This means that things that are just examples of normal resources (not profiles) should also use `exampleBoolean` now.  Sorry for the misdirection!

cmoesel:
Oh.  Better yet, I think you have the `Package`, right?  If so, then when you search `meta?.profile?`, instead of comparing against the canonical, use the URL to fish in the `Package` (using `Type.Profile`) and if you get back a result, that means it's a profile defined in the IG -- so you should use it for `exampleCanonical`.

src/fhirtypes/InstanceDefinition.ts
```diff
@@ -10,6 +10,7 @@ export class InstanceDefinition {
   resourceType: string;
   instanceName: string;
   id?: string;
+  meta?: any;
```
cmoesel: The type for this can be `Meta` (from `fhirTypes`).

kjmahalingam:
@jafeltra thanks for catching the error!

@cmoesel does that mean that when we set `exampleBoolean` to true instead of using the canonical for external profiles, then the example won't show up on the profiles page? I'm guessing that's okay, since we only want examples that are specific to our IG, but I want to make sure.

kjmahalingam:
Feedback from review has been incorporated, so this should be ready now. I verified that the IG I created showed the example in the correct tab.

@cmoesel In the process of incorporating the latest changes, I found that the test I edited to feature exampleCanonical functionality broke. I fixed it to use exampleBoolean again...but do you have a suggestion for another test that can utilize this functionality? Or is it possible that I implemented this incorrectly now?

cmoesel:
@kjmahalingam -- the IG Publisher produces and examples page that lists all examples -- no matter what they are an example of.  So the ones with `"exampleBoolean": true` will still show up there.  We don't put that page in the menu though, so you might have to dig for it.

cmoesel:
Thanks, @kjmahalingam.  Code looks good.  Just running it against mCODE definitions to verify.

Issue #121, opened 2020-01-22
Slicing anomaly in instance of US Core Organization


markkramerus:
There something wrong with the attached example. 

[Organization-mCODEOrganizationExample01.json.txt](https://github.com/FHIR/sushi/files/4098514/Organization-mCODEOrganizationExample01.json.txt)

The IG Publisher complains that:

`Practitioner | error | Unrecognised property '@identifier:NPI'`

Here's the FSH - it is pretty straightforward:

```
Instance: mCODEOrganizationExample01
InstanceOf: USCoreOrganization
* id = "mCODEOrganizationExample01"
* identifier[NPI].value = "1265714091"
* active = true
* name = "Foundation Medicine"
* contact.telecom.system = #phone
* contact.telecom.value = "617-418-2200"
* address.line = "150 Second Street"
* address.city = "Cambridge"
* address.state = "MA"
* address.postalCode = "02141"
* address.country = "USA"
```

markkramerus:
PS -- I added the .txt to the file name because a json file wasn't accepted.

cmoesel:
Oh.  Yeah, that instance JSON is definitely wrong.  @ngfreiter -- looks like the slicename support in instance generation might not be quite working as expected...  Perhaps it works with FSH-defined profiles but not external profiles?

markkramerus:
Here's another case with a different rendering. It comes from mCODECancerGeneticVariantExample01.json:
```
  "category:Laboratory": {
    "coding": [
      {
        "system": "http://terminology.hl7.org/CodeSystem/observation-category",
        "code": "laboratory"
      }
    ]
  },
```
The IG Publisher reports:
`Unrecognised property '@category:Laboratory'`

ngfreiter:
It looks to me like the issue arises specifically when a slice element is fixed on a Structure Definition, and we then make an instance of that Structure Definition. The function that is supposed to set fixed values based on the Structure Definition does not handle the case of a slice element correctly.

cmoesel:
Note that @jafeltra's work on another issue exposed this bug further.  Her work adds some validation where there previously was none.  As a result, these errors that used to fail silently are now logged.  The other result is that rather than putting the invalid properties into the JSON, they are now skipped.

These errors still exist and _are_ an error with SUSHI.  Now they're in the limelight being logged.  So when you run the next SUSHI release, don't be surprised by additional errors reported by SUSHI (and few errors reported by the IG Publisher).

PR #134, opened 2020-01-23
Add validateValueAtPath before setting properties

### LINKED ISSUE #114: Instance property must be an Array, not an Object
Elements that are defined as 0..* in resources must be JSON arrays in the instance, even if the cardinality has been constrained, or the array contains just one value. It appears array elements are being rendered as Objects rather than Arrays.

Example -- on master branch, at commit `https://github.com/standardhealth/fsh-mcode/commit/4948b6ebe0018ddf2f9d8d8c53a469de972f059d`

Example: `build\input\resources\Condition-mCODEComorbidConditionExample01.json`

IG Publisher reports:  `Condition.category | error | This property must be an Array, not an Object`

jafeltra:
Fixes #114 

The functionality for putting properties into arrays even if they are constrained is in the `validateValueAtPath` function. We call that before setting properties on instances that are defined by rules. However, we also set properties on instances based on the structure definition. I updated the function `setFixedValuesForDirectChildren` to also call `validateValueAtPath`. This will check the base cardinality and should fix the bug. I added two tests to check the two places the `setFixedValuesForDirectChildren` is called and check that the arrays are properly set.

### DIFF
--- a/src/export/InstanceExporter.ts
+++ b/src/export/InstanceExporter.ts
@@ -1,5 +1,5 @@
 import { FSHTank } from '../import/FSHTank';
-import { StructureDefinition, InstanceDefinition, ElementDefinition, PathPart } from '../fhirtypes';
+import { StructureDefinition, InstanceDefinition, ElementDefinition } from '../fhirtypes';
 import { Instance } from '../fshtypes';
 import { logger, Fishable, Type } from '../utils';
 import { setPropertyOnInstance, replaceReferences, replaceField } from '../fhirtypes/common';
@@ -37,7 +37,19 @@ export class InstanceExporter {
           // Add back non-numeric (slice) brackets
           pathPart.brackets?.forEach(b => (path += /^[-+]?\d+$/.test(b) ? '' : `[${b}]`));
           const element = instanceOfStructureDefinition.findElementByPath(path, this.fisher);
-          this.setFixedValuesForDirectChildren(element, pathParts.slice(0, i + 1), instanceDef);
+          // Reconstruct the part of the rule's path that we just got the element for
+          let rulePathPart = rule.path
+            .split('.')
+            .slice(0, i + 1)
+            .join('.');
+          rulePathPart += '.';
+
+          this.setFixedValuesForDirectChildren(
+            element,
+            rulePathPart,
+            instanceDef,
+            instanceOfStructureDefinition
+          );
         }
       } catch (e) {
         logger.error(e.message, rule.sourceInfo);
@@ -47,8 +59,9 @@ export class InstanceExporter {
     // Fix values from the SD for all elements at the top level of the SD
     this.setFixedValuesForDirectChildren(
       instanceOfStructureDefinition.findElement(instanceDef.resourceType),
-      [],
-      instanceDef
+      '',
+      instanceDef,
+      instanceOfStructureDefinition
     );
 
     // Remove all _sliceName fields
@@ -76,8 +89,9 @@ export class InstanceExporter {
    */
   private setFixedValuesForDirectChildren(
     element: ElementDefinition,
-    existingPath: PathPart[],
-    instanceDef: InstanceDefinition
+    existingPath: string,
+    instanceDef: InstanceDefinition,
+    instanceOfStructureDefinition: StructureDefinition
   ) {
     const directChildren = element.children(true);
     for (const child of directChildren) {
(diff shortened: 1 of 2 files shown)


test/export/InstanceExporter.test.ts
```diff
@@ -301,6 +325,26 @@ describe('InstanceExporter', () => {
       expect(exported.communication).toBeUndefined();
     });
 
+    it('should fix a nested element that has parents defined in the instance and fixed on the SD to an array even if constrained', () => {
+      const cardRule = new CardRule('communication');
```
mint-thompson: Should this `CardRule` apply to `contact` here? The test still passes when this change is made.

test/export/InstanceExporter.test.ts
```diff
@@ -185,6 +185,30 @@ describe('InstanceExporter', () => {
...
+      const condition = new Profile('TestCondition');
+      condition.parent = 'Condition';
+      doc.profiles.set(condition.name, condition);
+      const conditionInstance = new Instance('Bar');
+      conditionInstance.instanceOf = 'TestCondition';
+      doc.instances.set(conditionInstance.name, conditionInstance);
+      const fixedValRule = new FixedValueRule('category');
+      const fixedFshCode = new FshCode('foo', 'http://foo.com');
+      fixedValRule.fixedValue = fixedFshCode;
+      condition.rules.push(fixedValRule);
```
mint-thompson: Should `condition` have a `CardRule` added to constrain `category` to `1..1` here?

src/export/InstanceExporter.ts
```diff
@@ -76,8 +89,9 @@ export class InstanceExporter {
    */
   private setFixedValuesForDirectChildren(
     element: ElementDefinition,
-    existingPath: PathPart[],
-    instanceDef: InstanceDefinition
+    existingPath: string,
+    instanceDef: InstanceDefinition,
+    instanceOfStructureDefinition: StructureDefinition
```
mint-thompson: Since the function arguments changed, could you also updated the `@param` listing above the function?

test/export/InstanceExporter.test.ts
jafeltra: Oops yes! Good catch!

test/export/InstanceExporter.test.ts
jafeltra: For this one, I was trying to mirror the example that Mark mentioned in the original issue. It didn't look like that example was constraining the cardinality. It only has the rule `* category = SCT#398192003`, which I think we interpret as setting the value to the code (an object). When I undid my changes to the exporter, this test failed, so it seemed valid to me, but I'd be open to adding the `CardRule`. I was just trying to keep it as similar to the example as possible.

Do you have thoughts on which way to do this? I'm open to either.

jafeltra:
As for your top level comment about the string manipulation, if you have ideas about how to make it more clear, I'm happy to change it. From what I could see, I had to rebuild the string in that place because parsing the `rule.path` into the `PathPart[]` adds in extra indices to the path that weren't originally there, and I needed a string representation for the `validateValueAtPath` function to work properly since it outputs the path string in the error. If you think I should change/improve the manipulation, let me know!

test/export/InstanceExporter.test.ts
cmoesel: I doubt it makes a difference, but Mark's example is for a profile whose parent is USCoreCondition.  USCoreCondition constrains `category` to `1..*`.  So you could add that constraint if you want to be as close as possible -- although I doubt it will make a difference.

cmoesel:
I tried to think of a better way than the string rebuilding too, but couldn't come up with anything.  I think you've found the best way.  If there's something better, it probably involves a lot of refactoring.

So... if both of you are good with this, then I'm good with it!

test/export/InstanceExporter.test.ts
jafeltra: Ah good point! I just added it (locally) and it still passes and fails correctly, so I think it makes sense to add it. I can push that!

jafeltra:
I wasn't able to think of anything better than the string manipulation, so I'm good with that part if Mint is!

The new errors you are getting are because we are now validating the paths before setting the properties, so the errors from issue #121 are showing up now. I wrapped the calls for validating and setting in a try/catch so that the instance export will continue despite the error. You should now get the 22 instances, but they'll be missing the invalid slice. The last two commits add that try/catch and the `CardRule` to the test.

However, Chris also noticed that there is a value missing in the `mCODECancerRelatedRadiationProcedureExample01`, so I'm also going to look into why that's happening.

src/export/InstanceExporter.ts
mint-thompson: Ah, I should have been more specific: there's also the new `instanceOfStructureDefinition` that should be added to the `@param` listing.

jafeltra:
After an eventful afternoon, @cmoesel  and @kjmahalingam helped me figure out why the fixed value was missing from `mCODECancerRelatedRadiationProcedureExample01`. The root of the issue was the order in which we applied values to instances, so I swapped around the order we do this in. We now first set values from the Structure Definition the instance instantiates as necessary and then set values that are fixed directly on the instance. I also added the missing param that @mint-thompson pointed out.

I've compared the diffs between exporting `fsh-mcode` on this branch and on `master` and the only changes are properties in different orders (given the switch in how we export now) and the missing invalid slices (because we now skip over them instead of adding invalid slices, and we will fix the slices in issue #121).

In conclusion, I _think_ this branch should be set now.

PR #135, opened 2020-01-23
Prevent IG output from being sent to build/input/resources

### LINKED ISSUE #132: Entire IG generated to build/input/resources
If an `ig-data` folder exists (indicating you want to build an IG), then the entire IG gets generated to `build/input/resources` instead of just to `build`.  This was introduced when we changed the default output folder to `build/input/resources`.

We need to change the default folder back to `build` and then update the export to always export the resources to `${outDir}/input/resources`.

kjmahalingam:
Fixes #132


jafeltra:
I'm not entirely sure how we want to word this, but the README still says that the default output folder is `build/input/resources`. I think since the default is now just `build`, we should probably update this. However, I'm not sure if we also want to specify that resources eventually will end up in `{output_folder}/input/resources`, even if someone uses the `-o` flag to change the output location.

kjmahalingam:
I just changed the name back to `build`. I think that's probably the right solution given the situation.

jafeltra:
Oops sorry I noticed one other thing. It looks like the `package.json` is being written to `build/input/resources/` as well. Do we want that? It feels like since it is not a resource, it should just be at the root of `build`, or at least, that is where it used to be before this PR.

cmoesel:
@jafeltra - yes, `package.json` should be at the root of the out directory.

src/app.ts
```diff
@@ -83,8 +79,11 @@ async function app() {
 
   fs.ensureDirSync(program.out);
 
+  const resourceDir = path.join(program.out, 'input', 'resources');
+  fs.ensureDirSync(resourceDir);
+
   fs.writeFileSync(
-    path.join(program.out, 'package.json'),
+    path.join(resourceDir, 'package.json'),
```
cmoesel: As @jafeltra mentioned, this should really be `program.out`.

cmoesel:
Since it was such a simple fix, I went ahead and just fixed the `package.json` thing so we can get this merged ASAP.  As soon as it passes CI, I'll merge it.

PR #142, opened 2020-01-24
Show message on invalid FHIR id or name

### LINKED ISSUE #139: FHIR rules for name and id should be enforced
Currently SUSHI allows `name` and `id` to be any sequence of non-whitespace characters.  FHIR, however, is more specific:

* name: `[A-Z]([A-Za-z0-9_]){0,254}` (http://hl7.org/fhir/R4/structuredefinition-definitions.html#StructureDefinition.name)
* id: `[A-Za-z0-9\-\.]{1,64}`  (http://hl7.org/fhir/R4/datatypes.html#id)

We should probably catch names/ids that violate these rules and report them as errors in SUSHI.  While we could update the ANTLR grammar to reflect these requirements, it might be more user-friendly to keep the grammar as-is and do the check in code.  Grammar violations make ugly and sometimes-confusing error messages.

mint-thompson:
Fixes #139 

The id and name elements in code systems, structure definitions,
value sets, and instances are checked against the invariants from
the FHIR specification. Both invariants are regular expressions.
When an id or name is not valid, it is still used in the output,
but a message indicates to the user that it is invalid.

The validation of these elements occurs when exporting to FHIR.
Although it would be possible to check some of this earlier on
during FSH import, the validation step is based on specifications
from FHIR. Therefore, I think it makes the most sense to enforce
these rules at the time of creating the FHIR resources.

This pull request addresses https://standardhealthrecord.atlassian.net/browse/CIMPL-243.


src/errors/InvalidIdError.ts
```diff
@@ -0,0 +1,8 @@
+import { Annotated } from './Annotated';
+
+export class InvalidIdError extends Error implements Annotated {
+  specReferences = ['http://hl7.org/fhir/R4/datatypes.html#id'];
+  constructor(public badId: string) {
+    super(`The string "${badId}" does not represent a valid FHIR id.`);
```
cmoesel: Can we provide some more guidance here?  Perhaps repeating what the FHIR doc says.  So maybe something like:

> The string "${badId}" does not represent a valid FHIR id.  FHIR ids may contain any combination of upper- or lower-case ASCII letters ('A'..'Z', and 'a'..'z'), numerals ('0'..'9'), '-' and '.', with a length limit of 64 characters.

src/errors/InvalidNameError.ts
```diff
@@ -0,0 +1,12 @@
+import { Annotated } from './Annotated';
+
+export class InvalidNameError extends Error implements Annotated {
+  specReferences = [
+    'http://hl7.org/fhir/R4/structuredefinition-definitions.html#StructureDefinition.name',
+    'http://hl7.org/fhir/R4/valueset-definitions.html#ValueSet.name',
+    'http://hl7.org/fhir/R4/codesystem-definitions.html#CodeSystem.name'
+  ];
+  constructor(public badName: string) {
+    super(`The string "${badName}" does not represent a valid FHIR name.`);
```
cmoesel: Similar to above, I think it would be helpful to provide the user more guidance.  FHIR doesn't spell this one out in narrative as well as ID, but following the form of how they define id, maybe we could say something like:

> The string "${badName}" does not represent a valid FHIR name.  This name should be usable as an identifier for the module by machine processing applications such as code generation.  Valid names start with an upper-case ASCII letter ('A'..'Z') followed by any combination of of upper- or lower-case ASCII letters ('A'..'Z', and 'a'..'z'), numerals ('0'..'9') and '_', with a length limit of 255 characters.

src/export/CodeSystemExporter.ts
```diff
@@ -1,15 +1,26 @@
...
+    try {
+      validateFHIRName(codeSystem.name);
+    } catch (ex) {
+      logger.error(ex.message, fshDefinition.sourceInfo);
+    }
+    try {
+      validateFHIRId(codeSystem.id);
+    } catch (ex) {
+      logger.error(ex.message, fshDefinition.sourceInfo);
+    }
```
cmoesel: These take up a lot of lines of code to do validation.  Did you consider other alternatives?

For example, you could make `validateFHIRName` and `validateFHIRId` take an optional `SourceInfo` as well and return back the name/id that was passed in.  Then you can log the message in the validation function itself -- and that would also allow you to use it like this:
```ts
codeSystem.name = validateFHIRName(fshDefinition.name, fshDefinition.sourceInfo);
codeSystem.id = validateFHIRId(fshDefinition.id, fshDefinition.sourceInfo);
```

That would make a lot less noise and also allow you to remove those new Error classes (they wouldn't really be needed anymore).  I like having a place to put the refs to the docs, but we're not actually leveraging those right now, so you could just put them in as comments in the validate function for now.  Eventually we can figure out how to integrate them into the logging framework.

Whaddya think?  Like it or is it too weird?

src/fhirtypes/InstanceDefinition.ts
```diff
@@ -10,8 +11,8 @@ import { Meta } from './specialTypes';
 export class InstanceDefinition {
   resourceType: string;
   instanceName: string;
-  id?: string;
   meta?: Meta;
+  id?: FHIRId;
```
cmoesel: Is there a reason you swapped the order here?

src/fhirtypes/common.ts
```diff
@@ -176,3 +178,24 @@ export function replaceField(
...
+ * @throws {InvalidNameError} when the name does not match the regular expression
+ */
+
+export function validateFHIRName(name: string): void {
+  if (!nameRegex.test(name)) {
+    throw new InvalidNameError(name);
+  }
+}
+
+const nameRegex = /^[A-Z]([A-Za-z0-9_]){0,254}$/;
```
cmoesel: Organizationally, I think it might make sense to put this above the function that uses it.  I wasn't going to say anything, except that I noticed for the `validateFHIRId`, you _did_ put it above the function.  So might as well be consistent and I think above is the way to go!

src/fhirtypes/InstanceDefinition.ts
mint-thompson: Nope, that was just me moving quickly when resolving rebase conflicts. I'll put it back.

src/errors/InvalidNameError.ts
mint-thompson: I agree that providing additional information is the right decision here. The full explanation of a valid id or name is pretty long, so do you think it would suffice to provide a link to the relevant page in the FHIR documentation? On the other hand, it's nice to not have to look somewhere else for the answer and just have SUSHI provide the user with all the information they need. I'll consider the options over the weekend.

src/export/CodeSystemExporter.ts
mint-thompson: Hmm, how about this: there could be functions for `setName` and `setId` on the various FHIR classes. In those functions, the value would be assigned, and they would call the `validateFHIRName` and `validateFHIRId` functions. The validate functions would be changed to return a boolean result instead of throwing, so that would let me get rid of the two error classes as you mentioned. With this approach, the actual exporters would just call the set functions. This would also have the advantage that I know what FHIR resource type I'm working with, which lets me write more specific error messages or link to more specific documentation pages. The downside of this approach is that I'm throwing yet another function call on to the stack, and that the FHIR types would have nearly-identical set functions.

Maybe this could be a situation where it finally makes sense to make an abstract supertype for the various FHIR resources that get exported? Ugh, but Instances don't have names, which means that one way or another, things get a little ugly.

src/errors/InvalidNameError.ts
cmoesel: I think we can remove "This name should be usable as an identifier for the module by machine processing applications such as code generation" from above, but I think it's otherwise helpful to have the info right there, even if it is a little long.  That's my opinion anyway -- but I don't feel strongly enough about it that I'd object if you thought otherwise.

src/export/CodeSystemExporter.ts
cmoesel: Would the `setName`/`setId` functions also take in a `SourceInfo` so they can report the lines?  I don't think we want to lose that functionality.  If so, then I'm good with this approach as well!

mint-thompson: Upon further consideration, this seems like a reasonable time for me to try using mixins. The next commit will contain my attempt to do so.

cmoesel: Woah.  Looking forward to it!

src/fhirtypes/CodeSystem.ts
```diff
@@ -2,13 +2,16 @@ import { Meta } from './specialTypes';
...
+import { HasName, HasId } from './common';
+import { applyMixins } from '../utils';
 
 /**
  * Class representing a FHIR R4 CodeSystem
  * @see {@link https://www.hl7.org/fhir/codesystem.html}
  */
 export class CodeSystem {
-  id?: string;
+  id?: FHIRId;
```
cmoesel: My understanding is that we no longer need to declare this property as it gets mixed in, right?  Looking at the documentation you linked to, the base class does not declare the properties that are already defined in the mixed in classes.

One potential issue w/ leaving this here is if we change the property here or in the mixin (but not in the other), then I'm not sure what happens.  That said, if we remove it here, the fact that the property exists gets kind of hidden behind the mixin.

How would you feel about commenting it out and putting in a note?
```ts
// id?: FHIRId; // Provided by HasId mixin
```

cmoesel: Similar goes for `name`.

cmoesel: And... for all the other types that use these mixins.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -98,7 +99,7 @@ export class ElementDefinitionType {
  * @see {@link http://hl7.org/fhir/R4/elementdefinition.html}
  */
 export class ElementDefinition {
-  private _id: string;
+  private _id: FHIRId;
```
cmoesel: Technically, this is not correct.  `ElementDefinition` inherits the `id` field from `Element` -- which described `id` as a `string` (not an `id`):
image

It would actually be a significant problem if it was defined as an `id` because element ids can get very long and we'd definitely run into the character length limit on ids for deeply nested elements.

mint-thompson: Ah, thanks for catching this!

Issue #99, opened 2020-01-17
Choice slices are named repetitively in differential


ngfreiter:
When a choice is constrained to a single type, this is represented by slicing the choice. But to simplify things, this is represented in the differential using just the type of the sliced element, as shown below:
<img width="311" alt="Capture" src="https://user-images.githubusercontent.com/52747882/72627854-2cfe8780-391b-11ea-8766-9c7e4f2de60b.PNG">
When we represent this, we do not create just a single element, and we leave the slicing visible on the differential:
<img width="753" alt="Capture" src="https://user-images.githubusercontent.com/52747882/72627781-13f5d680-391b-11ea-965a-ee795cb7bc0e.PNG">
Instead of showing `value[x]` sliced by the element called `valueCodeableConcept:valueCodeableConcept`, we want to just have one single element called `valueCodeableConcept`, as the FHIR example has one single `valueQuantity` element.

PR #150, opened 2020-01-27
Fix differentials related to choice type slices

cmoesel:
1. Remove slicename from renamed choice type differential elements
2. Return root choice element ([x]) from differential when it's only there to introduce the type slice

This addresses #99.  Note that when you build the IG, you will often still see both `value[x]` and the specific choice (e.g., `valueString`) -- but you won't see the weird duplicate slice names anymore.  The reason you will usually still see `value[x]` is because we still must include `value[x]` in the differential if we constrained the choice to a smaller set of choices.  The only time we can exclude `value[x]` from the differential is if we _keep_ all the same types as valid choices.  See: https://blog.fire.ly/2019/09/13/type-slicing-in-fhir-r4/

If you build fsh-mcode, `CancerDiseaseStatus` shows instances of both cases.  

(a) The `EvidenceType` extension in `CancerDiseaseStatus` binds its `valueCodeableConcept` to a value set.  See the before and after in this screenshot.
image

(b) The `valueCodeableConcept` of `CancerDiseaseObservation` also binds to a value set.  Before and after here:
image

Note -- the after still looks weird (in a different way), but I believe it it actually technically correct based on the documentation.

### DIFF
--- a/src/fhirtypes/ElementDefinition.ts
+++ b/src/fhirtypes/ElementDefinition.ts
@@ -314,8 +314,14 @@ export class ElementDefinition {
         diff[prop] = cloneDeep(this[prop]);
       }
     }
-    // Path gets set automatically when setting id
+    // Set the diff id, which may be different than snapshot id in the case of choices (e.g., value[x] -> valueString)
+    // NOTE: The path also gets set automatically when setting id
     diff.id = diff.diffId();
+    // If the snapshot is a choice (e.g., value[x]), but the diff is a specific choice (e.g., valueString), then
+    // remove the slicename property from the diff (it is implied and not required in the diff)
+    if (this.path.endsWith('[x]') && !diff.path.endsWith('[x]')) {
+      delete diff.sliceName;
+    }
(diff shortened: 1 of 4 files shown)


ngfreiter:
So if there is a line like: `* value[x] only valueCodeableConcept`, we do keep the `value[x]` element around? But if we just directly modify the `valueCodeableConcept` element, we do not keep the `value[x]` element around, since the slicing on it can be inferred? Am I understanding how this works correctly?

cmoesel:
@ngfreiter:

> So if there is a line like: * value[x] only valueCodeableConcept, we do keep the value[x] element around?

Yes, and I got that from this:

--- start screenshot ---
image
--- end screenshot ---

and this:

--- start screenshot ---
image
--- end screenshot ---

> But if we just directly modify the valueCodeableConcept element, we do not keep the value[x] element around, since the slicing on it can be inferred?

Yes, and I got that from this:

--- start screenshot ---
image
--- end screenshot ---

PR #154, opened 2020-01-27
Add trailing newline to files without one

### LINKED ISSUE #145: Comment on the last line of a FSH file causes an error
If the last line of a FSH file is a comment or if the last line ends with an inline comment, the following error is reported: 

`error: extraneous input '//' expecting {<EOF>, KW_ALIAS, KW_PROFILE, KW_EXTENSION, KW_INSTANCE, KW_INVARIANT, KW_VALUESET, KW_CODESYSTEM}`

This may be related to the fix for #130 because I did not see this issue prior the the PR that addressed that bug.

Example FSH files that will report that error:

```
Instance: Example
InstanceOf: Patient
// Comment
```

```
Instance: Example
InstanceOf: Patient // Comment
```

kjmahalingam:
Fixes #145 

There was an issue with comments (line and block) causing a syntax error if they are at the end of a file, and there is no newline in the file. This PR addresses the issue by solving it the way some other languages do; adding a trailing newline to the end of the file content being processed if there isn't one already.

I'd like someone on Windows to test this out, since I know newlines can be kind of messy across operating systems.

I'm going to add a couple of tests to verify these work as well; stand by for that.

### DIFF
--- a/src/import/FSHImporter.ts
+++ b/src/import/FSHImporter.ts
@@ -111,6 +111,7 @@ export class FSHImporter extends FSHVisitor {
       docs.push(doc);
 
       // Create and store context for main import process
+      if (!rawFSH.content.endsWith('\n')) rawFSH.content += '\n';
       const ctx = this.parseDoc(rawFSH.content, rawFSH.path);
       contexts.push(ctx);
 
--- a/src/import/FSHImporter.ts
+++ b/src/import/FSHImporter.ts
@@ -111,8 +111,13 @@ export class FSHImporter extends FSHVisitor {
       docs.push(doc);
 
       // Create and store context for main import process
-      if (!rawFSH.content.endsWith('\n')) rawFSH.content += '\n';
-      const ctx = this.parseDoc(rawFSH.content, rawFSH.path);
+      // We are appending a newline to the file content if there is not one there already.
+      // This is being done because we discovered a syntax error that occurs if a comments
+      // ends a file, and there is no newline after it. Comments are only tokenized in our
+      // grammar if a newline follows it. In order to prevent this error from occurring,
+      // we add a newline to the content before we parse it so comments at EOF can be tokenized.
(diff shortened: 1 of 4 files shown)


kjmahalingam:
Tests added!

kjmahalingam:
Ah, that's a good call. I didn't think about that.
I will make those changes.

kjmahalingam:
All feedback has been addressed, @cmoesel . Let me know if there's anything else that needs to be changed.

If @ngfreiter or @mint-thompson is able to test this manually, as well, to make sure newlines don't cause issues on a Windows OS, that would be great!

ngfreiter:
I gave this a test run and did not see any issues with it working with Windows 👍 .

src/import/FSHImporter.ts
```diff
@@ -111,6 +111,7 @@ export class FSHImporter extends FSHVisitor {
       docs.push(doc);
 
       // Create and store context for main import process
+      if (!rawFSH.content.endsWith('\n')) rawFSH.content += '\n';
```
cmoesel: I'd really like to see a comment here indicating _why_ we're tacking on the newline.  A few years into the future, this line may look really suspicious to someone.

cmoesel: Also, we could avoid removing the `readonly` from `RawFSH` by changing this to:
```ts
const content = rawFSH.content.endsWith('\n') ? rawFSH.content + '\n';
const ctx = this.parseDoc(content, rawFSH.path);
```
This also means we're not mutating the original input.

src/import/RawFSH.ts
```diff
@@ -1,6 +1,6 @@
 export class RawFSH {
   public readonly path: string;
-  constructor(public readonly content: string, path?: string) {
+  constructor(public content: string, path?: string) {
```
cmoesel: You can now revert this change so it is `readonly` again.

PR #156, opened 2020-01-27
Set values from FHIR on slices

### LINKED ISSUE #121: Slicing anomaly in instance of US Core Organization

ngfreiter:
Fixes #121, which is also tracked by https://standardhealthrecord.atlassian.net/browse/CIMPL-245.

We were not correctly handling the case that a Structure Definition from which we are constructing an Instance would have existing values fixed to a slice element. This PR adds capability to handle when a value is fixed directly on a slice, and not on children of that slice.

To further test this, first build fsh-mcode on master using SUSHI on master. You will see a number of errors that look like: `error: Cannot resolve element from path: category:Laboratory`. Now build fsh-mcode using SUSHI on this branch. First, you should notice those errors go away. Second, you should notice that the `identifier` field in Organization-mCODEOrganizationExample01.json file is generated correctly as this:
```
"identifier": [
    {
      "system": "http://hl7.org/fhir/sid/us-npi",
      "value": "1265714091"
    },
    {
      "system": "urn:oid:2.16.840.1.113883.4.7"
    }
  ]
```
instead of this:
```
"identifier": [
    {
      "value": "1265714091",
      "system": "http://hl7.org/fhir/sid/us-npi"
    }
  ]
```
Finally, you should be able to go into CancerGenomicsExamples.fsh and remove all the lines that have been marked as a workaround for issue 148 (which was a duplicate of the issue addressed here), and you should still get the same output as when those lines were present (down to the ordering of elements in the generated SDs).

### DIFF
--- a/src/export/InstanceExporter.ts
+++ b/src/export/InstanceExporter.ts
@@ -102,7 +102,7 @@ export class InstanceExporter {
       const fixedValueKey = Object.keys(child).find(
         k => k.startsWith('fixed') || k.startsWith('pattern')
       );
-      if (fixedValueKey) {
+      if (fixedValueKey && child.min > 0) {
         // Get the end of the child path, this is the part that differs from existingPath
         const childPath = child
           .diffId()
@@ -126,6 +126,10 @@ export class InstanceExporter {
         } catch (e) {
           logger.error(e.message);
         }
+      } else if (fixedValueKey) {
+        logger.debug(
+          `Element ${child.id} is optional with min cardinality 0, so fixed value for optional element is not set on instance ${instanceDef.instanceName}`
+        );
       }
     }
   }
--- a/src/export/InstanceExporter.ts
+++ b/src/export/InstanceExporter.ts
@@ -104,14 +104,21 @@ export class InstanceExporter {
       );
       if (fixedValueKey) {
         // Get the end of the child path, this is the part that differs from existingPath
-        const childPathPart = child
+        const childPath = child
           .diffId()
           .split('.')
           .slice(-1)[0];
-
+        // Turn FHIR slicing (element:slicName/resliceName) into FSH slicing (element[sliceName][resliceName])
+        const colonSplitPath = childPath.split(':');
+        let fshChildPath = colonSplitPath[0];
+        const slicePathSection = colonSplitPath[1];
+        const sliceNames = slicePathSection?.split('/');
+        sliceNames?.forEach(s => {
+          fshChildPath += `[${s}]`;
+        });
         try {
           const { fixedValue, pathParts } = instanceOfStructureDefinition.validateValueAtPath(
-            existingPath + childPathPart,
+            existingPath + fshChildPath,
             child[fixedValueKey as keyof ElementDefinition],
             this.fisher
           );
--- a/src/fhirtypes/common.ts
+++ b/src/fhirtypes/common.ts
@@ -71,9 +71,14 @@ export function setPropertyOnInstance(
           if (j < current[key].length && j === index && current[key][index] == null) {
             current[key][index] = {};
           } else if (j >= current[key].length) {
-            // _sliceName is used to later differentiate which slice an element represents
-            const arrayFiller = sliceName ? { _sliceName: sliceName } : null;
-            current[key].push(j === index ? {} : arrayFiller);
+            if (sliceName) {
+              // _sliceName is used to later differentiate which slice an element represents
+              current[key].push({ _sliceName: sliceName });
+            } else if (j === index) {
+              current[key].push({});
+            } else {
+              current[key].push(null);
+            }
           }
         }
         // If it isn't the last element, move on, if it is, set the value
@@ -83,7 +88,11 @@ export function setPropertyOnInstance(
             current._sliceName = sliceName;
           }
         } else {
-          current[key][index] = fixedValue;
+          if (typeof fixedValue === 'object') {
+            Object.assign(current[key][index], fixedValue);
+          } else {
+            current[key][index] = fixedValue;
+          }
         }
       } else {
         // If it isn't the last element, move on, if it is, set the value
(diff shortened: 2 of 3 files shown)


cmoesel:
I don't know if we want to treat this as a separate bug or try to fix it in this PR, but in your example of the _right_ result, the 2nd identifier shouldn't be there:
image

It doesn't make sense for an instance to have an identifier w/ a system but no value.  Looking at the FSH source, this identifier isn't added by the instance -- so it must be coming from the profile.  Looking at US Core Organization, there it is:
image

Note that while the system is `1..1`, the slice above it is `0..1`, so as an optional slice in the element, it should not be automatically added to the instance.

ngfreiter:
Huh I would have figured that would work correctly, but something must be off. I'll try fixing it in this PR.

ngfreiter:
So after some further thought, I am not too sure about this. I _think_ that the rule we agreed upon was that if an element's parent is created, and that element has a fixed value, then we will fix that value, regardless of the cardinality. By this rule, it makes sense that `identifier:CLIA.system` gets set, since `identifier:CLIA`'s parent element is created (since its parent is the root), and `identifier:CLIA` does have a value that is fixed on it via `patternIdentifier`.

I still agree that it is wrong to set the system, but I think this issue is really related to an issue we have already logged: https://standardhealthrecord.atlassian.net/browse/CIMPL-200, which is just about how we don't really handle `pattern` in an intelligent way. My feeling is that solving CIMPL-200 would solve this issue, or at least make it clear how this issue should be handled.

cmoesel:
> I think that the rule we agreed upon was that if an element's parent is created, and that element has a fixed value, then we will fix that value, regardless of the cardinality.

OK... so the `identifier:CLIA` slice has a `patternIdentifier` on it?  Got it.  And for the statement you made above (quoted), we treat `pattern[x]` like a "fixed value". So since it has a `pattern[x]` on it, and since its parent is created, we created the "fixed value" automatically.

That's making me question our rule.  I wonder if we should only be doing it if it is 1..1 after all.  I mean, maybe there is a semantic difference?  Perhaps:

* a 1..1 fixed value means you _must_ include this property with this exact value
* a 0..1 fixed value means you _may_ include this property (or may not), but _if you do_, it must have this exact value.

That's starting to make sense to me and would mean that in the above, that rogue system would not be included.  But I wonder what else it would affect...

Also -- I don't think I exactly understand how CIMPL-200 would fix this, as its just about validating whether or not a value can be fixed somewhere -- and ensuring we're not overwriting something that was already fixed by a pattern above it.  Right?

ngfreiter:
> * a 1..1 fixed value means you _must_ include this property with this exact value
> * a 0..1 fixed value means you _may_ include this property (or may not), but _if you do_, it must have this exact value.

I agree with this. I thought I remembered discussing this before though, and I'm not sure why we didn't take this approach earlier, but I couldn't find the old conversation so maybe I am just remembering wrong.

In my head I am thinking this is similar to CIMPL-200, because if we have a way to check if an element is fixed by pattern[x] in one of its parent elements, I think that is relevant to this issue and to CIMPL-200. In this issue, we could use that functionality to see if we need to set values that are 1..1. We set them either if they are fixed directly, or if they are fixed by pattern[x] of their parents. In CIMPL-200, we want to be able to validate when we are fixing a value if that value is already fixed by pattern[x] in a parent element, so again we need to be able to check if an element is fixed by pattern[x] in its parents. That's my thinking for why CIMPL-200 is relevant, I'm not saying solving CIMPL-200 necessarily solves this, just that they seem closely tied to me.

ngfreiter:
@cmoesel ok the latest commit should fix it so that we only fix values that have min cardinality greater than 0 (please ignore the commit message which incorrectly says 1). It also updates tests that were previously operating on the assumption that all fixed values would be set, not only those with cardinality greater than 0.

cmoesel:
This mostly works as advertised, but there is a use case in mCODE that's not working as expected.  This may be one of the complex cases we discussed this morning, @ngfreiter.  Basically, in mCODE's `Examples.fsh`, we find this in one of the Practitioner examples:
```
* identifier[NPI].value = "1265714091"
// workaround for https://github.com/FHIR/sushi/issues/148: "Pattern in StructureDefinition not reflected in Instance" opened 1-25-2020 by MK
* identifier[NPI].system = "http://hl7.org/fhir/sid/us-npi"   // although URI is not a supported type in FSH yet, apparently you can get away with giving the URI as a string, and the json comes out correctly.
```

If we remove the workaround (i.e., delete the line fixing the system) then the system doesn't get set in the example at all:
```
  "identifier": [
    {
      "value": "9988776655"
    }
  ],
```
Although `indentifier[NPI]` slice is `0..1`, `identifier[NPI].system` is `1..1` -- so according to our basic rules, it *should* be set since we're going to `identifier[NPI]` to set the `value`.

Nick -- is this the case you wanted to treat as a separate issue?  Or should we be addressing this here?

ngfreiter:
You are correct, this is the case that is a separate issue, now tracked by https://github.com/FHIR/sushi/issues/160. I will not address that here, since it could be a bit complicated, and I don't want to hold this up to fix that bug, but I will work on it next.

PR #158, opened 2020-01-28
Start with cleaner plate

### LINKED ISSUE #157: Don't inherit potentially irrelevant metadata from parents
This issue stems from the conversation of #128.  The general discussion was around clearing inherited metadata that causes issues in the IG Publisher or is potentially irrelevant to the profile.  The main concern of #128 was the `"comment"` field, which is in `ElementDefinition`.  We've currently decided _not_ to address that because clearing comments would cause _many_ differential elements.

This issue is to point out that the top-level metadata of profiles and extensions still inherits many fields that are potentially irrelevant or wrong for the profile/extension.  We should remove those.  I proposed an approach on Zulip here: https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/Bad.20links.20on.20Detailed.20Description.20tab/near/186460459

I've also copied the approach in case something happens to the zulip:

* **id** - overwrite w/ new id
* **meta** - clear
* **implicitRules** - clear
* **language** - clear
* **text** - clear
* **contained** - clear
* **extension** - clear (see https://github.com/FHIR/sushi/issues/116)
* **modifierExtension** - clear
* **url** - overwrite w/ new url
* **identifier** - clear
* **version** - overwrite w/ IG version (author can override if maintaining per-profile versions)
* **name** - overwrite w/ new name
* **title** - clear or overwrite w/ new title (if applicable)
* **status** - currently defaults to `active` but FSH should add (required?) keyword
* **experimental** - clear
* **date** - clear
* **publisher** - clear (or overwrite w/ publisher from config?)
* **contact** - clear (or overwrite w/ contributors from config?)
* **description** - clear or overwrite w/ new description (if applicable)
* **useContext** - clear
* **jursidiction** - clear
* **purpose** - clear
* **copyright** - clear
* **keyword** - clear or keep? Would parent keywords still be applicable to child profiles?
* **fhirVersion** - set to target version (should match anyway)
* **mapping** - keep (assuming mappings are still valid and because removing them would require use to remove mappings from every element, creating tons of diffs)
* **kind** - keep (this should not change)
* **abstract** - keep or set to false.  If you set an abstract profile as a parent should you need to explicitly toggle abstractness to `false`?  Or should we default to `false`?
* **context** - keep (assuming context is still valid for child extensions)
* **contextInvariant** - keep (assuming context is still valid for child extensions)
* **type** - keep (this should not change for profiles or extensions)
* **baseDefinition** - overwrite w/ the parent URL of the new profile/extension
* **derivation** - set to constraint (so far, SUSHI only supports constraints)
* **snapshot** - keep, but elements will obviously be overwritten by constraints
* **differential** - clear, as the differential starts from scratch to reflect changes the profile has made to the parent

cmoesel:
When creating a profile, start from a cleaner slate, clearing potentially irrelevant metadata from the parent.  This PR fixes #157.

The unit tests show the basic functionality.  You can also verify by generating profiles/extensions using the `master` branch and then generating them with this branch.  A file compare will show the differences.

### DIFF
--- a/test/export/StructureDefinitionExporter.test.ts
+++ b/test/export/StructureDefinitionExporter.test.ts
@@ -91,7 +91,7 @@ describe('StructureDefinitionExporter', () => {
     expect(exported.version).toBe('0.0.1'); // provided by config
     expect(exported.name).toBe('Foo'); // provided by user
     expect(exported.title).toBeUndefined();
-    expect(exported.status).toBe('active'); // always active
+    expect(exported.status).toBe('draft'); // always active
     expect(exported.experimental).toBeUndefined();
     expect(exported.date).toBeUndefined();
     expect(exported.publisher).toBeUndefined();
@@ -197,7 +197,7 @@ describe('StructureDefinitionExporter', () => {
     expect(exported.version).toBe('0.0.1'); // provided by config
     expect(exported.name).toBe('Foo'); // provided by user
     expect(exported.title).toBeUndefined();
-    expect(exported.status).toBe('active'); // always active
+    expect(exported.status).toBe('draft'); // always active
     expect(exported.experimental).toBeUndefined();
     expect(exported.date).toBeUndefined();
     expect(exported.publisher).toBeUndefined();
--- a/test/export/StructureDefinitionExporter.test.ts
+++ b/test/export/StructureDefinitionExporter.test.ts
@@ -52,7 +52,7 @@ describe('StructureDefinitionExporter', () => {
   });
 
   // Profile
-  it('should set all metadata for a profile', () => {
+  it('should set all user-provided metadata for a profile', () => {
     const profile = new Profile('Foo');
     profile.id = 'foo';
     profile.parent = 'Observation';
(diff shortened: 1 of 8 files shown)


test/export/StructureDefinitionExporter.test.ts
```diff
@@ -69,7 +69,63 @@ describe('StructureDefinitionExporter', () => {
...
+    doc.profiles.set(profile.name, profile);
+    exporter.exportStructDef(profile);
+    const exported = pkg.profiles[0];
+
+    expect(exported.id).toBe('Foo'); // defaulted from user-provided name
+    expect(exported.meta).toBeUndefined();
+    expect(exported.implicitRules).toBeUndefined();
+    expect(exported.language).toBeUndefined();
+    expect(exported.text).toBeUndefined();
+    expect(exported.contained).toBeUndefined; // inherited from Observation
```
kjmahalingam: I'm not sure if this makes a difference (but I noticed it thanks to GitHub syntax highlighting). This is a `toBeUndefined` the property rather than `toBeUndefined()`. I can see that causing issues with some libraries, so could you either fix this or explain why it is this way if it should be?

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -69,7 +69,63 @@ describe('StructureDefinitionExporter', () => {
...
+      {
+        identity: 'sct-attr',
+        uri: 'http://snomed.org/attributebinding',
+        name: 'SNOMED CT Attribute Binding'
+      }
+    ]); // inherited from Observation
+    expect(exported.kind).toBe('resource'); // inherited from Observation
+    expect(exported.abstract).toBe(false); // always abstract
+    expect(exported.context).toBeUndefined; // inherited from Observation
+    expect(exported.contextInvariant).toBeUndefined; // inherited from Observation
```
kjmahalingam: Line 124 and 125 both have the same concern as line 86 above.

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -119,21 +167,66 @@ describe('StructureDefinitionExporter', () => {
...
+    doc.profiles.set(extension.name, extension);
+    exporter.exportStructDef(extension);
+    const exported = pkg.extensions[0];
+
+    expect(exported.id).toBe('Foo'); // defaulted from user-provided name
+    expect(exported.meta).toBeUndefined();
+    expect(exported.implicitRules).toBeUndefined();
+    expect(exported.language).toBeUndefined();
+    expect(exported.text).toBeUndefined();
+    expect(exported.contained).toBeUndefined; // inherited from patient-mothersMaidenName
```
kjmahalingam: Same as before.

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -119,21 +167,66 @@ describe('StructureDefinitionExporter', () => {
...
+    expect(exported.keyword).toBeUndefined();
+    expect(exported.fhirVersion).toBe('4.0.1'); // Inherited from patient-mothersMaidenName
+    expect(exported.mapping).toEqual([
+      { identity: 'v2', uri: 'http://hl7.org/v2', name: 'HL7 v2 Mapping' },
+      { identity: 'rim', uri: 'http://hl7.org/v3', name: 'RIM Mapping' }
+    ]); // inherited from patient-mothersMaidenName
+    expect(exported.kind).toBe('complex-type'); // inherited from patient-mothersMaidenName
+    expect(exported.abstract).toBe(false); // always abstract
+    expect(exported.context).toEqual([{ type: 'element', expression: 'Patient' }]); // inherited from patient-mothersMaidenName
+    expect(exported.contextInvariant).toBeUndefined; // inherited from patient-mothersMaidenName
```
kjmahalingam: Same as before.

test/export/StructureDefinitionExporter.test.ts
cmoesel: You are right.  I missed the `()`.  Multiple times (copy/paste).  I'll fix it.

cmoesel:
Someone from Firely just revealed the Forge strategy (which fields they keep, which they clear, and what default values they use.  It's extremely close to what I did in this PR, but there are a few minor differences.  Let's hold off on merging this until we can discuss.

cmoesel:
OK, PR is updated w/ changes from discussion on Zulip.  In short, we align better w/ Forge's approach now.  This also included changing the default value of `status` from `active` to `draft`.

PR #161, opened 2020-01-29
Replace references to local code system with url

### LINKED ISSUE #149: References to codes defined in locally-defined code system cause IG Pub error
I tried to define a local code system:
````
CodeSystem:  AJCCFairUse
Title: "AJCC Fair Use"
Description: "A small local subset of AJCC staging codes for example purposes."
* #3C
* #cM0
* #cT3
* #cN3
* #pM0
* #pT3
* #pN3
```
and then referenced it in an instance like this:

`* valueCodeableConcept = AJCCFairUse#3C "IIIC"`

That didn't work because the IG Pub complained that you must use a URL for a code system, not a local name. 

So I tried this:

`Alias:   FairUse = http://hl7.org/fhir/us/mcode/CodeSystem/AJCCFairUse`

and then tried to reference it in an instance like this:

`* valueCodeableConcept = FairUse#3C "IIIC"`

but that didn't work either. The IG Pub complained "Unknown Code System 'http://hl7.org/fhir/us/mcode/CodeSystem/AJCCFairUse'", and "URL value 'http://hl7.org/fhir/us/mcode/CodeSystem/AJCCFairUse' does not resolve"

So I'm wondering how one an define a code system in FSH and use codes from that system in the same FSH Tank.

Sorry I can't provide a more complete example because I pulled back my change.

mint-thompson:
Fixes #149 by way of task https://standardhealthrecord.atlassian.net/browse/CIMPL-247.

When an Instance or Structure Definition refers to a code system by name, and that code system is defined locally, replace the code system's name with the code system's url.

I didn't manage to exactly replicate every error mentioned in #149, but this pull request addresses a main portion of it. There may be further work required relating to references to locally defined code systems.

### DIFF
--- a/src/fhirtypes/common.ts
+++ b/src/fhirtypes/common.ts
@@ -1,6 +1,6 @@
 import { StructureDefinition, PathPart, ElementDefinition, InstanceDefinition } from '.';
 import { FixedValueRule } from '../fshtypes/rules';
-import { FshReference, Instance, SourceInfo } from '../fshtypes';
+import { FshReference, Instance, SourceInfo, FshCode } from '../fshtypes';
 import { FSHTank } from '../import';
 import { Type, Fishable } from '../utils/Fishable';
 import cloneDeep = require('lodash/cloneDeep');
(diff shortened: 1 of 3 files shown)


ngfreiter:
Ok @mint-thompson I've reviewed this and it looks good! But one question before I approve, do we want to close #149 once this is merged? Or are we going to keep that open since there are (potentially) other errors described there.

PR #172, opened 2020-01-31
Fix values in instances from parent patterns

### LINKED ISSUE #160: Set properties on instance from pattern[x] on parent element
When we create an instance, we copy over fixed and pattern values from the Structure Definition onto the instance. However, if an element's value is set by a pattern on that element's parent, the element does not get fixed, because we only account for fixed values that are directly on the element, not for pattern values that may be present on parents of the element. As an example, see the below:
```
* identifier[NPI].value = "1265714091"
// workaround for https://github.com/FHIR/sushi/issues/148: "Pattern in StructureDefinition not reflected in Instance" opened 1-25-2020 by MK
* identifier[NPI].system = "http://hl7.org/fhir/sid/us-npi"   // although URI is not a supported type in FSH yet, apparently you can get away with giving the URI as a string, and the json comes out correctly.
```
If we remove the workaround (i.e., delete the line fixing the system) then the system doesn't get set in the example at all:
```
  "identifier": [
    {
      "value": "1265714091"
    }
  ],
```
Although indentifier[NPI] slice is 0..1, identifier[NPI].system is 1..1 -- so according to our basic rules, it should be set since we're going to identifier[NPI] to set the value. The system element needs to not only be checked to see if it has any direct fixed values, it also needs to be checked to see if parents fix it as well. This is similar to https://github.com/FHIR/sushi/issues/84, and it is likely that solving this issue will make that issue easy to solve, but they are not quite duplicates.

ngfreiter:
This fixes #160 

Values are now fixed on an instance if they are fixed by a pattern[x] element in their parent. One possible way to test this is to build fsh-mcode using SUSHI on master. Then, in Examples.fsh, you should be able to comment out the lines that are labeled as part of a workaround. Then you can build fsh-mcode using SUSHI on this branch. You should get the same result, down to the ordering of the elements in the output. Additionally, you should notice that you do not see the problem that is described in #160

Note that this does not yet add validation to check if an element is fixed by pattern from its parent. This is only adding functionality to set values on an instance according to patterns on parent elements. I am open to adding validation to this PR if the reviewers think I should, but I felt like this could stand on its own.

Whoever reviews, if you can give a good look at the tests, I would appreciate that. I had trouble thinking through different possible test cases. As far as I could think of, my tests cases cover what we need, but I want to make sure this is well tested.

### DIFF
--- a/src/export/InstanceExporter.ts
+++ b/src/export/InstanceExporter.ts
@@ -20,18 +20,18 @@ export class InstanceExporter {
     instanceOfStructureDefinition: StructureDefinition
   ): InstanceDefinition {
     // Fix values from the SD for all elements at the top level of the SD
-    this.setFixedValuesForDirectChildren(
+    this.setFixedValuesFromStructureDefinition(
       instanceOfStructureDefinition.findElement(instanceDef.resourceType),
       '',
       instanceDef,
       instanceOfStructureDefinition
     );
 
-    // All rules will be FixValueRule
+    // Fix all values from the SD
     fshInstanceDef.rules.forEach(rule => {
       rule = replaceReferences(rule, this.tank, this.fisher);
       try {
-        const { fixedValue, pathParts } = instanceOfStructureDefinition.validateValueAtPath(
+        const validated = instanceOfStructureDefinition.validateValueAtPath(
           rule.path,
           rule.fixedValue,
           this.fisher
@@ -39,7 +39,7 @@ export class InstanceExporter {
 
         // For each part of that path, we add fixed values from the SD
         let path = '';
-        for (const [i, pathPart] of pathParts.entries()) {
+        for (const [i, pathPart] of validated.pathParts.entries()) {
           path += `${path ? '.' : ''}${pathPart.base}`;
           // Add back non-numeric (slice) brackets
           pathPart.brackets?.forEach(b => (path += /^[-+]?\d+$/.test(b) ? '' : `[${b}]`));
@@ -51,14 +51,27 @@ export class InstanceExporter {
             .join('.');
           rulePathPart += '.';
 
-          this.setFixedValuesForDirectChildren(
+          this.setFixedValuesFromStructureDefinition(
             element,
             rulePathPart,
             instanceDef,
             instanceOfStructureDefinition
           );
         }
+      } catch (e) {
+        logger.error(e.message, rule.sourceInfo);
+      }
+    });
 
+    // All rules will be FixValueRule
+    fshInstanceDef.rules.forEach(rule => {
+      rule = replaceReferences(rule, this.tank, this.fisher);
+      try {
+        const { fixedValue, pathParts } = instanceOfStructureDefinition.validateValueAtPath(
+          rule.path,
+          rule.fixedValue,
+          this.fisher
+        );
         // Fix value fom the rule
         setPropertyOnInstance(instanceDef, pathParts, fixedValue);
       } catch (e) {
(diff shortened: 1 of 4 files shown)


src/export/InstanceExporter.ts
```diff
@@ -51,14 +51,27 @@ export class InstanceExporter {
...
             instanceDef,
             instanceOfStructureDefinition
           );
         }
+      } catch (e) {
+        logger.error(e.message, rule.sourceInfo);
+      }
+    });
 
+    // All rules will be FixValueRule
```
cmoesel: I feel like the comments on line 30 and this line (60) maybe could be more specific about what's happening.  I.e.,

```
30: // Fix all values from the SD that are exposed when processing the instance rules

66: // Fix all values explicitly set in the instance rules (all rules will be FixValueRule)
```
Or something like that...

src/export/InstanceExporter.ts
```diff
@@ -51,14 +51,27 @@ export class InstanceExporter {
...
+    fshInstanceDef.rules.forEach(rule => {
+      rule = replaceReferences(rule, this.tank, this.fisher);
```
cmoesel: It's not a super big deal, but I noticed you make this function call as the first thing you do in both times you iterate over `fshInstanceDef.rules`.  I think you could eliminate this redundant processing by doing this before you iterate over any rules (i.e., around line 30):

```ts
const rules = fshInstanceDef.map(r => replaceReferences(r, this.tank, this.fisher));
```

Then iterating over `rules` instead of `fshInstanceDef.rules` in these two loops.

ngfreiter:
@cmoesel I fixed those initial points.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -846,6 +846,19 @@ export class ElementDefinition {
...
+   * @returns {any} the value the element is fixed to by its parent, undefined if value is not fixed
+   */
+  fixedByParent(): any {
+    const parent = this.parent();
+    const patternKey = parent ? Object.keys(parent).find(k => k.startsWith('pattern')) : null;
+    if (patternKey) {
+      const patternValue: any = parent[patternKey as keyof ElementDefinition];
+      return patternValue[this.path.replace(`${parent.path}.`, '')];
+    }
+  }
```
cmoesel: I noticed that this looks like it only goes up one parent to check.  Of course, the pattern could be several levels up (at a grandparent, for example).  So, does this only need to look up one level because other code (`validateValueAtPath`?) traverses the whole path -- so the grandparent values eventually make their way down to the child?

ngfreiter: When we set values on the instance from the StructureDefinition, the values are set based on paths that are used in the rules being used to create an instance. For each path that is contained in a rule, we traverse the whole path piece by piece. So if the path is `Patient.maritalStatus.coding.code`, we will set fixed values on the instance in the function `setFixedValuesFromStructureDefinition` using each part of that path, ie `Patient.maritalStatus`, then `Patient.maritalStatus.coding`, `Patient.maritalStatus.coding.code`. So yes if the pattern is on `Patient.maritalStatus` and that pattern fixes `Patient.maritalStatus.coding.code`, the `code` still gets fixed, even though the `fixedByParent` function only looks up one level.

src/export/InstanceExporter.ts
```diff
@@ -90,45 +103,48 @@ export class InstanceExporter {
...
         sliceNames?.forEach(s => {
-          fshChildPath += `[${s}]`;
+          fshElementPath += `[${s}]`;
         });
+        // Fix the value if we validly can
         try {
           const { fixedValue, pathParts } = instanceOfStructureDefinition.validateValueAtPath(
-            existingPath + fshChildPath,
-            child[fixedValueKey as keyof ElementDefinition],
+            fshElementPath === '' ? existingPath.slice(0, -1) : existingPath + fshElementPath,
```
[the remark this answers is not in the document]
ngfreiter: Good point, sorry for the lack of comments, I will add some to clarify.

PR #189, opened 2020-02-03
Valueset Escape Syntax

ngfreiter:
This fixes https://standardhealthrecord.atlassian.net/browse/CIMPL-218.

This adds the ability to use the `^` syntax on ValueSets. The `^` syntax can be used to set metadata attributes on the ValueSet such as "publisher". Unlike on StructureDefinitions, this syntax can only be applied at the root level of the ValueSet, so the rules follow the form:
`* ^caretPath = value`.


src/import/FSHImporter.ts
```diff
@@ -342,6 +343,9 @@ export class FSHImporter extends FSHVisitor {
           valueSet.components.push(vsComponent);
         }
       });
+    caretValueRuleCtx.forEach(caretValueRule =>
+      valueSet.rules.push(this.visitCaretValueRule(caretValueRule))
```
cmoesel: The `caretValueRule` rule in the parser grammar allows for an optional path:
```
caretValueRule:     STAR path? caretPath EQUAL value;
```

The path, however, is not supported in ValueSets.  As this is currently written, I think we would just ignore any path a user specifies and just try to set it at the root.  Instead, if the user specified a path, we should issue an error and reject the rule.  I don't know if we want to do that here (in the importer) or in the exporter.  Either would probably work, and there are good arguments for each, so I'll leave it to you to decide!

ngfreiter: Oh good call. I added a message to the importer, so it'll be logged, and rules with a path will be dropped.

PR #190, opened 2020-02-04
Provide separate error message for missing vs. invalid package.json

### LINKED ISSUE #177: Error in package.json leads sushi to claim there is no package.json
Reproduce:
Make an error in package.json, like leaving an extra comma:
```
[...]
"dependencies": {
    "hl7.fhir.r4.core": "4.0.1",
 },
[...]
```
Run sushi and see that it says there is no spoon:
```
PS C:\Users\Ward\Workspace\fsh-tutorial-master\FishExample> sushi .
error: No package.json in FSH definition folder.
```

cmoesel:
Fixes #177

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -43,12 +43,20 @@ async function app() {
     program.help();
   }
 
+  // Check that package.json exists
+  const packagePath = path.join(input, 'package.json');
+  if (!fs.existsSync(packagePath)) {
+    logger.error('No package.json in FSH definition folder.');
+    return;
+  }
+
+  // Parse package.json
   let config: any;
   try {
-    config = JSON.parse(fs.readFileSync(path.join(input, 'package.json'), 'utf8').toString());
-  } catch {
-    logger.error('No package.json in FSH definition folder.');
-    program.help();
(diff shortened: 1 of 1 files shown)


cmoesel:
Yes, it was intentional.  I was thinking that a missing package.json wasn't likely a misunderstanding of the CLI, but rather just not knowing that package.json had to exist.  The `program.help()` doesn't say anything about package.json, so I didn't see how it would be helpful.

So in what case would misuse of the CLI lead to this?  If you run sushi w/ no argument, you get:
```
error: Missing path to FSH definition folder.
```

If you run it with an argument to a non-existent folder you get:
```
error: Invalid path to FSH definition folder.
```

So the only way you get to the point where you get the package.json missing error is if you've used SUSHI w/ an argument to an existing folder.  In which case, I'm not sure how the `program.help()` is going to add something...  Or are you thinking someone might put the output folder instead of the input folder?

ngfreiter:
I guess even if someone is putting output, the output has a package.json. You're right it seems unlikely that someone would be confused enough to hit this error without knowing how the arguments to SUSHI are supposed to go. So I'm fine with not adding it.

PR #195, opened 2020-02-04
Pages and images in IGs

jafeltra:
This PR adds support for user defined pages and images in IGs. Pages can be put in `ig-data/input/pagecontent` and can be `.xml` or `.md` files. They get moved over to the exported IG folder `output_dir/input/pagecontent` and get added to the IG definition after the `index.md` file. Other file types there will be moved over, but not added to the IG definition and therefore not added to the table of contents.

Images can be put in `ig-data/input/images`. They will get moved over to the exported IG folder `ouput_dir/input/pagecontent`. Pages can reference images directly and the image will load in the page.

I also talked to @cmoesel about supporting -intro and -note files in `ig-data/input/resource-docs` which the IG documentation mentions in the page folders section [here](https://build.fhir.org/ig/FHIR/ig-guidance/using-templates.html#igroot-input). However, it doesn't seem that anything from `output_dir/input/resource-docs` gets moved over into `output_dir/output` when you generate the IG, so I'm not sure if we want to include the support for that. I can add in the support for just doing that move (I have it locally), but wanted to check to see if it was even worth it.


jafeltra:
Updated this PR to not include intro/notes files in the IG definition.

src/ig/IGExporter.ts
```diff
@@ -195,21 +196,64 @@ export class IGExporter {
...
+        const isIntroOrNotesFile = fileName.endsWith('-intro') || fileName.endsWith('-notes');
+        if (isSupportedFileType && !isIntroOrNotesFile) {
+          // If it is a valid file type and not an intro or note file, we will also add it to IG definition
+          this.ig.definition.page.page.push({
+            nameUrl: `${fileName}.html`,
+            title: `${fileName}`,
+            generation: fileType === 'md' ? 'markdown' : 'html'
+          });
+        } else {
+          invalidFileTypeIncluded = true;
```
cmoesel: This looks to me like it's going to flag intro and notes files as invalid (because they will fall into this `else` clause).  I don't think that's what we want.

src/ig/IGExporter.ts
```diff
@@ -195,21 +196,64 @@ export class IGExporter {
...
-  private checkForOtherPageContent() {
+  private addOtherPageContent(igPath: string) {
     const inputPageContentPath = path.join(this.igDataPath, 'input', 'pagecontent');
     if (fs.existsSync(inputPageContentPath)) {
-      const pages = fs.readdirSync(inputPageContentPath);
-      if (pages.length > 1 || (pages.length === 1 && pages[0] !== 'index.md')) {
-        logger.error('SUSHI does not yet support custom pagecontent other than index.md.', {
+      const pages = fs
+        .readdirSync(inputPageContentPath)
+        .filter(page => page !== 'index.md')
```
cmoesel: I think we probably need to add special handling for `index.xml` too.  So this should really be:
```
        .filter(page => page !== 'index.md' && page !== 'index.xml')
```

In addition, we should update the `addIndex` function to check for an `index.xml` too and not generate an index if it already exists.  And to make sure we put in the right `generation` value depending on if the index is `index.md` or `index.xml`.

cmoesel: Sorry for putting all that here, but since `addIndex` isn't changed in this PR, GitHub doesn't make it easy for me to comment on it there!

src/ig/IGExporter.ts
jafeltra: Ah good point. I'll update.

jafeltra:
Just pushed updates that should support `index.xml` files and not log a warning about the intro/notes files.

cmoesel:
Code looks good.  I'm updating the branch (to bring in sync w/ master) and will test locally.

src/ig/IGExporter.ts
```diff
@@ -183,33 +179,91 @@ export class IGExporter {
...
+        const fileType = page.slice(page.lastIndexOf('.') + 1);
+        const isSupportedFileType = fileType === 'md' || fileType === 'xml';
+        const isIntroOrNotesFile = fileName.endsWith('-intro') || fileName.endsWith('-notes');
+        if (isSupportedFileType) {
+          // Intro and notes files will be in supported formats but are not separate pages, so they should not be added to IG definition
+          if (!isIntroOrNotesFile) {
+            // Valid page files will be added to the IG definition
+            this.ig.definition.page.page.push({
+              nameUrl: `${fileName}.html`,
+              title: `${fileName}`,
```
cmoesel: I noticed that when the file names are lowercase (which is probably often the case), it looks kind of funny in the TOC and on the pages:
image
and a page for example:
image

I think it might be nice if we tried to be a little smart about the titles.  I just did a quick test w/ some libraries and I think something like this might be nice:
```js
const { titleCase } = require("title-case");
const _ = require('lodash');

[
  'myTestFileWithoutSpaces',
  'MyTestFileWithoutSpaces',
  'my-test-file-with-dashes',
  'my_test_file_with_underscores'
].forEach(fileName => {
  console.log(titleCase(_.words(fileName).join(' ')));
});

// My Test File Without Spaces
// My Test File Without Spaces
// My Test File with Dashes
// My Test File with Underscores
```

jafeltra: I like this! I'll add it in now!

PR #196, opened 2020-02-05
Add info messages during import and export

mint-thompson:
Addresses https://standardhealthrecord.atlassian.net/browse/CIMPL-208.

Informational messages are logged during FSH import and FHIR
export. These messages indicate progress through the import and
export processes. Export messages will only appear if there were
FSH resources of the appropriate type in the tank being exported.

The additional informational messages had an unpleasant side effect
upon the test cases that monitored logged messages, so the loggerSpy
class now retrieves log entries and messages of a specific logging
level. When the level argument is omitted, log entries of any level
will be returned.


mint-thompson:
@cmoesel that's a good point regarding the usefulness when troubleshooting. I think having the messages at the end of each process is still useful, though. Do you think it would be a message overload to have messages both at the start at end of these steps?

ngfreiter:
At the risk of adding too many messages, I think it would also be good to have some messages for package loading. Maybe a message along the lines of:
`package <package-name> loaded from <location>`
where the location would either be their local FHIR cache (~/.fhir), or the FHIR package server (packages.fhir.org). I think that could be helpful in letting people validate that they are loading the packages they think they are loading. Also loading something like all of R4 from the package server can take relatively long.

I personally prefer too many messages to too few, but if you both think this would be too many messages I'll obey democracy.

mint-thompson:
The latest commit adds more info messages to the start of tasks (thank you @cmoesel ) and during package loading (thank you @ngfreiter ).

src/app.ts
```diff
@@ -97,10 +106,12 @@ async function app() {
   );
 
   // If ig-data exists, generate an IG, otherwise, generate resources only
+  logger.info('Exporting FHIR resources...');
```
ngfreiter: I think here it could be more clear to say 'Exporting FHIR resources as JSON...', just so it is clear why this export step is necessary, since people might assume when we say "FHIR resources" that we already mean JSON.

src/export/CodeSystemExporter.ts
```diff
@@ -68,13 +68,17 @@ export class CodeSystemExporter {
...
+    const codeSystems = this.tank.getAllCodeSystems();
+    for (const cs of codeSystems) {
       try {
         this.exportCodeSystem(cs);
       } catch (e) {
         logger.error(e.message, cs.sourceInfo);
       }
     }
+    if (codeSystems.length > 0) {
+      logger.info('Finished exporting FSH for FHIR code systems.');
```
ngfreiter: I think this message, and all the similar messages, could be marked as "FSH Export: ", similar to how you marked some messages as "FSH Import: ".

cmoesel:
I really like the addition of the info messages and ability to specify log levels in the test utility.  Thanks @mint-thompson!  I just pushed up a commit that changes some of the text in the messages and also adds some messages for the IG Exporter.  Can you take a look at my last commit and let me know if it looks OK to you?

With my changes, running against the fsh-mcode IG produces these logs:
```
info: Checking local cache for hl7.fhir.r4.core#4.0.1...
info: Found hl7.fhir.r4.core#4.0.1 in local cache.
info: Checking local cache for hl7.fhir.us.core#3.1.0...
info: Found hl7.fhir.us.core#3.1.0 in local cache.
info: Importing FSH text...
info: Preprocessed 14 documents with 38 aliases.
info: Imported 58 definitions and 27 instances.
info: Loaded package hl7.fhir.r4.core#4.0.1
info: Loaded package hl7.fhir.us.core#3.1.0
info: Converting FSH to FHIR resources...
info: Converted 31 FHIR StructureDefinitions.
info: Converted 27 FHIR ValueSets.
info: Converted 27 FHIR instances.
info: Building FHIR Implementation Guide...
info: Exporting FHIR resources as JSON for Implementation Guide...
info: Exported 85 FHIR resources as JSON for Implementation Guide.
info: Generated default index.md.
info: Generated default ig.ini.
info: Generated default package-list.json
info: Generated ImplementationGuide-fhir.us.mcode.json
info: Built FHIR Implementation Guide; ready for IG Publisher.
info:
  Profiles:    25
  Extensions:  6
  Instances:   27
  ValueSets:   27
  CodeSystems: 0
  Errors:      0
  Warnings:    0
```

PR #197, opened 2020-02-05
Unfold Content Reference Elements

### LINKED ISSUE #184: Bug: error: Cannot read property 'length' of undefined
The following FSH file generates the error:
error: Cannot read property 'length' of undefined

But the same input with the extension on `compose.include.extension` instead works just fine.

```fsh
Profile: BoostedValueSet
Parent: ValueSet
Id: boostedvalueset
Title: "Boosted ValueSet"
Description: "Any ValueSet that uses the ValueSetBoost extension"
* compose.exclude.extension contains ValueSetBoost 0..* 

Extension: ValueSetBoost
Id: boost
Title: "ValueSet boosting"
Description: "boosting extension"
* value[x] only decimal
```

ngfreiter:
Fixes #184.

When we access the children of an element that has a contentReference field on it, we need to unfold from the element on the StructureDefinition that the contentReference refers to. A contentReference is used to refer to an element defined in that same StructureDefinition. Documentation on contentReferences is here https://www.hl7.org/fhir/elementdefinition-definitions.html#ElementDefinition.contentReference. 

This adds the capability to access children of an element that is a contentReference. Accessing a child will unfold all the children of the referenced element onto the element with the contentReference. The contentReference element will also have the contentReference field deleted, since it is no longer a contentReference. If you want to test on a real example, try adding the Profile on ValueSet described in #184 to a FSH file. You should see that the "exclude" element gets unfolded with the children of the "include" element, and constraints are correctly applied to it.

### DIFF
--- a/src/fhirtypes/ElementDefinition.ts
+++ b/src/fhirtypes/ElementDefinition.ts
@@ -1408,16 +1408,42 @@ export class ElementDefinition {
 
   /**
    * If the element has a single type, graft the type's elements into this StructureDefinition as child elements.
+   * If the element is sliced, unfold from the sliced element on the StructureDefinition
+   * If the elemnet is a content reference, unfold from the referenced element on the StructureDefintion
    * @param {Fishable} fisher - A fishable implementation for finding definitions and metadata
    * @returns {ElementDefinition[]} the unfolded elements or an empty array if the type is multi-value or type can't
    *   be resolved.
    */
   unfold(fisher: Fishable): ElementDefinition[] {
     if (
-      this.type.length === 1 &&
-      (this.type[0].profile == null || this.type[0].profile.length <= 1)
+      (this.type?.length === 1 &&
+        (this.type?.[0].profile == null || this.type?.[0].profile.length <= 1)) ||
+      this.contentReference
     ) {
       let newElements: ElementDefinition[] = [];
+      if (this.contentReference) {
+        // If the element is a content reference, we need to unfold from the referenced element
+        // Get the original resource JSON so we unfold unconstrained reference
+        const type = this.structDef.type;
+        const json = fisher.fishForFHIR(type, Type.Resource);
+        if (json) {
+          const def = StructureDefinition.fromJSON(json);
+          // Content references start with #, slice that off to id of referenced element
+          const referencedElement = def.findElement(this.contentReference.slice(1));
+          newElements = referencedElement?.children().map(e => {
+            const eClone = e.clone();
+            eClone.id = eClone.id.replace(referencedElement.id, this.id);
+            eClone.structDef = this.structDef;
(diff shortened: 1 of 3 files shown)


src/fhirtypes/ElementDefinition.ts
```diff
@@ -1408,16 +1408,42 @@ export class ElementDefinition {
...
    * @returns {ElementDefinition[]} the unfolded elements or an empty array if the type is multi-value or type can't
    *   be resolved.
    */
   unfold(fisher: Fishable): ElementDefinition[] {
     if (
-      this.type.length === 1 &&
-      (this.type[0].profile == null || this.type[0].profile.length <= 1)
+      (this.type?.length === 1 &&
+        (this.type?.[0].profile == null || this.type?.[0].profile.length <= 1)) ||
+      this.contentReference
```
cmoesel: No offense, but this `if` statement is kind of ugly.  I'm trying to think of how it can be simplified or made neater.

If the logic made it through `this.type?.length === 1`, then we know `this.type` exists, so we don't need to do `this.type?` in `and`ed phrases after that.  I also think that line 1420 can be simplified.  I *think* this might be equivalent to the above:

```ts
(this.type?.length === 1 && !(this.type[0].profile?.length > 1)) ||
this.contentReference
```

Is that better or is the inversion more confusing?

ngfreiter: Am I missing something, or is there a reason not to just do:

ngfreiter: Isn't this the same:
```
(this.type?.length === 1 && this.type[0].profile?.length <= 1) ||
this.contentReference
```
Were you just adding the inversion because it seems more clear?

cmoesel: That's not the same because in that one, if `profile` is null then it evaluates to `false`, whereas we want a null profile to evaluate to `true`.  If you wanted to do it that way, we'd have to do something more like:
```ts
(this.type?.length === 1 && (this.type[0].profile ?? []).length <= 1) ||
this.contentReference
```
I'd be fine with that too if you think it is more clear.

ngfreiter: Ah wait nevermind, they are different if `profile` is null

cmoesel: I actually think that one I just did is more clear than my original suggestion.  So it's got my vote for whatever that is worth. ;-)

ngfreiter: Alright I'm sold, I'll do that last one. And I'll add an instance test

ngfreiter:
Ok @cmoesel, I updated to use that cheeky if statement, and to add a test to the instance exporter.

PR #202, opened 2020-02-07
Fix resolution of "current" IG dependencies

### LINKED ISSUE #201: Resolution of "current" IG dependencies broken
Changes in the the CI Build infrastructure have caused the resolution of "current" IGs to fail.  For example, if your package.json declares the following dependencies:
```json
  "dependencies": {
     "hl7.fhir.r4.core": "4.0.1",
     "hl7.fhir.uv.vhdir":"current"
  },
```
Then running SUSHI will now issue these errors:
```
error: Failed to load hl7.fhir.uv.vhdir#current
error: 404 - <html>
<head><title>404 Not Found</title></head>
<body>
<center><h1>404 Not Found</h1></center>
<hr><center>nginx/1.17.8</center>
</body>
</html>

error: Failed to add hl7.fhir.uv.vhdir:current to ImplementationGuide instance.  Could not determine its canonical URL from the FHIR cache.
```

More on the changes that caused this:
https://chat.fhir.org/#narrow/stream/179165-committers/topic/Build.20Problem/near/187594182

cmoesel:
Changes in the CI Build infrastructure broke the way we were finding the "current" version of IG dependencies.  This commit fixes the approach as described on Zulip here: https://chat.fhir.org/#narrow/stream/179165-committers/topic/Build.20Problem/near/187610137

Fixes #201

### DIFF
--- a/test/fhirdefs/load.test.ts
+++ b/test/fhirdefs/load.test.ts
@@ -68,30 +68,30 @@ describe('#loadDependency()', () => {
       if (options.uri === 'http://build.fhir.org/ig/qas.json') {
         return [
           {
-            url: 'http://hl7.org/fhir/hspc/ImplementationGuide/hspc',
-            name: 'HSPC Implementation Guide',
-            'package-id': 'hl7.fhir.hspc',
-            'ig-ver': '1.0',
-            date: 'Thu, 11 Oct, 2018 11:00:14 -0600',
-            errs: 34,
-            warnings: 10,
-            hints: 0,
-            version: '3.0.1',
-            tool: '3.4.0-13844',
-            repo: 'nrdavis1/HSPCFHIRtest/branches/oldbranch/qa.json'
+            url: 'http://hl7.org/fhir/us/core/ImplementationGuide/hl7.fhir.us.core.r4-4.0.0',
+            name: 'USCoreR4',
+            'package-id': 'hl7.fhir.us.core.r4',
+            'ig-ver': '4.0.0',
+            date: 'Sat, 18 May, 2019 01:48:14 +0000',
+            errs: 538,
+            warnings: 34,
+            hints: 202,
+            version: '4.0.0',
+            tool: '4.1.0 (3)',
+            repo: 'HL7Imposter/US-Core-R4/branches/oldbranch/qa.json'
           },
           {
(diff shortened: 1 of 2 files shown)


test/fhirdefs/load.test.ts
```diff
@@ -148,7 +148,7 @@ describe('#loadDependency()', () => {
       }
     ]);
     expect(requestSpy.mock.calls[1][0].uri).toBe(
-      'http://build.fhir.org/ig/nrdavis1/HSPCFHIRtest/branches/package.tgz'
+      'http://build.fhir.org/ig/nrdavis1/HSPCFHIRtest/package.tgz'
```
ngfreiter: I feel like this no longer totally tests that the sorting by date is working anymore, since no matter which package you get, the `repo` starts with `nrdavis1/HSPCFHIRtest`. I think the date sort would only matter if there was a package that was being published with the same `package-id` from two different repos. Which I don't see an example of in qas.json. But if we are going to leave the date sort in, I feel like we should still test it here by changing the repo of the older qa entry.

cmoesel: Yeah, I mentioned the same thing to Grahame -- the date sort seems unnecessary at this point.  But in the off-chance that a package does switch repo names, I guess it might matter.

I take your point though.  If date sort was broken, we'd have no idea.  So I'll update the test.  I also realized this morning that I think my logic has an extra `slice(0,2)` that's totally not necessary since I'm using destructuring assignment (the slice is left over from an earlier version that did not use destructuring assignment).

cmoesel: OK, @ngfreiter -- this is fixed in the last commit.  I also switched up the test a little so that poor unsuspecting nrdavis1 is no longer in our testcases.

cmoesel:
> I don't know, I think anyone would be lucky to have their name randomly mentioned in our test cases.

I'll make sure I put `ngfreiter` all over my next set of test cases then.
```ts
const ngfreiter = 'ngfreiter://ngfreiter.ngfreiter/ngfreiter/';
switch (someone) {
  case 'ngfreiter': doHappyDance(); break;
  default: // who cares
}
// etc.
```

ngfreiter:
😊 I'm famous 😊

Issue #206, opened 2020-02-08
Required fields in Instance Validation


davidhay25:
When generating instances, sushi checks that the elements exist in the profile - would be good to check for required fields as well...

cmoesel:
Thanks for reporting this, @davidhay25.  As it so happens, @ngfreiter just started working on this on Friday -- so you should be able to expect it in the next release.

PR #209, opened 2020-02-11
Log errors for missing required elements

ngfreiter:
This fixes https://github.com/FHIR/sushi/issues/206, which is also tracked by https://standardhealthrecord.atlassian.net/browse/CIMPL-201.

An element is required if it's minimum cardinality is greater than 1. On an instance, any element that is at the top level of the instance that has min cardinality greater than 1 should be on the instance, and any nested element with min cardinality greater than 1 should be on the instance if its parent is on the instance. Additionally, these rules generalize to an array, which must have n >= m elements if it has min cardinality m.

This only checks that the cardinality is greater than the min cardinality, the functions that set values on an instance do not allow setting elements greater than the max cardinality, so that case is already handled. Also, this logs an error if a required element is missing, but the instance is still generated, even though it is invalid. This would be easy to change if we decide we want to avoid generating invalid instances.

I also refactored how some of the cleaning up happens on instances, and renamed a variable that had a confusing name.


src/export/InstanceExporter.ts
```diff
@@ -155,6 +142,98 @@ export class InstanceExporter {
...
+    children.forEach(child => {
+      // Get the last part of the path, A.B.C => C
+      const childPathEnd = child.path.split('.').slice(-1)[0];
+      let instanceChild = instance[childPathEnd];
+      // If the element is a choice, we will fail to find it, we need to use the choice name
+      if (!instanceChild && childPathEnd.endsWith('[x]')) {
+        const choiceSlices = children.filter(c => c.path === child.path && c.sliceName);
+        for (const choiceSlice of choiceSlices) {
+          instanceChild = instance[choiceSlice.sliceName];
+          if (instanceChild) {
```
cmoesel: What if the `instanceChild's` value is a boolean (e.g., `valueBoolean: false`)?  I think you should actually just be checking if it's defined, right?

src/export/InstanceExporter.ts
```diff
@@ -155,6 +142,98 @@ export class InstanceExporter {
...
+            break;
+          }
+        }
+      }
+      // Recursively validate children of the current element
+      if (instanceChild?.constructor === Array) {
```
cmoesel: Prefer:
```ts
if (Array.isArray(instanceChild)) {
```

src/export/InstanceExporter.ts
```diff
@@ -155,6 +142,98 @@ export class InstanceExporter {
...
+          this.validateRequiredChildElements(arrayEl, child, fshDefinition)
+        );
+      } else if (instanceChild) {
+        this.validateRequiredChildElements(instanceChild, child, fshDefinition);
+      }
+      // Log an error if:
+      // 1 - The child element is 1.., but not on the instance
+      // 2 - The child element is n..m, but it has k < n elements
+      if (
+        (child.min > 0 && !instanceChild) ||
```
cmoesel: Again -- if the `instanceChild` exists, but its value is `false` (or `0`), then this would be wrong, right?  I think we need to check that it's defined?

src/export/InstanceExporter.ts
```diff
@@ -155,6 +142,98 @@ export class InstanceExporter {
...
+        (instanceChild?.constructor === Array && instanceChild.length < child.min)
```
cmoesel: And another opportunity for `Array.isArray(...)`.

src/export/InstanceExporter.ts
```diff
@@ -155,6 +142,98 @@ export class InstanceExporter {
...
+      ) {
+        // Can't point to any specific rule, so give sourceInfo of entire instance
+        logger.error(
+          `Element ${child.id} has minimum cardinality ${child.min} but occurs ${
+            instanceChild ? instanceChild.length : 0
+          } time(s).`,
+          fshDefinition.sourceInfo
```
cmoesel: Very nice!

test/export/InstanceExporter.test.ts
```diff
@@ -897,6 +901,146 @@ describe('InstanceExporter', () => {
...
+      cardRule.min = 1;
+      cardRule.max = '1';
+      patient.rules.push(cardRule);
+      exportInstance(patientInstance);
+      expect(loggerSpy.getLastMessage('error')).toMatch(
+        /Patient.active.*File: PatientInstance\.fsh.*Line: 10 - 20/s
+      );
+    });
+
+    it('should log multiple errors when multiple required elemenets are not present', () => {
```
cmoesel: I think `elemenets` is kind of a fun word to say out loud.

test/export/InstanceExporter.test.ts
```diff
@@ -897,6 +901,146 @@ describe('InstanceExporter', () => {
...
+      const messages = loggerSpy.getAllMessages('error');
+      expect(messages[messages.length - 2]).toMatch(
+        /Patient.active.*File: PatientInstance\.fsh.*Line: 10 - 20/s
+      );
+      expect(messages[messages.length - 1]).toMatch(
+        /Patient.gender.*File: PatientInstance\.fsh.*Line: 10 - 20/s
+      );
+    });
+
+    it('should log an error when an element required by a fixed parent is not present', () => {
```
cmoesel: This confused me for a moment.  Maybe change the description to say:
> ... required by _an incomplete_ fixed parent...

or something like that?

test/export/InstanceExporter.test.ts
```diff
@@ -897,6 +901,146 @@ describe('InstanceExporter', () => {
...
+      expect(messages[messages.length - 3]).toMatch(
+        /DiagnosticReport.result.*File: LipidInstance\.fsh.*Line: 10 - 20/s
+      );
+      expect(messages[messages.length - 2]).toMatch(
+        /DiagnosticReport.result:Triglyceride.*File: LipidInstance\.fsh.*Line: 10 - 20/s
+      );
+      expect(messages[messages.length - 1]).toMatch(
+        /DiagnosticReport.result:HDLCholesterol.*File: LipidInstance\.fsh.*Line: 10 - 20/s
+      );
+    });
```
cmoesel: Can we add one test that ensures that it validates cardinalities that came from a parent SD not authored in FSH?  I think all of these tests are based on constraints in the direct profile -- but it would be good to have a test that proves it works when the constraint is inherited too.  `Observation.code` is `1..1` -- so that would be a pretty easy one.

Maybe two tests actually -- one that tests that an `InstanceOf: Observation` requires `code` -- and another that tests that an `InstanceOf: MyObservationProfile` requires `code`.  Then I think we have all our bases covered.

test/export/InstanceExporter.test.ts
ngfreiter: Pronounced: A-lemon-ets

cmoesel: It depends.  In some parts of the country they say: Ella-mennets.  The difference is mainly on whether the stress is on the first or second syllable.  Non-native speakers, however, sometimes place the stress on the 4th syllable, which gives it an entirely different feel.

ngfreiter:
Ok @cmoesel should be ready for re-review.

test/export/InstanceExporter.test.ts
ngfreiter: When in doubt the original French elleh-moh-nay always impresses

PR #211, opened 2020-02-12
Write resources out to type-based folders

cmoesel:
Instead of writing resources to "resources", write them to:
- profile
- extensions
- vocabulary
- examples
according to their type and purpose.

This also consolidates the code that actually writes the
files into a single location (instead of writing them in
app.js sometimes and IGExporter other times.


src/app.ts
```diff
@@ -94,40 +94,39 @@ async function app() {
...
-  const resourceDir = path.join(program.out, 'input', 'resources');
-  fs.ensureDirSync(resourceDir);
-
   fs.writeFileSync(
     path.join(program.out, 'package.json'),
     JSON.stringify(outPackage.config, null, 2),
     'utf8'
   );
 
+  logger.info('Exporting FHIR resources as JSON...');
```
cmoesel: Now we always write the JSON files here.  The IG Exporter is no longer responsible for doing that separately.

src/fhirtypes/CodeSystem.ts
```diff
@@ -10,6 +11,7 @@ import { applyMixins } from '../utils';
  * @see {@link https://www.hl7.org/fhir/codesystem.html}
  */
 export class CodeSystem {
+  readonly resourceType = 'CodeSystem';
```
cmoesel: You'll see I added a `resourceType` property to several of our classes.  This makes it easier to identify the type at run-time, and is used when writing out the files.

src/app.ts
```diff
@@ -94,40 +94,39 @@ async function app() {
...
+  let count = 0;
+  const writeResources = (
+    folder: string,
+    resources: { getFileName: () => string; toJSON: () => any }[]
```
mint-thompson: Aw yeah, there's that good anonymous interface.

PR #212, opened 2020-02-12
Include sliceName in StructureDefinition diff

mint-thompson:
When an element with a sliceName is included in a differential, the
sliceName attribute is included. This is done in order to meet the
needs of the IG publisher.

A skipped test is added to prevent the sliceName attribute from
being changed by a CaretValueRule.


src/fhirtypes/ElementDefinition.ts
```diff
@@ -323,6 +323,10 @@ export class ElementDefinition {
     if (this.path.endsWith('[x]') && !diff.path.endsWith('[x]')) {
       delete diff.sliceName;
     }
+    // If the original has a sliceName, the diff must include it, even if it didn't change
+    if (original.sliceName && diff.sliceName == null) {
+      diff.sliceName = original.sliceName;
+    }
```
ngfreiter: I think setting the sliceName misses the special case of when the element is a choice, but it has been constrained to a specific choice. In this case, we replace the `value[x]:valueString` element with just a `valueString` element, but only on the differential. It's kinda an annoying special case, but that is what FHIR likes us to do, based on this post: https://blog.fire.ly/2019/09/13/type-slicing-in-fhir-r4/.

So with the addition of this code, when the element is a choice that has been constrained to one element, the diff gets removed on line 324, but then added back on 328. This makes it so the IG looks like this:
image
Instead of this:
image

I tested out building with the IG publisher and manually removing those extra sliceNames, and I didn't run into any IG publisher errors.

mint-thompson: Augh! Okay, I think making those cases exclusive should avoid this. Update coming soon.

PR #213, opened 2020-02-12
Add support for description on instances

cmoesel:
Instances can now use the Description keyword.  The description will be put into the ImplementationGuide JSON and rendered in the documentation.

Other changes:
- the InstanceDefinition now carries metadata in _instanceMeta
- the ImplementationGuide will use an instance's Title as its name if it is set


cmoesel:
NOTE: I just tacked on a commit to update a couple of log messages, as requested in #214.

src/ig/IGExporter.ts
```diff
@@ -294,9 +294,10 @@ export class IGExporter {
     examples.forEach(example => {
```
ngfreiter: Thank you GitHub for not letting me comment on the line I want, but 293 is 
`const examples = sortBy(this.pkg.instances, instance => instance.id ?? instance.instanceName);`
This should be changed to use `instance._instanceMeta.instanceName` right?

src/fhirtypes/InstanceDefinition.ts
```diff
@@ -2,6 +2,7 @@ import cloneDeep = require('lodash/cloneDeep');
 import { Meta } from './specialTypes';
 import { HasId } from './common';
 import { applyMixins } from '../utils';
+import { isNull } from 'util';
```
ngfreiter: As GitHub has pointed out, 'isNull' is defined but never used.

src/ig/IGExporter.ts
cmoesel: Yes.  Not sure how that one escaped my global search!

src/fhirtypes/InstanceDefinition.ts
cmoesel: Ah, dang.  I remember that happening.  VS Code auto-completed something to `isNull`, which I did not intend -- so I fixed it, but hadn't realized that VS Code also auto-added an import for it.  Doh.

cmoesel:
@ngfreiter -- fixed!

PR #215, opened 2020-02-12
Infer simple or complex extensions

### LINKED ISSUE #183: Change extension card to 0..0 in simple extensions; value[x] to 0..0 in complex extensions
It seems FSHers have found it necessary to explicitly zero out the unused elements in extension (depending on whether it is complex or simple). Seems not to complete the publishing process unless they do this. But SUSHI should know which is which and zero out the appropriate element.

jafeltra:
This PR adds support for inferring if a user is creating a simple or complex extension. It fixes #183. This is accomplished by preprocessing Extensions.

If rules are only set on `value[x]` or any of it's elements and the user doesn't constrain the cardinality of `extension`, we add a new rule to constrain `extension` to 0..0, If rules are set only on `extension`, we add a new rule to constrain `value[x]`. If a user applies rules to both properties, we will log an error and not try to constrain either property and will allow the rules to be set.

### DIFF
--- a/src/export/StructureDefinitionExporter.ts
+++ b/src/export/StructureDefinitionExporter.ts
@@ -167,6 +167,32 @@ export class StructureDefinitionExporter implements Fishable {
     }
   }
 
+  /**
+   * Does any necessary preprocessing of extensions.
+   * @param {Extension} fshDefinition - The extension to do preprocessing on. It is updated directly based on processing.
+   */
+  private preprocessExtension(fshDefinition: Extension): void {
+    fshDefinition.rules.forEach(rule => {
+      if (rule.path.startsWith('extension')) {
+        // If Extension.extension is used, constrain cardinality of Extension.value[x] to 0..0
+        if (!(rule instanceof CardRule && rule.max === '0')) {
+          const valueCardRule = new CardRule('value[x]');
+          valueCardRule.min = 0;
+          valueCardRule.max = '0';
+          fshDefinition.rules.push(valueCardRule);
+        }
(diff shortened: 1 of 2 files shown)


ngfreiter:
Do we also want to apply these rules to subextensions? So for example, if I am making the following extension:
```
Extension: RaceExtension
* extension contains ombCategory 0..5 and detailed 0..* and text 1..1
* extension[ombCategory].value[x] only Coding
* extension[detailed].value[x] only Coding
* extension[text].value[x] only string
```
Then `Extension.value[x]` does correctly get zeroed out, but it also looks like `Extension.extension:text.extension` does not get zeroed out. In general it seems like this PR will work well when an Extension is defined using the `Extension:` keyword, but I don't think it will work when we are "inline" defining Extensions. For consistency, it would make sense that in the example above, the `Extension.extension:<name>` elements get the same treatment as Extensions being defined using `Extension:`.

cmoesel:
Ah.  Good call, @ngfreiter.  @jafeltra -- what do you think?  Merge this PR and add another story for inline extensions?  Or do you want to hold this up until you have inline extensions working too?  I'm fine either way.

jafeltra:
I'm fine either way as well!

ngfreiter:
Since both of you are fine either way, I'm going to make the call, let's not hold this up. I will review this again and we can add a task to handle inline Extensions.

ngfreiter:
Ok I will merge, I made this for inline extensions: https://standardhealthrecord.atlassian.net/browse/CIMPL-269.

PR #219, opened 2020-02-14
Account for pattern on parent

### LINKED ISSUE #84: Fixed codes don't account for parent elements w/ pattern[x]
Currently, when the user tries to fix a value, we check the Structure Definition to see if that value has already been set on the element by a pattern or fixed field. However, the pattern field can be used to specify which values children of the element must have. We do not account for this. For example, if Patient.maritalStatus has a field like:

```
"patternCodeableConcept": {
  "coding": [
    { "code": "foo" }
  ]
}
```

We currently still allow the user to re-fix the value of Patient.maritalStatus.coding[0].code. This behavior is incorrect, and should be fixed to not allow this value to be fixed to something other than "foo"

ngfreiter:
Fixes #84.

Previously when fixing values, we did not account for values being fixed via their parents by a pattern. This PR adds support for checking if a value is fixed by pattern from its parent, and we now throw an error if we try to re-set a value that is fixed by pattern.

If you'd like to test it out, fixing `CodeableConcept` elements works with pattern, so they provide a good test case. For example, the following will now generate an error, which was not the case with master:
```
Profile: myPatient
Parent: Patient
* maritalStatus = http://wedding.com#married
* maritalStatus.coding.code = #plentyoffish
```

### DIFF
--- a/test/fhirtypes/ElementDefinition.fixValue.test.ts
+++ b/test/fhirtypes/ElementDefinition.fixValue.test.ts
@@ -100,27 +100,105 @@ describe('ElementDefinition', () => {
         display: 'bar'
       });
     });
+
+    it('should throw ValueAlreadyFixedError when fixing a value fixed via parent pattern', () => {
+      const medicationForm = medication.elements.find(e => e.id === 'Medication.form');
+      medicationForm.fixValue(new FshCode('foo', 'http://thankYouForSettingMe.com'));
+      const medicationFormCodingSystem = medication.findElementByPath('form.coding.system', fisher);
+      expect(() => {
+        medicationFormCodingSystem.fixValue('http://ohManIWillNeverGetSet.sad');
+      }).toThrow(
+        'Cannot fix http://ohManIWillNeverGetSet.sad to this element; a different uri is already fixed: http://thankYouForSettingMe.com'
+      );
+    });
   });
-  describe('#fixedByParent', () => {
+
+  describe('#fixedByDirectParent', () => {
     it('should find a pattern value from the parent when it exists', () => {
       const statusReason = medicationRequest.elements.find(
         e => e.id === 'MedicationRequest.statusReason'
       );
       statusReason.fixValue(new FshCode('foo'));
       const statusReasonCoding = medicationRequest.findElementByPath('statusReason.coding', fisher);
-      const patternValue = statusReasonCoding.fixedByParent();
+      const patternValue = statusReasonCoding.fixedByDirectParent();
       expect(patternValue).toEqual([{ code: 'foo' }]);
     });
 
     it('should not find a pattern value from the parent when none is present', () => {
       const statusReasonCoding = medicationRequest.findElementByPath('statusReason.coding', fisher);
-      const patternValue = statusReasonCoding.fixedByParent();
+      const patternValue = statusReasonCoding.fixedByDirectParent();
       expect(patternValue).toBeUndefined();
     });
 
     it('should return undefined when being run on the root element', () => {
       const root = medicationRequest.elements.find(e => e.id === 'MedicationRequest');
-      const patternValue = root.fixedByParent();
+      const patternValue = root.fixedByDirectParent();
+      expect(patternValue).toBeUndefined();
+    });
+  });
+
+  describe('#fixedByAnyParent', () => {
+    it('should find a pattern value from a direct parent when it exists', () => {
+      const statusReason = medicationRequest.elements.find(
+        e => e.id === 'MedicationRequest.statusReason'
+      );
+      statusReason.fixValue(new FshCode('foo'));
+      const statusReasonCoding = medicationRequest.findElementByPath('statusReason.coding', fisher);
+      const patternValue = statusReasonCoding.fixedByAnyParent();
+      expect(patternValue).toEqual([{ code: 'foo' }]);
+    });
+
+    it('should find a pattern value from a grandparent when it exists', () => {
+      const identifier = medicationRequest.elements.find(
+        e => e.id === 'MedicationRequest.identifier'
+      );
+      identifier.max = '1';
+      // @ts-ignore
+      identifier.patternIdentifier = { period: { start: '2011-11-11' } };
+      const identifierPeriodStart = medicationRequest.findElementByPath(
+        'identifier.period.start',
+        fisher
+      );
+      const patternValue = identifierPeriodStart.fixedByAnyParent();
+      expect(patternValue).toBe('2011-11-11');
+    });
+
+    it('should find an array pattern value from a grandparent when it exists', () => {
+      const statusReason = medicationRequest.elements.find(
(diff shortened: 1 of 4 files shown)


test/fhirtypes/ElementDefinition.fixValue.test.ts
```diff
@@ -100,27 +100,105 @@ describe('ElementDefinition', () => {
         display: 'bar'
       });
     });
+
+    it('should throw ValueAlreadyFixedError when fixing a value fixed via parent pattern', () => {
+      const medicationForm = medication.elements.find(e => e.id === 'Medication.form');
+      medicationForm.fixValue(new FshCode('foo', 'http://thankYouForSettingMe.com'));
+      const medicationFormCodingSystem = medication.findElementByPath('form.coding.system', fisher);
+      expect(() => {
+        medicationFormCodingSystem.fixValue('http://ohManIWillNeverGetSet.sad');
```
cmoesel: Are you just checking to see if I actually read these PRs? ;-)

test/fhirtypes/ElementDefinition.fixValue.test.ts
```diff
@@ -100,27 +100,105 @@ describe('ElementDefinition', () => {
...
+
+      // Multiple not matching array elements
+      statusReason.patternCodeableConcept = { coding: [{ system: 'foo' }, { system: 'bar' }] };
+      patternValue = statusReasonCodingSystem.fixedByAnyParent();
+      expect(patternValue).toBeUndefined();
+
+      // Multiple matching array elements
+      statusReason.patternCodeableConcept = { coding: [{ system: 'foo' }, { system: 'foo' }] };
+      patternValue = statusReasonCodingSystem.fixedByAnyParent();
+      expect(patternValue).toBe('foo');
```
cmoesel: I'm trying to figure out how the implications of this test (and the behavior behind it) play out in the real world.  It kind of seems like this:

When I am trying to fix an array element:
* If a parent pattern sets a _single value_ in the array, I will get an error unless my value matches the single value.
* If a parent pattern sets _multiple, but equal, values_ in the array, I will get an error unless my value matches the value repeated in the pattern.
* If a parent pattern sets _mutliple different values_ in the array, I will not get an error?  SUSHI will let me set a new value?

If that's how it works, it seems a bit inconsistent to me -- but maybe you can explain the reasoning for that approach?  Or maybe I'm understanding it wrong...

ngfreiter: Ah you make a good point. I guess my original thinking was that when multiple different values are set in an array, the `fixedByAnyParent` function cannot determine any single fixed value, so we return undefined. But you're right, that is an issue because then there is no way to differentiate between undefined because no value is set, and undefined because conflicting values are set, and those cases need to be handled very differently.

So then the question is, what to return when there are multiple conflicting values? I'm actually  not totally sure, I need to think about that.

cmoesel: Is there a reason why you're flattening arrays into a single value before you return them?  If the path is fixed to an array, why not return the array?  And then let the code that calls `fixedByAnyParent` decide what to do based on context?  Or does that mess up the approach to instances?

ngfreiter: If the value itself is fixed to an array by a pattern, that does not get flattened. So as an example, let's say `code` is fixed as:
```
patternCodeableConcept: {
  coding: [ { system: 'foo' }, { system: 'bar' }]
}
```
Then `code.coding` is fixed to `[ { system: 'foo' }, { system: 'bar' }]`, and that is what gets returned by the `fixedByAnyParent` function. But when we run `fixedByAnyParent` on `code.coding.system` I didn't think there was a clear choice for what we should return. Since `system` is set to two different values in the different array elements. 

My thinking was, if all the `system`s in the array match, and they also match the new value you are trying to set, that is fine. If the `system`s in the array do not match, then you can't really set `system` to anything else, because no matter what you fixed it to, that would conflict with one of the values dictated by the pattern. So it was in this situation that I returned `undefined`, but I now see that doesn't entirely work.

cmoesel: Ah yes.  Right.  Got it.

I still think maybe we should think about returning an array.  I suppose we have two possibilities:

1) Always return an array of the collected values for the path -- so in the example above, `code.coding.system` would result in `['foo', 'bar']`.  In an example where the systems are the same, it would be `['foo', 'foo', 'foo']`.  In an example where only one is set, it would still be `['foo']` (since there is an array in its path).  Then the calling code handles it -- probably by checking that every element in the array matches the new fixed value.  That would handle all those cases correctly.

2) Still return a single value when there is only one or all the values are equal.  But in cases where they are different, return an array of the different values.  Then when the code checks to see if the new fixed value equals the old one, it will fail because an array never equals a single item.

Anyway, those are a couple of ideas.

ngfreiter: I think that makes sense, I don't like how in option 2, you wouldn't have any way to differentiate between an element that is fixed to multiple different values in different array elements on a parent, and an element that is actually fixed to an array. I think in option 1, these would be different, and element fixed to multiple values by an array on a parent would be `['foo', 'bar']`, where an element actually fixed to an array would be `[['foo', 'bar']]`. 

Let me know if that makes sense to you, so I know we are thinking of this the same way.

cmoesel: I was thinking we'd flatten to a single array, but I see your point.  I guess the question is if there is a case where we need to know the difference?  I guess maybe when generating instances?

As for `[['foo', 'bar']]`, I'm assuming that means that the tail of the path is an array, but so is one of it's parent elements?  In other words, however many dimensions the array is, that is how many array elements are in its path.  But if there is only one element in the path that is an array, then it is a single-dimension array (e.g., `[ 'foo', 'bar' ]` no matter where the array is introduced in the path, right?  If so, I'm down with that if you don't think it will be too hard to implement.

ngfreiter: Is it too hacky for you if I just return `[]` instead of `undefined` when multiple array elements exist that conflict? I think this will work fine because `[]` is never going to equal the value the user is trying to set, but it isn't `undefined`, so the user won't be able to reset the value. The approaches we have been discussing are less hacky, but complicated enough to implement that I'm not sure it's worth it.

ngfreiter: ^ Scrub that from your mind, won't work well

ngfreiter: > I don't like ... option 2
- Me

Yeah so I went with option 2 in the latest commit, I think it'll work well!

cmoesel: Ha!

PR #220, opened 2020-02-14
No snapshot

ngfreiter:
This PR stops snapshots from being generated by default. To generate snapshots, the user can use the `-s` or `--snapshot` flag.

My method for testing that this doesn't cause issues in the IG publisher was to run the IG publisher using output both with and without snapshots. I then compared the generated JSON output that the IG publisher created, and looked for any major divergence that wasn't just something like a changed timestamp. This was manual and not super efficient, so if any reviewers have a better way let me know what worked.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ Usage: sushi <path-to-fsh-defs> [options]
 Options:
   -o, --out <out>  the path to the output folder (default: "build")
   -d, --debug      output extra debugging information
+  -s, --snapshot   generate snapshot in Structure Definition output (default: false)
   -v, --version    print SUSHI version
   -h, --help       output usage information
 ```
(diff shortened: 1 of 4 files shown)


cmoesel:
What FSH did you run this test on?  Random sample FSH, fsh-mcode, or something else?

ngfreiter:
fsh-mcode for me.

cmoesel:
I was hoping you would say that!

jafeltra:
I did notice one additional thing after I had approved that I just wanted to double check on. When comparing the generated files `StructureDefinition-mcode-cancer-disease-status.json` in `output`, I noticed there were a couple differences between what was generated on master and what was generated on this branch. They don't seem too harmful, but I'm not entirely sure where/how they were changed, so I wanted to ask. The only difference in the corresponding SD in `input` is the missing `snapshot` on this branch.

I can also attach the two files if that's easiest for anyone.

ngfreiter:
At least some of these changes do seem to be positive changes. For example, the link http://hl7.org/fhir/us/mcode/extensibility.html is now http://hl7.org/fhir/R4/extensibility.html, changing it from a broken link into one that works.

PR #222, opened 2020-02-14
Support user provided IG menu.xml

jafeltra:
This PR adds support for a user provided `menu.xml` file. The file should be put in `ig-data/input/includes`. As part of this, it also supports copying over any other files that are placed in that folder. The menu and any additional files will be placed into `outputDir/input/includes/`.

### DIFF
--- a/src/ig/IGExporter.ts
+++ b/src/ig/IGExporter.ts
@@ -47,6 +47,7 @@ export class IGExporter {
     this.addIndex(outPath);
     this.addOtherPageContent(outPath);
     this.addImages(outPath);
+    this.addMenu(outPath);
     this.addIgIni(outPath);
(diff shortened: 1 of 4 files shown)


src/ig/IGExporter.ts
```diff
@@ -180,7 +181,7 @@ export class IGExporter {
    *
    * @param igPath {string} - the path where the IG is exported to
    */
-  private addIndex(igPath: string) {
+  private addIndex(igPath: string): void {
```
cmoesel: Oh, look at you adding return types!  Thanks for cleaning up after my slopping coding! ;-)

jafeltra: Well some of them were my sloppy coding ;)

PR #225, opened 2020-02-17
Add Invariant and ObeysRule

ngfreiter:
This PR adds an Invariant class and the corresponding ObeysRule. The Invariant class is based on the "constraint" field on an ElementDefinition https://www.hl7.org/fhir/elementdefinition-definitions.html#ElementDefinition.constraint. The ObeysRule is used to apply an invariant to an element. This PR doesn't do anything with the classes, just adds them.

### DIFF
--- a/src/fshtypes/Invariant.ts
+++ b/src/fshtypes/Invariant.ts
@@ -1,0 +1,19 @@
+import { FshEntity, FshCode } from '.';
+
+/**
+ * The Invariant class is used to represent the "constraint" field on ElementDefinition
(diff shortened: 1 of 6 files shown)


ngfreiter:
Yes you are correct. I was trying to model it after the `ValueSetRule`, where `ValueSetRule.valueSet` is used to match some ValueSet. In the case of Invariants, it doesn't really make sense to have them externally defined, since they aren't an actual entity in FHIR, so yes the only thing that would make sense there is `Invariant.name`.

PR #226, opened 2020-02-17
Sort pages in IG content

mint-thompson:
Numeric prefixes in IG content files enable sorting. The prefixes get removed during processing, unless doing so would cause problems. Because this code involves dealing with IG content files, there are some small content files added in to support the tests.

The case where two files have the same name other than their extension, for example `Something.xml` and `Something.md`, is not accounted for here.

This pull request is for task https://standardhealthrecord.atlassian.net/browse/CIMPL-268.


mint-thompson:
Local type defined! I also fixed up a few of my jsdoc comments that were in the wrong order.

Issue #217, opened 2020-02-13
Change error message color from red to something more visible


markkramerus:
It was agreed at the connectathon that red on black is a poor color combination, hard to read on some screens. I don't have a recommendation on the color, but it should be bright. 
I'm not sure if we need to simultaneously consider the possibility of white backgrounds.

PR #233, opened 2020-02-18
Error message color

ngfreiter:
Fixes https://github.com/FHIR/sushi/issues/217.

This PR removes the coloring from the text of error messages, and puts coloring on the background of the level of the error message. So that error messages now look like this:
image

I also had to change the regex in a lot of the tests. We were testing for line numbers like this `*Line: 1 - 4\D`, and the `\D` was matching any non-numeric character. When the error message itself was colored, it seems this would match, since there were some characters appended to the end of the string to handle coloring. Now that the messages are uncolored, the `\D` was not matching our messages. So these all got changed to the form `*Line: 1 - 4\D*`.

@jafeltra and @cmoesel at least one of you should probably test this to make sure it works and looks fine on a Mac.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -134,6 +134,28 @@
         "chalk": "^2.0.0",
         "esutils": "^2.0.2",
         "js-tokens": "^4.0.0"
+      },
+      "dependencies": {
+        "chalk": {
+          "version": "2.4.2",
+          "resolved": "https://registry.npmjs.org/chalk/-/chalk-2.4.2.tgz",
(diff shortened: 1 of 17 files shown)


ngfreiter:
Ok I made "warn" darker orange so that now it looks like this:
image
I tested it out and it seems like there is high contrast between the text and background color for color blindness too (thanks @mint-thompson!).

jafeltra:
The warn color looks better on the white-background terminal that I was looking at before.
image

PR #236, opened 2020-02-18
Recurse directories when gathering FSH definitions

peturv:
Added a flag to allow for recursion into subdirectories when gathering FSH definitions to allow for a directory based structure for larger projects.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -5,9 +5,9 @@ import fs from 'fs-extra';
 import program from 'commander';
 import { importText, FSHTank, RawFSH } from './import';
 import { exportFHIR } from './export';
-import { IGExporter } from './ig/IGExporter';
-import { logger, stats } from './utils/FSHLogger';
-import { loadDependency } from './fhirdefs/load';
+import { IGExporter } from './ig';
+import { logger, stats } from './utils';
+import { loadDependency } from './fhirdefs';
 import { FHIRDefinitions } from './fhirdefs';
 
 app();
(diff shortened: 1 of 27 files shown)


src/app.ts
```diff
@@ -155,3 +160,9 @@ function getVersion(): string {
   }
   return 'unknown';
 }
+
+function getFilesRecursive(dir: string): string[] {
+  return fs.statSync(dir).isDirectory()
+    ? [].concat(...fs.readdirSync(dir, 'utf8').map(f => getFilesRecursive(path.join(dir, f))))
+    : [dir];
```
cmoesel: This is stylistic, but it seems like there is a lot happening here in a single statement.  For readability, you might consider a more traditional `if/else` clause.

cmoesel:
Thanks again for the PR!  As noted in #239, I'd ask that you go ahead and remove the flag, making recursive traversal the norm.  I also made one comment that is more of a stylistic preference.  I'll accept the PR either way, but ask that you consider it.  Thanks!

src/app.ts
peturv: Would this be better?
```suggestion
    function getFilesRecursive(dir: string): string[] {
  if(fs.statSync(dir).isDirectory()) {
    const ancestors = fs.readdirSync(dir, 'utf8').map(f => getFilesRecursive(path.join(dir, f)));
    return [].concat(...ancestors);
  } else {
    return [dir];
  }
```

peturv: Would this be better? 
```
  if(fs.statSync(dir).isDirectory()) {
    const ancestors = fs.readdirSync(dir, 'utf8').map(f => getFilesRecursive(path.join(dir, f)));
    return [].concat(...ancestors);
  } else {
    return [dir];
  }
```

cmoesel:
I don't know why, but GitHub CI is stuck.  I've tested this locally and confirmed it is good.  Using my administrator super-powers to force the merge.

PR #238, opened 2020-02-18
Import Invariant and ObeysRule

mint-thompson:
An `ObeysRule` on a `Profile` or `Extension` references one or more `Invariant` instances. Messages are produced when:
* a system is specified for `ObeysRule.severity`, since the system is required to be `ConstraintSeverity`.
* a code other than `error` or `warning` is specified for `ObeysRule.severity`
See https://www.hl7.org/fhir/elementdefinition-definitions.html#ElementDefinition.constraint.severity, https://www.hl7.org/fhir/valueset-constraint-severity.html, and https://www.hl7.org/fhir/codesystem-constraint-severity.html for additional details. 

This pull request completes task https://standardhealthrecord.atlassian.net/browse/CIMPL-154.

### DIFF
--- a/src/import/FSHImporter.ts
+++ b/src/import/FSHImporter.ts
@@ -619,6 +619,12 @@ export class FSHImporter extends FSHVisitor {
       logger.warn('Do not specify a system for invariant severity.', concept.sourceInfo);
       concept.system = this.aliasAwareValue(system);
     }
+    if (code != 'error' && code != 'warning') {
+      logger.error(
+        'Invalid invariant severity code: code must be "#error" or "#warning".',
+        concept.sourceInfo
+      );
+    }
     return concept;
   }
 
--- a/src/import/FSHImporter.ts
+++ b/src/import/FSHImporter.ts
@@ -23,7 +23,8 @@ import {
   ValueSetFilter,
   VsOperator,
   ValueSetFilterValue,
-  FshCodeSystem
+  FshCodeSystem,
+  Invariant
 } from '../fshtypes';
 import {
   Rule,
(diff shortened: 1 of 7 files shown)


test/import/FSHImporter.Invariant.test.ts
```diff
@@ -0,0 +1,126 @@
+import { importSingleText } from '../testhelpers/importSingleText';
+import { FshCode } from '../../src/fshtypes';
+import { loggerSpy } from '../testhelpers/loggerSpy';
+
+describe('FSHImporter', () => {
+  describe('Invariant', () => {
+    describe('#invariantMetadata', () => {
+      it('should parse the simplest possible invariant', () => {
```
ngfreiter: I think this is actually technically invalid. It looks like on `constraint`, `key`, `severity` and `human` are all required, which corresponds to an Invariant that must have a `Description` and a `Severity`. So if someone is making an invalid Invariant, we should note that somewhere, either in the importer or the exporter. I don't know necessarily which makes more sense, do you have thoughts on that?

src/import/FSHImporter.ts
```diff
@@ -504,6 +586,48 @@ export class FSHImporter extends FSHVisitor {
...
 
+  visitExpression(ctx: pc.ExpressionContext): string {
+    return this.extractString(ctx.STRING());
+  }
+
+  visitXpath(ctx: pc.XpathContext): string {
+    return this.extractString(ctx.STRING());
+  }
+
+  visitSeverity(ctx: pc.SeverityContext): FshCode {
```
ngfreiter: Could `visitSeverity` use `visitCode` in some way? It seems like we could use `visitCode` to get the `FshCode` and then do the extra checking on `system` and the `code` that is required for severity. There definitely could be something I'm not seeing though that prevents this.

mint-thompson: Yes, I even made a note to myself that it would be appropriate to refactor those methods, since they share most of their functionality.

test/import/FSHImporter.Invariant.test.ts
mint-thompson: I think it would be reasonable to check for that and log an error in `visitInvariant`, since I don't think it is possible for additional metadata to be added once the invariant parser rule is finished.

Issue #194, opened 2020-02-04
Error message without line number


markkramerus:
[John Moerke]: I am getting an error from sushi without a line number
'''error: Cannot fix string value: working. Value does not match element type: code'''
seems this is caused by a constraint I have on an element of type code, AND when I have an instance example of that resource. If I remove the constraint OR remove the example then the error goes away.
In this case I am constraining the FHIR List resource, List.mode element to #working.

PR #243, opened 2020-02-19
Error message line number

ngfreiter:
Fixes https://github.com/FHIR/sushi/issues/194.

The error being emitted didn't have a line number because the error itself was incorrect. SUSHI was incorrectly thinking that a code value from a Structure Definition couldn't be fixed to a code on an instance, because the code on the Structure Definition gets serialized as a string. So to SUSHI, it seemed like we were trying to fix a string to a code. 

This is fixed by not validating values that are set on an Instance from a Structure Definition. We can do this, because the values have already been set on a Structure Definition, so we know they are valid to set on an Instance.


ngfreiter:
In my investigation of if this was a good approach, I realized that in `setFixedValuesFromStructureDefinition`, the `fixedByDirectParent` function was not actually doing anything. Since `setFixedValuesFromStructureDefinition` always gets called on the parent of an element before getting called on the element itself, there is no need to check for `pattern` on a parent, since that `pattern` would have already been accounted for when processing the parent. I'm now comfortable with this approach, but before merging I still want @jafeltra to confirm she is comfortable too.

jafeltra:
I took a look at this and had a question for Nick about using `fixedValue` vs `foundFixedValue`. I talked to him in person a bit and it sounded like he had additional thoughts while I was in a meeting that I'll circle back on early tomorrow morning. Pending that, I'm good to merge this.

Issue #232, opened 2020-02-17
Setting an extension's extension to 0..0 causes spurious error message


markkramerus:
I believe the following snippet should be legal, but it is generating an error message:
```
Extension:  CarePlanReview
Id: icare-care-plan-review
Description:    "Documentation of a re-assessment of the Care Plan"
* value[x] 0..0
* extension contains  ReviewedBy 0..1
* extension[ReviewedBy].value[x] only Reference
* extension[ReviewedBy].valueReference only Reference(Practitioner)
* extension[ReviewedBy].extension 0..0
```
I'm getting the follow error message, pointing at the final line: 
```
error: No individual slice may have max > max of sliced element, but max of slice undefined (*) > max of sliced element (0).
  File: C:\Users\mkramer\Documents\GitHub\fsh-icare\icare.fsh
```
(If I were to speculate, and I may be way off base, it seems like SUSHI is looking at the sub-extension under the extension that I'm trying to zero out, and for some reason, comparing its max card (*) to the max card of the parent (set to 0).)

jafeltra:
Fixed by PR #245 and will be included in the next release.

PR #245, opened 2020-02-20
Infer cardinalities for inline extensions

jafeltra:
This PR adds support for setting a `0..0` cardinality on `value[x]` or `extension` on extensions that are defined inline. Since extensions can be defined inline on Extensions or Profiles, this moves the preprocessing onto all the StructureDefinitions that are exported.

This should resolve the comment on PR #215 - comment [here](https://github.com/FHIR/sushi/pull/215#issuecomment-585749318).

### DIFF
--- a/src/export/StructureDefinitionExporter.ts
+++ b/src/export/StructureDefinitionExporter.ts
@@ -205,6 +205,10 @@ export class StructureDefinitionExporter implements Fishable {
               const relevantRule = `${basePath}value[x]`;
               inferredCardRulesMap.set(relevantRule, true);
             }
+          } else {
+            if (relevantContradictoryRuleMapEntry) {
+              inferredCardRulesMap.set(relevantContradictoryRule, false);
+            }
           }
         } else if (pathPart.startsWith('value') && isOnExtension) {
           const relevantContradictoryRule = `${basePath}value[x]`;
@@ -222,6 +226,10 @@ export class StructureDefinitionExporter implements Fishable {
               // If we don't already have a contradiction, add new rule to infer extension constraints
               const relevantRule = `${basePath}extension`;
               inferredCardRulesMap.set(relevantRule, true);
+            }
+          } else {
+            if (relevantContradictoryRuleMapEntry) {
+              inferredCardRulesMap.set(relevantContradictoryRule, false);
             }
           }
         }
--- a/src/export/StructureDefinitionExporter.ts
+++ b/src/export/StructureDefinitionExporter.ts
@@ -183,32 +183,36 @@ export class StructureDefinitionExporter implements Fishable {
           const relevantContradictoryRuleMapEntry = inferredCardRulesMap.get(
             relevantContradictoryRule
           );
-          if (relevantContradictoryRuleMapEntry) {
-            logger.error(
-              `Extension ${fshDefinition.name} cannot have both a value and sub-extensions`,
-              rule.sourceInfo
-            );
-            inferredCardRulesMap.set(relevantContradictoryRule, false);
-          } else if (!(rule instanceof CardRule && rule.max === '0')) {
-            // If we don't already have a contradiction, add new rule to infer value[x] constraints
-            const relevantRule = `${basePath}value[x]`;
-            inferredCardRulesMap.set(relevantRule, true);
+          if (!(rule instanceof CardRule && rule.max === '0')) {
+            if (relevantContradictoryRuleMapEntry) {
+              logger.error(
+                `Extension ${fshDefinition.name} cannot have both a value and sub-extensions`,
+                rule.sourceInfo
+              );
+              inferredCardRulesMap.set(relevantContradictoryRule, false);
+            } else {
+              // If we don't already have a contradiction, add new rule to infer value[x] constraints
+              const relevantRule = `${basePath}value[x]`;
+              inferredCardRulesMap.set(relevantRule, true);
+            }
           }
         } else if (pathPart.startsWith('value')) {
           const relevantContradictoryRule = `${basePath}value[x]`;
           const relevantContradictoryRuleMapEntry = inferredCardRulesMap.get(
             relevantContradictoryRule
           );
-          if (relevantContradictoryRuleMapEntry) {
-            logger.error(
-              `Extension ${fshDefinition.name} cannot have both a value and sub-extensions`,
-              rule.sourceInfo
-            );
-            inferredCardRulesMap.set(relevantContradictoryRule, false);
-          } else if (!(rule instanceof CardRule && rule.max === '0')) {
-            // If we don't already have a contradiction, add new rule to infer extension constraints
-            const relevantRule = `${basePath}extension`;
-            inferredCardRulesMap.set(relevantRule, true);
+          if (!(rule instanceof CardRule && rule.max === '0')) {
+            if (relevantContradictoryRuleMapEntry) {
+              logger.error(
+                `Extension ${fshDefinition.name} cannot have both a value and sub-extensions`,
+                rule.sourceInfo
+              );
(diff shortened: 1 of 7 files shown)


jafeltra:
I think I discovered an issue with inferring cardinalities for extensions and properly reporting errors when a user sets a cardinality to `0..0` directly. It is not introduced as part of this PR, but I have an idea of how to fix it, so I'm looking into it a bit. If I don't have a good solution by stand up in the morning, I may ask if you all think it is worth it to continue addressing as part of this PR.

jafeltra:
I mentioned this at stand up, but I found another issue with setting rules incorrectly on profiles. Since this is directly related to this PR, I'll continue working on it. But I don't think this PR is really ready to be reviewed in it's current state.

jafeltra:
Okay I think this is finally ready for re-review. The latest updates check that you are either on a FSH extension or you are setting a rule on an inline extension before doing any of the rule inferring. This allows you to add rules about `value[x]` elements on profiles and nested elements that have a `value[x]` on profiles. Also, you can add extensions on elements that are not already extensions without changing any rules about `value[x]`.

As part of this, I realized I was logging errors for invalid rules that were being inferred and then applied to StructureDefinitions, but the tests I had written weren't catching those logs because no errors were thrown and I only checked logs in the cases that were checking for errors. To fix this, I reset the `loggerSpy` before each test and check that it contains what is expected after each test (which should be an empty list for the majority of the tests).

src/export/StructureDefinitionExporter.ts
```diff
@@ -168,48 +168,75 @@ export class StructureDefinitionExporter implements Fishable {
...
-      if (rule.path.startsWith('extension')) {
-        if (hasRuleAppliedToValueX) {
-          logger.error(
-            `Extension ${fshDefinition.name} cannot have both a value and sub-extensions`,
-            rule.sourceInfo
+      const rulePathParts = splitOnPathPeriods(rule.path);
+      rulePathParts.forEach((pathPart, i) => {
+        const initialPath = rulePathParts.slice(0, i).join('.');
+        const basePath = `${initialPath}${initialPath ? '.' : ''}`;
+        const previousPathPart = rulePathParts.slice(i - 1, i)[0];
```
cmoesel: Isn't this just a really complicated way of saying `rulePathParts[i-1]`?

src/export/StructureDefinitionExporter.ts
```diff
@@ -168,48 +168,75 @@ export class StructureDefinitionExporter implements Fishable {
...
+        // A rule is related to an extension if it is directly on a FSH defined extension or if it is defined inline on a profile or extension.
+        // Check the section before the current pathPart to see if it is an inline extension.
+        const isOnExtension =
+          (isExtension && rulePathParts.length === 1) || previousPathPart?.startsWith('extension');
+
+        // See if we can infer any rules about an extension (inline or FSH defined)
+        if (pathPart.startsWith('extension') && isOnExtension) {
```
cmoesel: You check `isOnExtension` in the `if` and the `else if`.  Since there is no other logic in this loop, you could check `isOnExtension` as early as possible and immediately got to the next iteration if it's not on an extension.  E.g.,
```ts
for (let i=0; i < rulePathParts.length; i++) {
  // A rule is related to an extension if it is directly on a FSH defined extension or if it is defined inline on a profile or extension.
  // Check the section before the current pathPart to see if it is an inline extension.
  if (!(rulePathParts[i-1]?.startsWith('extension') || (isExtension && rulePathParts.length === 1))) {
    continue;
  }

  const pathPart = rulePathParts[i];
  const initialPath = rulePathParts.slice(0, i).join('.');
  // ...
}

Comment (cmoesel) [reply src/export/StructureDefinitionExporter.ts]: I consider that a suggestion that you can take or leave.  It would be a little more efficient (but not a lot) -- and I personally find it easier to follow, but that's totally subjective.

File: test/export/StructureDefinitionExporter.test.ts [hunk 3]
```diff
@@ -1404,6 +1405,7 @@ describe('StructureDefinitionExporter', () => {
 
     expect(valueElement.min).toEqual(0);
     expect(valueElement.max).toEqual('0');
+    expect(loggerSpy.getAllMessages()).toEqual([]);
```
cmoesel: I feel like this may be an overly brittle way of testing this.  If we add debug of info logs anywhere in the code path this exercises, these tests will start failing.  Assuming we really only care about errors, I'd check against the error logs only:
```ts
expect(loggerSpy.getAllLogs('error')).toEqual([]); // or .toHaveLength(0)
```

cmoesel: But maybe it would be even better to add an `assertNoErrorLogs` function to `testhelpers/asserts.ts`?

src/export/StructureDefinitionExporter.ts
jafeltra: Hm well yes. I think this shows the hole I truly dug myself into on this. I'll change it.

test/export/StructureDefinitionExporter.test.ts
jafeltra: Ah I like adding the assert function!

src/export/StructureDefinitionExporter.ts
jafeltra: I do like the idea of breaking out of the loop sooner. Do you know if there is a benefit to changing to use the `for` loop, rather than keeping the `forEach` loop and `return`ing out of the iteration you are in? Along the lines of:

```javascript
rulePathParts.forEach((pathPart, i) => {
  // A rule is related to an extension if it is directly on a FSH defined extension or if it is defined inline on a profile or extension.
  // Check the section before the current pathPart to see if it is an inline extension.
  const isOnExtension = /* existing definition...*/;
  if (!isOnExtension) {
    return;
  }

  const pathPart = rulePathParts[i];
  const initialPath = rulePathParts.slice(0, i).join('.');
  //...
});
```

cmoesel: I think which one is better is totally subjective.  For me, sometimes `return` inside a `forEach` loop is confusing -- I have to remember it is just returning early on a single iteration -- not returning from the entire `forEach`.  On the other hand, `continue` pretty much always means to go to the next iteration -- so I have to think less about it.  But I'm old -- so maybe the younger folk who grew up w/ ES6 don't get as hung up on that stuff. ;-)

jafeltra: Part of me still likes the `forEach`, but I don't know that my lack of confusion is due to growing up with ES6 😂

I don't have a really strong preference, just slightly leaning toward the `forEach`. If you have a strong preference for the `for`, I can be (easily) convinced to change it.

jafeltra: I stuck with the `forEach` for now, but will happily change it if you'd like.

test/export/StructureDefinitionExporter.test.ts
jafeltra: I ended up not creating a separate assert function, but used `getAllLogs('error') and checking the length in all the places I was checking the logger for consistency.

jafeltra:
I think I addressed all your comments @cmoesel.

I'm also realizing I forgot to post anything here earlier, but I was able to update the branch to not add any  duplicate rules if the user adds a `0..0` card rule themselves on an extension. I was also able to update the way we were searching for slices in ElementDefinition `constrainCardinality`, which was leading to the `InvalidMaxOfSliceError`s being logged, which Mark mentioned seeing in #232.

src/export/StructureDefinitionExporter.ts
cmoesel: It's fine.  Like I said, it's pretty subjective.  `forEach` it is!

Issue #185, opened 2020-02-03
Suggestion: Use $ to prefix aliases


markkramerus:
Suggestion by Nick George to distinguish aliases by prefixing with $. This would result in error messages if the alias were not resolved. With this change, if you say "STC" instead of "SCT" then SUSHI won't flag the error, and just put STC in the output.

cmoesel:
After some discussion, we've come to the following proposal:

1. Allow authors to _optionally_ use `$` as the first character in an alias name.
2. Disallow authors from defining other names, ids, or slices with a starting `$`.
3. Add smarter detection for contexts where only a URL/URI makes sense and log an error if there is a non-URL/non-URI string that does _not_ resolve to an aliased URL/URI.

Together, these allow for using `$` aliases as a stylistic preference for those who wish to do so, while also lowering the likelihood of alias typos from propagating without an error.

cmoesel:
I should add:
4. If SUSHI runs into something prefixed w/ `$` and can't resolve it, then it should issue an error.

PR #251, opened 2020-02-24
Check alias resolution

ngfreiter:
Fixes https://github.com/FHIR/sushi/issues/185.

The issue linked above had four requirements, those are addressed as follows:
1. Aliases could already start with `$`, nothing to change
2. Other names/ids are checked to see if they are valid FHIR names/ids, and valid FHIR names/ids cannot start with `$`
3. We now check that `SYS` resolves to a URL in the following cases:
* Fixing a code: `* code = SYS#somecode`
* VS binding: `* code from SYS`. Check that `SYS`
* VS from system components: `* codes from system SYS`
* VS from valueset components: `* codes from valueset SYS`
4. Now when we get a value in the importer that may be an alias, if the value starts with `$` and does not resolve as an alias, we log an error.


src/export/ValueSetExporter.ts
```diff
@@ -49,11 +50,19 @@ export class ValueSetExporter {
...
+          if (!isUri(composeElement.system)) {
+            throw new InvalidUriError(composeElement.system);
+          }
         }
         if (component.from.valueSets) {
           composeElement.valueSet = component.from.valueSets.map(vs => {
             return this.fisher.fishForMetadata(vs, Type.ValueSet)?.url ?? vs;
           });
+          composeElement.valueSet.forEach(vs => {
+            if (!isUri(vs)) {
```
cmoesel: Technically, `compose.include.valueSet` is type `canonical(ValueSet)`.  And according to FHIR doc:
> The canonical type differs from a uri in that it has special meaning in this specification, and in that it may have a version appended, separated by a vertical bar (|).

So... we need to support `|` here, but `isUri` rejects it since URIs typically don't allow `|`.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -798,6 +801,10 @@ export class ElementDefinition {
       }
     }
 
+    if (!isUri(vsURI)) {
```
cmoesel: This is another case where the type is `canonical(ValueSet)` so we need to support `|`.

src/import/FSHImporter.ts
```diff
@@ -1204,7 +1212,18 @@ export class FSHImporter extends FSHVisitor {
     }
   }
 
-  private aliasAwareValue(value: string): string {
+  private validateAliasResolves(value: string, parentCtx: ParserRuleContext): void {
+    const hasAlias = this.allAliases.has(value);
+    if (value.startsWith('$') && !hasAlias) {
```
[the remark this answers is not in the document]
ngfreiter: Easy change, why not be more efficient!

src/import/FSHImporter.ts
```diff
@@ -1204,7 +1212,18 @@ export class FSHImporter extends FSHVisitor {
...
+      logger.error(
+        `Value ${value} does not resolve as alias, values beginning with "$" must resolve`,
+        { location: this.extractStartStop(parentCtx), file: this.currentFile }
+      );
+    }
+  }
+
+  private aliasAwareValue(value: string, parentCtx: ParserRuleContext): string {
+    this.validateAliasResolves(value, parentCtx);
```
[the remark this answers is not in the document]
ngfreiter: I will do that second approach. Since almost every time the value is just `parentCtx.getText()`, I think that seems fair.

src/export/ValueSetExporter.ts
ngfreiter: Oh FHIR how tricky of you. I will fix this.

ngfreiter:
Ok @cmoesel, should be ready to take another look.

cmoesel:
So, I ran this against `fsh-mcode` and it reported these two errors:
```
error Resolved value "http://terminology.hl7.org/CodeSystem/v2-0360|2.7" is not a valid URI.
  File: /Users/cmoesel/dev/standardhealth/fsh-mcode/Examples.fsh
  Line: 146
error Resolved value "http://terminology.hl7.org/CodeSystem/v2-0360|2.7" is not a valid URI.
  File: /Users/cmoesel/dev/standardhealth/fsh-mcode/Examples.fsh
  Line: 146
```

Totally unrelated, but notice that the same exact error is logged twice?  It seems that we must be calling the function that spits out the error multiple times during processing.  We probably need to look at that, but it's not the fault of this PR.

But the error itself is somewhat related to this PR.  It is on this line:
```
* qualification.code = http://terminology.hl7.org/CodeSystem/v2-0360|2.7#MD "Doctor of Medicine"
```

Technically, according to the FSH documentation, this is allowed and *should* result in the code system version being set to `2.7`.  @ngfreiter and I discussed this topic recently and noted we don't support that yet in SUSHI.  But... if it is valid FSH, even if we don't do the right thing yet, we probably shouldn't flag it as an error.

So... I'm wondering if we should actually widen the instances where we accept `|` in a URL.  If it's too cumbersome to tease out all the allowed places, I think I'm totally comfortable with allowing it across the board in URLs.

ngfreiter:
I don't know, I'm not so sure about this. At this point, isn't it better to be emitting an error when the user uses the `|` to set a version? Because the user thinks from the FSH-docs that they are setting the URL and the version, when in reality SUSHI is setting an incorrect URL and no version at all.

I also wouldn't feel great about allowing `|` in URLs since it seems like we would only be doing that temporarily until we add the `|` syntax to be working as it is supposed to. Then after that is added, we would be incorrectly allowing people to put `|` in a URL where it doesn't belong.

cmoesel:
OK.  I hear you on that one.  Technically by allowing that URL through as-is, we're putting an invalid URL into the FHIR instance.  If nothing else, we should just strip the `|` suffix from the URL before using it and log a warning that the version is not currently supported.  But I bet that is almost the same amount of work as just implementing it. ;-)

My problem with treating it as an error is that it's not *wrong* from a FSH perspective and we're forcing the author to remove something that they'll just be wanting to add back later.  It doesn't seem very user-friendly to me.

As for supporting `|` across all URLs, I guess the appropriateness of that will depend on how and where we parse out the `|` from the URL.  If we do it early on (like in the importer) then yeah, it doesn't make sense to turn a blind eye in the exporter.  But if we don't parse it out until late in the game, then maybe it makes sense?  I'm thinking we want to parse it out as early as possible, but I haven't fully thought out the implications.

So I guess we have two choices:

1. Accept the PR as-is and people will have to back out any version-specific URLs they have until we add support for them, or
2. Hold the PR until we add support for version-specific URLs and merge it all at once.

ngfreiter:
I guess the way I see it is that either way, the `|` doesn't work, so in a sense whether we are emitting errors are not, author's should remove `|` until it does work. Because if we merge this PR, they will see errors about it, but also if we don't merge this PR, we are generating incorrect output. 

I don't think adding support for `|` will be too hard, so for that reason I guess we should wait and do that soon.

PR #253, opened 2020-02-25
Show a message when created choices are made obsolete with type constraints

### LINKED ISSUE #210: SUSHI allows authors to put constraints on choices that have been constrained out
The following is allowed by SUSHI:
```
Profile: MyProfile
Parent: Observation
* valueCodeableConcept MS
* value[x] only Quantity
```

Ideally, SUSHI should recognize and report this as an error and exclude the `valueCodeableConcept` constraints from the profile's SD.

mint-thompson:
Fixes #210 with task https://standardhealthrecord.atlassian.net/browse/CIMPL-271

The sample FSH in the issue gives an example of the case being handled: when a type constraint comes after a different rule that creates a choice, and the type constraint makes that choice obsolete, that's probably a user error. So, show a message to the user indicating this. When the type constraint comes first, existing code handles that case and logs an error.

To explain a bit about the `should not log an error when a type constraint is applied to a specific slice` test case: exciting things can happen when sliced elements contain choice elements. This test shows that type constraints on one slice's choice won't cause problems for other slices.

### DIFF
--- a/src/fhirtypes/StructureDefinition.ts
+++ b/src/fhirtypes/StructureDefinition.ts
@@ -560,7 +560,7 @@ export class StructureDefinition {
         // if we do not have a matching slice, we want to slice the first match
         const matchingXElement = matchingXElements[0];
         // If we find a matching [x] element, we need to slice it to create the child element
-        // NOTE: The spec is somewhat incosistent on handling choice slicing, we decided on this
+        // NOTE: The spec is somewhat inconsistent on handling choice slicing, we decided on this
         // approach per consistency with 4.0.1 observation-vitalsigns profiles and per this post
         // https://blog.fire.ly/2019/09/13/type-slicing-in-fhir-r4/.
         matchingXElement.sliceIt('type', '$this', false, 'open');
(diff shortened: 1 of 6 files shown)


src/fhirtypes/StructureDefinition.ts
```diff
@@ -598,6 +598,29 @@ export class StructureDefinition {
     return matchingSlice;
   }
 
+  findObsoleteChoices(baseElement: ElementDefinition, oldTypes: ElementDefinitionType[]): string[] {
+    // first, find all the elements representing choices for the same choice element
+    const baseId = baseElement.id.slice(0, baseElement.id.lastIndexOf('.'));
```
ngfreiter: `baseId` is just the id of the parent of `baseElement` right? I think you could use the `parent()` function on the ElementDefinition. It took me a bit to figure out what this was getting, and using `parent()` may make it more clear.

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -1019,6 +1019,89 @@ describe('StructureDefinitionExporter', () => {
...
+    secondType.types = [{ type: 'string' }];
+
+    profile.rules.push(
+      slicingType,
+      slicingPath,
+      slicingRules,
+      componentSlices,
+      firstType,
+      firstCard,
+      secondType
```
ngfreiter: One thing @jafeltra has started doing (that I should also be better about doing) is commenting her tests that have rules to look like this:
``` 
      const onlyRule = new OnlyRule('value[x]');
      onlyRule.types = [{ type: 'Quantity' }];
      observation.rules.push(onlyRule); // * value[x] only Quantity
      const fixedValRule = new FixedValueRule('valueQuantity');
      const fixedFshCode = new FshCode('foo', 'http://foo.com');
      fixedValRule.fixedValue = fixedFshCode;
      observation.rules.push(fixedValRule); // * valueQuantity = foo.com#foo
```
I think it could be useful here to edit this test and the test above to have comments like that, since these tests involve a lot of rules.

mint-thompson: Oh, I like this a lot. Definitely adding this in.

src/fhirtypes/StructureDefinition.ts
mint-thompson: Absolutely right, I've basically just copied the first line out of the `parent` method. This is what I get for not reading existing classes closely enough.

PR #259, opened 2020-02-26
Log error when fixing value on sliced element

### LINKED ISSUE #227: SUSHI should report error when fixing values on root path that is sliced and fixed in the slice
This is best explained by example:
```
Profile: MyProcedureLabProfile
Parent: http://hl7.org/fhir/us/core/StructureDefinition/us-core-observation-lab
* category = OBSCAT#procedure
```

US Core slices `category` and fixes the `Laboratory` slice to `laboratory`.  The rule `* category = OBSCAT#procedure`, however, implies a single allowed value (which is _not_ `laboratory`).  Therefore, SUSHI should report an error (after all, the IG Publisher does).

In addition, we probably should think in general about what it means for an author to fix a value on a multi-cardinality path.  Should we emit a warning (even when there is no conflict w/ an existing slice) since the intent probably isn't to fix every element in the array (and slicing would be more appropriate)?

mint-thompson:
Fixes #227 and completes task https://standardhealthrecord.atlassian.net/browse/CIMPL-277

When an element has slicing defined on it, log an error when a value is fixed on the element's root path. Fixing a value on the root path in this situation is occasionally legal, but almost never correct.

The issue being fixed, as well as the test cases, demonstrate what this scenario looks like. The scenario happening in the opposite order, where the parent fixes a value and the child creates a slice with a fixed value, will already log an error from existing code. Therefore, this pull request does not address that scenario.

The main thing I'd like some help with is the error message: maybe it needs to be made a little clearer that "what you have done is more or less guaranteed to be incorrect."


jafeltra:
This looks good to me! For the error message, I'm not sure if it's because I'm just not used to setting these types of values, but I think it would be helpful to say that what you've done is probably incorrect (like you mentioned). I feel like when I read that error, I don't quite recognize that statement as implying an issue. But again, that could be because I'm less used to doing something like this.

Maybe you could also add to your existing message "... which is likely fixing incorrect." 

Also, once the wording is settled, should the test also check that the message is logged, in addition to the source line?

jafeltra:
An additional thought - I tried seeing what the IG Publisher reported as an error since @cmoesel mentioned that it logs an error. I was hoping to draw inspiration for the wording for our error message from there, but it turns out, I don't see any issues logged in the IG publisher. Is that expected?

cmoesel:
I feel like this is a little too broad.  There are cases where fixing a value on a sliced element might be completely fine -- so I don't think it's appropriate to log an `error` just because someone fixed a value on a sliced element.  I'm not even sure I'd log a `warning` -- because, really, it's OK.

I think the code needs to check if the fixed value has any conflicts.  And I think that means there are two ways this can happen:

1. If someone fixes a value on a slice when the element it slices already has a fixed value.
```
Profile: FixedCategoryExample
Parent: Observation
* category ^slicing.discriminator[0].type = #pattern
* category ^slicing.discriminator[0].path = "$this"
* category ^slicing.rules = #open
* category = http://terminology.hl7.org/CodeSystem/observation-category#procedure
* category contains Lab 1..1
* category[Lab] = http://terminology.hl7.org/CodeSystem/observation-category#laboratory
```
Error on the last line (fixing `category[Lab]`)...

2. If someone fixes a value on a sliced element when one of the slices already has a fixed value.
```
Profile: FixedCategoryExample
Parent: Observation
* category ^slicing.discriminator[0].type = #pattern
* category ^slicing.discriminator[0].path = "$this"
* category ^slicing.rules = #open
* category contains Lab 1..1
* category[Lab] = http://terminology.hl7.org/CodeSystem/observation-category#laboratory
* category = http://terminology.hl7.org/CodeSystem/observation-category#procedure
```
Error on the last line (fixing `category`)...

Also note that this can apply to child elements of the slices too.  So, for example, this would also be bad:
```
Profile: FixedComponentExample
Parent: Observation
* component ^slicing.discriminator[0].type = #pattern
* component ^slicing.discriminator[0].path = "code"
* component.code = http://loinc.org#12345
* component contains foo 1..1
* component[foo].code = http://loinc.org#54321
```

In fact, I believe that already logs an error, so a better example would be:
```
Profile: FixedComponentExample
Parent: Observation
* component ^slicing.discriminator[0].type = #pattern
* component ^slicing.discriminator[0].path = "code"
* component contains foo 1..1
* component[foo].code = http://loinc.org#54321
* component.code = http://loinc.org#12345
```

src/export/StructureDefinitionExporter.ts
```diff
@@ -115,6 +115,12 @@ export class StructureDefinitionExporter implements Fishable {
             element.constrainCardinality(rule.min, rule.max);
           } else if (rule instanceof FixedValueRule) {
             const replacedRule = replaceReferences(rule, this.tank, this);
+            if (element.slicing) {
+              logger.error(
+                `Fixing value on sliced element ${element.id} in ${structDef.name}`,
+                rule.sourceInfo
+              );
+            }
```
cmoesel: This is too broad.  See my overall comment on the PR for details.

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -1198,6 +1198,55 @@ describe('StructureDefinitionExporter', () => {
     expect(loggerSpy.getLastMessage()).toMatch(/File: Fixed\.fsh.*Line: 4\D*/s);
   });
 
+  it('should log an error when a fixed value on a child interferes with a fixed slice on the parent', () => {
+    // Profile: ParentWithFixedSlice
+    // Parent: Observation
+    // * category ^slicing.discriminator[0].type = #pattern
+    // * category ^slicing.discriminator[0].path = "code"
```
cmoesel: It's just an example, but technically this path is not valid in this context.  The discriminator path should be relative to the element that is being sliced.  In this case, since `category` is a `CodeableConcept`, `"code"` doesn't work (because `CodeableConcept.code` isn't a thing).  Since the pattern is actually applied to `category` itself, you really want the path to be `"$this"`.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -820,15 +956,35 @@ export class ElementDefinition {
...
       if (strengths.indexOf(strength) < strengths.indexOf(this.binding.strength)) {
         throw new BindingStrengthError(this.binding.strength, strength);
       }
     }
 
+    const connectedElements = this.findConnectedElements();
+    connectedElements.forEach(ce => {
+      if (vsURI == ce.binding?.valueSet) {
+        try {
+          ce.bindToVS(vsURI, strength);
```
ngfreiter: Is it overly complicated to use `bindToVs` recursively here? Since you are not actually using the error checking that this function provides, does it make more sense to just set the `vsURI` and `strength` directly on `ce.binding`? Or is there something wrong with that approach?

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -1200,6 +1200,122 @@ describe('StructureDefinitionExporter', () => {
     expect(loggerSpy.getLastMessage()).toMatch(/File: Fixed\.fsh.*Line: 4\D*/s);
   });
 
+  it.skip('should log an error when a fixed value on a child interferes with a fixed slice on the parent', () => {
```
ngfreiter: Should these tests for fixed value be in the "Rules with Slices" describe block?

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -2294,6 +2410,620 @@ describe('StructureDefinitionExporter', () => {
...
+
+    // don't actually apply the rule to the slice, since it might be valid based on the contents of the value sets
+    // but if it's the same value set for both, _then_ we know it can't be a weaker binding, and that's what we should give a message about
+    // so this is ok:
+    // * category[Procedure] from ABC (extensible)
+    // * category from ABC (required)
+    // and it will update the strength to required
+    // this is not:
+    // * category from ABC (required)
+    // * category[Procedure] from ABC (extensible)
```
ngfreiter: Is this comment out of place?

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -2294,6 +2410,620 @@ describe('StructureDefinitionExporter', () => {
...
+      doc.profiles.set(observationWithSlice.name, observationWithSlice);
+      exporter.export();
+      const sd = pkg.profiles[0];
+      const rootInterpretation = sd.findElement('Observation.component.interpretation');
+      const labInterpretation = sd.findElement('Observation.component:Lab.interpretation');
+      expect(rootInterpretation.mustSupport).toBe(true);
+      expect(labInterpretation.mustSupport).toBe(true);
+    });
+
+    it('should not apply a flag rule that would disable a must support or is modifier flag on a slice when the list has that flag', () => {
```
ngfreiter: Ok so I see the different behavior of `isSummary` is definitely on purpose, but why?

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -2294,6 +2410,620 @@ describe('StructureDefinitionExporter', () => {
...
+        valueSet: 'http://example.com/MediocreObservationCodes',
+        strength: 'extensible'
+      };
+      expect(rootCategory.binding).toEqual(importantBinding);
+      expect(procedureCategory.binding).toEqual(mediocreBinding);
+    });
+
+    it('should apply ValueSetRules on a sliced element, then a slice, with different value sets', () => {
+      // * category[Procedure] from http://example.com/MediocreObservationCodes (extensible)
+      // * category from http://example.com/ImportantObservationCodes (required)
```
ngfreiter: This comment looks like it doesn't match what the rules are doing. This is the wrong order right?

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -2294,6 +2410,620 @@ describe('StructureDefinitionExporter', () => {
...
+      const labValue = sd.findElement('Observation.component:Lab.value[x]');
+      const expectedTypeQuantity = new ElementDefinitionType('Quantity');
+      expect(rootValue.type).toHaveLength(1);
+      expect(rootValue.type).toContainEqual(expectedTypeQuantity);
+      expect(labValue.type).toHaveLength(1);
+      expect(labValue.type).toContainEqual(expectedTypeQuantity);
+      expect(loggerSpy.getAllMessages('error')).toHaveLength(0);
+    });
+
+    it('should not apply an OnlyRule on a sliced element that would invalidate any of its slices', () => {
```
ngfreiter: Should this check for an error?

src/fhirtypes/ElementDefinition.ts
```diff
@@ -425,6 +428,39 @@ export class ElementDefinition {
       }
     }
 
+    const connectedElements = this.findConnectedElements();
+    if (connectedElements.length > 0) {
+      // check to see if the card constraint would actually be a problem for the connected element
+      // that is to say, if the new card is narrower than the connected card
+      connectedElements.forEach(ce => {
+        if ((ce.min == null && min > 0) || (ce.min != null && ce.min < min)) {
```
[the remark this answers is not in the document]
mint-thompson: So, working through this scenario: if this were allowed, then with one-sided cardinality rules, we could end up with something like:
```
* component[Lab].interpretation ..9
* component.interpretation 2..9
```
But, looking more closely at this case: since `component[Lab].interpretation.min` is `null`, then it should end up inheriting the minimum of 2 from `component.interpretation`, which is fine. So, I think you're right: if the min (or max, as you point out in your next comment) is null, then that's saying "we'll inherit the value," which is always fine.

It does mean it may be necessary to check for a scenario like this:
```
* component[Lab].interpretation ..9 // or other rule that creates this element
* component.interpretation 2..9
* component[Lab].interpretation 1..9
```
The third rule should not be applied in this case, since it would mean that going inside the `Lab` slice widens the cardinality of `interpretation`. If the first rule is skipped, then existing code should handle the error, since then the `component[Lab].interpretation` element will get created with the copied cardinality, I think.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -425,6 +428,39 @@ export class ElementDefinition {
...
+            min,
+            max,
+            ce.min ?? 0,
+            ce.max ?? '*'
+          );
+        }
+        // if the connected element's max is null or *, we can't constrain the max
+        // if the connected element's max is not null and is not *, we can't make the max smaller than its max
+        if (
+          ((ce.max == null || ce.max == '*') && maxInt != null) ||
```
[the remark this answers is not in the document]
mint-thompson: Similarly to above, the following should be fine:
```
* component[Lab].interpretation 2..
* component.interpretation 2..9
```

But the following error needs to be detected:
```
* component[Lab].interpretation 2.. // or other rule that creates this element
* component.interpretation 2..9
* component[Lab].interpretation 2..10
```

src/fhirtypes/ElementDefinition.ts
```diff
@@ -783,14 +913,20 @@ export class ElementDefinition {
...
+
+    const connectedElements = this.findConnectedElements();
+
     if (isBoolean(mustSupport)) {
       this.mustSupport = mustSupport;
+      connectedElements.forEach(ce => (ce.mustSupport = mustSupport || ce.mustSupport));
     }
     if (isBoolean(summary)) {
       this.isSummary = summary;
+      connectedElements.forEach(ce => (ce.isSummary = summary));
```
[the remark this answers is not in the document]
mint-thompson: Yeah, this is sort of a magical implementation detail that feels less necessary every time it comes up, since FSH has no way of explicitly trying to set a flag to false. Basically, MS and ?! can't be turned off once they're on, so I'll add a comment to that effect.

src/fhirtypes/ElementDefinition.ts
mint-thompson: The advantage of calling `bindToVS` recursively is that it already performs the binding strength checks that I want to perform. If I didn't do that, I'd need to duplicate those checks inside the `try` block.

test/export/StructureDefinitionExporter.test.ts
mint-thompson: These tests can probably be removed, actually. They're leftovers from a previous implementation, and I've added `todo` tests for FixedValueRule up on lines 2761-2766. Since supporting that rule is going to be handled in a different task, there's no need for them to stick around.

test/export/StructureDefinitionExporter.test.ts
mint-thompson: This just looks like a note I left for myself and forgot to remove. oops

test/export/StructureDefinitionExporter.test.ts
mint-thompson: Correct! I'll fix the comment to reflect the actual test being performed.

test/export/StructureDefinitionExporter.test.ts
mint-thompson: The justification is this:
* disabling `mustSupport` changes the software requirements of systems handling this resource, which means that you could end up with conflicting requirements (the parent resource says you have to support something, but the child resource says you don't.)
* disabling `modifier` changes the meaning of an element, which in my opinion is a way to make things more confusing.
* disabling `summary` changes whether an element appears in the resource's summary. The element still exists, and means the same thing, and still carries the same software requirements. If you need the element, you can directly ask for it. So, disabling `summary` seems less likely to cause issues.

Of course, all of this is still 1. up for discussion and 2. maybe meaningless, since there's no way in FSH to explicitly disable flags.

test/export/StructureDefinitionExporter.test.ts
mint-thompson: It should! I'll add that in 🍰

test/export/StructureDefinitionExporter.test.ts
mint-thompson: See also
* http://hl7.org/fhir/R4/elementdefinition-definitions.html#ElementDefinition.mustSupport 
* http://hl7.org/fhir/R4/elementdefinition-definitions.html#ElementDefinition.isModifier
* http://hl7.org/fhir/R4/elementdefinition-definitions.html#ElementDefinition.isSummary

Although, looking more closely at `isModifier`, it looks like there are even more detailed rules going on here that aren't being implemented correctly: `Only the definition of an element can set IsModifier true - either the specification itself or where an extension is originally defined. Once set, it cannot be changed in derived profiles.`

src/fhirtypes/ElementDefinition.ts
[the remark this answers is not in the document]
ngfreiter: I don't think this necessarily makes it so that your second example wouldn't be an issue. But one thing to consider is that one sided cardinalities don't set the cardinality to null, they just get it from the element's current cardinality. So in the second example I think `component[Lab].interpretation` would maintain whatever min cardinality was originally on `component.interpretation`.

test/export/StructureDefinitionExporter.test.ts
ngfreiter: Ok I buy that reasoning for `summary`, and good catch on `isModifier`, but that could be a separate task I think.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -558,6 +601,40 @@ export class ElementDefinition {
...
+          const obsoleteConnections = ce.structDef.findObsoleteChoices(ce, oldTypes);
+          if (obsoleteConnections.length > 0) {
+            logger.error(
+              `Type constraint on ${rule.path} makes rules in ${
+                ce.structDef.name
+              } obsolete for choices: ${obsoleteConnections.join(', ')}`,
+              rule.sourceInfo
+            );
+          } else {
+            throw new SliceTypeRemovalError(rule.path, ce.id);
```
jafeltra: I think the function documentation should include this error as well.

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -2294,6 +2410,620 @@ describe('StructureDefinitionExporter', () => {
...
+        valueSet: 'http://example.com/ImportantObservationCodes',
+        strength: 'required'
+      };
+      const mediocreBinding = {
+        valueSet: 'http://example.com/MediocreObservationCodes',
+        strength: 'extensible'
+      };
+      expect(rootCode.binding).toEqual(importantBinding);
+      expect(labCode.binding).toEqual(mediocreBinding);
+    });
```
jafeltra: On the non-child slices, it looks like we log an error when binding the same value set to a sliced element and the root but binding it more weakly on the sliced element (the test starting on line 2675). Should there be an error logging for a similar situation for children of slices? I'm thinking of something like:
```
* component.code from http://example.com/RegularObservationCodes (required)
* component[Lab].code from http://example.com/RegularObservationCodes (extensible)
```
There's a very good chance I'm missing something and you already tested this somewhere or this should be allowed.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -820,14 +940,19 @@ export class ElementDefinition {
...
     }
     if (summary === true) {
       this.isSummary = summary;
+      connectedElements.forEach(ce => (ce.isSummary = summary));
     }
     if (modifier === true) {
       this.isModifier = modifier;
+      connectedElements.forEach(ce => (ce.isModifier = modifier || ce.isModifier));
     }
     if (newStatusExtension) {
```
ngfreiter: One more question I have, should the standards status flags be applied to connected elements in the same way as the old flags?

mint-thompson: That's... a good question. They're allowed to be different, but they'll usually be the same, so I'd say yes, they should get applied to connected elements.

mint-thompson: Update: got a negative response when consulting an expert, so no, they will not be propagated.

ngfreiter: Ok so no prop.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -439,6 +475,38 @@ export class ElementDefinition {
...
+   * @param {string} postPath The path to append to the parent in order to try to find a connected element
+   * @returns {ElementDefinition[]} The elements at or inside of slices whose path matches the original element
+   */
+  private findConnectedElements(postPath = ''): ElementDefinition[] {
+    const connectedElements = this.getSlices()
+      .filter(e => e.max !== '0') // we don't need zeroed-out slices
+      .map(slice => {
+        return this.structDef.findElement(`${slice.id}${postPath}`);
+      })
+      .filter(e => e);
```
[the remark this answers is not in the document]
mint-thompson: `findElement` returns `undefined` if it can't find something at the provided path, so the filter removes those elements from the list.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -439,6 +475,38 @@ export class ElementDefinition {
...
+    if (this.parent()) {
+      return connectedElements.concat(
+        this.parent().findConnectedElements(
+          `.${this.path.slice(this.path.lastIndexOf('.') + 1)}${postPath}`
```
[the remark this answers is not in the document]
mint-thompson: Ah, absolutely yes!

test/export/StructureDefinitionExporter.test.ts
mint-thompson: Nope, I don't have this test yet! I'll add it in.

src/fhirtypes/ElementDefinition.ts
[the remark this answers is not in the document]
jafeltra: Ah, I hadn't thought of that! Nifty!

PR #260, opened 2020-02-26
Apply ObeysRules to Element.constraint

jafeltra:
This PR applies any `ObeysRule`s specified on structure definitions onto `Element.constraint`. Any fields specified on the Invariant are added to the constraint, which is then pushed onto the existing list of constraints.

If someone tries to apply an `ObeysRule` using an invariant that was not defined, it logs an error. If someone applies an `ObeysRule` that had previously logged an error (i.e. if the severity is not error), the invariant is still applied, since we have other places where we log errors and still set the technically incorrect values.

Bonus - removed some code that had expired according to @cmoesel's expiration date.


cmoesel:
I haven't reviewed the PR yet (only the comments), but I thought I'd remark on @mint-thompson's comment.  For that second example (with the rules reversed), Mint is right that the snapshot element for `Observation.value[x]:valueString` will technically be incorrect because it won't have the invariant.  But... this is an overarching problem for pretty much *all* our rules.  The same issue exists for cardinality rules, flag rules, only rules, etc.  So... I wouldn't hold up this PR for that reason; as it's probably best to resolve that on the systemic level-- probably by doing one of the following:
* re-ordering rules before they're applied such that all rules related to a sliced element come _before_ the rules related to specific slices of that element
* when applying rules to a sliced element (or its child), also apply those rules to any specific slices of that element (or their children).

The other wrinkle here (which is a bit of a saving grace) is that I think in most cases, the `differential` will actually be fine -- and when run through the IG Publisher, the publisher will correctly build the snapshot such that the rules get correctly applied to all slices.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -344,6 +344,24 @@ export class ElementDefinition {
...
+   */
+  applyConstraint(invariant: Invariant, source?: string): void {
+    const constraint: ElementDefinitionConstraint = {
+      ...(invariant.name && { key: invariant.name }),
+      ...(invariant.severity && { severity: invariant.severity.code }),
+      ...(invariant.description && { human: invariant.description }),
+      ...(invariant.expression && { expression: invariant.expression }),
+      ...(invariant.xpath && { xpath: invariant.xpath }),
+      ...(source && { source })
+    };
```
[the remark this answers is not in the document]
jafeltra: Google helped me out on this one, but I really like it!

jafeltra:
I hadn't considered @mint-thompson's test case, so I'm glad it was brought up. I think it seems like from @cmoesel's comment I won't add specific handling for that case as part of this PR, since it seems like something we should handle at a broader level.

src/export/StructureDefinitionExporter.ts
```diff
@@ -154,6 +155,16 @@ export class StructureDefinitionExporter implements Fishable {
             } else {
               structDef.setInstancePropertyByPath(rule.caretPath, rule.value, this);
             }
+          } else if (rule instanceof ObeysRule) {
+            const invariant = this.tank.getAllInvariants().find(i => i.name === rule.invariant);
```
ngfreiter: Does it make sense to use wrap this into the `fish` method on `FSHTank`? So we would get this like:
```
const invariant = this.tank.fish(i.name, Type.Invariant)
```

I am currently thinking about how to do this with `Mixins` and I'd want to be consistent. On the one hand since `Invariant`s should only be referred to by their name, maybe putting them in `fish` is overkill. On the other hand, maybe it is more consistent to always be finding stuff with `fish`.

jafeltra: Ah that's a good point! I hadn't thought of that, but I kind of like the idea of adding it into `fish`. I think I personally like the consistency and would be okay with having it look through the list only by name, but I also haven't touched this function a whole lot, so I'm open to others' opinions.

cmoesel: I like to `fish` (metaphorically).  So I'm in favor of adding it to the fisher.

jafeltra: Should I add the ability to fish by `id` as well as `name`? Presumably we'd only be using `name` right now, but I can add the other for consistency if we want.

cmoesel: I don't think invariants have an id, do they?  Don't they only have a name?
image
Or am I misunderstanding your question?

jafeltra: Ah no, you are correct. However, typescript gets mad at all the other places we are accessing `id` on the objects we've `fish`ed for because, until now, all the types that `fish` returns have an `id`. 

@cmoesel stopped by as I was writing this, and I'm going to add a read-only `id` property onto Invariant to make sure we can still access `.id` on anything we've `fish`ed for. I won't add any ability to `fish` by that `id` though.

jafeltra: The last commit should cover this!

PR #262, opened 2020-02-27
Copy over custom files when exporting IG

### LINKED ISSUE #257: Shouldn't ig-data/examples be copied over an installation
I put some files in ig-data/examples, but they weren't copied as expected.

Also, I'd love it if the copying of examples happened AFTER the generation of examples in input so that I could easily override FSH generated examples by stuffing a customized one in there.

ngfreiter:
Fixes #257 and also implements https://standardhealthrecord.atlassian.net/browse/CIMPL-256. 

Now when the user places valid JSON files in any of the following folders within the `ig-data` folder:
* capabilities (for CapabilityStatements/Conformance resources)
* examples
* extensions
* models (for logical models)
* operations
* profiles
* resources (for anything/everything or whatever doesn't fit one of the other folders)
* vocabulary (for code systems & value sets)

We add those files to the ImplementationGuide resource, and copy the file over into the corresponding folder in `build/input`. This happens after the FSH has been written to `build/input`, so custom files will overwrite FSH files.


src/ig/IGExporter.ts
```diff
@@ -405,6 +407,76 @@ export class IGExporter {
...
+  private addPredefinedResources(igPath: string): void {
+    const pathEnds = [
+      'capabilities',
+      'models',
+      'operations',
+      'profiles',
+      'resources',
+      'vocabulary',
+      'examples' // Must come last in case examples are of other resources
+    ];
```
jafeltra: Do we want to include `extensions` in this list here? It's a supported folder and you mentioned it in the PR description, but right now, it won't get copied over.

src/ig/IGExporter.ts
```diff
@@ -405,6 +407,76 @@ export class IGExporter {
...
+            reference: {
+              reference: `${resourceJSON.resourceType}/${resourceJSON.id}`
+            },
+            name: resourceJSON.title ?? resourceJSON.id,
+            description: resourceJSON.description
+          };
+          if (pathEnd === 'examples') {
+            const exampleUrl = resourceJSON.meta?.profile?.find(
+              url =>
+                this.fhirDefs.fishForFHIR(url, Type.Profile) ?? this.pkg.fish(url, Type.Profile)
```
jafeltra: This is just a question, but is there a reason to check the `fhirDefs` before the `pkg`? I know there must be, I'm just curious what it is.

src/ig/IGExporter.ts
ngfreiter: Oh most definitely we do!

src/ig/IGExporter.ts
ngfreiter: Ah you have a lot of trust in my reasoning. Maybe too much trust. So my thinking was that it shouldn't matter. Since we are finding based on a URL, a URL should be unique to the profile, so we shouldn't run into naming conflicts.

ngfreiter:
Ok copying over extensions too!

src/ig/IGExporter.ts
jafeltra: Ah! Good point! Thanks for clarifying!

cmoesel:
I'd like to review this one as well.  Will do so after our standup!

src/ig/IGExporter.ts
cmoesel: I think the reverse order (`pkg` then `fhirDefs`) might make more sense because there's a pretty good chance it's in the package, and the package is a _lot_ smaller than `fhirDefs`.  Although I _think_ they both use maps for lookup, so I guess maybe size doesn't matter.  I don't know... ?

src/ig/IGExporter.ts
```diff
@@ -405,6 +407,77 @@ export class IGExporter {
...
+            );
+            if (exampleUrl) {
```
cmoesel: This check isn't strict enough because we should only set `exampleCanonical` for examples of profiles _defined in this IG_.  Since you're searching through all of FHIRDefs, it's possible that it matched a profile from a dependency IG -- and in that case we should only use `exampleBoolean`, _not_ `exampleCanonical`.

src/ig/IGExporter.ts
```diff
@@ -48,6 +49,7 @@ export class IGExporter {
     this.addOtherPageContent(outPath);
     this.addImages(outPath);
     this.addIncludeContents(outPath);
+    this.addPredefinedResources(outPath);
```
cmoesel: Probably inconsequential, but I feel like this might make more sense coming immediately after `this.addResources();`.

src/ig/IGExporter.ts
```diff
@@ -405,6 +407,77 @@ export class IGExporter {
...
+            );
+            if (exampleUrl) {
+              resource.exampleCanonical = exampleUrl;
+            } else {
+              resource.exampleBoolean = true;
+            }
+          } else {
+            resource.exampleBoolean = false;
+          }
+          this.ig.definition.resource.push(resource);
```
cmoesel: IIRC, we agreed with Keith that any resources explicitly put in the `ig-data` should _overwrite_ FSH-generated resources (if it's, in fact, the same resource).  So `push`ing on the `resource` array will just create duplicate resource entries instead of overwriting the old one.  I think we're going to need to search through the `resource` array to see if the resource already exists and then act accordingly.

test/ig/IGExporter.test.ts
```diff
@@ -486,6 +486,165 @@ describe('IGExporter', () => {
...
+      const igContent = fs.readJSONSync(igPath);
+      expect(igContent.definition.resource).toContainEqual({
+        reference: {
+          reference: 'StructureDefinition/MyTitlePatient'
+        },
+        name: 'This patient has a title',
+        exampleBoolean: false
+      });
+    });
+
```
cmoesel: Let's get in a test or two to ensure that BYOJ resources overwrite FSH-generated resources.

src/ig/IGExporter.ts
```diff
@@ -405,6 +407,77 @@ export class IGExporter {
...
+      'examples' // Must come last in case examples are of other resources
+    ];
+    for (const pathEnd of pathEnds) {
+      const dirPath = path.join(this.igDataPath, 'input', pathEnd);
+      if (fs.existsSync(dirPath)) {
+        const files = fs.readdirSync(dirPath);
+        for (const file of files) {
+          let resourceJSON: InstanceDefinition;
+          try {
+            resourceJSON = JSON.parse(fs.readFileSync(path.join(dirPath, file), 'utf-8').trim());
```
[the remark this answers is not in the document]
ngfreiter: Do you think it is better to just not log an error? Let people put whatever they want in there and document that we will only copy over json? I'm thinking if someone does have files in there, maybe we should trust that they have a reason and not toss a bunch of errors in their face.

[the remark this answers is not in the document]
cmoesel: Just documenting our discussion -- let's follow @jafeltra's approach (as demonstrated in page content handling) and log a single warning if we encounter non-JSON files.  Furthermore, if we encounter XML files, log a more specific warning indicating that we don't currently support XML resources so we didn't update the IG JSON, but we did still copy them over.

src/ig/IGExporter.ts
```diff
@@ -425,45 +426,78 @@ export class IGExporter {
...
         for (const file of files) {
           let resourceJSON: InstanceDefinition;
-          try {
-            resourceJSON = JSON.parse(fs.readFileSync(path.join(dirPath, file), 'utf-8').trim());
-          } catch (e) {
-            logger.error(`Invalid input file: ${path.join(dirPath, file)}`);
+          if (file.endsWith('.json')) {
+            resourceJSON = fs.readJSONSync(path.join(dirPath, file));
+          } else {
+            xmlFile = file.endsWith('.xml');
```
cmoesel: This doesn't really aggregate the result, so `xmlFile` will always just be an indication of the last processed file.  I think you want it true as long as _any_ file was xml, right? If so, try:
```ts
xmlFile = xmlFile || file.endsWith('.xml');
```

src/ig/IGExporter.ts
```diff
@@ -475,6 +509,13 @@ export class IGExporter {
           );
         }
       }
+      if (invalidFile) {
+        let message = `Invalid file detected in directory ${dirPath}.`;
+        if (xmlFile) {
+          message += ' XML format not supported, input FHIR definitions must be JSON.';
+        }
```
cmoesel: Maybe you should add an `else` here that adds `Input FHIR definitions must be JSON.` to the message?  Otherwise `Invalid file detected in directory` is not a very helpful message (e.g. "why is it invalid?").

test/ig/IGExporter.test.ts
cmoesel: I don't see any new tests.  Did one of the existing tests cover this?  (Sorry -- haven't had time to closely review, but just noticed this comment hanging out here and no new tests).

src/ig/IGExporter.ts
ngfreiter: Ouch yeah fixed that here https://github.com/FHIR/sushi/pull/262/commits/7c3f17e30520a5f45ada23ce0b7e8f867904ed0f

src/ig/IGExporter.ts
ngfreiter: Improved here https://github.com/FHIR/sushi/pull/262/commits/7c3f17e30520a5f45ada23ce0b7e8f867904ed0f

test/ig/IGExporter.test.ts
ngfreiter: Well yeah the existing tests had been changed to cover this, but it's probably better to make it explicit in its own test anyway, so now it is explicit in its own test.

cmoesel:
So... I just thought of something else.  Should we support the use case where someone might add these additional definitions and _use them_ in the FSH files?  For example, if they add a BYOJ extension definition, should they be able to use it in a FSH profile?

This would mean bringing the definitions in earlier in the process (when we process the dependencies).  I think maybe we should support this.  What do you think, @ngfreiter?  If you think we should support it, does it make sense to do it as a separate PR or to just update this one?

ngfreiter:
If we do it I think it would make sense to do it in this PR. It seems to me that if we are storing these defs in memory earlier on that then changes how they get added to the IG resource. If we fully load them into the StructureDefinition class and so on, would that mean we wouldn't have to process them differently at all? At the point when we are exporting to the IG resource, there would just be a list of resources contained in our classes, and it wouldn't matter if they were originally loaded from FSH or from a bring your own JSON file. So I'm not sure it would make sense to do separately, since it feels to me like it would make most of this PR not totally relevant anymore.

ngfreiter:
I think it does make sense to do this. I think this offers significant advantages over the alternative of keeping a separate `dev` package that contains things that you want to load in. Those advantages being:
1) Simpler than keeping a `dev` package
2) What if you wanted resources that you wanted to depend on in your IG. You'd have to use this approach, not the `dev` package approach. Since in that approach you can depend on resources, but not actually have them be listed in the IG.

ngfreiter:
Ok @cmoesel I made this so that the definitions are loaded in earlier in the process, so that they can be used by someone in FSH. I did not use the approach I outlined above, as that approach is a little too tricky. Instead the custom resources are loaded in as if they are an externally defined package.

src/fhirdefs/load.ts
```diff
@@ -113,6 +113,51 @@ export async function loadDependency(
...
+  // Similar code for loading custom resources exists in IGExporter.ts addPredefinedResources()
+  const pathEnds = [
+    'capabilities',
+    'extensions',
+    'models',
+    'operations',
+    'profiles',
+    'resources',
+    'vocabulary',
+    'examples' // Must come last in case examples are of other resources
```
cmoesel: I thought we decided _not_ to load examples up front.  I'm guessing you need to do this since we also decided to put the file format checks up front?

src/fhirdefs/load.ts
```diff
@@ -113,6 +113,51 @@ export async function loadDependency(
...
+      for (const file of files) {
+        let resourceJSON: any;
+        if (file.endsWith('.json')) {
+          resourceJSON = fs.readJSONSync(path.join(dirPath, file));
+        } else {
+          xmlFile = xmlFile || file.endsWith('.xml');
+          invalidFile = true;
+          continue;
+        }
+        defs.add(resourceJSON);
```
cmoesel: Perhaps this should be:
```ts
if (pathEnd !== 'examples') {
  defs.add(resourceJSON);
}
```
I'm not sure if examples would cause any harm being in there -- but I'm 99.99% sure they would not cause any harm _not_ being in there!

src/ig/IGExporter.ts
```diff
@@ -426,95 +427,78 @@ export class IGExporter {
       'vocabulary',
       'examples' // Must come last in case examples are of other resources
```
cmoesel: Technically this comment isn't true anymore, right?  Since we preload into FHIRDefs, `examples` can be processed _whenever_ here...

src/fhirdefs/load.ts
cmoesel: Also -- I don't think that comment is relevant anymore.  It's fine for `examples` to be last, but we no longer need to document a specific reason for it.

src/ig/IGExporter.ts
ngfreiter: Yeah those comments are no longer relevant. I will remove them.

src/fhirdefs/load.ts
ngfreiter: Yeah I left it in for the checks of xml and such.

src/fhirdefs/load.ts
ngfreiter: They actually don't get loaded, `defs.add` only will add things of `resourceType` `StructureDefinition`, `ValueSet`, `CodeSystem`, or `ImplementationGuide`. But I guess that is not entirely clear, and it is possible the behavior of `defs.add` will change, so I will add that check just to make things more explicit.

cmoesel:
@ngfreiter -- I just pushed up a simple change related to something I noticed when testing out this PR locally.  Can you take a quick look?

ngfreiter:
Yeah that looks good, ordering is neater that way.

PR #263, opened 2020-03-02
Don't print logs during testing

cmoesel:
When the log statements are printed out to the console during testing, it creates a lot of clutter.  This suppresses log messages during testing.  They can be temporarily turned back on by setting the MUTE_LOGS constant to false.

BONUS: Also creates a type for the log level so I don't have to keep trying to remember if it's `'warn'` or `'warning'`.


cmoesel:
Test this by running the unit tests on master and noticing that there are a lot of log statements.  Then run them on this branch and notice there aren't.

And if you're thorough, you will also confirm that this did not affect the CLI (which should still print to the console).

Issue #182, opened 2020-02-02
Scope of extension names is global (should it be?)


markkramerus:
We have noticed that extensions can clash or be confused with extensions of the same name occurring elsewhere. This could be a problem, especially when importing externally-defined extensions that the FSHerfolk don't control. I'm not sure if the url could differentiate if the names are the same. But this leads to some unexpected behavior, especially when folks are assuming the in-line extension is locally scoped.

cmoesel:
Yes, David Hay has also noted this issue (#86).  It seems like maybe the best approach is to consider requiring that all external entities be referenced by their canonical URL (or an alias that points to their canonical URL).

cmoesel:
Fixed by #264.

PR #264, opened 2020-03-02
Extension contains syntax

### LINKED ISSUE #86: Issue creating complex extension definitions
When creating a complex extension using ginzu slicing, sushi attempts to locate an existing definition for each 'sub extension child' based on the name (as I understand it). I suspect that in most cases, the intention of the author is to create these child elements directly in the extension - rather than referencing some other.

That being the case, is the current behaviour the best? Would it be better to assume that the intention is to create an inline child, and have some other syntax when wanting to refer to an existing one?

As an example, the following would fail if anyone else used 'latitude' or 'longitude' as an extension name:

```
Extension: geolocation
Id: geolocation
Description:  "None supplied yet"

* extension contains
    latitude 0..1 and
    longitude 0..1

// definitions of sub-extensions
* extension[latitude].url = "latitude"
* extension[latitude] ^definition = "No description"
* extension[latitude].value[x] only decimal

* extension[longitude].url = "longitude"
* extension[longitude] ^definition = "No description"
* extension[longitude].value[x] only decimal
```

cmoesel:
Instead of treating slice name as both the slice name and a potential reference to an external extension, make an explicit syntax for declaring the slice name *and* type.  E.g.,
```
  Alias: MaxSizeExtension: http://hl7.org/fhir/StructureDefinition/maxSize
  // then within a profile definition:
  extension contains MaxSizeExtension named maxSize
```
See conversation: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Proposed.20Syntax.20Change.20for.20Extensions

Fixes #86 and #182.

**TO TEST**:
1. Run this branch against `fsh-mcode` `master` branch
2. Note the warnings; and note that the 1st warning is different than the others, because it concerns an extension that is defined outside the FSH Tank
3. Run this branch against `fsh-mcode` `new-extension-syntax` branch and note that the warnings go away
4. (Optional) Look at [new-extension-syntax](https://github.com/standardhealth/fsh-mcode/compare/new-extension-syntax) diff to see changes that resulted in removing warnings
5. Compare output of `sushi:master`/`fsh-mcode:master` run with `sushi:contains-extension-syntax`/`fsh-mcode:new-extension-syntax` run.  The results should be identical (well, there are two diffs in a *description* because I trimmed trailing whitespace off the description).


antlr/src/main/antlr/FSH.g4
```diff
@@ -97,6 +97,7 @@ KW_PREFERRED:       '(' WS* 'preferred' WS* ')';
 KW_EXTENSIBLE:      '(' WS* 'extensible' WS* ')';
 KW_REQUIRED:        '(' WS* 'required' WS* ')';
 KW_CONTAINS:        'contains';
+KW_NAMED:           'named';
```
ngfreiter: We will want to add this to the VSCode extension.

src/export/StructureDefinitionExporter.ts
```diff
@@ -178,6 +175,107 @@ export class StructureDefinitionExporter implements Fishable {
...
+          let card = 'm..n';
+          const cardRule = fshDefinition.rules.find(
+            r => r instanceof CardRule && r.path === `${rule.path}[${item.name}]`
+          );
+          if (cardRule) {
+            card = `${(cardRule as CardRule).min}..${(cardRule as CardRule).max}`;
+          }
+          // Customize the message based on whether or not extension comes from the tank
+          const fshMeta = this.tank.fishForMetadata(item.name);
+          if (fshMeta && fshMeta.url === external.url) {
```
ngfreiter: I'm a little confused by the `fshMeta.url === external.url` check. What is this check necessary for?

src/export/StructureDefinitionExporter.ts
```diff
@@ -178,6 +175,107 @@ export class StructureDefinitionExporter implements Fishable {
...
+            let altSliceName = words(fshMeta.name)
+              .join('-')
+              .toLowerCase();
+            if (altSliceName === fshMeta.name) {
+              altSliceName += '-ext';
+            }
+            logger.warn(
+              `Extension with slice name '${item.name}' will be treated as an inline extension, even ` +
+                'though the name can be resolved to an extension defined in the FSH Tank. Starting with ' +
+                ' SUSHI 0.10.0, extension slices using standalone extensions should explicitly declare a ' +
```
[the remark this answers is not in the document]
cmoesel: Dang.  But if that's the worst you found, I still call it a win.

src/export/StructureDefinitionExporter.ts
cmoesel: *In theory*, if `fshMeta` is not null, then the URLs should probably always match.  BUT... that assumes that when `item.name` was resolved, that the tank was searched _first_, before core FHIR or any external IGs.  I put the URL check in as insurance in case the tank is not searched first (particularly if any code in that area changes).  The URL is the only guaranteed unique identifier (as opposed to `name`, which can have multiple matches).

It's probably unnecessary 99.99% of the time, but... it's also harmless (as far as I can tell).

ngfreiter: Ok that's fair, if we change our resolution order then I could see this being meaningful, and you're right I can't see how it would hurt. So I'm alright with it!

src/export/StructureDefinitionExporter.ts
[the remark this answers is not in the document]
cmoesel: Fixed in 4903eaebcde39550524a065956a8c6a13a2d3b14.

cmoesel:
Documentation PR: https://github.com/HL7/fhir-shorthand/pull/32

src/export/StructureDefinitionExporter.ts
```diff
@@ -178,6 +175,107 @@ export class StructureDefinitionExporter implements Fishable {
...
+    }
+    rule.items.forEach(item => {
+      if (item.type) {
+        const extension = this.fishForMetadata(item.type, Type.Extension);
+        if (extension == null) {
+          logger.error(
+            `Cannot create ${item.name} extension; unable to locate extension definition for: ${item.type}.`,
+            rule.sourceInfo
+          );
+          return;
```
jafeltra: As I see you went with the `return` in the `forEach`. I hope I haven't swayed you out of your preferred ways.

test/import/FSHImporter.Profile.test.ts
```diff
@@ -845,6 +899,25 @@ describe('FSHImporter', () => {
...
+        const result = importSingleText(input);
+        const profile = result.profiles.get('ObservationProfile');
+        expect(profile.rules).toHaveLength(3);
+        assertContainsRule(profile.rules[0], 'component.extension', {
+          name: 'offset',
+          type: 'http://hl7.org/fhir/StructureDefinition/observation-timeOffset'
+        });
+        assertCardRule(profile.rules[1], 'component.extension[offset]', 0, 1);
+        assertFlagRule(profile.rules[2], 'component.extension[offset]', true, undefined, undefined);
+      });
```
jafeltra: Is there any need to add a test to check that the type is correctly imported from an existing FSH defined extension (one that has not been aliased)?

src/export/StructureDefinitionExporter.ts
cmoesel: Ha.  Yeah, I guess maybe I came at it with a more open mind than I previously would have.  That's not a bad thing!

test/import/FSHImporter.Profile.test.ts
cmoesel: Sure, I can do that!

cmoesel: Addressed in 65f02b6e679db12f7136d5e68c44c80a4f01ce93

PR #265, opened 2020-03-03
Import mixins

ngfreiter:
This adds importing of mixins as described here http://build.fhir.org/ig/HL7/fhir-shorthand/branches/master/index.html#defining-mixins. This addresses task https://standardhealthrecord.atlassian.net/browse/CIMPL-285.

Mixins are just treated as a group of rules, there is no support for externally defined mixins, they must be defined in FSH.


ngfreiter:
I realized the logs for `FSHImporter.Mixin.test.ts` were not muted, so I fixed that here https://github.com/FHIR/sushi/pull/265/commits/8b6f01a850dd99f8b82af19b58bc490f786ed87b.

antlr/src/main/antlr/FSH.g4
```diff
@@ -21,6 +21,8 @@ vsMetadata:         id | title | description;
 codeSystem:         KW_CODESYSTEM SEQUENCE csMetadata* (caretValueRule | concept)*;
 csMetadata:         id | title | description;
 
+mixin:              KW_MIXIN SEQUENCE sdRule*;
+
```
mint-thompson: Should it be legal to define a mixin with no rules? If not, this should be changed.

src/import/FSHImporter.ts
```diff
@@ -592,6 +621,17 @@ export class FSHImporter extends FSHVisitor {
...
+    if (ctx.COMMA_DELIMITED_SEQUENCES()) {
+      return ctx
+        .COMMA_DELIMITED_SEQUENCES()
+        .getText()
+        .split(/,\s*/);
+    } else {
+      return [ctx.SEQUENCE().getText()];
+    }
+  }
+
```
mint-thompson: This wasn't introduced in this PR, but looking at this code and the definition of `COMMA_DELIMITED_SEQUENCES`, I'm wondering if that lexer rule needs to be changed to allow whitespace between the `SEQUENCE` and the `,` tokens. The regex you're currently using is a correct fit for the current grammar, though.

antlr/src/main/antlr/FSH.g4
ngfreiter: Good point, a mixin with no rules really makes no sense. No longer allowed in https://github.com/FHIR/sushi/pull/265/commits/7d4aa648d5ef767027064efdf0b054332a39b81a.

src/import/FSHImporter.ts
ngfreiter: I updated the grammar and the regex to allow whitespace before the `,` in https://github.com/FHIR/sushi/pull/265/commits/7d4aa648d5ef767027064efdf0b054332a39b81a. I also updated the regex in the two other places where `COMMA_DELIMTED_SEQUENCES` are used, and updated some tests to prove that you can now have whitespace before the commas.

mint-thompson:
Latest changes look good! What are you thoughts about removing duplicates from the list in `mixins`?

ngfreiter:
> Latest changes look good! What are you thoughts about removing duplicates from the list in `mixins`?

Do you mean if someone did:
```
Profile: Prof
Parent: Patient
Mixins: Mix1, Mix2, Mix3, Mix1
```
get rid of the duplicated Mix1? Or am I misunderstanding?

mint-thompson:
> > Latest changes look good! What are you thoughts about removing duplicates from the list in `mixins`?
> 
> Do you mean if someone did:
> 
> ```
> Profile: Prof
> Parent: Patient
> Mixins: Mix1, Mix2, Mix3, Mix1
> ```
> 
> get rid of the duplicated Mix1? Or am I misunderstanding?

Yes, that's the situation I'm thinking about. I can't think of any situations where the user would want to apply a specific set of rules twice, so it seems more likely to me that this represents an input error.

ngfreiter:
Yeah I can't think of why someone would want to do that, so I'll add that.

mint-thompson:
Thank you! Would it also be possible to log something when duplicates are removed? I'm thinking this would help in a situation where two mixins have similar names, and the author accidentally lists one of them twice instead of listing both of them.

ngfreiter:
Yeah that's a good point, I added a warning when mixins are deduplicated.

ngfreiter:
Hey Nick, why'd you call it `Mixin`? We are going with `RuleSet`.

ngfreiter:
Ok @ngfreiter, I will update per your request.

ngfreiter:
This is updated to use `RuleSet` instead of `Mixin` now. It looks like @mint-thompson, your review has now been dismissed unfortunately.

antlr/src/main/antlr/FSH.g4
```diff
@@ -1,17 +1,17 @@
...
 
 profile:            KW_PROFILE SEQUENCE sdMetadata+ sdRule*;
 extension:          KW_EXTENSION SEQUENCE sdMetadata* sdRule*;
-sdMetadata:         parent | id | title | description;
+sdMetadata:         parent | id | title | description | mixins;
 sdRule:             cardRule | flagRule | valueSetRule | fixedValueRule | containsRule | onlyRule | obeysRule | caretValueRule;
 
 instance:           KW_INSTANCE SEQUENCE instanceMetadata* fixedValueRule*;
-instanceMetadata:   instanceOf | title | description;
+instanceMetadata:   instanceOf | title | description | mixins;
```
cmoesel: If `instanceMetadata` allows mixins, and if mixins allow _all_ rule types, then _somewhere_ we need to add validation of the mixedin ruleset to ensure it only contains rule types supported by instances (currently FixedValueRules only).  Logically, it makes sense (to me) for that check to be on the import side -- but it would likely need to be a post-processing step since it's possible that an instance might mix in a RuleSet that hasn't been imported yet.  What do you think?

ngfreiter: I have this happening on the export side. I see why it makes sense on the import side, but let me make the case for the export side. The `Mixins` on an `Instance` don't get resolved to actual `RuleSet`s until the export side. On the import side the `Mixins` are just tracked as a list of strings that may or may not resolve to some actual `RuleSet`s. So to check on the import side, I would need to post-process as you said, and do the resolution to `RuleSet`s that is already happening on the export side, just to check the `RuleSet` has the correct types of rules, not to actually apply them. If I do this check on the export side, the `Mixin` name only has to be resolved to a `RuleSet` in this one place.

Additionally, I convinced myself it made sense to log errors on the export side, because these errors are about invalid rules, and the export side is where all other errors about invalid rules get logged. But maybe I convinced myself because it was more convenient to code 😉 , so if I haven't convinced you, let me know and I'll add the check on the import.

test/import/FSHImporter.Profile.test.ts
```diff
@@ -54,11 +55,12 @@ describe('FSHImporter', () => {
         expect(profile.id).toBe('observation-profile');
         expect(profile.title).toBe('An Observation Profile');
         expect(profile.description).toBe('A profile on Observation');
+        expect(profile.mixins).toEqual(['Mixin1', 'Mixin2,Mixin3', 'Mixin4']);
```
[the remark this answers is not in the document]
ngfreiter: Yeah, I guess I just assumed it was that way for a reason, but if you don't know that reason, then I will happily change it.

antlr/src/main/antlr/FSH.g4
cmoesel: OK.  I'm convinced.  Export side it is!

ngfreiter:
Ok @cmoesel I addressed your comment and updated with the latest changes from master.

Issue #268, opened 2020-03-05
Template version not copied from ig.ini to build


markkramerus:
The ig-data ig.ini file has `template = hl7.fhir.template#0.1.0` but what ends up in the build is `template = hl7.fhir.template`

cmoesel:
We are using an `ini` parsing library -- and I think it is treating the `#` as a comment and discarding it.  We'll have to look into it.

PR #271, opened 2020-03-06
Ig data copy tweaks

### LINKED ISSUE #267: Need to preserve ignoreWarnings.txt if supplied
Running SUSHI doesn't transfer the contents to build

cmoesel:
Fixes #267 and #268.  See individual commits for details.

### DIFF
--- a/src/ig/IGExporter.ts
+++ b/src/ig/IGExporter.ts
@@ -52,6 +52,7 @@ export class IGExporter {
     this.addIncludeContents(outPath);
     this.addIgIni(outPath);
     this.addPackageList(outPath);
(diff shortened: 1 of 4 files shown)


ngfreiter:
@cmoesel As Travis is showing, this fails on windows. When changing the tests it looks like you introduced extra spaced which don't exist on the output:
image

ngfreiter:
More specifically it looks like it is the line endings, passed if you change to `\r\n` like so: 
image

cmoesel:
Ugh.

cmoesel:
But if I switch it to that, then it will probably fail on Mac.  I'll probably have to do OS-specific line-endings.  Ugh.  Stand by.

PR #273, opened 2020-03-09
Export RuleSets

ngfreiter:
This addresses https://standardhealthrecord.atlassian.net/browse/CIMPL-286 by adding exporting of `RuleSet`s. On an instance, a `RuleSet` can only apply `FixedValueRule`s, and we log an error when the rule is not of that type. To help the user track down errors that come from `RuleSet` rules, the rules have an added `appliedLocation` and `appliedFile` field in addition to the `location` and `file` fields that are on all `FSHEntity`s. Rules are added to the beginning of the `rules` array on the `Profile`, `Extension`, or `Instance` on which the `RuleSet` is applied. The rules are then processed as if they were normal rules.


src/fhirtypes/common.ts
```diff
@@ -205,6 +213,41 @@ export function replaceField(
...
+ * Adds Mixin rules onto a Profile, Extension, or Instance
+ * @param {Profile | Extension | Instance} fshDefinition - The definition to apply mixin rules on
+ * @param {FSHTank} tank - The FSHTank containing the fshDefinition
+ */
+export function applyMixinRules(
+  fshDefinition: Profile | Extension | Instance,
+  tank: FSHTank
+): void {
+  // Rules are added to beginning of rules array, so add the last mixin rules first
+  [...fshDefinition.mixins].reverse().forEach(mixinName => {
```
cmoesel: Would it be better just to create a top-level `mixedInRules` array, traverse the mixins in order pushing rules onto that array, and then prepend the `mixedInRules` array to the `fshDefinition.rules`?  The advantages are only minor but I think they are:
* It's easier to read/understand than going through the mixins backward
* If multiple errors are logged, they'll be logged in the same order as the author mixed things in
* _Maybe_ more efficient to only prepend an array once instead of multiple times?  (Not sure, I'm making an assumption that prepending requires a whole new allocation whereas pushing doesn't -- but it may be a totally invalid assumption).

Anyway, this is only a light suggestion.  If you disagree at all then keep it as is.

src/fhirtypes/common.ts
```diff
@@ -205,6 +213,41 @@ export function replaceField(
...
+    if (ruleSet) {
+      ruleSet.rules.forEach(r => {
+        // Record source information of Profile/Extension/Instance on which Mixin is applied
+        r.sourceInfo.appliedFile = fshDefinition.sourceInfo.file;
+        r.sourceInfo.appliedLocation = fshDefinition.sourceInfo.location;
+      });
+      const rules = ruleSet.rules.filter(r => {
+        if (fshDefinition instanceof Instance && !(r instanceof FixedValueRule)) {
+          logger.error(
+            'Mixin rules applied to an instance must fix a value. Other rules are ignored.',
```
cmoesel: Honest question -- should we refer to these as _Mixin rules_ or _Mixins rules_ (since that's the keyword)?  I'm not sure of the right answer.  Did you think about it?

src/fhirtypes/common.ts
```diff
@@ -205,6 +213,41 @@ export function replaceField(
...
+            r.sourceInfo
+          );
+          return false;
+        }
+        return true;
+      });
+      fshDefinition.rules = [...rules, ...fshDefinition.rules];
+    } else {
+      logger.error(`Unable to find definition for Mixin ${mixinName}.`, fshDefinition.sourceInfo);
```
cmoesel: Until we support other things, should this be: _Unable to find definition for RuleSet ${mixinName}_?

cmoesel:
Also... update your branch!

src/fhirtypes/common.ts
ngfreiter: I think for bullet point 2 alone that's worth doing. I'll update.

src/fhirtypes/common.ts
ngfreiter: I did think about it, and but nothing really satisfied me. I know `Mixins` is the keyword, but that is the keyword because it is the plural of `Mixin`, and I think `Mixin` is better grammar. I think I will change it to "Rules applied by mixins". Better?

src/fhirtypes/common.ts
ngfreiter: Yeah that makes sense, I'll change it.

src/fhirtypes/common.ts
cmoesel: Sure!

ngfreiter:
Ok @cmoesel updated from your comments and merged master

PR #275, opened 2020-03-11
Contained instances

### LINKED ISSUE #224: error: Cannot resolve element from path: contained.Questionnaire.id
Trying to create an instance of a contained resource but the path is not found

This should be enough to show how to trigger the error:

Instance:       ACDCExamplePHQ-2QuestionnaireResponse
InstanceOf:     ACDCQuestionnaireResponse
Title:          "IHE ACDC Example Questionnaire for PHQ-2"
* contained.Questionnaire.id = "contained-questionnaire"

ngfreiter:
This fixes #224 and addresses https://standardhealthrecord.atlassian.net/browse/CIMPL-288 by adding the ability to have instances contained in another instance. Any instance can be contained in another instance, and the syntax is:
```
Instance: MyPatient
InstanceOf: Patient
* name[0].family = "InstanceMan"

Instance: MyBundle
InstanceOf: Bundle
// set entry[0].resource to contain the MyPatient instance
```

This also adds the `Inline` keyword to the set of possible `Usage`s for an instance. An instance does not have to be declared `Inline` to be contained in another instance, but if an instance is declared `Inline`, then no output will be generated for that instance, the only way that instance will appear as output is if it is contained inside of another instance.


test/export/InstanceExporter.test.ts
```diff
@@ -1102,6 +1102,40 @@ describe('InstanceExporter', () => {
...
+        .withLocation([1, 2, 3, 4]);
+      inlineRule.fixedValue = 'MyFakePatient';
+      inlineRule.isResource = true;
+      patientInstance.rules.push(inlineRule); // * contained[0] = MyFakePatient
+
+      const exported = exportInstance(patientInstance);
+      expect(exported.contained).toBeUndefined();
+      expect(loggerSpy.getLastMessage('error')).toMatch(
+        /MyFakePatient.*File: FakeInstance.fsh.*Line: 1 - 3\D*/s
+      );
```
jafeltra: Should this also check for the logged error "Cannot find definition for Instance: MyFakePatient"?

test/utils/InstanceDefinitionUtils.test.ts
```diff
@@ -111,11 +120,29 @@ describe('InstanceDefinitionUtils', () => {
   });
 
   describe('Non-example instances', () => {
```
jafeltra: This is definitely my bad, but since these tests are not linked to each other, I feel like we don't actually need this separate `describe`.  How do you feel about getting rid of that wrapper around the rest of these tests?

test/export/InstanceExporter.test.ts
ngfreiter: I actually think that is the error that is being checked for here. I didn't check for the words themselves, just the content, in case we wanted to change the wording of the error messages. I can make it explicit though if you think this is too unclear.

test/utils/InstanceDefinitionUtils.test.ts
ngfreiter: Sounds good, I will change that!

test/export/InstanceExporter.test.ts
jafeltra: I think was was thinking of other places where we checked the wording of the error along with the source information. I agree that the source info is checking that the error is logged though, so if you feel like checking the wording isn't necessary, I'm okay with that.

ngfreiter: I'm going to check the wording. Why not be as clear as we can.

ngfreiter:
@jafeltra thanks for the comments, they should be fixed!

ngfreiter:
Yes @cmoesel it is only when the type is `Resource`. That's a good idea for a follow on, we will need to think about how we are going to validate that and how much we want to validate that.

PR #280, opened 2020-03-12
One sided cardinality statements

### LINKED ISSUE #258: FSH should allow one-sided cardinality statements
Currently, FSH requires you to specify both sides of a cardinality, e.g.:
```
Profile: MyProfileWithOneOptionalCategory
Parent: Observation
* category 0..1
```
but... we should allow authors to state only the side of the cardinality that changed, e.g.:
```
Profile: MyProfileWithOneOptionalCategory
Parent: Observation
* category ..1
```
This would also work the other way:
```
Profile: MyProfileWithAtLeastOneCategory
Parent: Observation
* category 1..
```
This would be an optional feature.  Authors still can specify both sides of the cardinality if they wish (and SUSHI will do the right thing).

This will help w/ tools that convert StructureDefinitions to FSH, as they can then work off the `differential` alone -- otherwise, if the `differential` contains only one side of the cardinality, they would need to look at the `snapshot` to get the other side in order to create a valid FSH cardinality rule with both sides of the cardinality.

jafeltra:
This PR fixes #258 and satisfies https://standardhealthrecord.atlassian.net/browse/CIMPL-289.

Both sides of a cardinality are now optional. A user can specify `* path 1..1`, `* path 1..`, or `* path ..1`. If one side of the cardinality is left off, the exporter will use the element's cardinality for that side. If neither side is specified (`* path ..`), an error will be logged and the element's current cardinality is used.

Since the `min` cardinality is of type `number`, when it parses the missing cardinality, it is parsed as an empty string, which is passed into `parseInt`. This results in `NaN`, which is why there are a few of those in the tests and an `isNaN` check. This felt a bit unexpected to me, but since `NaN` is still a type `number`, it felt the cleanest way to handle the absence of a `min` cardinality without updating types on the rule and in a bunch of functions. I'm open to other feedback as well.


src/import/FSHImporter.ts
```diff
@@ -795,8 +795,14 @@ export class FSHImporter extends FSHVisitor {
...
+  private parseCard(card: string, rule: CardRule): { min: number; max: string } {
     const parts = card.split('..', 2);
+    if (parts[0] === '' && parts[1] === '') {
+      logger.error(
+        `Neither side of the cardinality was specified on path "${rule.path}". A min, max, or both need to be specified.`,
+        rule.sourceInfo
+      );
+    }
     return {
       min: parseInt(parts[0]),
```
ngfreiter: What if here you did 
```
min: parseInt(parts[0]) || undefined,
```
Then the `min` would be undefined if no min was specified. This is only a suggestion and maybe there is some reason I don't see why this doesn't work. Even if this does work, I'm still willing to be convinced that leaving it as NaN is cleaner anyway, because it is not a number.

src/export/StructureDefinitionExporter.ts
```diff
@@ -118,6 +118,9 @@ export class StructureDefinitionExporter implements Fishable {
       if (element) {
         try {
           if (rule instanceof CardRule) {
+            // If only one side of the cardinality is set by the rule, use element's current cardinality
+            if (isNaN(rule.min)) rule.min = element.min;
+            if (rule.max === '') rule.max = element.max;
```
ngfreiter: Would it be better here to handle these cases inside `element.constrainCardinality` itself? To me that seems a little more self contained. What do you think? If you did make this change you'd also have to add some tests for when min, max, or both are undefined for `ElementDefinition.constrainCardinality`.

test/import/FSHImporter.Profile.test.ts
```diff
@@ -193,6 +193,48 @@ describe('FSHImporter', () => {
...
+        * category 1..
+        `;
+
+        const result = importSingleText(input);
+        const profile = result.profiles.get('ObservationProfile');
+        expect(profile.rules).toHaveLength(1);
+        assertCardRule(profile.rules[0], 'category', 1, ''); // Unspecified max
+      });
+
+      it('should parse card rule with only min', () => {
```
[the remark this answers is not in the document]
jafeltra: Whoops. Yes to both of these.

src/import/FSHImporter.ts
jafeltra: Hm that's a good idea! I actually think that if I do this for the `min` and `max` here, then both values will consistently be `undefined` if they have no value specified. I think I like the consistency for the two values, so I'm going to update this and the tests.

src/export/StructureDefinitionExporter.ts
jafeltra: I like that idea! I was thinking I wanted the true cardinalities that were being constrained be the ones passed into that function, but we do other logical checking within that function, so I can add these to the top of that as well.

src/import/FSHImporter.ts
jafeltra: After attempting this, it appears that `max` doesn't easily work having an `undefined` value. I'll keep exploring these options.

jafeltra:
Rebased the branch. 

Also moved the checks for unspecified min/max into constrain cardinality. I ended up not changing `min`'s default to `undefined` because I didn't see a huge benefit of that over `NaN` and there were too many small workarounds I would have to do to support `max`'s default being `undefined`, so I kept them both as is. Let me know your thoughts @ngfreiter.

PR #282, opened 2020-03-16
Import Mapping

ngfreiter:
This addresses https://standardhealthrecord.atlassian.net/browse/CIMPL-299 by adding the ability the import `Mapping` entities as described here https://build.fhir.org/ig/HL7/fhir-shorthand/reference.html#defining-mappings. Note that the documentation does not have a `Description` keyword, but that is added here. Note also that we have decided to update the syntax of mapping rules to include optional `language` and `comment` components, so mapping rules now have the syntax:
```
{path} -> {target} "{comment}" {language-code}
```
where the `{comment}` and `{language-code}` are optional.


test/import/FSHImporter.Mapping.test.ts
```diff
@@ -0,0 +1,203 @@
...
+      expect(mapping.id).toBe('my-map');
+      expect(mapping.source).toBe('Patient');
+      expect(mapping.target).toBe('http://some.com/mappedTo');
+      expect(mapping.description).toBe('This is a description');
+      expect(loggerSpy.getLastMessage('error')).toMatch(
+        /This is a description.*File: Mapping\.fsh.*Line: 10\D*/s
+      );
+    });
+
+    it('should only apply each metadata attribute the first time it is declared', () => {
```
jafeltra: What is the difference between this test and the one above besides checking the error that is logged in the test above? Do we need this test?

src/import/FSHImporter.ts
```diff
@@ -700,6 +772,14 @@ export class FSHImporter extends FSHVisitor {
     return concept;
   }
 
+  visitSource(ctx: pc.SourceContext): string {
+    return this.aliasAwareValue(ctx.SEQUENCE());
```
jafeltra: Should there be a test in `FSHImporter.Mapping.test.ts` checking that Alias's are supported on `Source`? Or is that covered by other Alias-related tests elsewhere?

src/fshtypes/rules/MappingRule.ts
```diff
@@ -0,0 +1,12 @@
+import { Rule } from './Rule';
+import { FshCode } from '../FshCode';
+
+export class MappingRule extends Rule {
+  map: string;
+  language: FshCode;
+  comment: string;
```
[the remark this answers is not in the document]
ngfreiter: Yup good point!

test/import/FSHImporter.Mapping.test.ts
ngfreiter: Nope, good point!

src/import/FSHImporter.ts
ngfreiter: Yup, good point!

ngfreiter:
> Also, I noticed that in our documentation, the first rule we have that does not have a path has a RHS that is not a string, `* -> Patient`. I'm guessing the code you have here making that RHS a string is correct and we want to update the documentation to `* -> "Patient"`, but I wanted to double check since I just noticed.

Yeah looks like that is just an error in the FSH Docs. The docs need to be updated in a few places in reference to mappings, so I will add that as something that has to be fixed.

ngfreiter:
Ok @jafeltra updated from your comments!

PR #283, opened 2020-03-16
Update IG scripts

jafeltra:
This PR satisfies https://standardhealthrecord.atlassian.net/browse/CIMPL-301.

This updates the `updatePublisher` and `genonce` scripts based on the updates in the sample-ig: https://github.com/FHIR/sample-ig.

I had to made two changes:
- In the `_genonce.sh` script, the sample-ig was using `wget` to check for an internet connection, but we don't want to impose another dependency on users and the `updatePublisher` script is already using `curl`, so I changed [line 6](https://github.com/FHIR/sushi/blob/update-ig-scripts/src/ig/files/_genonce.sh#L6) in our script. The original line is [here](https://github.com/FHIR/sample-ig/blob/master/_genonce.sh#L6).
- In the `_gencontinuous.sh` script, I changed the reference to the `genonce` script from `_genonce.sh` to `./_genonce.sh` because I got a "command not found" error with the original line they had in the sample IG repo. For reference, my line is [here](https://github.com/FHIR/sushi/blob/update-ig-scripts/src/ig/files/_gencontinuous.sh#L2) and the sample-ig file is [here](https://github.com/FHIR/sample-ig/blob/master/_gencontinuous.sh#L2). I also made the same change in the `_gencontinuous.bat` script ([here](https://github.com/FHIR/sushi/blob/update-ig-scripts/src/ig/files/_gencontinuous.bat#L2)), but haven't been able to test it.

If either or both of these changes make sense and seem useful, I can make a PR on the sample-ig repository with the changes.

This PR needs at least one reviewer who is using a Windows computer to test the `.bat` scripts. Also any feedback on the changes I made to the existing sample-ig scripts is appreciated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,6 @@ If the input folder (i.e., "FSH Tank") contains a sub-folder named "ig-data", th
 
 After running SUSHI, change to the output folder and run the `_updatePublisher` script and enter `y|Y` to download latest publisher when prompted. Then run the `_genonce` or `_gencontinuous` script.
 `_genonce` will generate the IG one time. `_gencontinuous` will continuously run the `genonce` script, which will regenerate the IG with any changes that were made to the input of the IG build.
-Note that `curl` is required to run these scripts.
 
 If the input folder does not contain a sub-folder named "ig-data", then only the resources (e.g., profiles, extensions, etc.) will be generated.
 
--- a/README.md
+++ b/README.md
@@ -46,7 +46,9 @@ If the input folder (i.e., "FSH Tank") contains a sub-folder named "ig-data", th
 * `ig-data/input/pagecontent/*`: If present, all other files of any type that do not match the above patterns will be copied into the IG input, but will not appear in the table of contents.
 * `ig-data/input/{supported-resource-input-folder}`: JSON files found in supported subfolders will be processed as additional resources in the IG, added to the IG JSON file, and copied to the corresponding locations in the IG input. This allows for a "Bring Your Own JSON" capability when you already have resources that were generated via another method. If a provided JSON file has the same URL as a FSH-defined resource, the provided JSON file will overwrite the FSH-defined one. Supported folders match those documented in the [Guidance for FHIR IG Creation](https://build.fhir.org/ig/FHIR/ig-guidance/using-templates.html#root.input): capabilities, examples, extensions, models, operations, profiles, resources, vocabulary.
 
-After running SUSHI, change to the output folder and run the `_updatePublisher` and `_genonce` scripts.
+After running SUSHI, change to the output folder and run the `_updatePublisher` script and enter `y|Y` to download latest publisher when prompted. Then run the `_genonce` or `_gencontinuous` script.
+`_genonce` will generate the IG one time. `_gencontinuous` will continuously run the `genonce` script, which will regenerate the IG with any changes that were made to the input of the IG build.
+Note that `curl` is required to run these scripts.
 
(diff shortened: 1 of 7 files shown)


src/ig/files/_updatePublisher.sh
```diff
@@ -1,9 +1,48 @@
...
+		jarlocationname="Input Cache"
+		upgrade=false
+	fi
+fi
+
+if "$upgrade"; then
+	message="Overwrite $jarlocation?"
+else
+	echo Will place publisher jar here: "$jarlocation"
+	message="Ok?"
```
cmoesel: Can we match the windows one by making this `"Ok? (Y/N)"`?

src/ig/files/_updatePublisher.sh
```diff
@@ -1,9 +1,48 @@
...
+	else
+		echo IG Publisher NOT FOUND in input-cache or parent folder...
+		jarlocation=$input_cache_path$publisher_jar
+		jarlocationname="Input Cache"
+		upgrade=false
+	fi
+fi
+
+if "$upgrade"; then
+	message="Overwrite $jarlocation?"
```
cmoesel: Can we match the windows one by making this `"Overwrite $jarlocation? (Y/N)"`?

src/ig/files/_updatePublisher.sh
```diff
@@ -1,9 +1,48 @@
...
+else
+	echo Will place publisher jar here: "$jarlocation"
+	message="Ok?"
+fi
+
+read -r -p "$message" response
```
cmoesel: Personally, I would like a space between the question and the input (rather than my typed response being the character immediately following the question -- currently `OK?Y`). How do you feel about putting a space in (e.g., `"$message "`)?

Does Windows have the same behaviour or does batch insert a space for you?

README.md
```diff
@@ -46,7 +46,9 @@ If the input folder (i.e., "FSH Tank") contains a sub-folder named "ig-data", th
...
 
-After running SUSHI, change to the output folder and run the `_updatePublisher` and `_genonce` scripts.
+After running SUSHI, change to the output folder and run the `_updatePublisher` script and enter `y|Y` to download latest publisher when prompted. Then run the `_genonce` or `_gencontinuous` script.
+`_genonce` will generate the IG one time. `_gencontinuous` will continuously run the `genonce` script, which will regenerate the IG with any changes that were made to the input of the IG build.
+Note that `curl` is required to run these scripts.
```
cmoesel: This is actually only true of the bash scripts.  The windows scripts use something else.  You can probably just leave off this sentence.

src/ig/files/_updatePublisher.sh
ngfreiter: Same behavior on windows, you get `(Y/N)Y`

README.md
jafeltra: I'll take that sentence off! I also added the second sentence about each script, but it's a pretty repetitive sentence, so I can also take it out if you think it's unnecessary.

src/ig/files/_updatePublisher.sh
jafeltra: @ngfreiter did you notice this missing space? I can easily add it in if it's missing.

README.md
cmoesel: I like the sentences explaining the difference between `_genonce` and `_gencontinuous`.  Keep those!

jafeltra:
da32a17 has changes requested by @cmoesel!

PR #288, opened 2020-03-17
Export mapping

ngfreiter:
This addresses https://standardhealthrecord.atlassian.net/browse/CIMPL-300 by adding exporting of mappings as described here: https://build.fhir.org/ig/HL7/fhir-shorthand/reference.html#defining-mappings.

This completes the functionality that was started in #282. When a `Mapping` is applied to a `Profile` or `Extension`, the `StructureDefinition` itself has data pushed to its `mapping` array element. Each individual element on the SD that is reference by a rule on the `Mapping` also has data pushed to its `mapping` array element.


src/errors/InvalidMappingError.ts
```diff
@@ -0,0 +1,10 @@
+import { Annotated } from './Annotated';
+
+export class InvalidMappingError extends Error implements Annotated {
+  specReferences = [
+    'https://www.hl7.org/fhir/elementdefinition-definitions.html#ElementDefinition.mapping'
+  ];
+  constructor() {
+    super('Invalid mapping, mapping.identity and mapping.map are 1..1 and must be set.');
```
mint-thompson: This error message focuses on the FHIR requirement being violated, but it might be more useful if it instead focused on what the FSH author would need to add in order to satisfy that requirement.

ngfreiter: Makes sense, I'll change that.

ngfreiter:
> I didn't see a test case for a mapping rule from the structure definition's root element. `* -> Patient` is an example from the FSH documentation. Is this supported?

Yes that is supported, I think for any rule if you do `* {somerule}` without specifying a path, you operate on the root element. Except for CaretValueRules. On CaretValueRules you operate on the StructureDefinition itself. This is overall a bit of a confusing inconsistency that we do need to address at some point. I'll add a test for it though.

ngfreiter:
After some discussion, we have decided that mappings need to have a `Title` keyword that maps to `mapping.name` on the StructureDefinition, since the `mapping.name` can be repeated across multiple different profiles. The `Mapping` keyword will still be used to create a globally unique identifier in FSH for the mapping, but the value of the `Mapping` keyword will not actually be used anywhere in the FHIR definition. So if you see this comment before I make that update, be aware that there are changes coming on this PR.

PR #289, opened 2020-03-19
Primitive slice on instance

ngfreiter:
This addresses https://standardhealthrecord.atlassian.net/browse/CIMPL-213 by adding handling of slicing on primitives to instances.

When we are adding slices to an instance, we need to track which value is being used to represent which slice. This could be done using the discriminator for that slicing, but that is difficult, so instead we chose to add a `_sliceName` field to the slice which contains the name of the slice that is being represented by that object. The `_sliceName` field would then be removed before final export of the Instance. This works well when the slices are objects, but this approach was failing when the slices are primitives, because we cannot add a `_sliceName` field on a primitive. This PR fixes this by temporarily turning primitives into objects, so that the `_sliceName` field can be added, and then turning the objects back into primitives in the same `cleanInstance` function that removes the extra `_sliceNames`.

You can test this by checking out 5816e2d648f53eb7b4c8e80faa9d9b6f070b5032 and running the tests. You will see that some of the tests added by this commit fail, since they test adding slicing on primitives on an instance. Next, check out 08c79677b74c5fec8c2915d4c72986556a65379d, which adds the functionality described above, and you should see that the tests pass now.


src/fhirtypes/common.ts
```diff
@@ -66,9 +66,15 @@ export function setPropertyOnInstance(
...
+          if (typeof fixedValue !== 'object') {
+            // When a fixedValue is a primitive but also a slice, we convert to an object so that
+            // the sliceName field can be tracked on the object. The _primitive field marks the object
+            // to later be converted back to a primitive by replaceField in cleanInstance
+            fixedValue = { fixedValue, _primitive: true };
+          }
           const sliceIndices: number[] = [];
           // Find the indices where slices are placed
-          current[key].forEach((el: any, i: number) => {
+          current[pathPart.base].forEach((el: any, i: number) => {
```
mint-thompson: I think I'm not quite understanding how this works when we have a primitive. Since the key will be `_${pathPart.base}` , then how is it that `current[pathPart.base]` even exists at this point?

ngfreiter: Hm yes, this is part of a larger overall problem that I now see. You can access the child of a primitive even if the primitive itself doesn't exist. For example, the following FSH snippet:
```
* name[0].prefix[0].id = "oops"
```
does not give you any errors, even if `name[0].prefix[0]` does not yet exist and you end up with output like:
```
"name": [
    {
      "_prefix": [
        {
          "id": "oops"
        }
      ]
    }
  ]
```
which clearly shouldn't be allowed. So I will add logic to check that `current[pathPart.base]` does indeed exist, or else we shouldn't be adding it's so called "child" to the instance.

ngfreiter: Actually I am not sure. Should that not be allowed? You can do:
```
* name[0].prefix[0] = "Dr."
```
And the name array will be populated. It isn't like you need to make the name array before setting its children. So I don't think you should have to make the primitive before making its children. I don't know what the expected behavior would be because primitives having children that are tracked as sibling properties with an `_` appended is already a bit weird. What do you think?

jafeltra: Hm it looks like you actually can have a primitive extension (so something like `_prefix`) without having a fixed value for it:
>Note: If the primitive has an id attribute or extension, but no value, only the property with the _ is rendered.

(from: https://www.hl7.org/fhir/json.html#primitive)

Given that , maybe this isn't an issue and can stay how it is?

ngfreiter: Ok well that's weird, but yeah I think that answers that question then. Thanks!

jafeltra:
I didn't notice anything else beyond what Mint pointed out and I think has been resolved! This looks good to me, but I'll let Mint review again, so I'll hold off on merging!

PR #294, opened 2020-03-24
Error if Profile defines itself as Parent

jafeltra:
This PR covers CIMPL-283: https://standardhealthrecord.atlassian.net/browse/CIMPL-283

I confirmed that we already resolve name/id in the order specified in the issue:
1. Local definitions (in the package and in the tank) are resolved first
2. External references are resolved second

There were actually tests that indicated this was the case, which can be seen in [one test](https://github.com/FHIR/sushi/blob/master/test/utils/MasterFisher.test.ts#L145) and [another test](https://github.com/FHIR/sushi/blob/master/test/utils/MasterFisher.test.ts#L160). (These are just links to our existing code since I haven't changed anything related to them.)

Since this was already covered, I implemented the additional error that was mentioned in the Jira. If a profile declares a Parent with the same name as itself, we now throw an error. This prevents a "Maximum call stack size exceeded" error on master. If the parent name resolves to a FHIR resource, we also include the URL that should be used in place of the resource name in the FHIR message.

### DIFF
--- a/src/export/StructureDefinitionExporter.ts
+++ b/src/export/StructureDefinitionExporter.ts
@@ -410,7 +410,7 @@ export class StructureDefinitionExporter implements Fishable {
 
     const parentName = fshDefinition.parent || 'Resource';
 
-    if (fshDefinition.name === fshDefinition.parent) {
+    if (fshDefinition.name === parentName) {
       const result = this.fishForMetadata(parentName, Type.Resource);
       throw new ParentDeclaredAsProfileNameError(
         fshDefinition.name,
--- a/src/export/StructureDefinitionExporter.ts
+++ b/src/export/StructureDefinitionExporter.ts
@@ -6,7 +6,7 @@ import {
 } from '../fhirtypes';
 import { Profile, Extension, Invariant } from '../fshtypes';
 import { FSHTank } from '../import';
-import { ParentNotDefinedError } from '../errors';
+import { ParentNotDefinedError, ParentDeclaredAsProfileNameError } from '../errors';
 import {
   CardRule,
   FixedValueRule,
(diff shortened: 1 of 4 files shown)


src/export/StructureDefinitionExporter.ts
```diff
@@ -408,6 +409,16 @@ export class StructureDefinitionExporter implements Fishable {
     }
 
     const parentName = fshDefinition.parent || 'Resource';
+
+    if (fshDefinition.name === fshDefinition.parent) {
```
ngfreiter: I think on line 413 the check should be against `parentName` and not `fshDefinition.parent`. Currently I could do:
```
Profile: Resource
* rules
```
and still get the maximum call stack error. I do also get an error from the grammar, because the grammar wants parent to be there, but if we bother to do the `parentName` default setting on line 411 I think we should use it here.

src/export/StructureDefinitionExporter.ts
```diff
@@ -408,6 +409,16 @@ export class StructureDefinitionExporter implements Fishable {
...
+      const result = this.fishForMetadata(parentName, Type.Resource);
+      throw new ParentDeclaredAsProfileNameError(
+        fshDefinition.name,
+        fshDefinition.sourceInfo,
```
ngfreiter: Do we actually need to track the `sourceInfo` here? It seems like the `sourceInfo` would get added on line 466 when the error is caught and logged. I'm not sure if there is a case where the `sourceInfo` tracked here would differ from the `sourceInfo` logged there, since it seems like they both just come from the `fshDefinition`.

src/export/StructureDefinitionExporter.ts
jafeltra: I think you're right that this should use `parentName`. I think I had originally switched back to using `fshDefinition.parent` because I thought it would help in the case when we export an Extension, but I don't think it will affect that case and we should support the case you mentioned above.

src/export/StructureDefinitionExporter.ts
jafeltra: I think you're right that it won't change. I was mirroring what it seems like the `ParentNotDefinedError` below this one does, but I don't think the source info is actually being logged from either of those errors now that I look more closely. Do you have a preference for removing it to use the info logged on line 466 or keep it for consistency's sake?

ngfreiter: I can't seem to find a reason why it needs to be there, so if you can't either then yeah I think that makes sense to be consistent. Makes me a little nervous because I trust there would have been some reason at the time, but it seems like perhaps times have changed.

jafeltra: I agree. Perhaps times have changed. I think I vote to keep the sourceInfo on both for consistency and for the sake of having more information on the errors.

PR #297, opened 2020-03-25
Usage as a code

jafeltra:
This PR addresses https://standardhealthrecord.atlassian.net/browse/CIMPL-308

It changes `Usage` on Instances to expect a code, more similar to `Severity` on Invariants.

The implementation here just changes the grammar and import side to expect and parse a code. Once the code is parsed, we just use the `FshCode`'s `code` property to do the same checking for allowed usages. The `usage` property on the instance class is still expecting a `string` type with one of the three supported values, so the parse function just returns the `code` string.

I did look around a bit to see if there was a clean way to check the specific value of `code` in the Instance class for a specific value. There didn't seem to be a good way to add additional requirements to an existing type without changing the type itself, which I didn't want to do, or created a new, slightly different type. If anyone thinks of a way to do this that I missed, let me know, although I don't mind how it works now, so hopefully you feel similarly!


src/import/FSHImporter.ts
```diff
@@ -741,15 +742,17 @@ export class FSHImporter extends FSHVisitor {
   }
 
   visitUsage(ctx: pc.UsageContext): InstanceUsage {
-    let usage = ctx.SEQUENCE().getText();
+    const usageConcept = this.parseCodeLexeme(ctx.CODE().getText(), ctx.CODE())
```
mint-thompson: good to see `parseCodeLexeme` getting reused!

PR #304, opened 2020-03-27
Fix handling of local templates in ig.ini

### LINKED ISSUE #299: Sushi writes local folder name with \
I start MY ig.ini in ig-data with:
template= #ig-template

Sushi writes ig.ini as
template= \#ig-template

Build fails on IG Builder (I have a workaround for my local build).

cmoesel:
Local templates start with # in the ig.ini file, but we were expecting name#version. This fixes it so it works when # is the first character.

Fixes #299.

### DIFF
--- a/test/ig/IGExporter.test.ts
+++ b/test/ig/IGExporter.test.ts
@@ -485,7 +485,7 @@ describe('IGExporter', () => {
     });
   });
 
-  describe('#customized-ig', () => {
+  describe('#customized-ig-with-local-template', () => {
     let pkg: Package;
     let exporter: IGExporter;
     let tempOut: string;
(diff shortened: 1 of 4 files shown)


test/ig/IGExporter.test.ts
```diff
@@ -485,6 +485,33 @@ describe('IGExporter', () => {
     });
   });
 
+  describe('#customized-ig', () => {
```
ngfreiter: Could you change this name? There is already a `customized-ig` describe block.

cmoesel: Whoops.  Missed that one.  Fixed in 2353baa.

PR #308, opened 2020-03-27
Generate path-history parameter for HL7 IGs

### LINKED ISSUE #286: Path-history parameter missing from IG resource
The IG Publisher will give a warning if the path-history parameters is not set. For mCODE, it should look like this:
```
    },
    "parameter": [
      {
        "code": "copyrightyear",
        "value": "2019+"
      },
      {
        "code": "releaselabel",
        "value": "STU1"
      },
      {
        "code": "show-inherited-invariants",
        "value": "false"
      },
      {
        "code": "path-history",
        "value": "http://hl7.org/fhir/us/mcode/history.html"
      }
    ]
  }
```

cmoesel:
Generate a path-history parameter for HL7 IGs.  We use the same approach to recognizing HL7 IGs as the publisher does -- looking at the canonical to see if it is in the `hl7.org` domain.

The publisher expects history paths to have the format `${canonical}/history.html`.

Unit tests show how it works, but if you want to test manually, run the master branch against an HL7 IG and then run the IG Publisher.  After the publisher completes, look at `output/qa.html` in a browser.  You will see something like this near the top of the page:
image

Now do the same thing using this branch.  After running the IG Publisher again, now the `qa.html` should have this instead:
image

Fixes #286

### DIFF
--- a/src/ig/IGExporter.ts
+++ b/src/ig/IGExporter.ts
@@ -108,7 +108,7 @@ export class IGExporter {
           generation: 'html',
           page: [] // index.[md|html] is required and added later
         },
-        // Parameter apparently required by IG Publisher (as of Jan 29, 2020)
+        // Parameters apparently required by IG Publisher (as of Jan 29, 2020)
         parameter: [
           {
             code: 'copyrightyear',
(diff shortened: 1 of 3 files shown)


jafeltra:
This looks good to me! I looked at the tests and ran them and they make sense. I'm waiting for the publisher jar to update, which is taking a surprisingly long time, but once it downloads (I've come this far), I'll run and approve!

PR #310, opened 2020-03-27
Log error when invariant has invalid name

### LINKED ISSUE #290: Invariant names are not validated as ids
The invariant name, so `myInvariant` in the snippet below:
```
Invariant: myInvariant
```
is used to fill the `constraint.key` field on a StructureDefinition. The `constraint.key` field is of type `id`, which has a special regex pattern that is must conform to https://www.hl7.org/fhir/datatypes.html#id. For things like Instances or Profiles, we check that `id`s correctly conform to this pattern, but we don't check that this pattern is matched for an Invariant. So we should check that the name of an Invariant correctly conforms to this pattern, and log an error if it doesn't.

mint-thompson:
Fixes #290 and completes task https://standardhealthrecord.atlassian.net/browse/CIMPL-311.

The constraint keys on an element are required to be valid FHIR ids. These keys come from the name of the invariant, so show an error message when they don't match the FHIR id regex. The invariant will
still be added to the list of constraints.

I know during standup I said that I was originally planning to not add the invalid id to the list of constraints, but I was convinced otherwise in the spirit of "SUSHI should do as much as it can."


src/export/StructureDefinitionExporter.ts
```diff
@@ -167,6 +168,12 @@ export class StructureDefinitionExporter implements Fishable {
                 rule.sourceInfo
               );
             } else {
+              if (!idRegex.test(invariant.name)) {
```
jafeltra: This might be a silly question, but is it possible to use the `validateId` function in common.ts?

mint-thompson: The catch with that function is that it's a member of the `HasId` class, which means I would need an instance of that or something that mixes it in. `HasId` gets mixed in to a few different FHIR types, but there's no FHIR type to represent an invariant. So I'd need a slightly different mixin (since I'm checking the invariant's `name`, not `id`) for this situation.

jafeltra: Ahh gotcha. I somehow managed to miss that it was a part of that class. It seems reasonable that we can't reuse that function and I agree that it doesn't quite make sense to use that class. I was just noticing that we had a very similar error message in a few places, so I was trying to cut down on the repetition we have there, but I can't really think of any other way to reuse that message.

src/export/StructureDefinitionExporter.ts
```diff
@@ -167,6 +168,12 @@ export class StructureDefinitionExporter implements Fishable {
...
+                logger.error(
+                  `Invariant ${invariant.name} has a name which does not represent a valid FHIR id. FHIR ids may contain any combination of upper- or lower-case ASCII letters ('A'..'Z', and 'a'..'z'), numerals ('0'..'9'), '-' and '.', with a length limit of 64 characters.`,
+                  rule.sourceInfo
```
ngfreiter: You could potentially use the `InvalidFHIRIdError` here. I know we decided that the constraint should still be applied even if the Id is invalid, but you could throw the error after applying the constraint. If you don't want to use the error and want to keep this more specific message, that's fine, but I wanted to make sure you knew of it.

PR #311, opened 2020-03-27
Bump version to 0.11.0

ngfreiter:


cmoesel:
This is a great PR.  Almost as great as #307 since it is also a single-character change, but not quite as great because GitHub diff renders it as a two-character change.  Sorry.

image

ngfreiter:
Yeah what the heck GitHub. Get your rendering together!

cmoesel:
Oh, and I totally missed the package-lock.json file, so really, not nearly as impressive as #307.  Sorry.

PR #324, opened 2020-03-31
Update deps

cmoesel:
Got tired of seeing the warning about 1000s of low vulnerability issues.  Now we have 3.

Also updated typescript to [3.8.3](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html).  Not using the new features yet, but now we *can* if we want.

And upgraded to Prettier 2.0 which has better support for TypeScript 3.8, is supposed to be faster, and has a better algorithm for splitting functions over lines.  Prettier 2.0 changed some defaults, so I updated the config to keep things mostly as-is (use trailing commas in arrays? really?).  There were a few things I didn't configure out -- so you'll see a few minor code formatting changes.

I could have upgraded a few more packages, but then they'd drop support for Node 8.  I think that's probably OK, but maybe not something to just do on a whim.  Maybe later.


cmoesel:
Oops.  Prettier doesn't support Node 8 anymore either.  I thought this wasn't a big deal, since we only use it in dev (so it doesn't affect users), but it's affecting our CI.  Doh.

cmoesel:
I'll update the GitHub actions later to only run lint and prettier once, on ubuntu w/ Node 12.  That's probably what we should have done from the start anyway!

cmoesel:
BONUS: Now with badges:
image

cmoesel:
Not entirely sure what's up w/ those last three tests that aren't coming back, but I suspect it has something to do w/ me changing the name of the job from "Node CI" to "Lint and Test".  Personally, I'd ignore it -- and I think future PRs will be ok.  We saw this once before.

cmoesel:
Well, doh.  @ngfreiter had to go and find the one edge case for which it would have made sense to run prettier on windows in CI too.  (Sigh).  I'll fix it, but I don't think I'm going to add a prettier run on Windows to CI.  Fair?

cmoesel:
OK.  @ngfreiter's comment is addressed.  I also found the setting to make those other checks (that were stuck before) go away.  So that's good.

PR #325, opened 2020-04-01
Fix root element of slice order

ngfreiter:
Two bugs are addressed by this PR. 

The first bug is that when an element is sliced, the `addElement` function would add children of that slice below the slices of that element. The correct behavior is to add the children of the element above the slices of the element.

The second bug fixed is related to #317. It turns out that a sliced element needs to be present in the differential if its children are changed alongside changes to the slices of that element.

This PR fixes those two bugs. One way to test it is using this snippet of FSH:
```
Extension: MyRace
Parent: http://hl7.org/fhir/us/core/StructureDefinition/us-core-race
* extension.id = "my-root-id"
* extension[ombCategory].id = "my-ombcat-id"
```
When you run SUSHI with this snippet on master, you will exhibits both of these bugs described above, and as a result causes errors in the IG Publisher. When you run SUSHI with this snippet on this branch, the output will be in the correct order, and have the correct elements added, so the IG Publisher will run without error, and produce the expected output.

### DIFF
--- a/src/fhirtypes/ElementDefinition.ts
+++ b/src/fhirtypes/ElementDefinition.ts
@@ -324,10 +324,10 @@ export class ElementDefinition {
         // @ts-ignore
         return prop && !isEqual(this[prop], original[prop]);
       }) ||
-      // When a slice has children that changed, we must treat the slice as if it
+      // When a slice or a sliced element has children that changed, we must treat the slice as if it
       // differs from the original. The IG Publisher requires slices with changed
       // children to be in the differential, or the snapshot is incorrectly generated
-      (this.sliceName && this.children().some(c => c.hasDiff()))
+      ((this.sliceName || this.getSlices().length > 0) && this.children().some(c => c.hasDiff()))
     );
   }
 
--- a/src/fhirtypes/ElementDefinition.ts
+++ b/src/fhirtypes/ElementDefinition.ts
@@ -324,8 +324,8 @@ export class ElementDefinition {
         // @ts-ignore
         return prop && !isEqual(this[prop], original[prop]);
       }) ||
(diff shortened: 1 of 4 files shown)


src/fhirtypes/StructureDefinition.ts
```diff
@@ -133,8 +133,9 @@ export class StructureDefinition {
       if (element.id.startsWith(`${currentId}.`) || element.id.startsWith(`${currentId}:`)) {
         lastMatchId = currentId;
       } else if (
-        !currentId.startsWith(`${lastMatchId}.`) &&
-        !currentId.startsWith(`${lastMatchId}:`)
+        (!currentId.startsWith(`${lastMatchId}.`) && !currentId.startsWith(`${lastMatchId}:`)) ||
+        // If element is not a slice at this level, and the currentId is a slice, break to add children before slices
+        (element.id.startsWith(`${lastMatchId}.`) && currentId.startsWith(`${lastMatchId}:`))
```
mint-thompson, line 138: Put another way, it's a preorder traversal of our element tree, but child nodes always get visited before slice nodes. Excellent!

src/fhirtypes/ElementDefinition.ts
```diff
@@ -324,10 +324,10 @@ export class ElementDefinition {
         // @ts-ignore
         return prop && !isEqual(this[prop], original[prop]);
       }) ||
-      // When a slice has children that changed, we must treat the slice as if it
+      // When a slice or a sliced element has children that changed, we must treat the slice as if it
```
mint-thompson, line 327: As a minor clarification: `...we must treat the slice or sliced element as if it...`

PR #326, opened 2020-04-01
Find slices of root element on slice children

ngfreiter:
This addresses https://standardhealthrecord.atlassian.net/browse/CIMPL-305.

This fixes the issue described in this comment: https://github.com/FHIR/sushi/pull/296#pullrequestreview-381986468. The `findElementByPath` function was unable to find an element when that element was a slice of an element that is itself a child of a slice. This is the FSH that I used to test this out:
```
Profile: FunWithSlicing
Parent: resprate
* category.coding ^slicing.discriminator.path = "$this"
* category.coding ^slicing.discriminator.type = #pattern
* category.coding ^slicing.rules = #open 
* category.coding contains SomeSlice 1..5
* category[VSCat].coding[SomeSlice] ^short = "DO I EXIST"
```
Be aware that to truly test this, you will want to wait until the fixes in https://github.com/FHIR/sushi/pull/325 have been merged. Although this PR does not explicitly depend on that one, the bugs in the IG Publisher that #325 fix will make it difficult to test this PR in the IG Publisher.

### DIFF
--- a/src/fhirtypes/StructureDefinition.ts
+++ b/src/fhirtypes/StructureDefinition.ts
@@ -616,14 +616,12 @@ export class StructureDefinition {
           const newSlice = matchingConnectedSlice.clone(false);
           newSlice.id = `${e.id}:${matchingConnectedSlice.sliceName}`;
           newSlice.structDef = this;
-          e.slicing = connectedSliceElement.slicing;
+          if (!e.slicing) e.slicing = connectedSliceElement.slicing;
           this.addElement(newSlice);
           return newSlice;
         }
       }
-    }
 
-    if (!matchingSlice && pathPart.brackets?.length === 1) {
       // If we don't find a match, search predefined extensions for a match
       const sliceDefinition = fisher.fishForFHIR(pathPart.brackets[0], Type.Extension);
       if (sliceDefinition?.url) {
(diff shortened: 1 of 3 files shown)


src/fhirtypes/StructureDefinition.ts
```diff
@@ -603,6 +603,26 @@ export class StructureDefinition {
     fisher: Fishable
   ): ElementDefinition {
     let matchingSlice = elements.find(e => e.sliceName === pathPart.brackets.join('/'));
+    if (!matchingSlice && pathPart.brackets?.length === 1) {
+      // If the current element is a child of a slice, the match may exist on the original
+      // sliced element, search for that here
+      for (const e of elements) {
+        const connectedSliceElement = e.findConnectedSliceElement();
+        const matchingConnectedSlice = connectedSliceElement
```
mint-thompson, line 611: This `if` block has the same condition as the next `if` block, so I think they can be combined.

src/fhirtypes/StructureDefinition.ts
```diff
@@ -603,6 +603,26 @@ export class StructureDefinition {
...
+          ?.getSlices()
+          .find(e => e.sliceName === pathPart.brackets.join('/'));
+
+        if (matchingConnectedSlice) {
+          const newSlice = matchingConnectedSlice.clone(false);
+          newSlice.id = `${e.id}:${matchingConnectedSlice.sliceName}`;
+          newSlice.structDef = this;
+          e.slicing = connectedSliceElement.slicing;
```
mint-thompson, line 619: Is it possible for `e` to already have a defined slicing at this point?

ngfreiter:
Ok @mint-thompson , both your comments should be fixed in the latest commit.

mint-thompson:
I'm still sort of wondering what kind of situation would lead to a case where the existing slicing exists, and is different from `connectedSliceElement.slicing`, but I think it's fine for now.

Issue #327, opened 2020-04-02
Forcing input from user during _upgradePublisher.sh breaks CI


peturv:
Hiya!

Tried running my automated pipeline after having it dormant for some time and realized that confirmation is now required before download in  _updatePublisher.sh

https://github.com/FHIR/sushi/blob/7d6163f5a75ae3ba14981a892df626a3c5cdfc6d/src/ig/files/_updatePublisher.sh#L41

It would be nice to be able to pass a flag to force the answer to the question.
Submitting a PR for this (only for the .sh as I don't have a PC to try out a .bat version)

ngfreiter:
Closing since #328 and #354 addressed this.

PR #328, opened 2020-04-02
Enable passing answer to download/overwrite question in _upgradePublisher.sh

peturv:
Adds the flags -f|--force to _updatePublisher.sh to allow for a forced download. Usable in non-interactive settings.

See issue #327


cmoesel:
Hi @peturv -- thanks for the PR!

We actually base our scripts on the ones in HL7's official [sample-ig](https://github.com/FHIR/sample-ig) -- so I would suggest you also make a PR for this there.  We're trying to stay in basic alignment.

I don't suppose you know how to do this in Windows `.bat` as well, do you?  We're also trying to keep parity between the Windows and Mac scripts.  If you don't, someone from my team might be able to do it.

peturv:
> Hi @peturv -- thanks for the PR!
> 
> We actually base our scripts on the ones in HL7's official [sample-ig](https://github.com/FHIR/sample-ig) -- so I would suggest you also make a PR for this there. We're trying to stay in basic alignment.
> 
> I don't suppose you know how to do this in Windows `.bat` as well, do you? We're also trying to keep parity between the Windows and Mac scripts. If you don't, someone from my team might be able to do it.

Thanks for the response!
Unfortunately I don't have a windows-based computer at my disposal at the moment since I am in isolation and not allowed in the office :D

I'll post the PR to the sample project instead

src/ig/files/_updatePublisher.sh
```diff
@@ -31,15 +45,17 @@ else
...
+  else
+    echo Will place publisher jar here: "$jarlocation"
+    message="Ok? (Y/N) "
+  fi
+  read -r -p "$message" response
 fi
 
-read -r -p "$message" response
-if [[ "$response" =~ ^([yY])$ ]]; then
+if [[ "$FORCE" ]] || [[ "$response" =~ ^([yY])$ ]]; then
```
jafeltra, line 58: Hi @peturv! I was reviewing this pull request and noticed one thing that should be changed. I think the check for the force flag here should be checking `"$FORCE" == true`. As it is now, if a user is not using the `--force` flag and chooses `N` at the prompt, the download will still occur. Adding this line should fix this issue.

peturv: Thanks! Feel a bit silly having missed that. Have posted an update.

jafeltra: No worries! Thanks for pushing a fix!

PR #331, opened 2020-04-02
Ig pub support

jafeltra:
This adds in additional support for running SUSHI from within an IG Publisher context. Based on the conversation in Slack, the following changes were made:

- The input path is no longer required in the `sushi` command. The path will default to the current directory if it is not provided.
- If a `fsh/` subdirectory is present in the input path/current directory, we will use that subdirectory as the path to the tank. However, someone can still directly call sushi on the subdirectory if they want to.
- If a `fsh/` subdirectory is present then we assume SUSHI is being run from a IG Publisher context. When running in this context, a few things change:
    - We default the output directory to the parent of the tank, rather than `build`. If we aren't in an IG publisher context (there is no `fsh/` subdirectory), the default output directory is still `build`. You can use `-o` to specify an output directory in either case.
    - We don't export any of the publisher related scripts during the IG export.

I tried to update the `--help` information to clarify some of these new cases. That is reflected in the README.

If we decide any of the changes add too much complexity, we can choose to remove any of them. SUSHI does successfully run within the publisher as it is now, so these are intended to make things simpler. If some of the inferring we now do is really to complex/sneaky, it doesn't have to stay. I'm interested to hear reviewers' thoughts on this.

Also, I added the `chmod` for `_gencontinous` and added a check for it in the tests, because apparently I forgot to do that when I added the script.


markkramerus:
I didn't review the code, but the functionality is 100% on target.

src/app.ts
```diff
@@ -45,10 +55,30 @@ async function app() {
 
   logger.info(`Running SUSHI ${getVersion()}`);
 
-  // Check that input folder is specified
+  // If no input folder is specified, set default to current directory
   if (!input) {
-    logger.error('Missing path to FSH definition folder.');
-    program.help();
+    input = path.join('.');
```
cmoesel, line 60: I think `path.join('.')` is basically a no-op, returning back `'.'`.

src/app.ts
```diff
@@ -45,10 +55,30 @@ async function app() {
...
+    logger.info('path-to-fsh-defs defaulted to current working directory');
+  }
+
+  // If a fsh subdirectory is detected, we are in an IG Publisher context
+  const isUsingFshDir = path.parse(input).base === 'fsh';
+
+  const fshSubdirectoryPath = path.join(input, 'fsh');
+  const hasFshDir = fs.existsSync(fshSubdirectoryPath);
+  const shouldUseFshDir = hasFshDir && input !== fshSubdirectoryPath;
```
cmoesel, line 69: Given `const fshSubdirectoryPath = path.join(input, 'fsh');`, how is it ever possible for `input !== fshSubdirectoryPath` to return anything but `true`?

src/app.ts
```diff
@@ -111,10 +141,18 @@ async function app() {
...
 
-  fs.ensureDirSync(program.out);
+  let outDir = program.out;
+  if (isIgPubContext && !program.out) {
+    // When running in an IG Publisher context, default output is the parent folder of the tank
+    outDir = path.join(input, '..');
+  } else if (!program.out) {
+    // Any other time, default output is just to 'build'
+    outDir = path.join('.', 'build');
+  }
```
cmoesel, line 151: I think in the cases above where we use a default output path (instead of provided, maybe we should log that?)

src/app.ts
```diff
@@ -45,10 +55,30 @@ async function app() {
...
+  if (isIgPubContext) {
+    logger.info(
+      'Current FSH tank conforms to an IG Publisher context. Default input and output will be adjusted accordingly.'
+    );
+  }
+
+  // Use fsh/ subdirectory if not already specified and present
+  if (shouldUseFshDir) {
+    input = path.join(input, 'fsh');
+    logger.info('fsh/ subdirectory detected and add to input path');
```
cmoesel, line 81: I feel like the above 15 lines of code or so can be simplified some... I'm not _sure_, but it seems like if we check the subfolder and modify the input first, then we can check if the path base is fsh after that? Then we'd have fewer similar variables to create and track.

src/ig/IGExporter.ts
```diff
@@ -193,15 +195,27 @@ export class IGExporter {
...
 
+    // If in an IG Publisher context, do not include any of the publisher scripts
+    if (this.isIgPubContext) {
+      removeSync(path.join(igPath, '_genonce.sh'));
+      removeSync(path.join(igPath, '_genonce.bat'));
+      removeSync(path.join(igPath, '_gencontinuous.sh'));
+      removeSync(path.join(igPath, '_gencontinuous.bat'));
+      removeSync(path.join(igPath, '_updatePublisher.sh'));
+      removeSync(path.join(igPath, '_updatePublisher.bat'));
+    }
```
cmoesel, line 206: I'm not sure we want to _remove_ these.  What if the user _wants_ these scripts at the root of their repo and copies them there themselves?  We don't want to delete the ones the user added themselves.  I think we need to make the copy smarter (by applying a filter or something)...

test/ig/IGExporter.test.ts
```diff
@@ -84,6 +84,8 @@ describe('IGExporter', () => {
     it('should copy over the static files', () => {
       expect(fs.existsSync(path.join(tempOut, '_genonce.bat'))).toBeTruthy();
       expect(fs.existsSync(path.join(tempOut, '_genonce.sh'))).toBeTruthy();
+      expect(fs.existsSync(path.join(tempOut, '_gencontinuous.bat'))).toBeTruthy();
+      expect(fs.existsSync(path.join(tempOut, '_gencontinuous.sh'))).toBeTruthy();
```
cmoesel, line 88: Ah.  Good call.

src/ig/IGExporter.ts
jafeltra: Ohh that's a good point. I'll look into this.

src/app.ts
jafeltra: I think you're right. The documentation says:

> Zero-length path segments are ignored. If the joined path string is a zero-length string then '.' will be returned, representing the current working directory.

Do you think it is safer to just assign `'.'` or to do `path.join()`? Is there ever a case where current directory would have anything else as a part of it?

src/app.ts
jafeltra: That seems reasonable. I'll add it

src/app.ts
jafeltra: Somehow I was thinking if it existed but wasn't used, but I don't think that's right anymore. But I'm also going to just re-evaluate this whole section and potentially change things around.

src/app.ts
cmoesel: I think just assigning `'.'` should be fine as I'm 99.99% sure that's what `path.join('.')` will return every time.  If you're looking for it to return the absolute path of the working directory, I think you need to do that a different way.

jafeltra: It doesn't need the relative path. It also seems like `path.join()` with no arguments passed in also returns `'.'`, but I'll just assign the string.

jafeltra:
@cmoesel all your comments should be addressed in 9d20dc6

cmoesel:
Thanks, @jafeltra -- I'm holding off on merging this until we get 0.11.3 out the door or until/unless we decide to call our next release 0.12.0.

PR #337, opened 2020-04-06
Recursively set Fixed Values on Instances

### LINKED ISSUE #300: SUSHI does not recursively set fixed values on instances
See this https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/shorthand.20.26.20TestScript.20instances for more details. SUSHI will set a fixed value on an instance in one of two cases:
1) The element is named in a path on the instance
2) The parent of the element is named in a path on an instance and the element is min 1.

SUSHI should be able to follow paths of `1..1` elements recursively and look for any fixed children at the end of the path. For example in this profile:
Testscript-profile

If `.test.name` is set in the Instance, SUSHI should be smart enough to recognize that the slices of `action` are `1..1`, and `action.assert` is `1..1` and `action.assert.description` is `1..1`, so the slices can be constructed with `description` set.

ngfreiter:
This fixes #300. 

This refactors the `setFixedValues` function on the `InstanceExporter`. The `setFixedValuesFromStructureDefinition` function is removed, and its functionality is replaced directly in the `setFixedValues` function. In addition to this refactor, this PR adds the ability to set values automatically that are deeply nested on the StructureDefinition. For example, if `a`, `a.b` , and `a.b.c` are `1..1` and `a.b.c = "somefixedvalue"`, then the instance will now automatically populate:
```
{
"a": {
       "b": {
              "c": "somefixedvalue"
               }
      }
}
```
You will also note that I removed a debug message that we used to be emitting when a `0..n` element was fixed. I didn't think this debug message is necessary, since automatically fixing elements that are optional would be unexpected behavior in my opinion. If we feel this message needs to be present, I can add it back.


ngfreiter:
Ok @mint-thompson, the latest commit should add a test explicitly showing a profile fixed value that does _not_ get included on the instance!

ngfreiter:
Ah my mistake, when I did the diff I saw that the `url` line had moved but didn't notice the `ombCategory` to `text` change. But either way, It should be fixed in https://github.com/FHIR/sushi/pull/337/commits/5758bc6d5fd629db1b1c71fd77be12f319577469.

PR #340, opened 2020-04-06
Wrap Jekyll comments in <!-- -->

### LINKED ISSUE #338: Generated jekyll comments interfere w/ IG Publisher parsing
SUSHI 0.11.2 introduced generated comments on SUSHI generated and copied files.  For xml files in pagecontent, we added Jekyll comments at the top of `.md` and `.xml` files.  While this works fine for `.md` files, the IG Publisher chokes on it for `.xml` files.

First it reports this error:
```
Error on line 1 column 1
  SXXP0003: Error reported by XML parser: Content is not allowed in prolog.
Exception generating page /Users/cmoesel/dev/fsh/simple-ig/build/temp/pages/_includes/index.xml: org.xml.sax.SAXParseException; lineNumber: 1; columnNumber: 1; Content is not allowed in prolog. (00:19.0789)
```

Then later on:
```
Jekyll has failed. Complete output from running Jekyll:   Liquid Exception: Could not locate the included file 'index.xml' in any of ["/Users/cmoesel/dev/fsh/simple-ig/build/temp/pages/_includes"]. Ensure it exists in one of those directories and, if it is a symlink, does not point outside your site source (00:23.0360)
Note: Check that Jekyll is installed correctly                                   (00:23.0360)
Publishing Content Failed: Process exited with an error: 1 (Exit value: 1)       (00:23.0361)
                                                                                 (00:23.0361)
Use -? to get command line help                                                  (00:23.0361)
                                                                                 (00:23.0361)
Stack Dump (for debugging):                                                      (00:23.0361)
org.apache.commons.exec.ExecuteException: Process exited with an error: 1 (Exit value: 1)
	at org.apache.commons.exec.DefaultExecutor.executeInternal(DefaultExecutor.java:404)
```

Thanks to Bob Milius for reporting this on Zulip: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/SUSHI.200.2E11.2E2/near/192873136

cmoesel:
The IG Publisher doesn't deal well w/ bare Jekyll comments in XML -- so always wrap Jekyll comments in .  Also add the source filename in the comment so the final HTML doesn't just have an empty comment.

Fixes #338.

You can test by building an IG w/ an `ig-data/input/pagecontent/index.xml` file using the OLD SUSHI -- and IG Publisher will crash.  With the new SUSHI, it succeeds.


cmoesel:
Ugh.  Just discovered this happens w/ the MD ones now:
image

Kind of ugly.  Do you think we should _not_ wrap .md in `<!-- -->` since bare Jekyll comments don't cause a problem for MD anyway?  I had wrapped the md for consistency with xml, but I don't really like all the blank lines in the final HTML source...

ngfreiter:
I verified that this fixes the issues with `xml` files. I am going to hold off on merging because I am guessing if you pointed that out then you want to fix it. What is interesting though is I was unable to produce that extra whitespace:
image

Is this some difference in our Jekyll? Maybe a Windows vs Mac issue? If at some point @mint-thompson or @jafeltra could try and see if they get output with or without a bunch of spaces that might be helpful to see.

cmoesel:
Updated the PR to use block comments on xml and md -- which renders more nicely in the final HTML and still maintains consistency between md and xml.

PR #341, opened 2020-04-07
Resolve Aliases used in Fixed Value Rules

### LINKED ISSUE #87: Aliases are limited
If you define `Alias: systemCPN = example.com`, then you can use that alias in some places, but not others. For example, you could use the alias in a code, like `someCode#systemCPN`, but you could not use the alias when fixing a value, `* identifier[0].system = systemCPN`. It may be better to expand the usefulness of aliases by allowing them to appear in more places.

jafeltra:
Fixes #87 

This PR adds support for resolving aliases that are used in Fixed Value rules. I also updated the check for if a rule is referencing a resource to not evaluate to true if we are resolving an alias. The two tests show two examples to test to ensure this works. These tests will fail on master, but pass on this branch.

I also looked through the other rules where aliases could potentially be used, and it looks like all other rules resolve aliases where possible.


jafeltra:
Hm that's a good question. So are you asking if we should support something like this?
``` 
Alias: EXAMPLE = http://example.org
....
* identifier[0].system = "EXAMPLE"
```
It feels to me like this would change how we reference aliases in all the other places, and I think I might prefer to keep it simpler and have only one way to use an alias (it needs to be a sequence, rather than a sequence or a string). That said, I'm not entirely sure and could definitely be persuaded to change my mind.

mint-thompson:
What I'm thinking about is a situation like this:
```
Alias: BREAKFAST = "A delicious breakfast."
...
* description = BREAKFAST
```
Currently, this is not supported.

cmoesel:
I think the spec currently defines alias values as being URL/OID only -- so supporting other types would be outside the spec.  See: https://build.fhir.org/ig/HL7/fhir-shorthand/reference.html#defining-aliases

jafeltra:
Ohhhh I see what you mean. That would kind of be cool. It's kind of like making an alias just a variable. I think that's outside of what I thought the scope of this task was, but I'd still be happy to include it here if we want! Also, I think since that changes syntax of FSH, I'd maybe want @cmoesel's opinion as well.

jafeltra:
If it is outside the spec, then maybe we should hold off on this for now at least, and we can evaluate if it is something we want to add later on.

jafeltra:
> Looks good! I also regression tested w/ mCODE and verified that nothing breaks.

Oh thank you! I didn't even think to do that this time around.

PR #342, opened 2020-04-07
Non-breaking space character counts as whitespace

### LINKED ISSUE #315: Parser doesn't recognize non-breaking spaces
The parser's definition for whitespace does not include non-breaking spaces. As a result, if FSH contains a non-breaking space (possibly from cut-and-paste from a web page), the parser will fail parsing.  We should add non-breaking space to the lexer rule for `WS`.

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/use.20identifier

mint-thompson:
Fixes #315 and completes task https://standardhealthrecord.atlassian.net/browse/CIMPL-327

The character with decimal value 160 represents a non-breaking space, and is now grouped in with other whitespace characters in the grammar. This helps in situations where FSH is copied from a web page that adds some non-breaking space characters as part of the formatting.


antlr/src/main/antlr/FSH.g4
```diff
@@ -156,7 +156,7 @@ UNIT:               '\'' (~[\\'])* '\'';
 CODE:               SEQUENCE? '#' (SEQUENCE | CONCEPT_STRING);
 
 
-CONCEPT_STRING:      '"' (~[ \t\r\n\f\\"] | '\\"' | '\\\\')+ (WS (~[ \t\r\n\f\\"] | '\\"' | '\\\\')+)* '"';
+CONCEPT_STRING:      '"' (~[ \t\r\n\f\u00A0\\"] | '\\"' | '\\\\')+ (WS (~[ \t\r\n\f\u00A0\\"] | '\\"' | '\\\\')+)* '"';
```
cmoesel, line 159: Since we have the `WS` fragment, can this be simplified to:
```
CONCEPT_STRING:      '"' (~(WS|[\\"]) | '\\"' | '\\\\')+ (WS (~(WS|[\\"]) | '\\"' | '\\\\')+)* '"';
```

cmoesel: I'm actually not sure -- so that's a real question: _can_ it be simplified to that?

antlr/src/main/antlr/FSH.g4
```diff
@@ -171,7 +171,7 @@ CARD:               ([0-9]+)? '..' ([0-9]+ | '*')?;
 REFERENCE:          'Reference' WS* '(' WS* SEQUENCE WS* ('|' WS* SEQUENCE WS*)* ')';
 
                  //  ^  NON-WHITESPACE
-CARET_SEQUENCE:     '^' ~[ \t\r\n\f]+;
+CARET_SEQUENCE:     '^' ~[ \t\r\n\f\u00A0]+;
```
cmoesel, line 174: Can this be simplified to:
```
CARET_SEQUENCE:     '^' ~WS+;
```

antlr/src/main/antlr/FSH.g4
```diff
@@ -183,12 +183,12 @@ COMMA_DELIMITED_CODES: (CODE (WS+ STRING)? WS* COMMA WS+)+ CODE (WS+ STRING)?;
 COMMA_DELIMITED_SEQUENCES: (SEQUENCE WS* COMMA WS*)+ SEQUENCE;
 
                  // NON-WHITESPACE
-SEQUENCE:           ~[ \t\r\n\f]+;
+SEQUENCE:           ~[ \t\r\n\f\u00A0]+;
```
cmoesel, line 186: Can this be simplified to:
```
SEQUENCE:           ~WS+;
```

cmoesel: Since this is the second time we'd use `~WS+`, should we create another fragment for `NONWS`?

antlr/src/main/antlr/FSH.g4
mint-thompson: Unfortunately, no: it seems that antlr won't allow the negation of the `WS` fragment like that, even though that fragment represents a single character from a set.

antlr/src/main/antlr/FSH.g4
mint-thompson: This, we can do. In places where we need exactly `~WS`, the new `NONWS` fragment will be used. But it has to be exact: we can't use these nice fragments if we need some extra characters to be matched.

PR #343, opened 2020-04-07
FHIR Spot Check

jafeltra:
This PR adds support for two early checks in the SUSHI process, addressing https://standardhealthrecord.atlassian.net/browse/CIMPL-221.

1. We check if the user specified FHIR R4 in their dependencies in `package.json`. If they did not, we log an error and exit.
2. We check if the StructureDefinition resource exists and has a basic `version` property on it. If something has gone in the FHIR definition download, this property likely won't exist. If there is something wrong with the resource, we log an error and exit.

Since these changes just live in `app.ts`, there are no tests. To test this PR locally, I did the following:

1. To check the FHIR R4 version, I changed the version specified in the `hl7.fhir.r4.core` dependency in package.json of the FSH tank, and I also removed that dependency entirely from the package.json.
2. To test that the StructureDefinition check worked, I moved the `StructureDefinition-StructureDefinition.json` file in my FHIR cache in r4 core package out of its usual directory. If you don't want to mess around with your FHIR cache, you can feel free to just look at the code or I can show you it working (and breaking) on my computer.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -100,13 +100,14 @@ async function app() {
     return;
   }
 
-  const fhirR4Dependency = config.dependencies['hl7.fhir.r4.core'];
-  if (!(fhirR4Dependency && fhirR4Dependency === '4.0.1')) {
+  // Ensure FHIR R4 is added as a dependency
+  const fhirR4Dependency = config.dependencies?.['hl7.fhir.r4.core'];
+  if (fhirR4Dependency !== '4.0.1') {
     logger.error(
       'The package.json must specify FHIR R4 as a dependency. Be sure to' +
         ' add "hl7.fhir.r4.core": "4.0.1" to the dependencies list.'
     );
-    program.help();
+    return;
   }
 
   // Load external dependencies
@@ -145,11 +146,13 @@ async function app() {
 
   // Check for StructureDefinition
   const structDef = defs.fishForFHIR('StructureDefinition', Type.Resource);
-  if (!(structDef && structDef.version === '4.0.1')) {
+  if (structDef?.version !== '4.0.1') {
     logger.error(
-      'StructureDefinition resource not found for 4.0.1. Your FHIR package may be corrupt.'
+      'StructureDefinition resource not found for v4.0.1. The FHIR R4 package in local cache' +
+        ' may be corrupt. Local FHIR cache can be found at <home-directory>/.fhir/packages.' +
+        ' For more information, see https://wiki.hl7.org/FHIR_Package_Cache#Location.'
     );
-    program.help();
+    return;
   }
 
   logger.info('Converting FSH to FHIR resources...');
--- a/src/app.ts
+++ b/src/app.ts
@@ -7,7 +7,7 @@ import cloneDeep from 'lodash/cloneDeep';
 import { importText, FSHTank, RawFSH } from './import';
 import { exportFHIR, Package } from './export';
 import { IGExporter } from './ig';
-import { logger, stats } from './utils';
+import { logger, stats, Type } from './utils';
 import { loadDependency, loadCustomResources } from './fhirdefs';
 import { FHIRDefinitions } from './fhirdefs';
 import {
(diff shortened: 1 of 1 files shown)


src/app.ts
```diff
@@ -100,6 +100,15 @@ async function app() {
     return;
   }
 
+  const fhirR4Dependency = config.dependencies['hl7.fhir.r4.core'];
```
ngfreiter, line 103: I think you may want to make this `config.dependencies?.['hl7.fhir.r4.core']`. That way if someone doesn't specify a dependencies list at all, we get your nice graceful error instead of an unhandled promise rejection.

src/app.ts
```diff
@@ -100,6 +100,15 @@ async function app() {
...
+  if (!(fhirR4Dependency && fhirR4Dependency === '4.0.1')) {
```
ngfreiter, line 104: Does this need the check that `fhirR4Depency` exists? Doesn't the equality check that for us? Not a huge deal either way.

Also do we necessarily not allow using `4.0.0`?

src/app.ts
```diff
@@ -100,6 +100,15 @@ async function app() {
...
+    logger.error(
+      'The package.json must specify FHIR R4 as a dependency. Be sure to' +
+        ' add "hl7.fhir.r4.core": "4.0.1" to the dependencies list.'
+    );
+    program.help();
```
ngfreiter, line 109: I could so an argument for using `program.help()` here and below, but we don't use it in other places, like if there is no package.json or an invalid package.json. So I'm not necessarily sure it makes sense in this case. Feel free to convince me otherwise though.

src/app.ts
```diff
@@ -133,6 +142,16 @@ async function app() {
 
   const tank = new FSHTank(docs, config);
   await Promise.all(dependencyDefs);
+
+  // Check for StructureDefinition
+  const structDef = defs.fishForFHIR('StructureDefinition', Type.Resource);
+  if (!(structDef && structDef.version === '4.0.1')) {
+    logger.error(
+      'StructureDefinition resource not found for 4.0.1. Your FHIR package may be corrupt.'
```
ngfreiter, line 150: It may be helpful to point the user to the `.fhir` cache here.

src/app.ts
```diff
@@ -133,6 +142,16 @@ async function app() {
 
   const tank = new FSHTank(docs, config);
   await Promise.all(dependencyDefs);
+
+  // Check for StructureDefinition
```
ngfreiter, line 146: It looked like the task noted that we should check for ElementDefinition as well. Did you want that included here too?

src/app.ts
ngfreiter: I take back the second question I think. From what I can tell, you can't even get `4.0.0` off of the fhir package server. If you go to http://packages.fhir.org/hl7.fhir.r4.core only the `4.0.1` version is listed. But if you go to http://packages.fhir.org/hl7.fhir.us.core, you can see a bunch of different versions. So if they aren't posting `4.0.0` as a package, then people probably shouldn't be using it I'm thinking?

src/app.ts
jafeltra: Ah good idea! I'll add that in.

src/app.ts
jafeltra: Exactly. At first I was checking that they specified a `4.0.x` version, but then noticed the same thing you noticed on packages.fhir.org. So I mentioned it to Chris at my 1 on 1 yesterday, and he said we should just require that they use `4.0.1`.

src/app.ts
jafeltra: Actually, you're probably right that it should just `return` similar to those other two places. There's nothing wrong with how they're invoking the command, so the `help` information doesn't give them any additional useful information.

src/app.ts
jafeltra: Good idea!

src/app.ts
jafeltra: I talked about this with Chris yesterday too, and he said that he thought that checking for StructureDefinition was sufficient. I think his thinking was that if the StructureDefinition is fine, the ElementDefinition is also probably fine. If you can think of a reason that one might end up there and the other might not, I can add it back in. It isn't difficult.

ngfreiter: Nah, I'm good with leaving it out. Just wanted to double check that it was a conscious decision.

jafeltra:
@ngfreiter you comments should all be addressed in e1b61a4

PR #348, opened 2020-04-09
Set discriminator value cardinality

### LINKED ISSUE #301: SUSHI should set fixed values that are used in discriminators to 1..1
See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Creating.20Examples.20with.20slicing/near/191818082

ngfreiter:
This fixes #301 

Now when a part of a slice gets a value fixed on it, and that value is used by the `path` in the relevant slice discriminator which discriminates based on `value` or `pattern`, if the slice has a minimum cardinality of 0, we change that minimum cardinality to be 1.


cmoesel:
This looks like it works well for the nearest discriminator -- but for completeness, we really ought to be looking at any discriminators in the parent paths -- as it's possible for a sliced element to have children that are also sliced -- and a discriminator could be at either one of the sliced elements.

PR #349, opened 2020-04-09
Update app.ts

markkramerus:


src/app.ts
```diff
@@ -295,14 +295,28 @@ const MESSAGE_MAP: { [key in ResultStatus]: string[] } = {
...
-    'Warnings... Water those about?'
+    'Warnings... Water those about?',
+    'Looks like you are casting about.',
+    'A bit pitchy, but tuna-ble'
   ],
-  errors: ['Ick! Errors!', 'Some-fin went wrong...', 'Unfor-tuna-tely, there are errors.']
+  errors: ['Ick! Errors!', 'Some-fin went wrong...', 'Unfor-tuna-tely, there are errors.',
+'That really smelt.',
+'You spawned some errors.',
+'Just keep swimming, Dory.',
```
cmoesel, line 316: I don't get this one... I guess if I watched one of those Nemo movies I would?

markkramerus: Yep, it's Disney. The exact quote it longer, but we only have so much real estate.

PR #358, opened 2020-04-13
Fix bug with matching parts of path names

### LINKED ISSUE #357: Unable to find certain paths on instances
See this zulip thread for details: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/SUSHI.200.2E12.2E0.20(Ballot.20Edition). 

It looks like there is an issue with finding certain types of paths on instances. My suspicion is that this relates to the fact that the paths being confused in this scenario have matching endings.

ngfreiter:
See https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/SUSHI.200.2E12.2E0.20(Ballot.20Edition) for more details. This fixes #357. 

It looks like this issue was that when finding out how much two paths overlap at the beginning, we were accidentally counting all overlap, at any level of the path, not just overlap at the beginning of the path. This fixes that. 

To test, you can check out https://github.com/FHIR/sushi/commit/fe8170cff6859836e7124f0e81a7c136ecb4e6bf and see that the tests fail, and then check out https://github.com/FHIR/sushi/commit/d0efb741bd44aaa672c42d235058e03a07dd6683 and see that the failing test now passes. Another good way to test is to copy the example in the linked zulip thread, and try running it on master, for which you'll see the incorrect error, and on this branch, for which you won't see an error, and the output will be correct.


src/export/InstanceExporter.ts
```diff
@@ -92,9 +92,13 @@ export class InstanceExporter implements Fishable {
...
-            element.id.split('.').forEach((p, i) => {
-              if (p === associatedEl.id.split('.')[i]) overlapIndex++;
-            });
+            for (const [i, p] of element.id.split('.').entries()) {
+              if (p === associatedEl.id.split('.')[i]) {
+                overlapIndex++;
+              } else {
+                break;
+              }
+            }
```
cmoesel, line 101: Probably not a big deal, but you're splitting `associateEl.id` on every iteration.  It would be more efficient to pull it out.  Also, noticing that I think `i` and `overlapIndex` are pretty closely related -- and the `break` can happen automatically in a traditional for loop.  Is this too fancy?

```ts
const elParts = element.id.split('.');
const assocElParts = associatedEl.id.split('.');
let overlapIdx = 0;
for (; overlapIdx < elParts.length && elParts[overlapIdx] === assocElParts[overlapIdx]; overlapIdx++);
```

ngfreiter: Not too fancy, in fact I'd say just fancy enough. So it's added in https://github.com/FHIR/sushi/pull/358/commits/a5b68fd8aba2a34bc822abe127daa6f624a371bf.

src/export/InstanceExporter.ts
```diff
@@ -91,17 +91,21 @@ export class InstanceExporter implements Fishable {
...
+            );
             // We must keep the relevant portion of the beginning of path to preserve sliceNames
             // and combine this with portion of the associatedEl's path that is not overlapped
-            const pathStart = splitOnPathPeriods(path).slice(0, overlapIndex - 1);
+            const pathStart = splitOnPathPeriods(path).slice(0, overlapIdx - 1);
             const pathEnd = associatedEl
               .diffId()
               .split('.')
-              .slice(overlapIndex)
+              .slice(overlapIdx)
```
mint-thompson, line 108: I know you just updated this part to be cleaner, but: you could also consider using [takeWhile](https://lodash.com/docs/4.17.15#takeWhile) from lodash.

ngfreiter: Ah cool, that looks like the function that I felt like should've existed but I could not find, thanks! In this case I don't know how I would go about using it, because I find the overlap from the **FHIR** paths on the ElementDefinitions of the elements, and then use that overlap to slice up the corresponding **FSH** versions of the paths. So I'm not sure how I'd simplify things with `takeWhile` here.

mint-thompson: Ah, that's a good point. It might be more troublesome than useful. I think it would look something like this:
```
const pathStart = _.takeWhile(splitOnPathPeriods(path), (val, ind) => _.nth(elParts, ind) == _.nth(assocElParts, ind));
const pathEnd = associatedEl.diffId().split('.').slice(pathStart.length()).map(...);
```

Also, looking at the indices, shouldn't the slice for `pathEnd` start at `overlapIdx - 1`? I know different arrays are being sliced, but this doesn't use the path part at index `overlapIdx - 1` from either path array.

ngfreiter: The `pathStart` is taken from `path`, which is a pure FSH path, so it doesn't include the extra resourceType at the beginning of the path (so just `a.b.c`, not `Patient.a.b.c`). The `pathEnd` is taken from a FHIR id that id getting converted into a FSH path, so it does have the resourceType appended to the beginning of the path (`Patient.a.b.c`, not `a.b.c`). So let's say the original path was `a.b.c`, and we get element `Patient.a.b.c` from this. Then the associated element path could be `Patient.a.b.d`, so the overlap is 3. From the original path (`a.b.c`), we want `a.b`, and from the associated element path, we want `d`. So to get `a.b` from `a.b.c`, we need to subtract 1 from the overlap index, since there is no `Patient` at the beginning of the `path`.

cmoesel:
It looks like the _node 8 on macos-latest (pull_request)_ is jammed.  It hasn't even initialized the VM yet and all the other checks are down.  Merging this now as all the other checks pass and there is nothing different that should cause any issues on this stalled check.

PR #359, opened 2020-04-13
Inline instances on non resources

### LINKED ISSUE #347: Fix inline Instances to type constrained elements
See this thread for more details: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Issue.20creating.20Bundle.20instance.

Currently, an inline instance can be fixed to an element of type Resource. However, users should also be able to fix an instance of a Patient to an element of type Patient, or an instance of Observation to an element of type Observation. Additionally, users shouldn't be able to fix an instance of Observation to an element of type Patient.

ngfreiter:
This fixes #347. 

When setting some inline instance A on another instance B, we were only allowing A to be set on B if the element on B was of type Resource. However, instance B could be an instance of a profile that constrains the type of an element on B from Resource to something more specific, like Patient. If this is the case, then we should still be able to set A as an inline instance to the element on B if A is of type Patient, but we did not allow that. This PR fixes this by allowing inline instances to be set on elements that are sub-types of Resource. The rules for adding are:
* If the element is of type Resource, any instance can be set there, regardless of resourceType
* If the element is of type DomainResource, any instance that is not of resourceType Bundle, Parameters, or Binary can be set. These are the only 3 resources that do not inherit from DomainResource, https://www.hl7.org/fhir/domainresource.html#bnr.
* If the element is of type X, any instance of type X can be set on it.

To test this, you can look at the test cases added in https://github.com/FHIR/sushi/commit/6171c5d94d48a1f474d59a6f3ac99bcd788796a1 and see that they fail, then test them on https://github.com/FHIR/sushi/commit/3e90b7e22f7485d492b70a8911111b1fa91424cd and see that they pass. You can also construct an example set of FSH definitions out of the test case added to InstanceExporter.test.ts, and see that when running SUSHI on that example on master, you will get errors and setting the inline instance won't work, but when running SUSHI on this branch, the errors will go away and setting the inline instance will work correctly.


test/fhirtypes/StructureDefinition.test.ts
```diff
@@ -1083,11 +1083,69 @@ describe('StructureDefinition', () => {
       }).toThrow('Cannot resolve element from path: extension[fake-extension].value[x]');
     });
 
-    it('should allow fixing an InstanceDefinition to a resource element', () => {
-      const instanceDef = new InstanceDefinition();
-      instanceDef.resourceType = 'Patient';
-      const { fixedValue } = respRate.validateValueAtPath('contained[0]', instanceDef, fisher);
-      expect(fixedValue.resourceType).toBe('Patient');
+    describe('#Inline Instances', () => {
```
mint-thompson, line 1086: Should there be an additional test for fixing a `Bundle`-type instance definition to a `Resource`? The tests right now show what `Bundle` can't be bound to, but not what it can be bound to.

ngfreiter: Yeah! I will add that.

ngfreiter: Added in https://github.com/FHIR/sushi/pull/359/commits/e35a0219dcfb9496d06c6a86ad73c392d460e863!

test/fhirtypes/StructureDefinition.test.ts
```diff
@@ -1083,11 +1083,77 @@ describe('StructureDefinition', () => {
...
+        const instanceDef = new InstanceDefinition();
+        instanceDef.id = 'OfJoy';
+        instanceDef.resourceType = 'Bundle';
+        const { fixedValue } = respRate.validateValueAtPath('contained[0]', instanceDef, fisher);
+        expect(fixedValue.resourceType).toBe('Bundle');
+      });
+
+      it('should not allow fixing a Bundle type InstanceDefinition to a DomainResource element', () => {
+        const instanceDef = new InstanceDefinition();
+        instanceDef.id = 'OfJoy';
```
cmoesel, line 1142: Awww... it's a little bundle of joy!

ngfreiter: 👶

PR #360, opened 2020-04-14
Fix length of puns

ngfreiter:
One of the puns added to app.ts was too long, so the box would get offset:
image
This fixes that.


mint-thompson:
Good catch! Thanks for scaling down this message. Assuming a maximum of three digits each for the numbers of warnings and errors, it looks like the allowable message length is 32 characters. Anything longer than that will have to be reeled in.

PR #361, opened 2020-04-15
Config YAML type

cmoesel:
This PR renames the old `Config` type to `PackageJSON` and adds a new `Configuration` type.  This will hopefully reduce confusion and make it easier to know which config we're talking about (while they both exist).

The new `Configuration` type is not hooked into anywhere -- it is just an unused type at this point.  I wanted to get this into the code now so that the rest of the team can begin implementing exporters for it.  This type is not a 1:1 mapping to the YAML format, as I felt that would make the file harder for us to use with the exporters; so instead it's designed to be a little friendlier with the exporters.

I'm working on a `YAMLConfiguration` type that resembles the exact JSON you would receive from parsing a YAML file directly and also working on an importer that will convert the `YAMLConfiguration` format to the friendlier `Configuration` format.  As it turns out, it's way more tedious than expected and TypeScript is maybe not my friend right now.  All that will come in a separate PR. But the rest of the team shouldn't need that to get started on exporters and unit tests!

Also note that this PR is set to merge into a `config-yaml` branch.  Since the new config is kind of a big feature, I don't want to release parts of it before all of it is ready -- so I propose that all config-related features get merged into `config-yaml`, and once everything is done, we'll merge `config-yaml` into `master`.  Of course, if you think this is a bad idea, let me know!

### DIFF
--- a/src/fhirtypes/dataTypes.ts
+++ b/src/fhirtypes/dataTypes.ts
@@ -36,9 +36,9 @@ export type Coding = {
  * @see {@link http://hl7.org/fhir/R4/datatypes.html#ContactPoint}
  */
 export type ContactPoint = {
-  system?: string;
+  system?: 'phone' | 'fax' | 'email' | 'pager' | 'url' | 'sms' | 'other';
   value?: string;
-  use?: string;
+  use?: 'home' | 'work' | 'temp' | 'old' | 'mobile';
   rank?: number;
   period?: Period;
 };
--- a/src/fshtypes/Configuration.ts
+++ b/src/fshtypes/Configuration.ts
@@ -1,0 +1,168 @@
+import {
+  ContactDetail,
+  UsageContext,
+  ImplementationGuideDefinitionTemplate,
+  Meta,
+  Narrative,
+  Extension,
+  ImplementationGuideDependsOn,
+  ImplementationGuideGlobal,
+  ImplementationGuideDefinitionResource,
+  ImplementationGuideDefinitionPage,
+  ImplementationGuideDefinitionParameter,
+  ImplementationGuideStatus,
+  CodeableConcept
+} from '../fhirtypes';
+
+/**
+ * Configuration is related to the proposed YAML configuration format for FSH, but eliminates
+ * ambiguities (like singular vs list representations) and normalizes closer to the intended
+ * export formats.
+ *
+ * @see {@link http://hl7.org/fhir/R4/implementationguide.html}
+ * @see {@link https://build.fhir.org/ig/FHIR/ig-guidance/using-templates.html#igroot}
+ * @see {@link https://confluence.hl7.org/display/FHIR/NPM+Package+Specification}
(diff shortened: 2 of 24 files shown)


src/fshtypes/Configuration.ts
```diff
@@ -0,0 +1,168 @@
...
+  // The pages property corresponds to IG.definition.page. SUSHI can auto-generate the page list,
+  // but if the author includes pages in this file, it is assumed that the author will fully manage
+  // the pages section and SUSHI will not generate any page entries. If title is not provided, then
+  // the title will be generated from the file name.  If a generation value is not provided, it
+  // will be inferred from the file name extension.
+  pages?: ImplementationGuideDefinitionPage[];
+
+  // The parameters property represents IG.definition.parameter. For a partial list of allowed
+  // parameters see: https://confluence.hl7.org/display/FHIR/Implementation+Guide+Parameters
+  parameters?: ImplementationGuideDefinitionParameter[];
```
ngfreiter, line 95: So this is where `copyrightYear` and `releaseLabel` go right?

src/fhirtypes/dataTypes.ts
```diff
@@ -36,9 +36,9 @@ export type Coding = {
  * @see {@link http://hl7.org/fhir/R4/datatypes.html#ContactPoint}
  */
 export type ContactPoint = {
-  system?: string;
+  system?: 'phone' | 'fax' | 'email' | 'pager' | 'url' | 'sms' | 'other';
   value?: string;
-  use?: string;
+  use?: 'home' | 'work' | 'temp' | 'old' | 'mobile';
```
ngfreiter, line 41: Why the change here?

src/fshtypes/Configuration.ts
cmoesel: Yes.  Sorry, that's a bit less clear here without the YAMLConfiguration and importer stuff I held back from this PR.  But you've got it-- if the YAML has `copyrightYear`, `copyrightyear`, `releaseLabel`, or `releaselabel`, they'll get added as `copyrightyear` and/or `releaselabel` parameters (as appropriate).

src/fhirtypes/dataTypes.ts
cmoesel: More "typey"?  In the FHIR definitions, these are _required_ bindings with those values.  Since it's a fixed set of values, I thought I'd more closely reflect that in the TypeScript code.  Is it absolutely necessary?  No -- but it does help us be more "typey" and potentially avoid typos if we ever use that data in code.

If you were to reword this question as "Why the change here, but not there?" (pointing to a similar bit of code where I did not do this), then the answer is just that I did it as I encountered it.  So I ran into this code (actually while working on the import part) and updated it.

cmoesel:
I'm going to go ahead and merge this since there are two approvals.  As a reminder, it's not merging to `master`; it's merging to `config-yaml`.

Issue #362, opened 2020-04-15
Unable to fix values on slices without discriminators


ngfreiter:
In the following situation:
```
Profile:        OurNewProfile
Parent:         Patient
Id:             our-new-profile

* telecom 2..
* telecom ^slicing.rules = #closed
* telecom contains phone 1..1 and email 1..1
* telecom[phone].system = #phone
```
SUSHI emits an error 
```
Cannot read property 'some' of undefined
  File: /[...]/Patient.fsh
  Line: 9
```
This looks to be caused by the lack of discriminator on `telecom.slicing`. If a discriminator is added with the following two lines
```
* telecom ^slicing.discriminator.type = #value
* telecom ^slicing.discriminator.path = "system"
```
then the error message goes away, and the output looks correct. However, a discriminator is not a required element on slicing, so we should not have any issues when the discriminator is not present, so this is a bug that needs to be fixed.

ngfreiter:
See this thread for any more details; https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Problem.20assigning.20values.20to.20specific.20slice

PR #363, opened 2020-04-15
Allow fixing values on slices with no discriminators

ngfreiter:
This fixes https://github.com/FHIR/sushi/issues/362. 

This was a new bug introduced in 0.12.0 that causes SUSHI to log an error when trying to fix a value on a slice for which the corresponding slicing does not have a discriminator. The issue contains a fully detailed example of this situation.

To test this, you can use the example shown in the issue to compare running SUSHI on that example on this branch, versus running SUSHI on that example on master. On master you will see the incorrect error message, and on this branch you should not. I also added a test of this situation.


cmoesel:
It's irrelevant to the bug, but note that technically the example in the GH issue is invalid anyway -- because if there is no `slicing.discriminator` then there _must_ be a `slicing.description`:
image

cmoesel:
But I agree, the PR looks good!

PR #367, opened 2020-04-16
Remove inline instances from IG.json

### LINKED ISSUE #366: Inline Instances added to IG.json
It seems that even when Instances are specified as `#inline`, they are still added to the ImplementationGuide.json file. For example, in the following case:
```
Instance: MyPatient
InstanceOf: Patient
Usage: #inline 
* name[0].family = "InstanceMan"

Instance: MyBundle
InstanceOf: Bundle
* type = #batch
```
no MyPatient resource is generated, because it is labeled as `#inline`, but a reference to MyPatient is still incorrectly added to ImplementationGuide.json, causing the IG Publisher to fail.

ngfreiter:
Fixes #366 

When Instances were specified as `#inline`, they were correctly not being written to JSON files, but references to these resources were incorrectly still being added to the IG.json. Then the IG Publisher would crash, because it expects the resources defined in IG.json to exist, but no JSON file for these resources exists. This fixes that by not adding `#inline` resources to the IG.json. One way to test is to checkout https://github.com/FHIR/sushi/commit/2b5bddfb5bf4ecfa5057cb8501b39fb64afae448 and run the tests in IGExporter.test.ts that now test with an `#inline` resource. The tests will fail, because a reference to the Instance does get added to the IG.json. Then checkout this branch, and you will see that the test passes because the instance is not added.

Another good test is to use the example referenced in the issue, and ensure that you can build that example on this branch, and successfully build using the IG Publisher. On master, you should be able to successfully build the same example using SUSHI, but the IG Publisher step should fail, because the reference to the `#inline` resource will still be added to IG.json.

### DIFF
--- a/src/ig/IGExporter.ts
+++ b/src/ig/IGExporter.ts
@@ -455,26 +455,28 @@ export class IGExporter {
       this.pkg.instances,
       instance => instance.id ?? instance._instanceMeta.name
     );
-    instances.forEach(instance => {
-      const resource: ImplementationGuideDefinitionResource = {
-        reference: {
-          reference: `${instance.resourceType}/${instance.id ?? instance._instanceMeta.name}`
-        },
-        name: instance._instanceMeta.title ?? instance._instanceMeta.name,
-        description: instance._instanceMeta.description
-      };
-      if (instance._instanceMeta.usage === 'Example') {
-        const exampleUrl = instance.meta?.profile?.find(url => this.pkg.fish(url, Type.Profile));
-        if (exampleUrl) {
-          resource.exampleCanonical = exampleUrl;
+    instances
+      .filter(instance => instance._instanceMeta.usage !== 'Inline')
+      .forEach(instance => {
+        const resource: ImplementationGuideDefinitionResource = {
+          reference: {
+            reference: `${instance.resourceType}/${instance.id ?? instance._instanceMeta.name}`
+          },
+          name: instance._instanceMeta.title ?? instance._instanceMeta.name,
+          description: instance._instanceMeta.description
+        };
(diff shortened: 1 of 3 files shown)


cmoesel:
How did we miss this before?  Did the IG Publisher previously allow this?  It seems to me like me _must_ have tested this before (and would have noticed a crash)?

ngfreiter:
I don't know how to get an older version of the IG Publisher. If you know I can test it out, but I would be surprised if it works with an older version since either way we reference a resource that doesn't exist. I agree though, I can't believe I didn't see this when first adding inline instances.

cmoesel:
I don't think we need to investigate any further.  I was just surprised we hadn't hit it already and trying to figure out why.  But at this point, it doesn't really matter why.

cmoesel:
I looked at the code diff, but haven't reviewed it in detail.  That said, @jafeltra's review is good enough for me, so let's get this merged!

Which reminds me, by the way, when you see the acronym "LGTM" in pull request reviews, do you interpret it as "Looks Good To Me" _or_ "Let's Get This Merged"?  Or, as the urban dictionary helpfully points out, it could also mean "Literally Garbage To Me."

jafeltra:
> Which reminds me, by the way, when you see the acronym "LGTM" in pull request reviews, do you interpret it as "Looks Good To Me" _or_ "Let's Get This Merged"? Or, as the urban dictionary helpfully points out, it could also mean "Literally Garbage To Me."

This is interesting. I always read it as "Looks Good To Me" but I rarely use the acronym myself. However, now that I know that third option is a potential resolution of the acronym, maybe I'll spice things up and start using it, just to cause minor chaos.

PR #368, opened 2020-04-17
Override inline instances

ngfreiter:
This addresses https://standardhealthrecord.atlassian.net/browse/CIMPL-342.

This PR enables the user to do something like this on an instance:
```
Instance: MyBundle
InstanceOf: Bundle
* entry[0].resource.name.given = "Bob"
```
An instance `MyPatient` is placed on another instance, `MyBundle` in this case, and then values within `MyPatient` are modified within the context of `MyBundle`. The original values of the `MyPatient` instance should remain unchanged. 

This also made it clear that users need to be able to set `resourceType` on an element of type resource. For example, the following alternative to what is shown above could be done to get the same results:
```
Instance: MyBundle
InstanceOf: Bundle
* entry[0].resource.resourceType = "Patient"
* entry[0].resource.id = "SomePatientId"
* entry[0].resource.name.given = "Bob"
```
The `resourceType` property is a special field that has to be added because of the serialization to JSON, https://www.hl7.org/fhir/json.html#xml, so this needed special handling. 

The user should now be able to inline assign valid elements a `resourceType` or a predefined inline instance, and then modify the fields on that instance in the same way as if that instance was being defined directly and not inline.

### DIFF
--- a/src/errors/InvalidResourceTypeError.ts
+++ b/src/errors/InvalidResourceTypeError.ts
@@ -1,0 +1,8 @@
+import { Annotated } from './Annotated';
+
+export class InvalidResourceTypeError extends Error implements Annotated {
+  specReferences = ['https://www.hl7.org/fhir/resourcelist.html'];
+  constructor(public resourceType: string, public elementType: string) {
+    super(`A resourceType of ${resourceType} cannot be set on an element of type ${elementType}.`);
+  }
+}
--- a/src/errors/index.ts
+++ b/src/errors/index.ts
@@ -39,3 +39,4 @@ export * from './MultipleStandardsStatusError';
 export * from './InvalidMappingError';
 export * from './InvalidFHIRIdError';
 export * from './ParentDeclaredAsProfileNameError';
+export * from './InvalidResourceTypeError';
--- a/src/export/InstanceExporter.ts
+++ b/src/export/InstanceExporter.ts
@@ -50,7 +50,7 @@ export class InstanceExporter implements Fishable {
       if (r.isResource && r.fixedValue instanceof InstanceDefinition) {
         inlineResourcePaths.push({
           path: r.path,
-          instanceOf: r.fixedValue.resourceType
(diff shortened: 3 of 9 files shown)


cmoesel:
I assume the `,` is a typo below and should be `.` -- right?
```
* entry[0].resource,resourceType = "Patient"
```

ngfreiter:
> I assume the `,` is a typo below and should be `.` -- right?
> 
> ```
> * entry[0].resource,resourceType = "Patient"
> ```

Typo indeed, I fixed it in the description now.

cmoesel:
This approach doesn't seem to mix well w/ the optional `[0]` indexer.  It's probably rare, but something like this does not work:
```
Instance: MyBundle
InstanceOf: Bundle
* entry.resource = MyPatient
* entry[0].resource.name.given = "Bob"
```
nor the reverse:
```
Instance: MyBundle
InstanceOf: Bundle
* entry.resource.name.given = "Bob"
```

Is that something that could be easily fixed in this PR or is it better to address it separately?

ngfreiter:
Ah yep that would make sense. The way I would think to fix it is just to remove all `[0]` from any path on an instance, so that all paths are the using the optional syntax and we should therefore not hit this problem. That doesn't strike me as too hard to do within this PR, unless you think that's a bad approach for some reason.

cmoesel:
Yeah, I think that sounds like a good approach.  At first I was going to say to normalize the syntax to always use `[0]`, but that requires you to look at every part of the path to see if it is an array.  Normalizing _out_ the `[0]` makes a lot more sense since you don't even have to look at the types to do that.  Good idea.

ngfreiter:
Okey doke, I will add/test that on this PR then.

cmoesel:
Other than what I already noted and @ngfreiter is working on, it all looks good to me.  I took it through a few paces and it worked as expected.  I also verified that mCODE comes out the same with or without this branch (as expected).  So... looks good once the `[0]` bit is fixed.

PR #378, opened 2020-04-22
Fix contained choice bug

### LINKED ISSUE #377: Cannot fix resource to choice element
See this thread for details: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/contained.20resources.20of.20different.20types.

If the `contained` element is constrained to a choice of several types, such as `Patient` or `Observation`, then we should still allow any resource of type `Patient` or `Observation` to be placed on an instance in the `contained` array. Trying to do this currently causes an error, because we require that an element have a single type if any value is going to be set on it.

ngfreiter:
This fixes #377 and the bug that is referenced in the zulip thread linked on #377. It turns out they are actually two separate problems.

We used to be unable to set an element that is a choice of several resources to a specific instance of one of the resources on an Instance. Now if an element on an instance has a choice of several resource types, such as `Patient` or `Observation`, that element can be set to any instance of type `Patient` or `Observation`.

This also fixes a bug with arrays of inline instances. When several inline instances are defined in an array:
```
* contained[0] = medicalHistoryEmpty
* contained[1] = adverseReactionEmpty
* contained[2] = immunizationEmpty
* contained[3] = medicineListEmpty
```
There was a bug with elements in the array beyond the first. The `[0]` index on the first element of the array would be normalized out, so the first path becomes `contained`. Then, because of this bug, we would interpret subsequent elements as children of this path, so we would view them as attempting to override the value of `contained`. This of course caused issues, which are fixed in this PR.

### DIFF
--- a/src/errors/FixingNonResourceError.ts
+++ b/src/errors/FixingNonResourceError.ts
@@ -1,0 +1,7 @@
+export class FixingNonResourceError extends Error {
+  constructor(public resourceType: string, public id: string) {
+    super(
+      `Instance ${id} of type ${resourceType} is not an Instance of a Resource. Only Instances of Resources may be assigned to other Instances.`
+    );
+  }
+}
--- a/src/errors/FixingNonResourceError.ts
+++ b/src/errors/FixingNonResourceError.ts
@@ -1,7 +1,7 @@
 export class FixingNonResourceError extends Error {
   constructor(public resourceType: string, public id: string) {
     super(
-      `Instance ${id} of type ${resourceType} is not an Instance of a Resource. Only Instances of Resources may be assigned to other Instances.`
+      `Instance ${id} of type ${resourceType} is not an Instance of a Resource or Profile of a Resource. Only Instances of Resources or Profiles of Resources may be assigned to other Instances.`
     );
   }
 }
--- a/src/fhirtypes/ElementDefinition.ts
+++ b/src/fhirtypes/ElementDefinition.ts
@@ -27,7 +27,8 @@ import {
   FixedToPatternError,
   MultipleStandardsStatusError,
   InvalidMappingError,
-  InvalidFHIRIdError
+  InvalidFHIRIdError,
+  FixingNonResourceError
 } from '../errors';
 import { setPropertyOnDefinitionInstance, splitOnPathPeriods, isInheritedResource } from './common';
 import { Fishable, Type, Metadata, logger } from '../utils';
@@ -1366,14 +1367,22 @@ export class ElementDefinition {
    * @param {InstanceDefinition} value - The resource to fix
    * @param {Fishable} fisher - A fishable implementation for finding definitions and metadata
    * @throws {MismatchedTypeError} when the ElementDefinition is not of type Resource
+   * @throws {FixingNonResourceError} when the value being fixed is non-Resource
    * @returns {InstanceDefinition} the input value when it can be fixed
    */
   checkFixResource(value: InstanceDefinition, fisher: Fishable): InstanceDefinition {
     if (this.type?.some(t => isInheritedResource(value.resourceType, t.code, fisher))) {
       return value;
     } else {
-      const typesString = this.type?.map(t => t.code).join(', ');
-      throw new MismatchedTypeError(value.resourceType, value.id, typesString);
+      const resource = fisher.fishForFHIR(value.resourceType, Type.Resource);
+      if (resource) {
+        // In this case the value is actually a resource, but not one that inherits from this.type
+        const typesString = this.type?.map(t => t.code).join(', ');
+        throw new MismatchedTypeError(value.resourceType, value.id, typesString);
+      } else {
+        // In this case the value is not actually a resource
+        throw new FixingNonResourceError(value.resourceType, value.id);
+      }
     }
   }
 
--- a/src/fhirtypes/ElementDefinition.ts
+++ b/src/fhirtypes/ElementDefinition.ts
@@ -1365,19 +1365,15 @@ export class ElementDefinition {
    * Checks if a resource can be fixed to this element
    * @param {InstanceDefinition} value - The resource to fix
    * @param {Fishable} fisher - A fishable implementation for finding definitions and metadata
-   * @throws {NoSingleTypeError} when the ElementDefinition does not have a single type
    * @throws {MismatchedTypeError} when the ElementDefinition is not of type Resource
    * @returns {InstanceDefinition} the input value when it can be fixed
    */
   checkFixResource(value: InstanceDefinition, fisher: Fishable): InstanceDefinition {
-    if (!this.hasSingleType()) {
-      throw new NoSingleTypeError('Resource');
-    }
-    const type = this.type[0].code;
-    if (isInheritedResource(value.resourceType, type, fisher)) {
+    if (this.type?.some(t => isInheritedResource(value.resourceType, t.code, fisher))) {
       return value;
     } else {
-      throw new MismatchedTypeError(value.resourceType, value.id, type);
+      const typesString = this.type?.map(t => t.code).join(', ');
+      throw new MismatchedTypeError(value.resourceType, value.id, typesString);
     }
   }
 
--- a/test/fhirtypes/ElementDefinition.checkFixResource.test.ts
+++ b/test/fhirtypes/ElementDefinition.checkFixResource.test.ts
@@ -39,7 +39,7 @@ describe('ElementDefinition', () => {
       expect(value.resourceType).toBe('Patient');
     });
 
-    it('should throw MismatchedTypeError when the value is fixed to a non-Resource', () => {
+    it('should throw MismatchedTypeError when a Resource is fixed on a non-Resource element', () => {
       const status = observation.elements.find(e => e.id === 'Observation.status');
       expect(() => {
         status.checkFixResource(inlineInstance, fisher);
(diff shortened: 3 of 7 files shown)


src/fhirtypes/ElementDefinition.ts
```diff
@@ -1365,19 +1365,15 @@ export class ElementDefinition {
...
    * @throws {MismatchedTypeError} when the ElementDefinition is not of type Resource
    * @returns {InstanceDefinition} the input value when it can be fixed
    */
   checkFixResource(value: InstanceDefinition, fisher: Fishable): InstanceDefinition {
-    if (!this.hasSingleType()) {
-      throw new NoSingleTypeError('Resource');
-    }
-    const type = this.type[0].code;
-    if (isInheritedResource(value.resourceType, type, fisher)) {
+    if (this.type?.some(t => isInheritedResource(value.resourceType, t.code, fisher))) {
```
cmoesel, line 1372: I'm trying to think of cases where maybe we still do want the single type check...  If someone created an instance that was an instance of a datatype (not a resource), then `checkFixResource` would result in a `MismatchedTypeError` since it doesn't have a `resourceType` field, but the error would be confusing.  For example:
```
Instance: MyConcept
InstanceOf: CodeableConcept
Usage: #inline
* coding.code = #bar
* coding.system = "http://foo.org"

Instance: AnObservation
InstanceOf: Observation
* value[x] = MyConcept
```
results in this error:
> error Cannot fix CodeableConcept value: MyConcept. Value does not match element type: Quantity, CodeableConcept, string, boolean, integer, Range, Ratio, SampledData, time, dateTime, Period

which is confusing since it _is_ one of those types!

Anyway, I think that all cases where a choice _should_ be split out to different properties (e.g., `valueQuantity` or `valueCodeableConcept` or...) always have an element path that ends with `[x]`.  So maybe this function should still throw a `NoSingleTypeError` if the element path ends in `[x]` but it is not a single type?  Or maybe just setting anything on an element that ends in `[x]` (even if it is constrained to a single type) is bad?

ngfreiter: Wait...are we allowing people to make an Instance of a datatype? If you are doing that I don't think you would ever be able to set something to an instance of a datatype, because the datatype never has a resourceType.

ngfreiter: Yeah so like this:
```
Instance: MyConcept
InstanceOf: CodeableConcept
* coding.code = #bar
* coding.system = "http://foo.org"

Instance: AnObservation
InstanceOf: Observation
* valueCodeableConcept = MyConcept
```
Is also not allowed. Should it be?

cmoesel: I don't know.  I don't recall what we say in the FHIR Shorthand spec... but it seems that if it doesn't work, we should either produce an error outright as soon as someone tries to create and instance like that, or we should add support for it.  I guess as a starting point (since it's easier) we should probably just produce an error when someone tries to create an instance like that.

cmoesel: But either way, I think maybe the check for `[x]` is a good thing anyway in case there is any other situation this might come up?

ngfreiter: The spec kinda assumes that things will be a Resource I think:
image

I will add the error message and explicitly disallow making an instanceOf anything that is not a resource. 

I'm sorry I'm not getting what the check for `[x]` would check in this situation. Assuming that I add the error discussed above so that instances always are Resources and therefore have some resourceType.

ngfreiter: image
I'm timid now about removing the ability to make an instance of anything that isn't a resource, since the code shown above that looks for json to make an instance of explicitly includes `Type.Type` and `Type.Extension`.

cmoesel: > I'm sorry I'm not getting what the check for [x] would check in this situation. 

Two things:

1. I don't know if it exists anywhere (probably not?) but if there is a type that is something like `value[x]` and a choice of `Resource` or `CodeableConcept` or...  then even if the thing is a resource, it would need to be assigned against `valueResource`.  (Admittedly, this is likely a stretch).

2. Future-proofness.  If we support instances of datatypes in the future (and I think there *are* use cases for it), then this will continue to work if we have the check for `[x]` in it.  If we don't have the check for `[x]` in it, I'm thinking we'll probably forget to add it at that point and it will be a bug waiting to be discovered.

That said, if you feel strongly against adding it, I won't push for it any further.

ngfreiter: That makes sense, I just didn't understand how it would fit in, so I will add it.

ngfreiter: Unfortunately we have tests in the InstanceExporter which use Instances of Extensions. The tests don't necessarily rely on the fact that the Instance is an Extension, just happens to be the case. It doesn't seem like it makes sense to change all those tests because we are taking out support for something that we plan to add back in support for. Additionally it does feel kinda weird to me to put the `value[x]` check in there when I can't test it on a situation for which it would be relevant. So it kinda feels like the only sensible way forward is to add full support for inline Extensions and Datatypes now.

cmoesel: If you add the `[x]` check now, but _don't_ disallow instances of datatypes, then can't you test it by replicating the example I posted above -- and the test ensures that the correct error is logged (instead of the misleading one)?  Then we can make adding support for inlining datatype instances a separate story.

ngfreiter: I guess my issue with that is that if someone tries to do this:
```
Instance: MyConcept
InstanceOf: CodeableConcept
Usage: #inline
* coding.code = #bar
* coding.system = "http://foo.org"

Instance: AnObservation
InstanceOf: Observation
* value[x] = MyConcept
```
and they get an error about value[x] not having a single type, I think their next logical step is then to try this:
```
Instance: AnObservation
InstanceOf: Observation
* valueCodeableConcept = MyConcept
```
but that will give them the following error message, which is maybe even worse:
image

That error message would really only be fixed by fully supporting instances of datatypes and extensions.

cmoesel: OK, fair enough.  How about checking for `resourceType` and if it's not there, issuing an error that assigning non-resource instances is not currently supported?  I'm just trying to find a way to be able to merge this without having to wait for code to support datatype instances.

ngfreiter: Ok that seems to make the most sense at this stage. Then later on if people are clamoring for it they will be clamoring for a new feature that has a clear message than a bug with a confusing one.

cmoesel: And when we implement it, we'll be forced to look at this section of the code since we'll need to remove the check for `resourceType` -- so it's less likely we'll forget to address multiple types at that point.

ngfreiter: So bad news is we can't use `resourceType` because `resourceType` is set on non-Resources too, because non-Resources aren't handled in a special way by the InstanceExporter. I think the solution I put in place accomplishes something similar. I don't love it, but I couldn't think of a better way to put a band-aid on this for now until we want to truly address it.

src/errors/FixingNonResourceError.ts
```diff
@@ -0,0 +1,7 @@
+export class FixingNonResourceError extends Error {
+  constructor(public resourceType: string, public id: string) {
+    super(
+      `Instance ${id} of type ${resourceType} is not an Instance of a Resource. Only Instances of Resources may be assigned to other Instances.`
```
cmoesel, line 4: Should we add "or profile of a Resource" just to be more clear that it is allowed?

test/fhirtypes/ElementDefinition.checkFixResource.test.ts
```diff
@@ -34,21 +39,21 @@ describe('ElementDefinition', () => {
...
       }).toThrow(
-        'Cannot fix Resource value on this element since this element does not have a single type'
+        'Cannot fix Patient value: MyInlineInstance. Value does not match element type: code'
       );
     });
 
-    it('should throw MismatchedTypeError when the value is fixed to a non-Resource', () => {
-      const status = observation.elements.find(e => e.id === 'Observation.status');
+    it('should throw FixingNonResourceError when the value is fixed to a non-Resource', () => {
+      const status = observation.elements.find(e => e.id === 'Observation.category');
```
cmoesel, line 52: Super minor, but this should be `const category`.

test/fhirtypes/ElementDefinition.checkFixResource.test.ts
```diff
@@ -34,21 +39,21 @@ describe('ElementDefinition', () => {
       expect(value.resourceType).toBe('Patient');
     });
 
-    it('should throw NoSingleTypeError when element has multiple types', () => {
-      const valueX = observation.elements.find(e => e.id === 'Observation.value[x]');
+    it('should throw MismatchedTypeError when the value is fixed to a non-Resource', () => {
```
cmoesel, line 42: I think this should be `when the value is fixed to a Resource`, right?

ngfreiter:
Ok @cmoesel, https://github.com/FHIR/sushi/pull/378/commits/fb30cbe5fd7a4cb5d51ab700aa7720a72209aacb, should fix those issues.

PR #381, opened 2020-04-23
Config YAML export package-list

ngfreiter:
This adds exporting of `package-list.json` from the `config.yaml` file to the IGExporter.

### DIFF
--- a/src/ig/IGExporter.ts
+++ b/src/ig/IGExporter.ts
@@ -40,6 +40,7 @@ import { FHIRDefinitions } from '../fhirdefs';
 export class IGExporter {
   private ig: ImplementationGuide;
   private readonly packagePath: string;
+  private readonly configPath: string;
   private readonly outputLog: Map<string, outputLogDetails>;
   constructor(
     private readonly pkg: Package,
@@ -48,7 +49,12 @@ export class IGExporter {
     private readonly isIgPubContext: boolean = false
   ) {
     this.packagePath = path.resolve(this.igDataPath, '..', 'package.json');
+    this.configPath = path.resolve(this.igDataPath, '..', 'config.yaml');
     this.outputLog = new Map();
+  }
+
+  getOutputLogDetails(file: string) {
+    return this.outputLog.get(file);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/ig/IGExporter.ts
```diff
@@ -690,78 +696,22 @@ export class IGExporter {
...
+  addPackageList(igPath: string): void {
+    if (this.pkg.config?.history) {
+      const outputPath = path.join(igPath, 'package-list.json');
+      outputJSONSync(outputPath, this.pkg.config.history, { spaces: 2 });
+      logger.info('Generated package-list.json');
+      this.updateOutputLog(outputPath, [this.configPath], 'generated');
+    } else if (/^https?:\/\/hl7.org\//.test(this.pkg.config?.url)) {
+      logger.warn(
+        'HL7 IGs must have a package-list.json. Please define "history" in config.yaml to generate one.'
+      );
```
jafeltra, line 713: If someone had an existing `package-list.json` file, it will eventually be imported into `this.pkg.config.history`, right? So eventually we will only reach this warning if there is no package-lock information defined anywhere, right?

test/ig/IGExporter.test.ts
```diff
@@ -10,7 +10,7 @@ import { loggerSpy } from '../testhelpers/loggerSpy';
 import { FHIRDefinitions, loadFromPath, loadCustomResources } from '../../src/fhirdefs';
 import { TestFisher } from '../testhelpers';
 
-describe('IGExporter', () => {
+describe.skip('IGExporter', () => {
```
jafeltra, line 13: Is there a reason to skip all of the IGExporter tasks from the start rather than skipping individual tests as they break?

src/ig/IGExporter.ts
ngfreiter: Yes this should happen only if there is no `package-list.json` or information in the `this.pkg.config.history`, and even then only if it is an HL7 IG.

test/ig/IGExporter.test.ts
ngfreiter: I was thinking that at some point all these tests are going to have to be refactored/removed, so may as well just skip em. But I think it may be a better idea to only skip the tests I need to skip, so that this PR only concerns itself with `package-list`. So I think I will change that.

src/ig/IGExporter.ts
jafeltra: Okay! That makes sense, thanks!

ngfreiter:
As of https://github.com/FHIR/sushi/pull/381/commits/8c4aab943c50f84efc3f0769bd53614525905d8a, only the tests pertaining to package-list.json are skipped.

PR #383, opened 2020-04-23
Return non-zero exit code

### LINKED ISSUE #375: Uncaught errors exit process w/ `0` exit code
SUSHI should exit with a non-zero exit code if there are any errors in the SUSHI build.  This happens correctly as long as SUSHI is not interrupted by an unexpected thrown error.  In the case of a throw error that bubbles to the top, Node reports this:
> (node:15111) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
> (node:15111) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.

Note that it appears that Node intentionally stops a non-zero exit code from being returned. We will need to handle the rejection (via a top-level catch), log the error, and exit the process w/ a non-zero exit code.

jafeltra:
This PR fixes #375 (https://standardhealthrecord.atlassian.net/browse/CIMPL-356).

The async `app` function will return a promise. We can catch unexpected errors in a `catch` block on the function. Since the fix for this is in `app.ts`, I'm not sure I can write a test for this. However, to test it manually, you can throw an error in `app` before the export begins. On master, the error in the GitHub issue is reported. On this branch, the error message is logged and the process exits with code `1`, which is the code we return if there are any errors in the export.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -84,14 +84,15 @@ async function app() {
     files = getFilesRecursive(input);
   } catch {
     logger.error('Invalid path to FSH definition folder.');
-    program.help();
+    program.outputHelp();
+    process.exit(1);
   }
 
   // Check that package.json exists
   const packagePath = path.join(input, 'package.json');
   if (!fs.existsSync(packagePath)) {
     logger.error('No package.json in FSH definition folder.');
-    return;
+    process.exit(1);
   }
 
   // Parse package.json
@@ -100,7 +101,7 @@ async function app() {
     config = fs.readJSONSync(packagePath);
   } catch (e) {
     logger.error(`The package.json file is not valid JSON: ${packagePath}`);
-    return;
+    process.exit(1);
   }
 
   // Ensure FHIR R4 is added as a dependency
@@ -110,7 +111,7 @@ async function app() {
       'The package.json must specify FHIR R4 as a dependency. Be sure to' +
         ' add "hl7.fhir.r4.core": "4.0.1" to the dependencies list.'
     );
-    return;
+    process.exit(1);
   }
 
   // Load external dependencies
@@ -155,7 +156,7 @@ async function app() {
         ' may be corrupt. Local FHIR cache can be found at <home-directory>/.fhir/packages.' +
         ' For more information, see https://wiki.hl7.org/FHIR_Package_Cache#Location.'
     );
-    return;
+    process.exit(1);
   }
 
   logger.info('Converting FSH to FHIR resources...');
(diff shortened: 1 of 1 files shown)


src/app.ts
```diff
@@ -23,7 +23,10 @@ import {
 import { pad, padStart, sample, padEnd } from 'lodash';
 import chalk from 'chalk';
 
-app();
+app().catch(e => {
+  logger.error(e.message);
```
ngfreiter, line 27: Do you think it might make sense here to just say something about how SUSHI exited unexpectedly along with the `e.message`? In theory we shouldn't ever get here anyway. But I'm thinking that if an error actually does sneak its way up here, it's possible the message associated with it is confusing/not helpful.

jafeltra: Sure! That seems reasonable. How do you feel about
> SUSHI encountered the following unexpected error: ${error.message}

ngfreiter: Looks good to me!

jafeltra:
I updated the error message in 924f004.

But now I think I might have a question for my own PR. In some of the other error cases that we already handle (like a missing `package.json`), we either `return` or call `program.help`, but I believe both of those exit with an exit code `0`. Is that okay? Or should all the situations where we `return` out of `app()` return a non-zero code?

jafeltra:
Okay, actually I found the following definition for exit code 1:

```
1 Uncaught Fatal Exception: There was an uncaught exception, and it was not handled by a domain or an 'uncaughtException' event handler.
```

So I'm thinking that since those are places that we've caught the error, maybe we don't want them to all just default to using exit code 1. However, I'm really not sure. @cmoesel do you think I should update the other places as well?

cmoesel:
Wow, I didn't know that there were defined meanings for exit codes, @jafeltra!  I just picked `1` because it's not `0`.  Little did I know it has meaning.

You bring up a good point.  Ideally, I think we *should* return a non-zero exit code on any early exit.  So in those places where there is missing config data (or files), I think we should change `return` to `process.exit(1)` (or a more appropriate number if you can find one).  In the spot where we call `program.help()` we should change to `program.outputHelp()` and then exit with a non-zero exit code.

jafeltra:
I just found this list here! https://nodejs.org/api/process.html#process_exit_codes

But okay that makes sense! I'll update those other places as well. Thanks!

jafeltra:
Okay c0d0b94 changes the other places we were exiting to return exit code `1` as well!

PR #384, opened 2020-04-23
Export implementation guide using YAML config

mint-thompson:
`IGExporter` now uses the configuration that lives in the `Package` it receives in order to assemble and export the implementation guide JSON file. A few additions to the IG that used to live in functions other than `initIG` have been moved to `initIG`. All the tests in `IGExporter.test.ts` have been turned off, and a new file `IGExporter.IG.test.ts` contains tests specific to the exported implementation guide.

Apart from all that, the biggest change is to the `page` attribute. When a configuration is specified for pages, that configuration gets obeyed. If no configuration is specified for pages, the existing automatic generation steps are performed.


src/ig/IGExporter.ts
```diff
@@ -513,61 +626,86 @@ export class IGExporter {
...
             };
+            const configResource = (this.config.resources ?? []).find(
+              resource => resource.reference?.reference == referenceKey
+            );
 
-            if (pathEnd === 'examples') {
-              const exampleUrl = resourceJSON.meta?.profile?.find(
-                url =>
-                  this.pkg.fish(url, Type.Profile) ?? this.fhirDefs.fishForFHIR(url, Type.Profile)
+            if (!(configResource?.omit === true)) {
```
cmoesel, line 639:
```ts
if (configResource?.omit !== true) {
```

src/ig/IGExporter.ts
```diff
@@ -513,61 +626,86 @@ export class IGExporter {
...
+              const existingIndex = this.ig.definition.resource.findIndex(
+                r => r.reference.reference === referenceKey
               );
-              if (exampleUrl) {
-                resource.exampleCanonical = exampleUrl;
-              } else {
-                resource.exampleBoolean = true;
+              const existingResource =
+                existingIndex >= 0 ? this.ig.definition.resource[existingIndex] : null;
```
cmoesel, line 644: I was a little confused when first reviewing this.  I think I understand now that this code about `existingResource` is to support the use case where someone wants to override a FSH-generated resource with a user-provided resource (which might be helpful to get around a known bug in the FSH generation).  Maybe put a comment somewhere indicating the use case we're trying to handle?

src/ig/IGExporter.ts
```diff
@@ -513,61 +626,86 @@ export class IGExporter {
...
               }
-            } else {
-              resource.exampleBoolean = false;
-              // On some resources (Patient for example) these fields can be objects, avoid using them when this is true
-              const title = typeof resourceJSON.title === 'string' ? resourceJSON.title : null;
-              const name = typeof resourceJSON.name === 'string' ? resourceJSON.name : null;
-              if (title || name) {
-                resource.name = title ?? name;
+              if (configResource?.groupingId) {
+                newResource.groupingId = configResource.groupingId;
```
cmoesel, line 655: Again, need to also look at `config.groups`.

src/ig/IGExporter.ts
```diff
@@ -513,61 +626,86 @@ export class IGExporter {
               continue;
             }
 
-            const resource: ImplementationGuideDefinitionResource = {
+            const referenceKey = `${resourceJSON.resourceType}/${resourceJSON.id}`;
```
cmoesel, line 629: This is the third time we're seeing a block of logic like this with a lot of similarities -- but I expect it's one of those cases where things are just slightly-different enough that it can't easily be extracted to a common function?  Assuming that to be the case, we may want to put in a block of comments before each section like this saying the it shares very similar code with two other sections and that when changes are made here, the developer should also refer to the other sections to see if changes need to be made there as well.  What do you think?

test/ig/IGExporter.IG.test.ts
```diff
@@ -0,0 +1,548 @@
...
+      const config: Configuration = {
+        filePath: path.join(fixtures, 'config.yml'),
+        id: 'sushi-test',
+        url: 'http://hl7.org/fhir/sushi-test',
+        version: '0.1.0',
+        name: 'sushi-test',
+        title: 'FSH Test IG',
+        description: 'Provides a simple example of how FSH can be used to create an IG',
+        dependencies: [
+          { packageId: 'hl7.fhir.r4.core', version: '4.0.1' },
```
cmoesel, line 38: In the new Configuration, the FHIR core package shouldn't be listed as a dependency.

test/ig/IGExporter.IG.test.ts
```diff
@@ -0,0 +1,548 @@
...
+          { packageId: 'hl7.fhir.us.core', version: '3.1.0' },
+          { packageId: 'hl7.fhir.uv.vhdir', version: 'current' }
+        ],
+        status: null,
```
cmoesel, line 42: Since I think we shouldn't default status anymore, and instead we should require status in the config, I think maybe we should just set status to a real value here.

test/ig/IGExporter.IG.test.ts
```diff
@@ -0,0 +1,548 @@
...
+        template: null,
```
cmoesel, line 43: Unless you're specifically testing handling of null template, I'd set it to a real value here.  I think we will soon be assuming that if no template is provided that means you don't want an IG generated.

test/ig/IGExporter.IG.test.ts
```diff
@@ -0,0 +1,548 @@
...
+      };
+      pkg = new Package(packageJSON, config);
+      exporter = new IGExporter(pkg, defs, path.resolve(fixtures, 'ig-data'), false);
+    });
+
+    afterAll(() => {
+      temp.cleanupSync();
+    });
+
+    it('should provide defaults for copyrightyear and releaselabel parameters', () => {
```
cmoesel, line 313: I think maybe it shouldn't anymore.  I'll generate out some defaults to config.yaml in the case that the user has _no_ config.yaml in their project.  But if a config.yaml exists, we should just expect the user to set copyright year and releaselabel since any guesses we take will be wrong in some cases.

test/ig/IGExporter.IG.test.ts
```diff
@@ -0,0 +1,548 @@
...
+        },
+        {
+          nameUrl: 'nothing.html',
+          title: 'Nothing',
+          generation: 'markdown'
+        }
+      ]);
+      expect(loggerSpy.getLastMessage('error')).toMatch(/nothing\.md not found/s);
+    });
+  });
```
cmoesel, line 547: I feel like there should be more tests around `config.resources` -- as these tests only cover when the config doesn't specify resources.  But we need to test use cases like omitted resources, partial resource info, etc.

cmoesel: We also need tests for `config.groups` when you've implemented that.

src/ig/IGExporter.ts
```diff
@@ -75,50 +82,27 @@ export class IGExporter {
...
-    // TODO: Consider adding a generated file warning when the IG Publisher supports JSON5-style comments
+  private initIG() {
+    // first, properties that can be directly used without much trouble
     this.ig = {
       resourceType: 'ImplementationGuide',
-      id: packageJSON.name,
-      url: `${packageJSON.canonical}/ImplementationGuide/${packageJSON.name}`,
-      version: packageJSON.version,
+      id: this.config.id,
+      url: `${this.config.url}/ImplementationGuide/${this.config.name}`,
```
ngfreiter, line 94: Doesn't the format for the url have to be `<canonical>/<resourceType>/<resource.id>`? So if we are setting `id` to `this.config.id`, don't we need to set `url` to `${this.config.url}/ImplementationGuide/${this.config.id}`?

src/ig/IGExporter.ts
```diff
@@ -131,57 +115,68 @@ export class IGExporter {
...
-            value: '' // Gets set when ig.ini is processed
+            value: this.config?.parameters?.find(p => p.code == 'releaselabel')?.value ?? 'CI Build'
           },
-          {
-            code: 'show-inherited-invariants',
-            value: 'false' // TODO: Make this configurable
-          }
+          ...(this.config?.parameters?.filter(
+            p => ['copyrightyear', 'releaselabel'].indexOf(p.code) == -1
+          ) ?? [])
```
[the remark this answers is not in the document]
cmoesel: Does it crash?  If so, then I vote we make it required in the config.  I just don't want to guess at something we don't actually know.

test/ig/IGExporter.IG.test.ts
```diff
@@ -0,0 +1,548 @@
...
+      loadFromPath(
+        path.join(__dirname, '..', 'testhelpers', 'testdefs', 'package'),
+        'testPackage',
+        defs
+      );
+      const fixtures = path.join(__dirname, 'fixtures', 'simple-ig');
+      // several parts of the IG exporter still need packageJSON to function
+      const packageJSON: PackageJSON = fs.readJSONSync(path.join(fixtures, 'package.json'));
+      const config: Configuration = {
+        filePath: path.join(fixtures, 'config.yml'),
```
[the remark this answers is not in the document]
cmoesel: The importer looks for both `config.yaml` and `config.yml`, preferring the former.

src/ig/IGExporter.ts
cmoesel: Yes.  Whoops.  I missed that one.  I expect this is just a remnant of the fact that with `packageJSON`, `packageJSON.name` _was_ the `id` of the IG.

test/ig/IGExporter.IG.test.ts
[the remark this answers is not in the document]
ngfreiter: Wait it does currently? Or is that an update that is coming with your current task?

src/ig/IGExporter.ts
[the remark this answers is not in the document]
ngfreiter: image

I got the following when removing the `copyrightyear` parameter. This is just using `fhir.base.template`.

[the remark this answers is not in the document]
ngfreiter: And similar for `releaselabel`:
image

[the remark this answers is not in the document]
ngfreiter: Although I do not get a crash when removing `show-inherited-invariants`, even though this comment here kinda makes me think I should:
image

test/ig/IGExporter.IG.test.ts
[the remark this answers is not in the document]
cmoesel: Oh, yeah. Looks like that's only in my local branch. ;-)

src/ig/IGExporter.ts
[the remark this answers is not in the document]
cmoesel: So I think we should make `copyrightyear` and `releaselabel` required then.  I'll do that in my branch.  I think that comment (above `parameter: [`) was written when `parameter` only had the `copyrightyear` and `releaselabel` items.  I added `show-inherited-invariants` later and so that comment doesn't really apply.  That was more of a "user friendly default" at the time.

src/ig/IGExporter.ts
```diff
@@ -131,57 +115,68 @@ export class IGExporter {
...
+    if (this.config.global?.length) {
+      this.ig.global = this.config.global;
+    }
+
+    // Add the path-history, if applicable (only applies to HL7 IGs)
+    if (/^https?:\/\/hl7.org\//.test(this.config.url)) {
+      this.ig.definition.parameter.push({
+        code: 'path-history',
+        value: `${this.config.url}/history.html`
+      });
```
jafeltra, line 179: I believe the use of `config.url` here should now be replaced with `config.canonical`.

src/ig/IGExporter.ts
jafeltra, line 94: I believe this can now just be `this.config.url` because the `url` will be set to the default Nick described on the import side.

src/ig/IGExporter.ts
```diff
@@ -513,61 +626,86 @@ export class IGExporter {
...
-            const resource: ImplementationGuideDefinitionResource = {
+            const referenceKey = `${resourceJSON.resourceType}/${resourceJSON.id}`;
+            const newResource: ImplementationGuideDefinitionResource = {
               reference: {
-                reference: `${resourceJSON.resourceType}/${resourceJSON.id}`
-              },
-              name: resourceJSON.id, // will be overwritten w/ title or name where applicable
-              description: resourceJSON.description
+                reference: referenceKey
+              }
```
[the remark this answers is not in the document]
cmoesel: It's fine as is.  Who knows, prettier might force it anyway! ;-)

test/ig/IGExporter.test.ts
```diff
@@ -28,7 +28,36 @@ describe('IGExporter', () => {
...
+      const config: Configuration = {
+        filePath: path.join(fixtures, 'config.yml'),
+        id: 'sushi-test',
+        canonical: 'http://hl7.org/fhir/sushi-test',
+        version: '0.1.0',
+        name: 'sushi-test',
+        title: 'FSH Test IG',
+        description: 'Provides a simple example of how FSH can be used to create an IG',
+        dependencies: [
+          { packageId: 'hl7.fhir.r4.core', version: '4.0.1' },
```
cmoesel, line 40: I noticed this test is skipped now anyway, but if we bring it back, we shouldn't have FHIR R4 listed as a dependency since there is now a dedicate `fhirVersion` property.

test/ig/IGExporter.test.ts
```diff
@@ -28,7 +28,36 @@ describe('IGExporter', () => {
...
+          { packageId: 'hl7.fhir.us.core', version: '3.1.0' },
+          { packageId: 'hl7.fhir.uv.vhdir', version: 'current' }
+        ],
+        status: null,
+        template: null,
```
cmoesel, line 45: I noticed this test is skipped now anyway, but if we bring it back, we should set real values for status and template.

src/ig/IGExporter.ts
```diff
@@ -81,50 +90,33 @@ export class IGExporter {
...
-      id: packageJSON.name,
-      url: `${packageJSON.canonical}/ImplementationGuide/${packageJSON.name}`,
-      version: packageJSON.version,
+      id: this.config.id,
+      meta: this.config.meta,
+      implicitRules: this.config.implicitRules,
+      language: this.config.language,
+      text: this.config.text,
+      contained: this.config.contained,
+      modifierExtension: this.config.modifierExtension,
```
ngfreiter, line 107: Should `extension` be in here?

src/ig/IGExporter.ts
```diff
@@ -81,50 +90,33 @@ export class IGExporter {
...
-      }),
-      description: packageJSON.description,
-      packageId: packageJSON.name,
-      license: packageJSON.license,
-      fhirVersion: ['4.0.1'],
-      dependsOn: [],
+      name: this.config.name.replace(/[^A-Za-z0-9_]/g, ''),
+      title: this.config.title,
+      status: this.config.status,
+      publisher: this.config.publisher,
```
ngfreiter, line 114: I don't see where `experimental` and `date` are handled. I think they are simple copy overs and would go here?

src/ig/IGExporter.ts
```diff
@@ -133,61 +125,54 @@ export class IGExporter {
...
+        value: `${this.config.canonical}/history.html`
       });
     }
-
-    // Add the dependencies
-    if (this.pkg.packageJSON.dependencies) {
+    // check for some optional properties and add them if present
+    if (this.config.useContext?.length) {
+      this.ig.useContext = this.config.useContext;
+    }
```
ngfreiter, line 148: Why can't `useContext` be added the same was as the other properties that are being copied from `this.config` to `this.ig`? What is unique about it in this case?

src/ig/IGExporter.ts
```diff
@@ -231,90 +216,171 @@ export class IGExporter {
...
       logger.info('Generated default index.md.');
       const warning = warningBlock('<!-- index.md {% comment %}', '{% endcomment %} -->', [
         'This index.md file was generated from the "description" in package.json. To provide',
         'your own index file, create an index.md or index.xml in the ig-data/input/pagecontent',
-        'folder.',
+        'or ig-data/input/pages folder.',
         'See: https://build.fhir.org/ig/FHIR/ig-guidance/using-templates.html#root.input'
       ]);
-      const outputPath = path.join(igPath, 'input', 'pagecontent', 'index.md');
+      const outputPath = path.join(pageContentExportPath, 'index.md');
```
ngfreiter, line 282: I think in the spirit of not automatically generating stuff that could overwrite user content. Do we perhaps not generate an index.md by default here? This relates to some of the conversation happening in #411 about menu.xml.

cmoesel: Yeah, I agree it seems to go against the new philosophy... The index.md we generate is so simple... maybe instead of generating it, we just log a warning that it didn't exist and the user should create it?  For some people, this warning will be a little overkill (if they are managing index.md elsewhere), but as a temporary solution, maybe it's not bad?

ngfreiter: I think we came to a solution for handling this at the end of this comment thread https://github.com/FHIR/sushi/pull/409#pullrequestreview-407642325

cmoesel:
Woot!  It's good!  Thanks for this very impressive PR, @mint-thompson!  There's some really good stuff in here -- and this full control over the IG JSON will be *very* helpful to our users!  Thanks!

BTW -- I merged in the latest updates from `config-yaml` and it caused one slight issue: the new code to load dependencies from the config added FHIR R4 to the config dependencies.  This isn't really what we want (it's not valid for FHIR ImplementationGuide), so I pushed a very small commit (36c1b27) to stop adding R4 to the config.dependencies.

Issue #374, opened 2020-04-20
dev dependencies should fall back to current if not found


cmoesel:
According to [IG Publisher doc](https://confluence.hl7.org/pages/viewpage.action?pageId=35718627#IGPublisherDocumentation-DependencyList):

> dev: use the latest local build of the IG. in the absence of a local build, this falls back to using the current version

Zulip: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/dev.20dependencies

PR #386, opened 2020-04-23
Load #dev packages from #current

ngfreiter:
Fixes https://github.com/FHIR/sushi/issues/374.

This makes it so now when a dependency is specified as `dev`, if it is not locally cached, it will be loaded from build.fhir.org in the same way as if it was specified as `current`.

I debated on this PR if I should call the newly loaded package that is being placed in the cache `x.y#dev` or `x.y#current`. In the end I chose `x.y#dev`, but there may be an argument for `x.y#current`. When the IG Publisher runs, if it can't find a package specified as `dev`, it loads the package from build.fhir.org, but it writes it to the cache as `x.y#current`. So in this way our approach would be inconsistent with the IG Publisher. But I think there is a reason to be inconsistent in this case. 

If we wrote the package to the cache as `x.y#current`, then each time SUSHI runs, it will not be able to find the `x.y#dev` package, so it will fall back to `current`, and reload the `x.y#current` package into the cache. The IG Publisher seems to avoid this scenario by not overwriting a `current` package every time. I assume the IG Publisher must somehow check to see if a `current` package has changed, and only overwrite it if there has been a change. But I'm not sure how the Publisher goes about this. SUSHI does not have smart checking like that, so it overwrites a `current` package every time.

Even if SUSHI had this smart checking, I still think it makes sense to write the package to the cache as `x.y#dev`. If the dependency is specified as `dev`, I would assume a user wouldn't want their cache to update even _if_ the current version on build.fhir.org changes, so the IG Publisher way of doing things confuses me a bit. To me it seemed to make the most sense to download the package one time and save it as `x.y#dev`, so now the user has a local copy that will not update unless they purposefully change their dependency version from `dev`. Hopefully this made it somewhat clear why I chose to write the package as `x.y#dev`, but if you think writing the package as `x.y#current` or some other approach makes more sense, suggest away!


ngfreiter:
Hmm, not sure what is going on with GitHub actions here though.

cmoesel:
So... I kind of feel like we should write it as `#current` in the cache.  I feel this way because:
* it is more consistent with the IG publisher
* the docs describe `#dev` as _"the latest local build of the IG"_, so I feel like anything marked `#dev` in the cache _should_ be a _local build_.

If someone has a dependency marked as `#dev`, but no local build, I think it's fair to log an info message indicating that `#dev` was not found and it will use `#current` instead.  Then at least they know why it's doing a fresh download.

I think it might also be worth taking 10-15 minutes to see if there is an easy way to determine if a build is current... Maybe it is easier than we think.

ngfreiter:
Okey doke, I'll change to use `current` then.

I think if we are going to come up with a way to determine current builds, we should try to figure out what the IG Publisher does and do that, so we don't get into trouble at some point.

ngfreiter:
What do you think makes sense to put in the `dependsOn` of the ImplementationGuide. I could see wanting to switch it to `#current` there as well, to indicate that because `#dev` is not present, we have given them the `#current` package. But I could also see leaving it as `#dev`, because that is the value they entered (presumably intentionally), and because the IG Publisher will see the `#dev`, look for `#dev` in the cache, not find it, and then fall back to `#current` just as we did. As a point of reference, when I have an IG Resource with `#dev` in the `dependsOn`, and I run it through the IG Publisher without having the package cached, the `#current` package is downloaded to the cache, but `#dev` still comes out the other end in the IG Resource. So my default would be to follow what the Publisher does and leave the dependency as `#dev` on the IG Resource, but call the package `#current` in the cache. But this time I'm asking before coding 😉 .

cmoesel:
I agree with your assessment.  Keep it `#dev`.

ngfreiter:
Ok so https://github.com/FHIR/sushi/pull/386/commits/9fd40d96971f118a587256a883e6f1fade65fcfa makes it so that it is written to the cache as `#current`.

I wasn't able to figure out how the IG Publisher handles updating `#current`. I think this could be a good task on its own to investigate further, but for now I wasn't able to figure out how to make the IG Publisher update the `#current` folder. I tried changing the dates of the package.json, the IG.json resource, and the ig.ini stored in the package folder, but none of that seemed to work to force an update. So as it stands SUSHI will just update `#current` every time.

ngfreiter:
I've added it as a task to our backlog to investigate further: https://standardhealthrecord.atlassian.net/browse/CIMPL-364. Since at least as far as I can tell it isn't just a simple thing to add.

PR #389, opened 2020-04-24
Make copyrightYear and releaseLabel required config parameters

cmoesel:
Submitting this mini-PR since it affects one of @ngfreiter's tasks...

Note that these really should only be required if `template` is populated.  That's part of what Nick will be working on.


cmoesel:
Yeah, let's just leave it for now.  Since the importer doesn't halt on missing copyrightYear/releaseLabel, then it is actually still possible to get a `config.parameters` that is null anyway.

ngfreiter:
Ah good point, merging then!

PR #396, opened 2020-04-28
Add canonical field to configuration

jafeltra:
This addresses https://standardhealthrecord.atlassian.net/browse/CIMPL-362.

It adds a `canonical` field to `config.yaml` that is used in place of the current `url` field. The `url` field now represents the ImplementationGuide url. `canonical` becomes a minimally required field and `url` is no longer minimally required. `url` is defaulted during the import and uses the value specified for `canonical`. 

This also updates the existing tests to reflect the change in how `canonical` and `url` are used in a `config.yaml` file and it adds tests for the new `canonical` field in the `importConfiguration` tests.


src/fshtypes/Configuration.ts
```diff
@@ -36,7 +36,8 @@ export type Configuration = {
   contained?: any[];
   extension?: Extension[];
   modifierExtension?: Extension[];
-  url: string;
+  canonical: string;
+  url?: string;
```
ngfreiter, line 40: Sorry if this is an annoying request, but could you put canonical directly below `filePath`? Or nearer the end of the definition of `Configuration`. I ask because right now these properties match up against the properties of the IG resource perfectly in the same order.

src/import/YAMLConfiguration.ts
```diff
@@ -39,7 +39,8 @@ export type YAMLConfiguration = {
   contained?: ImplementationGuide['contained']; // any[]
   extension?: ImplementationGuide['extension']; // Extension[]
   modifierExtension?: ImplementationGuide['modifierExtension']; // Extension[]
-  url: ImplementationGuide['url']; // string
+  canonical: string;
+  url?: ImplementationGuide['url']; // string
```
ngfreiter, line 43: And if you change the order of canonical up there, can you change it here too?

src/import/importConfiguration.ts
```diff
@@ -91,7 +91,8 @@ export function importConfiguration(yaml: YAMLConfiguration | string, file: stri
     contained: yaml.contained,
     extension: yaml.extension,
     modifierExtension: yaml.modifierExtension,
-    url: yaml.url, // minimum config property
+    canonical: yaml.canonical, // minimum config property
+    url: yaml.url ?? `${yaml.canonical}/ImplementationGuide/${yaml.id}`,
```
ngfreiter, line 95: And if you change the order there, could it change here too!

test/import/importConfiguration.test.ts
```diff
@@ -365,23 +367,38 @@ describe('importConfiguration', () => {
...
   });
 
+  describe('#url', () => {
+    it('should import url as-is if provided', () => {
+      minYAML.url = 'http://foo.org/some-url/ImplementationGuide/my.guide';
+      const config = importConfiguration(minYAML, 'test-config.yaml');
+      expect(config.url).toBe('http://foo.org/some-url/ImplementationGuide/my.guide');
+    });
+    it('should default url based on canonical if url is not provided', () => {
+      // loggerSpy.reset(); // JULIA
```
ngfreiter, line 394: JULIA

jafeltra: OMG NOOOO

src/import/importConfiguration.ts
jafeltra: If you give a mouse a cookie!

test/import/importConfiguration.test.ts
jafeltra: I will be force pushing to remove all evidence of this.

src/import/importConfiguration.ts
jafeltra: But also yes.

jafeltra: I think I'm going to move it to the top of the list because it seems odd to bury the required property at the bottom of the long list. That does put it above `id`, which seems a little bold, so let me know if you'd prefer it at the bottom.

ngfreiter: Top is good with me!

PR #403, opened 2020-05-01
Allow setting fixed by fixed[x] values on an Instance

### LINKED ISSUE #398: Cannot set fixed[x] values on Instances
Details of this issue are here https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/fixedString.20in.20a.20profile.20causing.20errors.20in.20instance. As an example, consider this rule on an instance of this profile http://build.fhir.org/ig/hl7au/au-fhir-base/StructureDefinition-au-immunization.html.
```
* performer[administeredBy].function.text = "Administering Provider" // a fixed string in the profile
```
The user gets the following error:
>Cannot fix this element using a pattern; as it is already fixed in the StructureDefinition using fixedString. Since fixed[x] requires exact matches, while pattern[x] allows for variation in unspecified properties, fixed[x] cannot be replaced by pattern[x] since it would loosen the constraint.

This seems to be because when we test if a value on an instance can be fixed, we test if fixing it by `pattern[x]` on the profile would succeed. My suspicion is that we need to either always test if fixing by `fixed[x]` rather than `pattern[x]` would succeed. Or we need to determine which field is set on the element on the profile, and test using the corresponding strategy.

ngfreiter:
Fixes #398.

The issue here can be most clearly illustrated with the following example:
```
Profile: MyPatient
Parent: Patient
* name.family = "Smith" (exactly)

Instance: PatientInstance
InstanceOf: MyPatient
* name.family = "Smith"
```
SUSHI on master will allow you to do this, but an error will be emitted like:
>Cannot fix this element using a pattern; as it is already fixed in the StructureDefinition using fixedString. Since fixed[x] requires exact matches, while pattern[x] allows for variation in unspecified properties, fixed[x] cannot be replaced by pattern[x] since it would loosen the constraint.

With the error referencing the line setting `name.family` on the instance. This is incorrect, because when setting a value on an instance, fixed[x] and pattern[x] do not matter, you are just setting the value. This error happens because to test if we can fix a value on an instance, we try to fix the same value on the corresponding profile. But by mistake, we by default would always test fixing the value on the profile using pattern. This PR fixes that by now always testing if we can fix a value using fixed. On this branch, the above error should go away when testing the snippet above. There are also several other examples you can test on, in the linked issue and zulip chats. Finally, you can also run the tests on https://github.com/FHIR/sushi/commit/1cde85da1455ea332b521b727e44d76b95ea8591 and see that they fail, then run the tests on https://github.com/FHIR/sushi/commit/6bdd93eec1765dc82db3a9661cb520b63ef994a6 and see that they pass.

### DIFF
--- a/src/fhirtypes/StructureDefinition.ts
+++ b/src/fhirtypes/StructureDefinition.ts
@@ -575,7 +575,7 @@ export class StructureDefinition {
       // fixValue will throw if it fails, but skip the check if value is null
       if (value != null) {
         // since we don't actually keep the clone, argument for exactly should not matter
-        clone.fixValue(value, false, units);
+        clone.fixValue(value, true, units);
       }
       // If there is a fixedValue or patternValue, find it and return it
       const key = Object.keys(clone).find(k => k.startsWith('pattern') || k.startsWith('fixed'));
--- a/src/fhirtypes/StructureDefinition.ts
+++ b/src/fhirtypes/StructureDefinition.ts
@@ -574,7 +574,7 @@ export class StructureDefinition {
     } else {
       // fixValue will throw if it fails, but skip the check if value is null
       if (value != null) {
-        // since we don't actually keep the clone, argument for exactly should not matter
+        // exactly must be true so that we always test fixing with the more strict fixed[x] approach
         clone.fixValue(value, true, units);
       }
       // If there is a fixedValue or patternValue, find it and return it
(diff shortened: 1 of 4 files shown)


src/fhirtypes/StructureDefinition.ts
```diff
@@ -575,7 +575,7 @@ export class StructureDefinition {
       // fixValue will throw if it fails, but skip the check if value is null
       if (value != null) {
         // since we don't actually keep the clone, argument for exactly should not matter
```
jafeltra, line 577: Question - is this comment not really true anymore? Should we update it to say why we need `exactly` to be `true`?

ngfreiter: Oops! Good catch. I remember seeing that and thinking I should fix it, but here we are.

jafeltra: I've been there countless times, so I support you.

PR #407, opened 2020-05-05
Respect urls for local value sets and code systems

### LINKED ISSUE #380: SUSHI (probably) assumes that locally defined things have canonical-based URLs
It's possible for someone to use `^url = ...` to override the URL for any profile, extension, value set, code system, etc.  But I think there are places in our code (Fisher?) that probably assume the generated URL for these instead of considering that it might be overridden.  We should look into this and fix it as appropriate.

mint-thompson:
Fixes #380 and completes task [CIMPL-355](https://standardhealthrecord.atlassian.net/browse/CIMPL-355)

Locally defined value sets and code systems may have their urls specified with caret value rules. When this is the case, exported FHIR resources should use those urls when referring to those value sets and code systems.

As far as I could tell, the situations where this is potentially a problem are when a StructureDefinition or ValueSet refers to a ValueSet or CodeSystem. So, those four combinations each have a test case.


src/import/FSHTank.ts
```diff
@@ -204,18 +205,41 @@ export class FSHTank implements Fishable {
...
   }
 
+  /**
+   * Determines the URL to use to refer to a fished-up FHIR entity.
+   * If a caret value rule has been applied to the entity's url, use the
+   * value specified in that rule.
+   * Otherwise, use the default url based on the configured canonical url.
+   *
+   * @param result - The FHIR entity that was fished up
+   * @param fhirType - The entity's type as a string for building the default URL
```
ngfreiter, line 227: Small thing, but could you add the types to this description?

PR #411, opened 2020-05-07
BYO menu.xml

jafeltra:
This PR supports specifying a file path to a `menu.xml` file in the `config.yaml` `menu` attribute.

A user can now either define a menu within the configuration file or they can point to a specific file with a file path relative to the config file. If a file doesn't exist at the path specified, we log an error. 

There is a little bit of complexity with the various warnings we log due to the fact that we are still looking to see if there is a `menu.xml` file within `ig-data/input/includes` by default. So for example, if we find a menu.xml file there and a user specified something in `config.yaml` that does not point to that location, we log a warning. I think this is helpful and I believe all the cases are covered, but if we think we're doing too much inferring about what the user wants, I can try to take some of that extra handling out.


cmoesel:
If we do that then I can update my config generator to put in a menu (in the config) that is roughly equivalent to what we're generating now.  That way existing users who are happily using our generated menu will still get the generated menu, but it will be right in the config and we're essentially passing ownership of it to them at that point.

So... I think that means we have these possibilities:
* **no config.menu:** don't create or copy a menu
* **config.menu is  path to menu.xml:** copy over _that_ menu
* **config.menu is an object:** generate a menu based on the object

I'd be down with that.

jafeltra:
Okay yeah I like that plan too! I think it will get rid of some of the complexity/confusing warning cases and messages. One question I have is if there's a case where someone does have a menu.xml file in ig-data/input/includes and doesn't point to it in the config, do we just treat it like any other file in that folder and copy it over? Or do we try to be smart and still log a warning that it exists, and then either skip over it or copy it anyway?

ngfreiter:
I think copying it over violates the principle that if it isn't in the config, we are not going to make one appear. So maybe the best thing would be to issue a warning saying that that menu.xml is unused, so either point to it or get rid of it?

cmoesel:
Or we simplify things more:
* **no config.menu:** don't create a menu; but if you find one in `ig-data/input/includes`, copy it
* **config.menu is an object:** generate a menu based on the object

To me, I think this actually makes sense...  Then we'd do the same for `ig.ini`, and `package-list.json`.  If any of those exist in `ig-data`, then assume they want us to copy it -- and don't force them to put a path in the config where there is only one valid path anyway.

What do you think?

jafeltra:
Ah I like that plan and I'll start updating this PR to reflect them.

jafeltra:
@ngfreiter and @cmoesel 693a0b8 updates `menu.xml` as Chris last described it. These are the following scenarios:

- If config.menu is undefined and no menu is in ig-data, do nothing
- If config.menu does not exist but menu is in ig-data, copy it over
- If config.menu exists _and_ menu is in ig-data, log a warning but use the config to build a menu
- If config.menu exists and no menu in ig-data, build a menu XML file

There is still a disconnect between this and `package-list.json` (which I think I made worse when I tried to align the two earlier (the alignment was different then than what it should be now)). I can either align them here or it can be a separate task.

jafeltra:
Oh also @cmoesel I can remove the default menu.xml we use now in this PR, unless it is helpful for you to have while building the config generator.

ngfreiter:
> There is still a disconnect between this and `package-list.json` (which I think I made worse when I tried to align the two earlier (the alignment was different then than what it should be now)). I can either align them here or it can be a separate task.

I'd vote separate task on this.

cmoesel:
> Oh also @cmoesel I can remove the default menu.xml we use now in this PR, unless it is helpful for you to have while building the config generator.

Go ahead!  I just implemented it in the config generator.

jafeltra:
> And there is already a task that covers this

Ah I had missed that task. Given that and the fact that I just removed the SUSHI default menu, I think this PR is set to be re-reviewed.

src/ig/IGExporter.ts
```diff
@@ -441,52 +440,43 @@ export class IGExporter {
...
-          'be used to generate the IG menu. Please remove the menu.xml file from the ig-data folder. ' +
-          'To provide your own menu.xml, remove the "menu" attribute in config.yaml.'
-      );
+    if (existsSync(menuXMLDefaultPath) && this.pkg.config?.menu) {
+      const warningMessage =
+        'An IG menu is configured in config.yaml and provided in ig-data/input/includes/menu.xml. ' +
+        'Only the menu configured by config.yaml will be used to build the IG menu. ' +
+        'Remove the menu in ig-data/input/includes or remove the "menu" in config.yaml.';
+      logger.warn(warningMessage);
     }
```
ngfreiter, line 459: This could probably just be in the `logger.warn` now right?

src/ig/IGExporter.ts
```diff
@@ -441,52 +440,43 @@ export class IGExporter {
...
 
     // Always use config menu if defined
     if (this.pkg.config?.menu) {
+      // type is ConfigurationMenuItem[] so build XML file
```
ngfreiter, line 463: Was this comment leftover from when this could also be a string?

PR #418, opened 2020-05-11
Load dependencies from config.yaml

ngfreiter:
This addresses https://standardhealthrecord.atlassian.net/browse/CIMPL-378.

We now look at config.yaml instead of package.json to determine which packages to load.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -114,7 +114,7 @@ async function app() {
   }
 
   // Ensure FHIR R4 is added as a fhirVersion
-  const dependencies = config.dependencies;
+  const dependencies = config.dependencies ?? [];
   if (!config.fhirVersion.includes('4.0.1')) {
     logger.error(
       'The config.yaml must specify FHIR R4 as a fhirVersion. Be sure to' +
--- a/src/app.ts
+++ b/src/app.ts
@@ -113,28 +113,31 @@ async function app() {
     return;
   }
 
-  // Ensure FHIR R4 is added as a dependency
-  const fhirR4Dependency = packageJSON.dependencies?.['hl7.fhir.r4.core'];
-  if (fhirR4Dependency !== '4.0.1') {
+  // Ensure FHIR R4 is added as a fhirVersion
+  const dependencies = config.dependencies;
+  if (!config.fhirVersion.includes('4.0.1')) {
     logger.error(
-      'The package.json must specify FHIR R4 as a dependency. Be sure to' +
-        ' add "hl7.fhir.r4.core": "4.0.1" to the dependencies list.'
+      'The config.yaml must specify FHIR R4 as a fhirVersion. Be sure to' +
+        ' add "fhirVersion: 4.0.1" to the config.yaml file.'
     );
     return;
+  } else {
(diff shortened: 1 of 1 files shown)


src/app.ts
```diff
@@ -113,28 +113,31 @@ async function app() {
     return;
   }
 
-  // Ensure FHIR R4 is added as a dependency
-  const fhirR4Dependency = packageJSON.dependencies?.['hl7.fhir.r4.core'];
-  if (fhirR4Dependency !== '4.0.1') {
+  // Ensure FHIR R4 is added as a fhirVersion
+  const dependencies = config.dependencies;
```
mint-thompson, line 117: `config.dependencies` will be `undefined` if there were no dependencies listed in the configuration file, so an empty list should be used as a backup.

ngfreiter: Thanks for catching that, fixed in https://github.com/FHIR/sushi/pull/418/commits/751800e0d5398398eff0f617aab1bcc8c1a451d3

src/app.ts
```diff
@@ -113,28 +113,31 @@ async function app() {
...
     logger.error(
-      'The package.json must specify FHIR R4 as a dependency. Be sure to' +
-        ' add "hl7.fhir.r4.core": "4.0.1" to the dependencies list.'
+      'The config.yaml must specify FHIR R4 as a fhirVersion. Be sure to' +
+        ' add "fhirVersion: 4.0.1" to the config.yaml file.'
     );
     return;
+  } else {
+    // Add hl7.fhir.r4.core so that it will be loaded as dependency
+    dependencies.push({ packageId: 'hl7.fhir.r4.core', version: '4.0.1' });
```
jafeltra, line 126: I don't think there is really any harm in this, but I think it is technically possible to have someone specify R4 in the dependencies and then add it again here. Is that okay?

ngfreiter: You're right that would be possible, it wouldn't break anything, just load the files two times. Do you think it makes more sense to:
* Leave it as it currently is and let files be loaded twice.
* Check if `hl7.fhir.r4.core` so that it is only loaded once.
* Do the bullet point above, but also give a warning message saying the `hl7.fhir.r4.core` in the dependency list is not doing anything and should be removed.

ngfreiter: Actually I take that back, I believe the files don't actually even get loaded twice, because we cache the definitions.

jafeltra: Oh true so it just finds it in local cache the second time (if it isn't also there the first time)? I feel like since it really doesn't do any harm, I'm fine without a warning. I'd even be fine with the first bullet point (so just leave this as it is), I just wanted to confirm that it wasn't harmful, which it seems like it isn't!

ngfreiter: Yeah not only the local cache, as in `/.fhir` but I mean actually loading into memory. As in the first time, the 4.0.1 definitions will be loaded into memory, and the second time, we find that the definitions are already loaded into memory, so the files aren't even loaded a second time from `/.fhir`. And I've confirmed this is indeed how it works.

jafeltra: Ah gotcha! Okay I'm fine with leaving this as it is then!

PR #420, opened 2020-05-12
Copy package-list.json from ig-data

ngfreiter:
This PR aligns package-list.json with the behavior of menu.xml, and the planned behavior of ig.ini.
* If `history` is in config.yaml, that is used to generate package-list.json. If an ig-data/package-list.json also exists, a warning is emitted.
* If an ig-data/package-list.json exists and there is no `history` in config.yaml, then the ig-data/package-list.json is copied to the output.
* If neither exist, then we do nothing (although if the IG is an HL7 IG, we do emit a warning, since HL7 IGs should have a package-list.json)


src/ig/IGExporter.ts
```diff
@@ -1065,27 +1065,33 @@ export class IGExporter {
...
+        );
+      }
+    } else if (isIgDataPackageList) {
+      this.copyAsIs(packageListPath, path.join(igPath, 'package-list.json'));
+      logger.info('Copied ig-data/package-list.json.');
     } else if (/^https?:\/\/hl7.org\//.test(this.config.canonical)) {
       logger.warn(
-        'HL7 IGs must have a package-list.json. Please define "history" in config.yaml to generate one.'
+        'HL7 IGs must have a package-list.json. Please define "history" in config.yaml to generate one, ' +
+          'or add a package-list.json at ig-data/package-list.json'
```
cmoesel, line 1094: Do we want to warn on this?  For users who maybe don't use ig-data, but rely on package-list.json already being in the target folder, then even though they don't have it in ig-data, the overall workflow _will_ work...

ngfreiter: Yeah I guess that would be annoying for us to be warning someone who is still doing things right. On the other hand, a warning isn't an error, we are just saying we think you might be wrong, but maybe you know better.

But I think it's better to just not warn. I don't think this warning does much for us anyway. I'll remove it.

ngfreiter: To be clear though, I am going to not warn at all, not try to also detect if they have a package-list.json already in their target, because I don't think that's a road we want to head down.

ngfreiter: Warning gone in https://github.com/FHIR/sushi/pull/420/commits/aea4e46d9b92477f69631401906fe99ed8f3478d

src/ig/IGExporter.ts
```diff
@@ -1065,28 +1065,29 @@ export class IGExporter {
...
       this.updateOutputLog(outputPath, [this.configPath], 'generated');
-    } else if (/^https?:\/\/hl7.org\//.test(this.config.canonical)) {
-      logger.warn(
-        'HL7 IGs must have a package-list.json. Please define "history" in config.yaml to generate one.'
-      );
+      if (isIgDataPackageList) {
+        logger.warn(
+          'A package-list.json file is configured with "history" in config.yaml and provided in ig-data/package-list.json. ' +
+            'Only the package-list configured by config.yaml will be used to generate a package-list.json. ' +
+            'Remove the "history" in config.yaml to use the package-list.json in ig-data/package-list.json.'
```
cmoesel, line 1085: I feel like this message is a little awkward sounding.  I'm definitely biased, but I kind of like the general feel of the one I did for template/ig.ini: https://github.com/FHIR/sushi/blob/f59d3e196e2da19e7f61ec985c2b5c91a14bffcf/src/ig/IGExporter.ts#L986-L989

I won't hold up the PR for this, but I do feel like the message could be improved.

ngfreiter:
This should use `path.sep` instead of `/` in comments as @cmoesel does in #421.

src/ig/IGExporter.ts
ngfreiter: I agree, and am changing it currently.

ngfreiter: Ok in https://github.com/FHIR/sushi/pull/420/commits/e1b96568538bc6b2c3a675224ec51665d0416f93 I made the warning more closely align with the one for template/ig.ini, and I also did the same for the relevant warning for menu/menu.xml, so now all three warnings will be very similar.

jafeltra: Oh thank you! I was thinking of doing that also, so I'm glad you did it!

PR #423, opened 2020-05-12
Rules can be set on children of choice elements

mint-thompson:
Completes task [CIMPL-363](https://standardhealthrecord.atlassian.net/browse/CIMPL-363)

Every type available on a given choice element will have some common ancestor, so any elements of that ancestor should be allowed in the path for a rule. Therefore, when a choice element is encountered on a rule path, allow it to be unfolded to gain access to any elements that are guaranteed to exist.

All types share `Element` as an ancestor, but it's possible that a more specific type is a common ancestor of all the choices. This can give access to additional paths.


mint-thompson:
The tests are working on my machine, but they're not working here, so I will investigate why.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1613,6 +1613,50 @@ export class ElementDefinition {
...
+        const ancestry: string[] = [sd.url];
+        let ancestryWalker = sd;
+        while (ancestryWalker.baseDefinition) {
+          ancestryWalker = StructureDefinition.fromJSON(
+            fisher.fishForFHIR(ancestryWalker.baseDefinition)
+          );
+          ancestry.push(ancestryWalker.url);
+        }
+        return ancestry;
+      });
```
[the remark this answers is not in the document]
cmoesel: Actually, we probably want to consider profiles if they are there.  Thanks for pair-coding this with me, @mint-thompson!
```ts
const types: string[] = [];
this.type.forEach(t => {
  if (t.profile?.length) {
    types.push(...t.profile);
  } else {
    type.push(t.code);
  }
});

const allTypeAncestry = types.map(t => this.getTypeLineage(t, fisher).map(l => l.parent))
```

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -1434,6 +1448,28 @@ describe('StructureDefinitionExporter', () => {
       }
     ]);
   });
+  it('should apply a fixedValueRule on the child of a choice element with constrained choices', () => {
+    // Quantity is the first ancestor of Duration and Age
+
+    // * value[x] only Duration, Age
```
cmoesel, line 1454: Technically the comma isn't allowed in real FSH (it should be `Duration or Age`).  Not a big deal since it's in a comment though.

cmoesel: BTW -- thanks for this example.  It actually causes errors in IG Publisher, which I *think* is a bug in IG Publisher.  So you helped surface that one.

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -1705,6 +1741,35 @@ describe('StructureDefinitionExporter', () => {
     expect(fooSlice).toBeDefined();
   });
 
+  it('should apply a containsRule on the child of a choice element with a common ancestor of Element', () => {
+    // Element contains extension and id
+    // * value[x].extension contains my-inline-extension 0..1
```
cmoesel, line 1746: BTW -- this example also surfaced an issue w/ IG Publisher.  I'm not sure if it's our issue or the publishers, but it seems that the publisher does not like "sparse differentials" in definitions like this.  The publisher crashed unless I added elements for `Observation` and `Observation.value[x]` in the differential.  If that's our bug to fix, we'll fix it in a different PR though.

PR #424, opened 2020-05-12
config.yaml remove packageJSON

ngfreiter:
This PR removes any reliance on a `packageJSON` object loaded from a package.json file. All required configuration information is now obtained from the `Configuration` object loaded from config.yaml. I also removed any place where package.json was used or mentioned. This reaches a lot of files, but the changes are simple, mostly in the tests, and mostly find and replace type stuff.


src/ig/IGExporter.ts
```diff
@@ -968,16 +966,16 @@ export class IGExporter {
...
     logger.info(`Generated ImplementationGuide-${this.ig.id}.json`);
   }
 
   /**
-   * Creates an ig.ini file based on the package.json and exports it to the IG folder.
-   * If the user specified an igi.ini file in the ig-data folder, then use its values
-   * as long as they don't conflict with values already in package.json.
+   * Creates an ig.ini file based on the "template" in config.yaml and exports it to the IG folder.
+   * If the user specified an igi.ini file in the ig-data folder, and no "template" in config.yaml is specified,
+   * the file from ig-data is used instead.
```
cmoesel, line 978: Whoops. Guess I missed updating this comment, didn't I?

ngfreiter:
Oh and also note that I deleted IGExporter.test.ts. I don't think it makes sense to do the `config.yaml` release with that file still existing with all its skipped tests. I see the point that we may want to look at those tests in the future, but since we can still access them by just checking out an older commit, I felt like removing them made more sense then doing something hacky so that they could stay.

src/ig/IGExporter.ts
ngfreiter: Yeah you're right, `exampleConfig` could've been easier. But the old sunk cost got to me when I was halfway through changing things. And now this way we don't have unnecessary inconsistency across the tests.

PR #425, opened 2020-05-12
Support index file in config

jafeltra:
This PR makes the `index.[md | xml]` file work more similarly to `ig.ini`, `package-list.json`, and `menu.xml`.

- If the `indexPageContent` property is set in `config.yaml`, we use the value there to create an `index.md` file.
- If the `indexPageContent` property is not set and the user provided an `index.md` or `index.xml` file in either `pages` or `pagecontent`, then we copy that file over.
- If there is a file in any of those locations and `indexPageContent` is set, we use the config property and log a warning.

This also adds functionality to support setting a default `indexPageContent` in config if no file is provided. This allows us to support the default that a user would get on master now.

I do have two outstanding observations/questions:
- I didn't add a test case in `ensureConfiguration.test.ts` to check for every possible name/location of an index file. I can definitely do that, similar to the IG export tests, if reviewers think it is necessary.
- I kept the functionality to add the index to the list of `ig.pages` within the `addIndex` file, but this did require making `initIG` and `addImplementationGuide` public functions. I've used [rewire](https://github.com/jhnns/rewire) to use private functions in tests before (although not in a TypeScript project), or I could separate out that functionality into a different test, or it can stay as it is.


jafeltra:
Oh also if this gets in after Nick's PR #424 (which I'm guessing is likely), I'll need to resolve a few conflicts.

src/ig/IGExporter.ts
```diff
@@ -251,7 +251,38 @@ export class IGExporter {
...
+    if (existsSync(inputIndexMarkdownPagesPath)) filePath = inputIndexMarkdownPagesPath;
+    if (existsSync(inputIndexXMLPagesPath)) filePath = inputIndexXMLPagesPath;
+
+    if (this.config.indexPageContent) {
+      ensureDirSync(pageContentExportPath);
+      logger.info('Generated index.md based on "indexPageContent" in config.yaml.');
+      const warning = warningBlock('<!-- index.md {% comment %}', '{% endcomment %} -->', [
+        'To change the contents of this file, edit the "indexPageContent" attribute in the tank config.yaml file',
+        'or provide your own index file in the ig-data/input/pagecontent or ig-data/input/pages folder.',
+        'See: https://build.fhir.org/ig/FHIR/ig-guidance/using-templates.html#root.input'
```
ngfreiter, line 266: Could this use `path.sep` instead of `/` to give some love to us Windows users?

src/ig/IGExporter.ts
```diff
@@ -251,7 +251,38 @@ export class IGExporter {
...
     let generation: ImplementationGuideDefinitionPageGeneration = 'markdown';
-    if (existsSync(inputIndexMarkdownPageContentPath)) {
+    let filePath: string;
+    if (existsSync(inputIndexMarkdownPageContentPath)) filePath = inputIndexMarkdownPageContentPath;
+    if (existsSync(inputIndexXMLPageContentPath)) filePath = inputIndexXMLPageContentPath;
+    if (existsSync(inputIndexMarkdownPagesPath)) filePath = inputIndexMarkdownPagesPath;
+    if (existsSync(inputIndexXMLPagesPath)) filePath = inputIndexXMLPagesPath;
+
+    if (this.config.indexPageContent) {
+      ensureDirSync(pageContentExportPath);
```
ngfreiter, line 261: So we prefer to use the `pagecontent` folder instead of `pages`, is that correct? What is the reasoning for that?

src/ig/IGExporter.ts
```diff
@@ -251,7 +251,38 @@ export class IGExporter {
...
+      logger.info('Generated index.md based on "indexPageContent" in config.yaml.');
+      const warning = warningBlock('<!-- index.md {% comment %}', '{% endcomment %} -->', [
+        'To change the contents of this file, edit the "indexPageContent" attribute in the tank config.yaml file',
+        'or provide your own index file in the ig-data/input/pagecontent or ig-data/input/pages folder.',
+        'See: https://build.fhir.org/ig/FHIR/ig-guidance/using-templates.html#root.input'
+      ]);
+      const outputPath = path.join(pageContentExportPath, 'index.md');
+      outputFileSync(outputPath, `${warning}${this.config.indexPageContent}`);
+      this.updateOutputLog(outputPath, [this.configPath], 'generated');
```
ngfreiter, line 270: Since below we have logs for when we copy `index.[md | xml]`, I think it would be fair here to have a log that we have generated `index.md`. And to be clear I mean a `logger` log, not a `updateOutputLog` log.

src/ig/IGExporter.ts
```diff
@@ -559,13 +580,13 @@ export class IGExporter {
     const menuXMLOutputPath = path.join(igPath, 'input', 'includes', 'menu.xml');
 
     // If user provided menu file in input/includes and no config, copy over the file.
-    if (existsSync(menuXMLDefaultPath) && !this.pkg.config?.menu) {
+    if (existsSync(menuXMLDefaultPath) && !this.config.menu) {
```
ngfreiter, line 583: Good catch on these!

src/import/ensureConfiguration.ts
```diff
@@ -283,6 +283,45 @@ function generateConfiguration(root: string, allowFromScratch: boolean): string
...
+    !fs.existsSync(inputIndexMarkdownPagesPath) &&
+    !fs.existsSync(inputIndexXMLPagesPath)
+  ) {
+    // no index file found, create our default
+    setPairs.push(
+      new YAMLPair('indexPageContent', packageJSON.description ?? '')
+        .withCommentBefore(
+          getBoxComment(
+            'index.md',
+            'To use a provided index.md or index.xml file, delete the "indexPageContent" property below.'
```
ngfreiter, line 316: I think here it would be good to specify the location of the `index.md` file, aka `ig-data/input/pagecontent/index.[md | xml]`.

ngfreiter:
Oh and in response to your bullet points, my feelings are:
* No need
* We've made other functions along the export line public so it doesn't really bother me to do it for those as well.

src/ig/IGExporter.ts
jafeltra: No love for the Windows users.

But also, yeah of course.

src/import/ensureConfiguration.ts
jafeltra: I can add that!

src/ig/IGExporter.ts
jafeltra: Yes, that's correct. But I don't have much of a reasoning beyond that is where we were putting the generated file before, which really isn't much of a great answer..

src/ig/IGExporter.ts
jafeltra: There actually is one that I think fulfills this case, it's just up on line 262 rather than at the end. Is that log what you were thinking? If so, I can move it down to mirror the other cases.

test/ig/IGExporter.index.test.ts
```diff
@@ -0,0 +1,272 @@
...
+      expect(fs.existsSync(igPath)).toBeTruthy();
+      const igContent = fs.readJSONSync(igPath);
+      expect(igContent.definition.page.page).not.toContainEqual({
+        nameUrl: 'index.html',
+        title: 'Home',
+        generation: 'markdown'
+      });
+
+      // No warnings or errors logged
+      expect(loggerSpy.getAllMessages()).toHaveLength(loggerSpy.getAllMessages('info').length);
```
mint-thompson, line 59: I'm going to have to remember this one for later.

src/ig/IGExporter.ts
ngfreiter: That's good enough for me though.

src/ig/IGExporter.ts
ngfreiter: Oh yeah oops, no need to move!

jafeltra: Now I want them to match though, so I'm going to move it.

test/ig/IGExporter.index.test.ts
cmoesel: Or we could add a method to loggerSpy: `isClean()` (or something like that).

PR #427, opened 2020-05-13
Export config.templates to IG.definition.template

cmoesel:
Also... a few tweaks to affect order of properties in IG JSON...


cmoesel:
Yeah, it kinda bugged me when I did regression tests w/ mCODE and the only difference between the old and new version was that `dependsOn` was at the bottom of the file in the new version.  So, I figured, why not just make it do what I want it to do. ;-)

PR #429, opened 2020-05-13
Configuration-based IGExporter integration tests

mint-thompson:
Completes task [CIMPL-361](https://standardhealthrecord.atlassian.net/browse/CIMPL-361)

These tests begin with reading in configuration from a yaml file and test to make sure that expected default and configured content is produced in output files. Existing yaml files are used in these tests.

### DIFF
--- a/test/ig/IGExporter.test.ts
+++ b/test/ig/IGExporter.test.ts
@@ -5,12 +5,14 @@ import { Package } from '../../src/export';
 import { IGExporter } from '../../src/ig';
 import { importConfiguration } from '../../src/import';
 import { FHIRDefinitions, loadFromPath } from '../../src/fhirdefs';
+import { loggerSpy } from '../testhelpers';
 
 describe('IGExporter', () => {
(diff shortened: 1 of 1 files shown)


ngfreiter:
> The one odd thing I noticed is that there are info/error messages logging in my terminal. Is that normal and I'm just blanking? Or are these new to these tests? I feel like they're new, but I might be forgetting something obvious.

I'm also seeing the logs! Should've ran the tests locally, now I will continue to have GitHub Actions trust issues.

mint-thompson:
Correct: there are lots of errors getting logged due to missing files (and there are tests for these messages in `IGExporter.IG.test.ts`), since the tests don't actually load any of the files specified in the configuration.

cmoesel:
Can you load the logger spy to suppress the logging messages when the tests get run?

ngfreiter:
I think though we don't want the logs to be outputted to the terminal when we run `npm run test`. And in this case they do. I believe it is because logger isn't loaded

PR #454, opened 2020-05-20
Validate fixed resources

### LINKED ISSUE #395: SUSHI should validate fixed references
This originally came about from zulip here: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Slicing.20by.20profile.

The idea is that when Instance 1 has a `Reference` type element, and it fixes that `Reference` type element to a `Reference` to Instance 2, we should check that Instance 2 has a matching type. So if the `Reference` is `Reference(Practitioner | Organization)`, we should not allow Instance 2 to be a Patient.

However, we probably still want to allow `Reference`s to Instances that don't exist. This is only in the case where the `Reference` is resolved to an Instance that is defined in FSH.

jafeltra:
Fixes #395 

This PR validates references when fixing them to a value. For simplicity, I added an `sdType` property to the `FshReference` in order to have easier access to the type when checking if it is a valid type. The first commit (1f57717) adds two tests that try to add invalid references which should fail on that commit because there was nothing stopping you from adding the invalid reference. The second commit (27f3902) adds functionality to prevent invalid types from being added and all tests should pass.


test/export/InstanceExporter.test.ts
```diff
@@ -924,6 +924,24 @@ describe('InstanceExporter', () => {
       });
     });
 
+    it('should log a warning when an invalid reference is fixed', () => {
```
mint-thompson, line 927: This situation logs an error, not a warning.

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -1432,6 +1432,27 @@ describe('StructureDefinitionExporter', () => {
     });
   });
 
+  it('should not apply a Reference FixedValueRule with invalid type and log an error', () => {
```
mint-thompson, line 1435: The test description says that this logs an error, but nothing in the test case checks to see if an error is logged.

test/export/InstanceExporter.test.ts
jafeltra: I changed the wording of this! I had written the test before I decided which I was going to log. Oops!

test/export/StructureDefinitionExporter.test.ts
jafeltra: Added the check for the error! Good catch!

jafeltra:
I think this PR should be ready to be reviewed again. I updated the way I checked for valid references so that even if a reference is based on a profile, it checks if the type is a valid type, even if it isn't specifically based on the profile referenced.

test/export/InstanceExporter.test.ts
```diff
@@ -924,6 +924,64 @@ describe('InstanceExporter', () => {
...
+      const fixedRefRule = new FixedValueRule('subject');
+      fixedRefRule.fixedValue = new FshReference('Bar');
+      profiledInstance.rules.push(fixedRefRule);
+      doc.instances.set(profiledInstance.name, profiledInstance);
+      doc.instances.set(patientInstance.name, patientInstance);
+      const exported = exportInstance(profiledInstance);
+      expect(exported.subject).toEqual({
+        reference: 'Patient/Bar'
+      });
+    });
```
ngfreiter, line 944: I am not sure I understand what this test is up to, but that could definitely just be me not understanding. So could you give an example of what this would look like in FSH?

ngfreiter: Ok wait I think I've figured it out. But two things I don't understand. One, can't we just directly set the `instanceOf` on the `profiledInstance` to be `http://hl7.org/fhir/us/core/StructureDefinition/us-core-observation-lab`? Instead of creating `profiledObservation`. And two, what is the function of `doc.instances.set(patientInstance.name, patientInstance)`?

test/export/InstanceExporter.test.ts
```diff
@@ -924,6 +924,64 @@ describe('InstanceExporter', () => {
...
+      loggerSpy.reset();
+      const profiledObservation = new Profile('USCoreObservation');
+      profiledObservation.parent =
+        'http://hl7.org/fhir/us/core/StructureDefinition/us-core-observation-lab';
+      doc.profiles.set(profiledObservation.name, profiledObservation);
+
+      const profiledInstance = new Instance('MyExampleObservation');
+      profiledInstance.instanceOf = 'USCoreObservation';
+      const fixedRefRule = new FixedValueRule('subject');
+      fixedRefRule.fixedValue = new FshReference('Pow'); // Pow is instanceOf Bundle and not an allowed subject reference type
```
ngfreiter, line 974: Isn't `Bundle` not allowed on an `Observation.subject` anyway? Would it make more sense here to test something that is allowed on `Observation.subject` but is then disallowed by the profile? To me it looks like this would be `Group`, `Device`, or `Location`.

test/export/InstanceExporter.test.ts
```diff
@@ -924,6 +924,64 @@ describe('InstanceExporter', () => {
...
+      profiledInstance.rules.push(fixedRefRule);
+      doc.instances.set(profiledInstance.name, profiledInstance);
+      doc.instances.set(patientInstance.name, patientInstance);
```
ngfreiter, line 977: I am confused here too about what `patientInstance` does in this context.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1162,6 +1162,15 @@ export class ElementDefinition {
...
         value = value as FshReference;
+        if (value.sdType) {
+          let lineage: Metadata[] = [];
+          this.type.forEach(t =>
+            t.targetProfile.map(tp => (lineage = lineage.concat(this.getTypeLineage(tp, fisher))))
+          );
+          if (!lineage.find(md => md.sdType === (value as FshReference).sdType)) {
+            throw new InvalidTypeError(`Reference(${value.sdType})`, this.type);
+          }
+        }
```
ngfreiter, line 1173: I'm not sure this is handled correctly when it comes to inheritance. Easiest to explain with an example:
```
Instance: MyObservation
InstanceOf: Observation
* subject = Reference(MyResource)

Instance: MyResource
InstanceOf: Resource
* id = "hmm"
```
The FSH above will not get any `InvalidTypeError`s in my testing, but I think it should. The `subject` field is supposed to have a Reference to a Patient, Group, Device, or Location. Resource is a parent of these four, not a child, so shouldn't it be an invalid type here? 

I honestly may just be overly confusing myself, so definitely double check me on this. But it seems like the type checking is almost happening backwards here. I think we would want to get the lineage of the `value.sdType`. Then if anything in the lineage of that `sdType` matches something in `this.type`, then we are good, because that means the `value.sdType` is, or is a child of, something in `this.type`.

jafeltra: Hm this is a good question. I sort of was thinking that the example of `* subject = Reference(MyResource)` would be allowed. I talked to @ngfreiter about this and we're still not sure of what is allowed in FHIR. @cmoesel do you know if this should be allowed?

test/export/InstanceExporter.test.ts
jafeltra: You're right, I'll switch to just directly set the `instanceOf` and I think the `patientInstance` line was a copy and paste mistake so I'll remove it. I'll also see if I can maybe add some comments to clarify what the test is doing to make it easier to understand.

test/export/InstanceExporter.test.ts
jafeltra: That's a good idea. I had added this test knowing that `Bundle` was not allowed but thought it made sense for completeness sake, but I think you're right that it can test one of those other three types. I'll update.

test/export/InstanceExporter.test.ts
jafeltra: Nothing. I think it's a copy and paste error. I'll remove.

src/fhirtypes/ElementDefinition.ts
cmoesel: I can see why you might think an instance of `Resource` would be fine, since `Patient`/`Group`/`Device` are all `Resource`s and any data in a `Resource` instance would be valid in a `Patient`/`Group`/`Devices` instances.  But... there are cases where I think this could be problematic.

For example, imagine that `subject` could only be a `Group` (maybe we profiled it that way).  A `Group` is a `Resource`, but `Group` _requires_ `1..1 type` and `1..1 actual` -- so an instance of `Resource` is _not_ a valid instance of `Group` (since `Resource` does not have `type` or `actual` fields).

Similarly, `Resource` has a few fields that _could_ be profiled out by a sub-type.  So what if a sub-type profiled out `language` to be `0..0`?  If something was supposed to be a reference to that sub-type, then some references to `Resource` would be invalid instances of the sub-type (if they had values for `language`).

So... I think Nick is probably on the right path about getting the lineage of `value.sdType` and checking if anything in there is a match for the declared type of the field.

jafeltra: Okay that makes sense! That's what we were going back and forth on. In that case, I'll update to support this and add a test to check that it works!

jafeltra:
I updated some of the tests and added comments to address @ngfreiter's comments and hopefully make the intent of the tests clearer. I also updated the approach for referencing parent types, so the example that Nick specified above will now throw an InvalidTypeError.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1162,6 +1162,17 @@ export class ElementDefinition {
         break;
       case 'Reference':
         value = value as FshReference;
+        if (value.sdType) {
+          const validTypes: string[] = [];
+          this.type.forEach(t =>
+            t.targetProfile.map(tp => validTypes.push(fisher.fishForMetadata(tp).sdType))
```
ngfreiter, line 1168: Although it would be weird, I think it is possible that the `targetProfile` is something that the `fisher` fails to find, if that is the case here, we'd get an undefined error instead of a nicer InvalidTypeError. So I think adding some `?` here could help that. 

Also sorry for this maybe pedantic comment, but wouldn't this line make more sense as a `forEach` than a `map`?

jafeltra: I updated the inner loop here to use a `forEach` and to only add types that are found.

jafeltra:
@ngfreiter 806a81f addresses your last set of comments. I updated the loop and added a test that checks both of the cases you described by checking that a profiled Patient instance can be fixed to the DocumentReference.context.related field, which is a Reference to `Any`.

PR #457, opened 2020-05-21
Clean up StructureDefinition.extension during export

### LINKED ISSUE #412: Setting an extension on a StructureDefinition
We're trying to set the value of an extension on the underlying StructureDefinition for a profile we're defining by using the "escape syntax" (^):
```
Profile: NewOnsetOfT2DMDataElementCPG3
Parent: Observation
* ^extension[inferenceExpression].valueExpression.language = #text/cql
* ^extension[inferenceExpression].valueExpression.expression =
  "exists([Condition in \"T2DMConditionVS\"] c where c.onset is within last 2 months)"
* code = ExampleCdsCS#t2dm
```

Which builds fine, but it seems to be treating the `.valueExpression.language` part as constraints, rather than instance assignments...
```
"extension": [
    {
      "_sliceName": "inferenceExpression",
      "valueExpression": {
        "language": {
          "fixedValue": "text/cql",
          "_primitive": true
        },
        "expression": {
          "fixedValue": "exists([Condition in \"T2DMConditionVS\"] c where c.onset is within last 2 months)",
          "_primitive": true
        }
      }
    }
  ]
```

mint-thompson:
Fixes #412 and completes [CIMPL-384](https://standardhealthrecord.atlassian.net/browse/CIMPL-384)

When rules are applied to a StructureDefinition's `extension`, they should look the same way that they would as if those rules were on an instance: that is, they should look like regular values, and not include the extra properties that get added during processing.

Because cleaning is now being applied to StructureDefinitions as well as Instances, this PR moves a lot of the cleaning-up functions to the shared `common.ts`.


cmoesel:
> but I can't imagine why someone would need to put contained resources in an SD. So practically I want to appeal to @cmoesel here...

Funny you should ask.  If you had asked me a few days ago, I probably would have said the same.  But since then I have learned that in some cases, if people have a value set that will only be used in one SD, they inline it in the SD's `contained`!  WHAT?!?!

Also, if we are adding support for instances of datatypes, I expect that opens up a bunch more viable use cases.  For example `contact` and `useContext` are both fields with complex values that I could imagine someone wanting to define as re-usable instances.

ngfreiter:
I should've known they wouldn't let us get away with it! Well in that case I think that could be included in this PR, or it could be out of scope for this PR and something we should address later. I'd personally vote we don't do this yet. But @mint-thompson if you want to do it now I won't stand in the way.

cmoesel:
If this is a regression (i.e., it works on master but not in this PR) then it should be fixed in this PR.  Otherwise, I'm fine with making it a separate task if this PR adds value on its own without it (and I think it does).

mint-thompson:
I ran some tests of the cases y'all pointed out, and it seems like I have more work to do. This one worked on master:
```
Profile: NewOnsetOfT2DMDataElementCPG3
Parent: Observation
* ^url.id = "my-id"
```
The resulting StructureDefinition has an element at `_url` with value `{id: "my-id"}`. So, I'll need to account for this one in this PR.

However, inlining resources did not work on master:
```
Profile: NewOnsetOfT2DMDataElementCPG3
Parent: Observation
* ^contained = myResource
```
This gives an error of trying to fix a `string` to a `Resource`.

mint-thompson:
Ah, I misunderstood your comment. Thank you for clarifying. I'll take a look at toJSON, but since I didn't change that function at all, I wouldn't expect the behavior to be different. For StructureDefinition, it only checks specific property names, so the underscored primitive-with-children elements all get lost.

ngfreiter:
I made an issue here to track the task to add support for `^` working with contained resources and children of primitives: https://standardhealthrecord.atlassian.net/browse/CIMPL-397. Made it one task for now, but it can be split into two tasks later on when we do it if we decide to split.

PR #461, opened 2020-05-22
Sanitize ids that are valid names

### LINKED ISSUE #441: Should SUSHI default id to name even when it is an invalid id
If authors don't provide an explicit `Id:` for profiles, extensions, value sets, or code systems, SUSHI defaults the `id` to be the same as the `name`.  It's possible, however, that the name contains characters that are invalid for an id.  In this case, we should probably do one of the following:
* sanitize the id to remove invalid characters
* log an error indicating that the id cannot safely default to the name value

If we sanitize the id, we could optionally log a warning or info indicating that we did that.

This is probably something that should be governed by the spec.  @markkramerus -- do you have an opinion on this?  I'd lean toward sanitizing the id and logging an info message -- but I don't have a strong opinion.

(For reference, see: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Naming.20issues)

ngfreiter:
Fixes #441.

Now when a FHIR id is specified that is an invalid id, but a valid name, that name will be turned into an id by replacing `_` with `-` and by truncating to 64 characters. A warning will be emitted, since the users really should just specify the id directly if they are running into this case. If the id is not a valid name, then we can't be sure we can turn it into a valid id, so we do nothing. This will happen even if the id is not being set via a name. Aka even if the user does:
```
CodeSystem: CodeSystem_MaritalStatus
Id: My_Invalid_Id
```

I thought this would be fine, since the old approach would be to log an error and export using `My_Invalid_Id`. In this approach, we would fix `My_Invalid_Id` to be `My-Invalid-Id`, which is not a valid id, and only log a warning. Either way the user will probably want to fix it, but this way the user at least gets a valid id in their output.

### DIFF
--- a/test/export/CodeSystemExporter.test.ts
+++ b/test/export/CodeSystemExporter.test.ts
@@ -165,6 +165,39 @@ describe('CodeSystemExporter', () => {
     expect(loggerSpy.getLastMessage('error')).toMatch(/File: Strange\.fsh.*Line: 3 - 8\D*/s);
   });
 
+  it('should sanitize the id and log a message when a valid name is used to make an invalid id', () => {
+    const codeSystem = new FshCodeSystem('Not_good_id')
+      .withFile('Wrong.fsh')
+      .withLocation([2, 8, 5, 18]);
+    doc.codeSystems.set(codeSystem.name, codeSystem);
+    const exported = exporter.export().codeSystems;
+    expect(exported[0].name).toBe('Not_good_id');
+    expect(exported[0].id).toBe('Not-good-id');
+    expect(loggerSpy.getLastMessage('warn')).toMatch(
+      /The string "Not_good_id" represents a valid FHIR name but not a valid FHIR id.*The id will be exported as "Not-good-id"/s
+    );
+    expect(loggerSpy.getLastMessage('warn')).toMatch(/File: Wrong\.fsh.*Line: 2 - 5\D*/s);
+  });
+
+  it('should sanitize the id and log a message when a long valid name is used to make an invalid id', () => {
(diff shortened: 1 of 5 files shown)


test/export/CodeSystemExporter.test.ts
```diff
@@ -165,6 +165,39 @@ describe('CodeSystemExporter', () => {
...
+  });
+
+  it('should sanitize the id and log a message when a long valid name is used to make an invalid id', () => {
+    let longId = 'Toolong';
+    while (longId.length < 65) longId += 'longer';
+    const codeSystem = new FshCodeSystem(longId).withFile('Wrong.fsh').withLocation([2, 8, 5, 18]);
+    doc.codeSystems.set(codeSystem.name, codeSystem);
+    const exported = exporter.export().codeSystems;
+    expect(exported[0].name).toBe(longId);
+    expect(exported[0].id).toBe(longId.slice(0, 64));
```
mint-thompson, line 189: For the tests with long names that get shortened to make them valid ids, you could store the expected id after slicing once. It's not a particularly expensive operation to do twice, but I think it would make it easier to read the expected warning.

ngfreiter: Ok, will do!

ngfreiter:
> Something I thought of when changing ids: does SUSHI check to make sure that two resources of the same type have different ids? If so, that check would need to come after possibly changing the id to make it valid.

I don't think SUSHI yet has a check like this. This would fall under the umbrella of https://standardhealthrecord.atlassian.net/browse/CIMPL-375 right? Which is in progress.

Issue #391, opened 2020-04-24
Inability to create example data referencing sub extension


bwalsh:
## Problem

Following guide https://build.fhir.org/ig/HL7/fhir-shorthand/reference.html#extension-paths
and comments in #108 and #136 

All attempts to access the sub-extension fail with `Cannot resolve element from path:`
See AnvilResearchStudyExample below

## Input 

```

Extension:      AnvilDocumentReference
Id:             anvil-document-reference
Title:          "AnVIL document reference"
Description:    "An association to a document."
* value[x] only DRSAttachment


Profile:        AnvilResearchStudy
Parent:         ResearchStudy
Id:             anvil-research-study
Title:          "AnVIL ResearchStudy"
Description:    "An AnVIL ResearchStudy."
* extension contains 
    AnvilDocumentReference named attachments 0..*

Profile:        AnvilResearchSubject
Parent:         ResearchSubject
Id:             anvil-research-subject
Title:          "AnVIL ResearchSubject"
Description:    "An AnVIL ResearchSubject."
* extension contains 
    AnvilDocumentReference named attachments 0..*


Profile:        AnvilSpecimen
Parent:         Specimen
Id:             anvil-specimen
Title:          "AnVIL Specimen"
Description:    "An AnVIL Specimen."
* extension contains 
    AnvilDocumentReference named attachments 0..*


Instance: DRSAttachmentExampleMinimal
InstanceOf: DRSAttachment
Description: "An example representation of a DRSAttachment"
* extension[drs].id = "any-fhir-attachment-id"
* extension[drs].extension[id].valueString = "any-drs-id"
* extension[drs].extension[name].valueString = "any-file-name"
* extension[drs].extension[self_uri].valueString = "drs://url-here"
* extension[drs].extension[created_time].valueDateTime = "1985-04-12T23:20:50.52Z"
* extension[drs].extension[updated_time].valueDateTime = "1985-04-12T23:20:50.52Z"
* extension[drs].extension[size].valueInteger = 12345
* extension[drs].extension[version].valueString = "0.0.0"
* extension[drs].extension[mime_type].valueString = "application/json"
* extension[drs].extension[checksums].extension[checksum].valueString = "abcdef0123456789"
* extension[drs].extension[checksums].extension[type].valueString = "etag"
* extension[drs].extension[access_methods].extension[type].valueString = "s3"
* extension[drs].extension[access_methods].extension[access_url].valueString = "s3://some-url-here"
[...]

PR #464, opened 2020-05-28
Instances of Non-Resources

### LINKED ISSUE #385: Problem w/ example: 1)Unrecognised property '@meta'   2)Inconsistency with IG
## QA Error
image

## Input

```
Profile:        DRSAttachment
Parent:         Attachment
Id:             drs-attachment
Title:          "DRS Attachment"
Description:    "An attachment with DRS Object attributes."
// https://github.com/ga4gh/data-repository-service-schemas/blob/master/openapi/data_repository_service.swagger.yaml#L190-L304

// adds DRSObject to Attachment
* extension contains DRSObject named drs 0..1

// remove hash from Attachment definition
* hash 0..0

// Make DRSObject MS
* extension[DRSObject] MS

Extension:  DRSObject
Id: drs-object
Title: "DRS Object"
Description: "The drs object"
* extension contains
    id 1..1 MS and
    name 0..* 

// inline definition of sub-extensions
* extension[id] ^short = "An identifier unique to this `DrsObject`."
* extension[id].value[x] only string
* extension[name] ^short = "A string that can be used to name a `DrsObject`."
* extension[name].value[x] only string


Instance: DRSAttachmentExample
InstanceOf: DRSAttachment
* extension[drs-object].extension[id].valueString = "any-id"
* extension[drs-object].extension[name].valueString = "any-file-name"

```

## Output
 build/input/examples/Attachment-DRSAttachmentExample.json

```
{
  "resourceType": "Attachment",
  "id": "DRSAttachmentExample",
  "meta": {
    "profile": [
      "http://theanvil.io/fhir/anvil/StructureDefinition/drs-attachment"
    ]
  },
  "extension": [
    {
      "extension": [
        {
          "url": "id",
          "valueString": "any-id"
        },
        {
          "url": "name",
          "valueString": "any-file-name"
        }
      ],
      "url": "http://theanvil.io/fhir/anvil/StructureDefinition/drs-object"
    }
  ]
}
```
As you can see, the meta tag QA is complaining about is present.

## IG

**However**, if we navigate to the IG artifact summary, we see this.  Note that meta tag is missing.

image

ngfreiter:
Fixes #385 and #391 

This adds support for creating Instances of non-resources. Originally, it was possible to create instances of non-resources, but they would incorrectly have a `resourceType` element, and they would be incorrectly exported to JSON files. Then we realized this was wrong, so we added an error and prevented non-resources from being created at all. 

This PR removes that error message, and adds the ability to add non-resources which do not incorrectly have a `resourceType`. A non-resource must always be `#inline`, because they should only exist within an instance of a resource. If the non-resource instance is declared and is non `#inline`, a warning is logged. Besides not having a `resourceType` and having to be `#inline`, a non-resource instance can be generated in the same way as a normal resource instance.

I've added several tests, but to further test, the example in #391 gives a good test case. When you run that example on master, you should see an error like:
```
Instance DRSAttachmentExample of type Attachment is not an Instance of a Resource or Profile of a Resource. Only Instances of Resources or Profiles of Resources may be assigned to other Instances.
```
And an `Attachment-DRSAttachmentExample.json` file will incorrectly be generated. On this branch, you should only see a warning about DRSAttachmentExample being inline (which goes away if you specify `Usage: #inline`), and the DRSAttachmentExample will be correctly inlined, without any DRSAttachmentExample.json file being generated.


src/fhirtypes/ElementDefinition.ts
```diff
@@ -1368,22 +1367,22 @@ export class ElementDefinition {
...
-      const resource = fisher.fishForFHIR(value.resourceType, Type.Resource);
-      if (resource) {
-        // In this case the value is actually a resource, but not one that inherits from this.type
-        const typesString = this.type?.map(t => t.code).join(', ');
-        throw new MismatchedTypeError(value.resourceType, value.id, typesString);
-      } else {
-        // In this case the value is not actually a resource
-        throw new FixingNonResourceError(value.resourceType, value.id);
-      }
+      // In this case the value is actually a resource, but not one that inherits from this.type
```
cmoesel, line 1380: Now it's not necessarily a _resource_, right?

ngfreiter: It wasn't a big deal, but I couldn't help fixing it now that I see it. So it's fixed in https://github.com/FHIR/sushi/pull/464/commits/c14eea953273941b738d1421af2192da8c56b20e, sorry to dismiss your review 😬

PR #465, opened 2020-05-28
Enforce uniqueness for FSH names and FHIR ids

### LINKED ISSUE #152: SUSHI should validate that names and IDs are unique
Within an IG / FSH Tank, names and ids should be unique.  Technically I think that uniqueness is only required per type by FHIR (e.g., you *could* have a profile and extension w/ the same name) --- but for now I think we should be more stringent in FSH.  I'm not sure SUSHI would handle duplicate names/ids well, even if they were for different types.

So, SUSHI should issue an error when encountering definitions that introduce a name or id that has already been used in the FSHTank.

mint-thompson:
Fixes #152 and completes [CIMPL-375](https://standardhealthrecord.atlassian.net/browse/CIMPL-375).

FSH entities are tracked by name, so a reused name for the same entity type is an error. FHIR resources must have [unique ids](https://www.hl7.org/fhir/resource.html#id), so a reused id for the same FHIR type is an error. Mappings are not resources, but multiple mappings with the same id on the same StructureDefinition is confusing, so that is also an error.

The checks for imported FSH entities occur during import, and entities that reuse a name are skipped. This is different than the previous behavior, where the new entity would overwrite the previous entity. Because parsing is skipped for the new entity, errors that would have occurred while parsing it are never actually encountered.

The checks for exported FHIR resources are mostly similar to one another, but there are small differences based on when exactly the resource is added to the package.


src/export/InstanceExporter.ts
```diff
@@ -264,6 +264,24 @@ export class InstanceExporter implements Fishable {
...
+
+    // check for another instance of the same type with the same id
+    // see https://www.hl7.org/fhir/resource.html#id
+    // instanceDef has already been added to the package, so it's fine if it matches itself
+    if (
+      this.pkg.instances.some(
+        instance =>
+          instanceDef.resourceType === instance.resourceType &&
+          instanceDef.id === instance.id &&
+          instanceDef !== instance
```
ngfreiter, line 276: I think this could become a problem with #464. In that PR, we are adding support for instances of Extensions and Types (like a Quantity or CodeableConcept). But if the user is making an instance of a Quantity, I did not want to force them to put an id on that instance, since they likely don't want an id by default, and that instance is not actually being exported to JSON. So in that case I think two inline instances of Quantity which both would not have a `resourceType` and would likely not have an `id` (unless the user chose to set one for some reason) could fall into this case, both those fields on both instances would be null.

So I think here you want to exclude Inline instances, since they are not standing alone as resources.

src/export/StructureDefinitionExporter.ts
```diff
@@ -484,6 +484,21 @@ export class StructureDefinitionExporter implements Fishable {
...
     structDef.inProgress = false;
 
+    // check for another structure definition with the same id
+    // see https://www.hl7.org/fhir/resource.html#id
+    // the structure definition has already been added to the package, so it's fine if it matches itself
+    if (
+      this.pkg.profiles.some(profile => structDef.id === profile.id && structDef !== profile) ||
+      this.pkg.extensions.some(
+        extension => structDef.id === extension.id && structDef !== extension
+      )
```
ngfreiter, line 494: I don't think you can just search the `pkg` in this case. Because it is possible that I am using the bring your own JSON functionality to add an existing StructureDefinition with id `fish-patient`. Then I could declare a Profile in my FSH with id `fish-patient`. Based on my testing, that does not get caught, and the FSH exported profile is silently overwritten by the BYO-JSON profile. I think this could also be a problem in the ValueSetExporter, CodeSystemExporter, and InstanceExporter as well.

src/import/FSHImporter.ts
```diff
@@ -352,8 +373,15 @@ export class FSHImporter extends FSHVisitor {
     const valueSet = new FshValueSet(ctx.SEQUENCE().getText())
       .withLocation(this.extractStartStop(ctx))
       .withFile(this.currentFile);
-    this.parseValueSet(valueSet, ctx.vsMetadata(), ctx.vsComponent(), ctx.caretValueRule());
-    this.currentDoc.valueSets.set(valueSet.name, valueSet);
+    if (this.currentDoc.valueSets.has(valueSet.name)) {
+      logger.error(`Skipping value set: a value set named ${valueSet.name} already exists.`, {
```
ngfreiter, line 377: I think we should consider calling this a `ValueSet` in the importer. Same would go for all of the other entities in FSH, I think it makes sense to refer to them in the message by the keyword that the user is seeing. On the export side I could go either way, but on the import side when we are actually referring the the FSH entities, I think the keyword makes sense.

src/export/InstanceExporter.ts
```diff
@@ -281,6 +281,26 @@ export class InstanceExporter implements Fishable {
...
     cleanResource(instanceDef);
+
+    // check for another instance of the same type with the same id
+    // see https://www.hl7.org/fhir/resource.html#id
+    // instanceDef has already been added to the package, so it's fine if it matches itself
+    // inline instances are not written to their own files, so those can be skipped in this
+    if (
+      this.pkg.instances.some(
+        instance =>
+          instance._instanceMeta.usage !== 'Inline' &&
```
ngfreiter, line 292: I'm wondering about one more case:
```
Instance: MyPatient1
InstanceOf: Patient
* id = "testing"
* name.family = "hello"

Instance: MyPatient2
InstanceOf: Patient
Usage: #inline
* id = "testing"
* name.family = "hello there"
```
In this case, `MyPatient2` gets flagged for having the same id as `MyPatient1`. I *think* that this should not be the case right? Since `MyPatient2` would not be existing as its own entity. You'd have to be a pretty cruel user to go around doing this and then getting mad that it doesn't work, so sorry for this edge case on an edge case. But what do you think? Should this case still be an error or not?

mint-thompson: This case should not be an error, since there will only be one resource on the server with that id, meaning that the FHIR rules about id uniqueness is satisfied. And importantly, it currently does not throw an error when the instances are in the opposite order. I'll handle this case and add a test.

src/export/StructureDefinitionExporter.ts
cmoesel: Allowing BYOJ to overwrite FSH-generated definitions is currently an intentional feature.  This gives authors a way to "patch" files that SUSHI might generate incorrectly.  The idea is that once SUSHI fixes whatever bug, the BYOJ patch can be removed and it's all good.  I'm not sure this is the best feature ever, but it's the way things are meant to work for now anyway.  So it's fine not to issue an error on that.

PR #469, opened 2020-06-01
Update CodeSystem count when appropriate

### LINKED ISSUE #419: Set the count on FSH-authored CodeSystems
If SUSHI can determine the count for a FSH-authored CodeSystem, it should set `count`.  The `content` property should probably be considered as SUSHI may not be able to set a valid count for CodeSystems whose content is  not `complete` (e.g., `fragment`).

cmoesel:
If the count is not explicitly set AND the content is "complete" AND there is at least one concept, SUSHI will set the count to the length of the concept array.

Fixes #419


src/export/CodeSystemExporter.ts
```diff
@@ -55,6 +55,18 @@ export class CodeSystemExporter {
     }
   }
 
+  private updateCount(codeSystem: CodeSystem): void {
+    // Only update the count when it isn't explicitly set, when the content is 'complete',
+    // and there is at least one concept.
+    if (
+      codeSystem.count == null &&
+      codeSystem.content === 'complete' &&
+      codeSystem.concept?.length
```
ngfreiter, line 64: I'm wondering if we should only do this if the `codeSystem.count` is null. Specifically what if the `codeSystem.content` is `complete`, and we can clearly count that there are 4 concepts, and the user enters 5 as a count. Should we correct that since we do know better? Since we are already inferring the count for them if they don't set it, it doesn't feel like a big stretch to expect we would fix things if they do set it and we are certain they set it wrong.

ngfreiter: Yeah that sounds like a good plan, that way the user has ultimate power to ignore the warning if they really feel they need to.

cmoesel: OK, updated in abbfe28.

src/export/CodeSystemExporter.ts
```diff
@@ -55,15 +55,24 @@ export class CodeSystemExporter {
...
+      if (codeSystem.count == null && actualCount != null) {
+        codeSystem.count = actualCount;
+      } else if (codeSystem.count !== actualCount) {
+        const countRule = fshDefinition.rules.find(
+          r => r instanceof CaretValueRule && r.caretPath === 'count'
+        );
+        const sourceInfo = countRule?.sourceInfo ?? fshDefinition.sourceInfo;
+        logger.warn(
+          `The user-specified ^count (${codeSystem.count}) does not match the specified number of concepts ` +
+            `(${actualCount}). If this is not a "complete" ValueSet, set the ^content property to the appropriate ` +
```
ngfreiter, line 71: Two comments on this line:
1. If they specify a count but no concepts, then `actualCount` could be undefined here, so I think if it is undefined you want to default to printing `0`, not `undefined`.
2. Shouldn't "ValueSet" be "CodeSystem"?

cmoesel: 1. Good point.  I didn't want to default `undefined` `actualCount` to `0` above because I want `undefined` `actualAccount` to `===` `undefined` `count`.  But when I print it, I _should_ default it to `0`.  Thanks.
2. Ugh.

cmoesel: 1176b64.  There.  Happy now? Or would you like to point out any other embarrassingly obvious issues with this PR? ;-)

ngfreiter: I would've loved to, but I couldn't find any 😞

PR #470, opened 2020-06-01
Only load current packages when newer version is available

ngfreiter:
Addresses https://standardhealthrecord.atlassian.net/browse/CIMPL-364.

We used to always load a new package to the cache if the version was specified as `#current`. Now, if we are loading a `#current` package, and that package is already cached, we try to be intelligent about if we need to download that package or not. Current packages are downloaded from build.fhir.org, and each current package has a `package.manifest.json` file which lists the `date` for that package. When loading a current package that is already locally cached, we will now compare the date on the local copy with the date on the `package.manifest.json` hosted on build.fhir.org. If the dates match, we do nothing. If the dates do not match, we trigger a download.

All other loading behavior should remain the same, but adding this change required a bit of refactoring, so the `loadDependency` code has changed quite a bit. Additionally, I realized that our tests for `loadDependency` could be a little more comprehensive, so I tried to flesh out those tests more in this PR.

I would suggest that reviewers also do a test run our two to ensure that loading is working as expected on their machine on this branch. The tests rely on mocking out certain functions, so it is nice to have further confirmation that things are working when the non-mocked functions are called.


ngfreiter:
>First, I noticed that the directory structure of "current" packages is different than the directory of versioned packages. The "current" packages put everything in a package subfolder. The versioned packages (e.g. hl7.fhir.us.core#3.1.0) only put JSON in package and other folders are siblings to package. I assume this is a known thing? And that if the IG Publisher downloads current, it does it the same way?

I don't think this is something that is consistent across all packages. From what I have found, spot checking some packages, most packages just have everything in the package subfolder. Which is the documented structure of packages: https://confluence.hl7.org/pages/viewpage.action?pageId=35718629#NPMPackageSpecification-Format.

So it kinda just seems like a weird issue with the 3.1.0 version of US Core hosted on packages.fhir.org. If you visit packages.fhir.org/hl7.fhir.us.core/3.1.0 you will download the package from there, and you can see it does indeed have directories *not* within the package subfolder. What is weird though is that when the IG Publisher downloads US Core 3.1.0, it gets a version that *does* follow the correct structure, and has everything nested in a package subfolder. I had been made to beleive when first implementing this that the IG Publisher also got its packages from packages.fhir.org, so I don't know if that has changed or I was under the wrong impression. But it seems to be specific to this US Core package.

cmoesel:
Hmmm... I guess this behavior is probably not new to this PR -- but I wonder if this might cause issues...  If the IG publisher somehow installs the (correct?) expected folder structure, but SUSHI installs the oddball folder structure, might the publisher have problems processing the cached US Core package that SUSHI installed?  So far, it seems like it hasn't been a problem, but I actually find that... a little surprising!

ngfreiter:
So I don't *think* it will be a problem because
1. It hasn't been a problem yet.
2. The definitions themselves are still in the same place, and isn't that all that should matter.

That being said it is a bit concerning, so I've done some further investigation, and the IG Publisher is doing something that it did not used to do. There is now a `packages-client` folder in my `.fhir` cache: 
image

And that folder contains the `.tgz` file from packages.fhir.org for hl7.fhir.us.core:
image

But then what is weird is that this zip file does have the incorrect directory structure, aka folders are at the same level as `package`, not nested within it. So is the IG Publisher converting this package into the proper form? If so, why is it doing that and not just requiring that packages are in the documented format? My investigations have not yet gone that far. But I don't think we want to get in the business of converting packages into the proper format if they are in some improper format.

You will also notice that I chose to keep actual dates out of the informational messages, and limit the number of messages. I feel like we are getting to a place where there is a bit of information overload coming from package loading, and I don't know if we want to add more info about why certain things are happening, when I imagine most users don't really care why stuff is happening, but what is actually happening. The other reason not to add dates is that the IG Publisher uses a weird non-documented format for its dates, so we are then tied to that format changing on a whim.

cmoesel:
Perhaps we should just raise it w/ Grahame to ensure we understand the issue and that our assumption (that it's harmless) is correct?  I notice that `examples` and `spec.internals` aren't in the `package` folder -- but as long as the IG Publisher never tries to load those for dependencies, I guess we're probably OK.

ngfreiter:
> Perhaps we should just raise it w/ Grahame to ensure we understand the issue and that our assumption (that it's harmless) is correct? I notice that `examples` and `spec.internals` aren't in the `package` folder -- but as long as the IG Publisher never tries to load those for dependencies, I guess we're probably OK.

Alright, we can keep an eye on this thread https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/dev.20dependencies for any further explanation of what is going on here.

ngfreiter:
@cmoesel I felt like I was too quick to think messages with a date wouldn't be helpful, so I added them as debug messages in https://github.com/FHIR/sushi/pull/470/commits/4d94a5aa8e59dafc1e88e3facf1772872421c3e5. That way people won't have to see them, but if they are confused about what is going on and use the debug flag, they can get more detailed info. But this way it's also easy for me to make them info if you feel strongly they should be, so let me know what you think.

ngfreiter:
@cmoesel Moment is no more in https://github.com/FHIR/sushi/pull/470/commits/97a340fecb2d468286d546112459594487a37109.

src/fhirdefs/load.ts
```diff
@@ -75,33 +74,33 @@ export async function loadDependency(
...
-              'YYYYMMDDHHmmss'
-            ).format(
-              'YYYY-MM-DDTHH:mm:ss'
-            )}) does not match last build date on build.fhir.org (${moment(
-              manifest?.data?.date,
-              'YYYYMMDDHHmmss'
-            ).format('YYYY-MM-DDTHH:mm:ss')})`
+            `Cached package date for ${fullPackageName} (${cachedPackageJSON.date?.replace(
+              dateRegex,
+              dateFormatString
```
cmoesel, line 85: You're probably ready for this PR to just be done, but I feel like it would be cleaner to put the regex and replacement logic in a `formatDate` function so that in the four places you use it, you can just do it like `${formatDate(cachedPackageJSON.date)}`.  Nit-picky, maybe -- but it would make these long debug statements easier to read.

ngfreiter:
Alright @cmoesel I added a formateDate function. Good suggestion, not nit-picky! 

There has also been a response on our relevant zulip thread. I'll let you read it for yourself. I'm not sure what is the best way forward. The way I see it we can either do nothing and wait to see if this ever is a problem, or we can align with the IG Publisher approach and fix packages that have the old format.

Issue #466, opened 2020-05-29
SUSHI sometimes generated invalid characters in `ImplementationGuide.dependsOn.id`


cmoesel:
While SUSHI replaces `.` with `_`, there are other disallowed characters that are not replaced.  Per Grahame, that id should be
> a..z, A..Z, 0..9, and _-. and it must start with a..z | A..Z

I'm not sure if `-` should really be allowed after all.  It seems like it could be misinterpreted, but I'm waiting on a reply for that.

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/SUSHI.200.2E12.2E7/near/199193333

cmoesel:
Update.  The recommended guidance has been updated to:
> a..z, A..Z, 0..9, and _ and it must start with a..z | A..Z

PR #472, opened 2020-06-02
More carefully validate ig.dependsOn.id

ngfreiter:
This fixes https://github.com/FHIR/sushi/issues/466. 

Now we will replace both `-` and `.` with `_`, and if the packageId does not start with `[a-zA-z]`, we make it start with `[a-zA-z]` by prepending `id_`. That is kind of arbitrary, but this `id` is just a token that is used interally, so I think it is ok to be a little arbitrary.

### DIFF
--- a/test/ig/IGExporter.test.ts
+++ b/test/ig/IGExporter.test.ts
@@ -172,7 +172,7 @@ describe('IGExporter', () => {
             packageId: 'hl7.fhir.uv.vhdir',
             version: 'current'
           },
-          // test-wonky-id-1 tests that we are correctly turning - in packageId to - in id
+          // test-wonky-id-1 tests that we are correctly turning - in packageId to _ in id
           {
             id: 'test_wonky_id_1',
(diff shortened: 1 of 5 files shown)


test/ig/IGExporter.test.ts
```diff
@@ -171,6 +171,20 @@ describe('IGExporter', () => {
             uri: 'http://hl7.org/fhir/uv/vhdir/ImplementationGuide/hl7.core.uv.vhdir',
             packageId: 'hl7.fhir.uv.vhdir',
             version: 'current'
+          },
+          // test-wonky-id-1 tests that we are correctly turning - in packageId to - in id
```
jafeltra, line 175: This is the smallest comment I have ever made, but I think the second `-` in this comment should be `_`

ngfreiter: I can't believe I've created such a flawed PR. Shame on me, shame!

PR #476, opened 2020-06-03
Menu Configuration Keywords

### LINKED ISSUE #440: Consider supporting additional menu features
Two common features of IG menus are:
* ability to launch in a new tab
* indicator for external web site

We could potentially support them via keywords like this:
```yaml
menu:
  Home: index.md
  FHIR Spec: external new-tab {{site.data.fhir.path}}index.html
```

Before we implement this, however, we should probably decide if the "menu" feature in the config.yaml beta will make it through to the mainline release.

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/SUSHI.200.2E13.2E0-beta.2E1/near/197489659

jafeltra:
This PR fixes #440 and addresses https://standardhealthrecord.atlassian.net/browse/CIMPL-393.

It adds support for two keywords in menu configuration. `new-tab` allows a user to configure that the menu item will open in a new tab (using `target="_blank"`) and `external` allows a user to add the external link icon in the menu item.

I also updated the default menu configuration that we add to include these two keywords on the FHIR Spec menu link to more closely reflect what the current `menu.xml` looks like. Any provided `menu.xml` that we try to translate into a configuration object will now look to use these keywords where applicable.

### DIFF
--- a/src/ig/IGExporter.ts
+++ b/src/ig/IGExporter.ts
@@ -653,15 +653,16 @@ export class IGExporter {
       menuItem += `${prefixSpaces}</li>${EOL}`;
     } else {
       menuItem += `${prefixSpaces}<li>${EOL}${prefixSpaces}${'  '}`;
-      if (item.url || item.openInNewTab) {
+      if (item.url) {
         menuItem += '<a ';
         if (item.openInNewTab) menuItem += 'target="_blank" ';
-        if (item.url) menuItem += `href="${item.url}"`;
-        menuItem += '>';
+        menuItem += `href="${item.url}">`;
       }
       menuItem += item.name;
-      if (item.isExternal) menuItem += '<img src="external.png" style="text-align: baseline"/>';
-      if (item.url || item.openInNewTab) menuItem += '</a>';
+      if (item.url) {
+        if (item.isExternal) menuItem += ' <img src="external.png" style="text-align: baseline"/>';
+        menuItem += '</a>';
+      }
       menuItem += `${EOL}${prefixSpaces}</li>${EOL}`;
     }
     return menuItem;
--- a/src/ig/IGExporter.ts
+++ b/src/ig/IGExporter.ts
@@ -653,9 +653,15 @@ export class IGExporter {
       menuItem += `${prefixSpaces}</li>${EOL}`;
     } else {
       menuItem += `${prefixSpaces}<li>${EOL}${prefixSpaces}${'  '}`;
-      if (item.url) menuItem += `<a href="${item.url}">`;
+      if (item.url || item.openInNewTab) {
+        menuItem += '<a ';
+        if (item.openInNewTab) menuItem += 'target="_blank" ';
+        if (item.url) menuItem += `href="${item.url}"`;
+        menuItem += '>';
+      }
       menuItem += item.name;
-      if (item.url) menuItem += '</a>';
+      if (item.isExternal) menuItem += '<img src="external.png" style="text-align: baseline"/>';
+      if (item.url || item.openInNewTab) menuItem += '</a>';
       menuItem += `${EOL}${prefixSpaces}</li>${EOL}`;
     }
     return menuItem;
--- a/src/import/ensureConfiguration.ts
+++ b/src/import/ensureConfiguration.ts
@@ -402,6 +402,11 @@ function getBoxComment(title: string, comment: string): string {
   return boxComment;
 }
 
+// Helper functions for converting menu XML file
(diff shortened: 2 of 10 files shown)


src/ig/IGExporter.ts
```diff
@@ -653,9 +653,15 @@ export class IGExporter {
       menuItem += `${prefixSpaces}</li>${EOL}`;
     } else {
       menuItem += `${prefixSpaces}<li>${EOL}${prefixSpaces}${'  '}`;
-      if (item.url) menuItem += `<a href="${item.url}">`;
+      if (item.url || item.openInNewTab) {
```
cmoesel, line 656: Why `|| item.openInNewTab`?  If there is no URL, then what is the point of opening in a new tab?  What do browsers even do in that case? Open a blank page?

src/ig/IGExporter.ts
```diff
@@ -653,9 +653,15 @@ export class IGExporter {
...
+        menuItem += '<a ';
+        if (item.openInNewTab) menuItem += 'target="_blank" ';
+        if (item.url) menuItem += `href="${item.url}"`;
+        menuItem += '>';
+      }
       menuItem += item.name;
-      if (item.url) menuItem += '</a>';
+      if (item.isExternal) menuItem += '<img src="external.png" style="text-align: baseline"/>';
+      if (item.url || item.openInNewTab) menuItem += '</a>';
```
cmoesel, line 664: As above, I'm not sure about adding a hyperlink when there might not be a URL (even if `new-tab` was specified).

src/import/ensureConfiguration.ts
```diff
@@ -453,26 +453,32 @@ function getMenuObjectFromMenuXML(menuXML: string): YAMLConfigurationMenuTree {
         if (li.a && !Array.isArray(li.a)) {
           const name = li.a._text;
           const link = li.a._attributes?.href;
+          const inNewTab = li.a._attributes?.target === '_blank' ? 'new-tab ' : '';
+          const isExternal = li.a.img?._attributes?.src === 'external.png' ? 'external ' : '';
+          const menuLinkWithKeywords = `${inNewTab}${isExternal}${link}`;
```
cmoesel, line 458: If you make functions instead, you can re-use them in the sub-menu too and have less duplicate code.  E.g.,
```ts
const inNewTab = (item: any): string => item.a._attributes?.target === '_blank' ? 'new-tab ' : '';
const isExternal = (item: any): string => item.a.img?._attributes?.src === 'external.png' ? 'external ' : '';
const menuLinkWithKeywords = `${inNewTab(li)}${isExternal(li)}${link}`;
```
_NOTE: I did `item: any` because I was too lazy to lookup what the real type is..._

src/ig/IGExporter.ts
```diff
@@ -653,9 +653,15 @@ export class IGExporter {
...
-      if (item.url) menuItem += `<a href="${item.url}">`;
+      if (item.url || item.openInNewTab) {
+        menuItem += '<a ';
+        if (item.openInNewTab) menuItem += 'target="_blank" ';
+        if (item.url) menuItem += `href="${item.url}"`;
+        menuItem += '>';
+      }
       menuItem += item.name;
-      if (item.url) menuItem += '</a>';
+      if (item.isExternal) menuItem += '<img src="external.png" style="text-align: baseline"/>';
```
cmoesel, line 663: I think we need a space before the image tag.  Right now, there is no space between the menu item text and the img tag, so it looks crowded:
image

Adding a space will improve that.  I looked at the [Sample IG](https://build.fhir.org/ig/FHIR/sample-ig/) and it puts a space before the image as well.

src/ig/IGExporter.ts
jafeltra: I was thinking that since URL wasn't strictly required to be present, I would support the weird case for if there was a new-tab keyword and no URL. If there is no URL, the menu item isn't clickable and you get the text cursor (the one that looks like an "I"). Since that doesn't actually get you any extra functionality, I can remove the extra check for `item.openInNewTab` and make the code simpler. I think a related change should also be to only add the `external` image if there is a URL specified as well.

src/import/ensureConfiguration.ts
jafeltra: That's a good idea and I added the functions. `item: any` actually seems to be the real type we have specified for the list items.

src/ig/IGExporter.ts
jafeltra: Added the space back!

Issue #463, opened 2020-05-28
SUSHI should not force un-sparse differentials


cmoesel:
In short, we should roll back the fix for #442, as we've found cases where *non-sparse* differentials cause errors -- and we've received confirmation that the IG Publisher *should* fully support sparse differentials.  So SUSHI should not jump through hoops to avoid that.

See: https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/The.20link.20.22XXX.22.20for.20.22coding.22.20cannot.20be.20resolved

PR #478, opened 2020-06-04
Bring back sparse differentials

ngfreiter:
Fixes https://github.com/FHIR/sushi/issues/463.

In https://github.com/FHIR/sushi/pull/453 we got rid of sparse differentials. Now sparse differentials are back! I mostly just removed the code and tests that were added in #453. So if you compare the contents of StructureDefinition.ts on this branch to the contents on the commit before #453 (https://github.com/FHIR/sushi/commit/b9bd6c78f377a83ff69833ed334b95022ea3e589), things should look very similar.


cmoesel:
I realize this is a rather unconventional thing to do, but in reviewing your PR, I realized that there are *some* things about the approach I took for sparse differentials that I kind of like -- and personally prefer over the old way I wrote that code.  So... I issued a PR against your PR with the newish approach minus the special code for sparse differentials.  All the tests still pass.  Let me know what you think.  If you don't like it better, that's OK.

See: #479

PR #479, opened 2020-06-04
Refactor approach to populating the diff elements

cmoesel:
Instead of applying multiple filters, just iterate through the elements. This is a slightly more straight-forward representation.


cmoesel:
See my note on #478.  This PR is for your consideration, as an alternate approach to fully reverting all of the sparse differential code.

cmoesel:
Doh!

cmoesel:
OK, I remembered to run `npm run check` this time!  3c36912

Also, that is a pretty awesome commit hash counting by threes at the end there.

cmoesel:
I intentionally left in the bad lint because I knew my next commit would be that awesome.

PR #484, opened 2020-06-09
Fix type overwrite error

### LINKED ISSUE #462: Setting ^type on a Profile breaks path resolution
In the following example:
```
Profile: MyBundle
Parent: Bundle
* ^type = "history"
* meta.extension contains myslice 0..1
```
The `StructureDefinition.type` for the `MyBundle` SD is getting set to `history`. However, in several places, we rely on the fact that the `type` for a Profile on Bundle should be Bundle. Setting the type to something other than Bundle causes SUSHI to fail to resolve paths for subsequent rules. Also the type of any SD has to come from this Value Set: http://www.hl7.org/fhir/valueset-defined-types.html, so `history` should not be allowed. I'm not sure of any use case as to why someone would make a Profile and then change the `StructureDefinition.type` on that Profile intentionally, so we should likely disallow this and log an error.

SamCitron:
Fixed bug where user is able to overwrite Parent's type with caret syntax. This fixes #462 an example for testing could be found in that issue.

### DIFF
--- a/src/errors/InvalidAccessError.ts
+++ b/src/errors/InvalidAccessError.ts
@@ -1,0 +1,5 @@
+export class InvalidAccessError extends Error {
+  constructor() {
+    super('Cannot directly change type. Elements will naturally inherit their Parent type.');
+  }
+}
--- a/src/fhirtypes/StructureDefinition.ts
+++ b/src/fhirtypes/StructureDefinition.ts
@@ -8,7 +8,8 @@ import {
   CannotResolvePathError,
   InvalidElementAccessError,
   MissingSnapshotError,
-  InvalidResourceTypeError
+  InvalidResourceTypeError,
+  InvalidAccessError
 } from '../errors';
 import {
   getArrayIndex,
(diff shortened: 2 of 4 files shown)


src/fhirtypes/StructureDefinition.ts
```diff
@@ -286,6 +287,10 @@ export class StructureDefinition {
     if (path.startsWith('snapshot') || path.startsWith('differential')) {
       throw new InvalidElementAccessError(path);
     }
+    const parentName = this.type || 'Resource';
```
ngfreiter, line 290: I don't think you need the `|| 'Resource'` part here. The `this` we are using in this context is a StructureDefinition. If you look at the documentation for StructureDefinition found here http://www.hl7.org/fhir/structuredefinition.html, you will see that type is 1..1:
image

So we can safely assume `this.type` is going to exist, and directly compare `value !== this.type` on the line below.

src/fhirtypes/StructureDefinition.ts
```diff
@@ -286,6 +287,10 @@ export class StructureDefinition {
...
+    if (path == 'type' && value != parentName) {
```
ngfreiter, line 291: Using `==` does work fine, but I think it is more conventional in this case to use `===` and `!==`. The triple equals checks for equality of value and type. The double equals can still return true even when things types don't match, for example, something like `'1' == 1` is true, but `'1' === 1` is false. In this case the distinction would probably never matter, but it's still probably better to be safe and use `===`.

src/errors/InvalidAccessError.ts
```diff
@@ -0,0 +1,5 @@
+export class InvalidAccessError extends Error {
+  constructor() {
+    super('Cannot directly change type. Elements will naturally inherit their Parent type.');
```
ngfreiter, line 3: Here I think we'd want to say "StructureDefinitions will naturally..." instead of "Elements will naturally...", since StructureDefinition is the FHIR terminology for what we are exporting.

PR #489, opened 2020-06-12
Use expected ig-data path even when it doesn't exist on the filesystem

### LINKED ISSUE #488: SUSHI errors and exits when `ig-data` folder is missing
SUSHI 0.13.0 introduced a new bug that causes SUSHI to exit with the following error if the `ig-data` folder is missing:

> error SUSHI encountered the following unexpected error: The "path" argument must be of type string. Received null

This appears to be related to the final merge of changes in _master_ (particularly the `app.ts` changes) and the `config-yaml` branch.

cmoesel:
Fixes #488 

The IGExporter expects a path to be passed in, even if the path does not exist on the file system. If it receives null, it throws.  Fix app.ts to never pass in a null path to the IGExporter.

Since we only get the path in one place, and since it is now a single line of code, we no longer have need for a separate getIgDataPath function.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -19,8 +19,7 @@ import {
   loadExternalDependencies,
   fillTank,
   writeFHIRResources,
-  getRawFSHes,
-  getIgDataPath
+  getRawFSHes
 } from './utils/Processing';
 
 app().catch(e => {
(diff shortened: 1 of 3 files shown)


cmoesel:
BTW -- couldn't easily write a test since we don't currently have a framework/harness for testing app.ts.  But... reproduce the error by running SUSHI 0.13.0 on a FSH Tank with _no_ ig-data folder.  It will fail. Then run this branch on the same FSH Tank.  It will succeed.

Issue #447, opened 2020-05-19
SUSHI exports choice specializations incorrectly


cmoesel:
`Observation.value[x]` allows the following choices: `Quantity`, `CodeableConcept`, `string`, `boolean`, `integer`, `Range`, `Ratio`, `SampledData`, `time`, `dateTime`, `Period`.

FSH allows authors to say this:
```
* value[x] only Duration
```

Duration is a *specialization* (not a *profile*) of `Quantity`.  This is something like a concrete subclass relationship.  Unlike profiles, specializations are in fact a distinct *type* in the FHIR type system.  SUSHI currently exports the rule above like so:
```json
{
  "id": "Observation.value[x]",
  "path": "Observation.value[x]",
  "type": [ { "code": "Duration" } ]
}
```

When I first implemented it, the above worked.  IG Publisher did not complain and it rendered `value[x]` as expected. (At least that's how I remember it).  I assumed it was legal.

It appears, however, that it is not actually legal and now the IG Publisher complains:
> illegal constrained type Duration from Quantity, CodeableConcept, string, boolean, integer, Range, Ratio, SampledData, time, dateTime, Period in http://hl7.org/fhir/StructureDefinition/Observation

The only proper way to support this is via a FHIRPath.  SUSHI needs to do one of the following.  Either:

* generate and insert an invariant using the valid FHIRPath to represent the constraint (in which case it will need to generate a unique constraint id)
* log an error and provide the user detailed guidance on how to add the invariant to their FSH

The first is more user friendly, but might do some unexpected things for unsuspecting users.  The second requires action from the user, but they become aware of the issue and also have control over its precise details (e.g., they can use an invariant id consistent w/ other invariant ids in their IG).

We also need to decide if we are going to further shield the user by allowing them to use the path `valueDuration` in rules even though it really should be `valueQuantity`.  I think we probably need to enforce `valueQuantity` because if they constrained a quantity to `Duration or Age`, it would not be valid to specify different constraints on `valueDuration` and `valueAge` at the same time.

cmoesel:
See: https://chat.fhir.org/#narrow/stream/179177-conformance/topic/Narrow.20Quantity.20to.20Duration.20and.20Age

ngfreiter:
Oh the GitHub magic doesn't work

PR #494, opened 2020-06-16
Log an error when constraining non-abstract type to specialization

ngfreiter:
A type that is not abstract cannot be constrained to any specializations of that type. When the type is not abstract, the type is final, and so it can only be constrained to profiles of that type. The best example is `Quantity`, which is not abstract, and has both specializations, such as `Duration`, and profiles, such as `SimpleQuantity`. So an element of type `Quantity`, say `valueQuantity`, can have the following rule:
```
* valueQuantity only SimpleQuantity
```
but cannot have:
```
* valueQuantity only Duration
```

SUSHI used to allow the second rule, but the IG Publisher does not, so it would have errors. Now the second rule is disallowed, and an error is logged. The justification for this is in https://chat.fhir.org/#narrow/stream/179177-conformance/topic/Narrow.20Quantity.20to.20Duration.20and.20Age. I don't know if there is a location in the spec that really dives into this distinction. The only example I can find of a non-abstract type being specialized is `Quantity`, but the code shouldn't be specific to `Quantity`.

### DIFF
--- a/test/export/StructureDefinitionExporter.test.ts
+++ b/test/export/StructureDefinitionExporter.test.ts
@@ -1595,7 +1595,7 @@ describe('StructureDefinitionExporter', () => {
     // Quantity is the first ancestor of Duration and Age
 
     // * value[x] only Duration or Age
-    // * value[x].comparator = #>
+    // * value[x].comparator = #>=
 
     const extension = new Extension('QuantifiedExtension');
 
--- a/test/export/StructureDefinitionExporter.test.ts
+++ b/test/export/StructureDefinitionExporter.test.ts
@@ -1595,22 +1595,21 @@ describe('StructureDefinitionExporter', () => {
     // Quantity is the first ancestor of Duration and Age
 
     // * value[x] only Duration or Age
-    // * value[x].comparator = #>=
+    // * value[x].comparator = #>
 
-    const profile = new Profile('QuantifiedObservation');
-    profile.parent = 'Observation';
+    const extension = new Extension('QuantifiedExtension');
 
     const onlyRule = new OnlyRule('value[x]');
     onlyRule.types = [{ type: 'Duration' }, { type: 'Age' }];
-    profile.rules.push(onlyRule);
+    extension.rules.push(onlyRule);
 
     const fixedValueRule = new FixedValueRule('value[x].comparator');
     fixedValueRule.fixedValue = new FshCode('>=');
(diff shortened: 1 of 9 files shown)


ngfreiter:
Fixes #447

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -1595,22 +1595,21 @@ describe('StructureDefinitionExporter', () => {
     // Quantity is the first ancestor of Duration and Age
 
     // * value[x] only Duration or Age
-    // * value[x].comparator = #>=
+    // * value[x].comparator = #>
```
mint-thompson, line 1598: Very minor, but you can undo this change, as the comparator is still `>=`

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -1595,22 +1595,21 @@ describe('StructureDefinitionExporter', () => {
...
     fixedValueRule.fixedValue = new FshCode('>=');
-    profile.rules.push(fixedValueRule);
+    extension.rules.push(fixedValueRule);
 
-    exporter.exportStructDef(profile);
-    const sd = pkg.profiles[0];
-    const fixedElement = sd.findElement('Observation.value[x].comparator');
+    exporter.exportStructDef(extension);
+    const sd = pkg.extensions[0];
+    const fixedElement = sd.findElement('Extension.value[x].comparator');
```
mint-thompson, line 1612: Just to make sure I understand: the type on `value[x]` would be `Quantity`. Would the type have profiles for `Age` and `Duration`, or would it have no profiles?

ngfreiter: In this case the type on `value[x]` does actually get set to `Duration` or `Age`. `Duration` and `Age` are specializations of `Quantity`, but they are treated just as if they were any other type, because `Quantity` is not an abstract type, so it cannot be replaced by either `Duration` or `Age`. So on the `Observation.value[x]`, you cannot constrain the type to `Duration`, because `Duration` is not one of the choices for that `value[x]` element. `Quantity` is, but we cannot replace `Quantity` with `Duration`. That is why I had to change this test to use `Extension`, since `Duration` and `Age` are both explicitly listed in the list of types the `Extension.value[x]` choice can take on.

mint-thompson: Ahh, okay. Thank you for clarifying!

PR #495, opened 2020-06-17
Contained resources in StructureDefinitions, export primitives with children

mint-thompson:
Completes task [CIMPL-397](https://standardhealthrecord.atlassian.net/browse/CIMPL-397). See also #457 

Primitive elements can have child elements. They were already being added during import, but were getting lost during export in the `toJSON` method.

StructureDefinitions can contain resources on their `contained` element. This element can be accessed using a CaretValueRule. The current implementation checks to see if the element's type is `Resource`, but this may be too restrictive in cases that I haven't thought of yet. Because the resource to add may be an inline Instance, these rules are deferred until after the InstanceExporter finishes.

### DIFF
--- a/src/export/StructureDefinitionExporter.ts
+++ b/src/export/StructureDefinitionExporter.ts
@@ -11,7 +11,8 @@ import {
   ParentDeclaredAsProfileNameError,
   InvalidFHIRIdError,
   InvalidExtensionParentError,
-  ParentDeclaredAsProfileIdError
+  ParentDeclaredAsProfileIdError,
+  MismatchedTypeError
 } from '../errors';
 import {
   CardRule,
@@ -37,6 +38,8 @@ import { Package } from './Package';
  * The operations and structure of both exporters are very similar, so they currently share an exporter.
  */
 export class StructureDefinitionExporter implements Fishable {
+  deferredRules = new Map<StructureDefinition, CaretValueRule[]>();
+
   constructor(
     private readonly tank: FSHTank,
     private readonly pkg: Package,
@@ -173,7 +176,23 @@ export class StructureDefinitionExporter implements Fishable {
             if (rule.path !== '') {
               element.setInstancePropertyByPath(rule.caretPath, rule.value, this);
             } else {
-              structDef.setInstancePropertyByPath(rule.caretPath, rule.value, this);
+              try {
+                structDef.setInstancePropertyByPath(rule.caretPath, rule.value, this);
+              } catch (e) {
+                if (
+                  e instanceof MismatchedTypeError &&
+                  e.valueType === 'string' &&
+                  e.elementType === 'Resource'
+                ) {
+                  if (this.deferredRules.has(structDef)) {
+                    this.deferredRules.get(structDef).push(rule);
+                  } else {
+                    this.deferredRules.set(structDef, [rule]);
+                  }
+                } else {
+                  throw e;
+                }
+              }
             }
           } else if (rule instanceof ObeysRule) {
             const invariant = this.tank.fish(rule.invariant, Type.Invariant) as Invariant;
@@ -199,6 +218,25 @@ export class StructureDefinitionExporter implements Fishable {
         );
       }
     }
+  }
+
+  applyDeferredRules() {
+    this.deferredRules.forEach((rules, sd) => {
+      for (const rule of rules) {
+        if (typeof rule.value === 'string') {
+          const fishedValue = this.fishForFHIR(rule.value);
+          if (fishedValue) {
+            try {
+              sd.setInstancePropertyByPath(rule.caretPath, fishedValue, this);
+            } catch (e) {
+              logger.error(e, rule.sourceInfo);
+            }
+          } else {
+            logger.error(`Could not find a resource named ${rule.value}.`, rule.sourceInfo);
+          }
+        }
+      }
+    });
   }
 
   /**
--- a/src/export/StructureDefinitionExporter.ts
+++ b/src/export/StructureDefinitionExporter.ts
@@ -2,7 +2,8 @@ import {
   StructureDefinition,
   ElementDefinition,
   ElementDefinitionBindingStrength,
-  idRegex
+  idRegex,
+  InstanceDefinition
 } from '../fhirtypes';
 import { Profile, Extension, Invariant } from '../fshtypes';
 import { FSHTank } from '../import';
@@ -225,11 +226,25 @@ export class StructureDefinitionExporter implements Fishable {
       for (const rule of rules) {
         if (typeof rule.value === 'string') {
           const fishedValue = this.fishForFHIR(rule.value);
+          // an inline instance will fish up an InstanceDefinition, which can be used directly.
+          // other instances will fish up an Object, which needs to be turned into an InstanceDefinition.
+          // an InstanceDefinition needs a resourceType, so check for that property.
           if (fishedValue) {
-            try {
-              sd.setInstancePropertyByPath(rule.caretPath, fishedValue, this);
-            } catch (e) {
-              logger.error(e, rule.sourceInfo);
+            if (fishedValue instanceof InstanceDefinition) {
+              try {
+                sd.setInstancePropertyByPath(rule.caretPath, fishedValue, this);
+              } catch (e) {
+                logger.error(e, rule.sourceInfo);
+              }
+            } else if (fishedValue instanceof Object && fishedValue.resourceType) {
+              const fishedInstance = InstanceDefinition.fromJSON(fishedValue);
+              try {
+                sd.setInstancePropertyByPath(rule.caretPath, fishedInstance, this);
+              } catch (e) {
+                logger.error(e, rule.sourceInfo);
+              }
+            } else {
+              logger.error(`Could not find a resource named ${rule.value}.`, rule.sourceInfo);
             }
           } else {
             logger.error(`Could not find a resource named ${rule.value}.`, rule.sourceInfo);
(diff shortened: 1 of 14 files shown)


src/import/FSHImporter.ts
```diff
@@ -1232,6 +1232,8 @@ export class FSHImporter extends FSHVisitor {
     // Get the caret path, but slice off the starting ^
     caretValueRule.caretPath = this.visitCaretPath(ctx.caretPath()).slice(1);
     caretValueRule.value = this.visitValue(ctx.value());
+    caretValueRule.isInstance =
```
ngfreiter, line 1235: I see that this is set here, but is this used anywhere?

src/export/StructureDefinitionExporter.ts
```diff
@@ -201,6 +221,39 @@ export class StructureDefinitionExporter implements Fishable {
...
   }
 
+  applyDeferredRules() {
+    this.deferredRules.forEach((rules, sd) => {
+      for (const rule of rules) {
+        if (typeof rule.value === 'string') {
+          const fishedValue = this.fishForFHIR(rule.value);
+          // an inline instance will fish up an InstanceDefinition, which can be used directly.
+          // other instances will fish up an Object, which needs to be turned into an InstanceDefinition.
+          // an InstanceDefinition needs a resourceType, so check for that property.
```
ngfreiter, line 231: I'm a bit confused about this comment. Is an inline instance represented differently than a non-inline instance? I don't think that is the case, but I feel like I am missing something.

ngfreiter: Ok I think I understand, the distinction is between instances defined in FSH and those defined externally? Is that correct?

src/export/StructureDefinitionExporter.ts
```diff
@@ -173,7 +177,23 @@ export class StructureDefinitionExporter implements Fishable {
...
               element.setInstancePropertyByPath(rule.caretPath, rule.value, this);
             } else {
-              structDef.setInstancePropertyByPath(rule.caretPath, rule.value, this);
+              try {
+                structDef.setInstancePropertyByPath(rule.caretPath, rule.value, this);
+              } catch (e) {
+                if (
+                  e instanceof MismatchedTypeError &&
+                  e.valueType === 'string' &&
+                  e.elementType === 'Resource'
```
ngfreiter, line 186: This seems like a safe approach when the user is just setting the `contained` array, but this won't work for something like:
```
* ^identifier = MyIdentifier
```
Which is something to consider. I'm not saying that use case needs to be included in this PR necessarily. But maybe this approach could be refactored so that if we do add that, it is not a huge lift to add. You set the `IsInstance` flag on the caret value rule in the importer. Wouldn't that be a sufficient condition for knowing if a rule should be deferred?

ngfreiter:
One other thing I am noticing is that if you put a CodeableConcept in place of a Resource like so:
```
Instance: MyCodeable
InstanceOf: CodeableConcept
Usage: #inline

Profile: MyProf
Parent: Patient
* ^contained[0] = MyCodeable
```
Then you get a TypeError:
```
 TypeError: Cannot read property 'sdType' of undefined
```

I think this is because in the `checkFixInlineInstance` function on ElementDefinition, we search into the instanceMeta, but the instanceMeta is stripped out when you use `fishForFHIR`.

src/export/StructureDefinitionExporter.ts
```diff
@@ -37,6 +39,8 @@ import { Package } from './Package';
  * The operations and structure of both exporters are very similar, so they currently share an exporter.
  */
 export class StructureDefinitionExporter implements Fishable {
+  deferredRules = new Map<StructureDefinition, CaretValueRule[]>();
```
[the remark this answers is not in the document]
mint-thompson: That seems reasonable, I'll just need to add a getter so that I can still test that rules get added to it.

src/import/FSHImporter.ts
mint-thompson: Nope! I think I was planning to use this as the condition for whether to put a rule into `deferredRules`, but maybe that's not necessary? I thought I needed it for the same reason that FixedValueRule needs it, but now I'm thinking about the case where an alias is used for the URL to an instance that will be present in the loaded FHIR definitions. So, in that case, this condition would actively work against me.

src/export/StructureDefinitionExporter.ts
mint-thompson: Yes, that's precisely it. An instance defined in FSH will get us an InstanceDefinition (since that's what the InstanceExporter creates), but an instance defined externally is just some JSON we read in, and never gets turned into anything more specific than an Object.

src/export/StructureDefinitionExporter.ts
mint-thompson: That's a good point, and I think that case is within scope for this PR. So maybe I want to keep that `isInstance` flag around, and just check for that? It only gets checked if assigning the string fails, so that shouldn't get me into too much trouble.

mint-thompson:
Thanks for the detailed comments and for pointing out the case with an Instance of CodeableConcept!

mint-thompson:
@ngfreiter After some additional exploration and research, I'm not so sure about the case you pointed out with an Instance of CodeableConcept. A CodeableConcept is not a Resource, so I don't think it could be assigned there? Regardless, I should be doing a bit better than a TypeError in that case.

ngfreiter:
Ah yes I agree, it should be an error about setting CodeableConcept onto a type that does not match, but it should not be allowed.

src/export/StructureDefinitionExporter.ts
```diff
@@ -173,7 +177,19 @@ export class StructureDefinitionExporter implements Fishable {
             if (rule.path !== '') {
               element.setInstancePropertyByPath(rule.caretPath, rule.value, this);
             } else {
-              structDef.setInstancePropertyByPath(rule.caretPath, rule.value, this);
+              try {
+                structDef.setInstancePropertyByPath(rule.caretPath, rule.value, this);
+              } catch (e) {
+                if (e instanceof MismatchedTypeError && rule.isInstance) {
```
ngfreiter, line 183: Sorry I didn't think of this before, but now that I'm looking again, is it possible that here we could just immediately know to defer a rule if it has `IsInstance` true? And then we don't have to try to `setInstancePropertyByPath` and catch the error.

test/import/FSHImporter.Profile.test.ts
```diff
@@ -1483,6 +1483,17 @@ describe('FSHImporter', () => {
...
+
+      it('should add resources to the contained array using a CaretValueRule', () => {
+        const input = `
+        Profile: ObservationProfile
+        Parent: Observation
+        * ^contained = myResource
+        `;
+        const result = importSingleText(input);
+        const profile = result.profiles.get('ObservationProfile');
+        assertCaretValueRule(profile.rules[0], '', 'contained', 'myResource');
```
ngfreiter, line 1495: I think this assert should check that `IsInstance` is true right?

src/export/StructureDefinitionExporter.ts
```diff
@@ -201,6 +217,43 @@ export class StructureDefinitionExporter implements Fishable {
...
+        if (typeof rule.value === 'string') {
+          const fishedValue = this.fishForFHIR(rule.value);
+          // an inline instance will fish up an InstanceDefinition, which can be used directly.
+          // other instances will fish up an Object, which needs to be turned into an InstanceDefinition.
+          // an InstanceDefinition of a resource needs a resourceType, so check for that property.
+          // if we can't find a resourceType or an sdType, we have a non-resource Instance, which is no good
+          if (fishedValue) {
+            if (
+              fishedValue instanceof InstanceDefinition &&
+              (fishedValue.resourceType || fishedValue._instanceMeta?.sdType)
```
ngfreiter, line 232: I'm not sure if the `fishedValue._instanceMeta?.sdType` comes into play here. As far as I can tell, `this.fishForFHIR` calls `fisher.fishForFHIR`, which in turn calls `pkg.fishForFHIR`, which strips instanceMeta on an instance by calling `toJSON` I believe. At least that is how I am tracing it, but it's a confusing thing to track down so I may be missing something.

src/export/StructureDefinitionExporter.ts
mint-thompson: You might be correct? This code is intended to handle the situation where `rule.isInstance` is true (which means `rule.value` was parsed from a SEQUENCE) and the rule gets applied successfully. This can happen when the caretPath points at an element with a type that gets handled in `ElementDefinition.fixString`.

test/import/FSHImporter.Profile.test.ts
mint-thompson: Oh! I'd better add `isInstance` to `assertCaretValueRule`.

src/export/StructureDefinitionExporter.ts
ngfreiter: So the case would be something like this?
```
* ^version = MyInstance
```

That should be an error right? I think that if they want to set a string we should require the usage of "".

src/export/StructureDefinitionExporter.ts
mint-thompson: Ah, that's a good point. I was trying to make sure it would be ready for `ElementDefinition.checkFixInlineInstance`, but when I removed the check for `sdType`, none of the tests broke. So either my tests have missed something, or that check isn't necessary.

src/export/StructureDefinitionExporter.ts
mint-thompson: Ah, I think I understand now. So if we're willing to say "quotes are required when setting a string", then yes, anything caretValueRule with `isInstance` true should get deferred, I think. I was a bit hesitant, but I tried it out with mCODE and nothing broke, so I'll go for it.

PR #498, opened 2020-06-22
Use "or" when listing references

mint-thompson:
Completes task [CIMPL-421](https://standardhealthrecord.atlassian.net/browse/CIMPL-421). See also [this FHIR spec feedback](https://jira.hl7.org/browse/FHIR-27265) and #483.

References should no longer be listed with a "|" character. This is still allowed, but will result in a warning. This change occurs in parallel to deprecating "," in favor of "and" when listing other items.


src/import/FSHImporter.ts
```diff
@@ -1144,18 +1144,41 @@ export class FSHImporter extends FSHVisitor {
...
-      this.aliasAwareValue(ctx.REFERENCE(), this.parseReference(ctx.REFERENCE().getText())[0])
-    )
-      .withLocation(this.extractStartStop(ctx))
-      .withFile(this.currentFile);
+    let ref: FshReference;
+    if (ctx.OR_REFERENCE()) {
+      ref = new FshReference(
+        this.aliasAwareValue(
+          ctx.OR_REFERENCE(),
+          this.parseOrReference(ctx.OR_REFERENCE().getText())[0]
```
ngfreiter, line 1152: I think this is more the fault of whoever first implemented fixed references (it was me), but we are getting just the first element in the list of References, so if it is `Reference(MyPatient1 or MyPatient2)`, we just set the Reference to `MyPatient1`. This makes sense, since the `visitReference` function is called when we are fixing a value, when fixing a Reference you have to pick the specific thing you want to reference to. You can't fix a reference to multiple values. So I'm wondering if here it makes more sense to just have an error if they use `or` or `|`, since really neither one should be used here.

mint-thompson: The correct usage here of indicating only a single reference type will be parsed as an `OR_REFERENCE` token, but I can check the length of the result from `parseOrReference` and `parsePipeReference` to see if that error should be logged.

PR #500, opened 2020-06-24
Insert Rules

### LINKED ISSUE #279: Consider adding RuleSets anywhere within a list of rules
Since rules can be order-dependent, consider the possibility of inserting rule sets at arbitrary points in a set of rules, by simply stating the rule set name, as follows:

```
Profile: XYZ
* rule-XYZ-1
* rule-set-1
* rule-XYZ-2
* rule-set-2
* rule-XYZ-3
etc.
```

ngfreiter:
This fixes #279 by addressing https://standardhealthrecord.atlassian.net/browse/CIMPL-404, https://standardhealthrecord.atlassian.net/browse/CIMPL-396, and https://standardhealthrecord.atlassian.net/browse/CIMPL-332. This adds an `insert` rule to the grammar which allows for applying a `RuleSet` anywhere within a list of rules. The `insert` rule can be applied onto any entity in FSH that is defined by a list of rules. Upon export, the rule is replaced with the list of rules defined on the `RuleSet` referred to by the `insert` rule. As an example:
```
RuleSet: rs1
* valueString = "hello"

Instance: ob1
InstanceOf: Observation
* insert rs1
```
is equivalent to:
```
Instance: ob1
InstanceOf: Observation
* valueString = "hello"
```
An `insert` rule can be applied on a RuleSet itself, so they are recursively expanded when expanding an `insert` rule.

A lot of these changes are refactoring changes. To make these rules work nicely on `ValueSet`s and `CodeSystem`s, I changed those classes to only have a list of rules, instead of a list of rules and a list of `ValueSetComponent`s, as was the case for `ValueSet`. So now everything that is defined in FSH with a `*` is represented in SUSHI as a subclass of `Rule`.

The `insert` rule replaces the `Mixins` keyword, although the usage of the `Mixins` keyword is not fully removed, it is deprecated with a warning.


cmoesel:
#500!  You lucky duck you!

src/fshtypes/Extension.ts
```diff
@@ -19,4 +21,13 @@ export class Extension extends FshEntity {
...
   }
+
+  ruleIsAllowed(rule: Rule) {
+    return !(
+      rule instanceof InsertRule ||
+      rule instanceof MappingRule ||
+      rule instanceof ValueSetComponentRule ||
+      rule instanceof ConceptRule
+    );
+  }
```
mint-thompson, line 32: Although the list of disallowed `Rule` types is shorter, I would prefer for this to be implemented directly as checking a list of allowed types. With an allow-list implementation, new `Rule` types are automatically permitted, which seems more error prone than requiring the new type to be explicitly allowed.

src/fshtypes/Profile.ts
```diff
@@ -1,18 +1,52 @@
...
   }
+
+  ruleIsAllowed(rule: Rule) {
+    return !(
+      rule instanceof InsertRule ||
+      rule instanceof MappingRule ||
+      rule instanceof ValueSetComponentRule ||
+      rule instanceof ConceptRule
+    );
+  }
```
mint-thompson, line 40: See previous comment regarding allow vs deny list of types.

test/fhirtypes/applyInsertRules.test.ts
```diff
@@ -0,0 +1,302 @@
+import { FSHTank, FSHDocument } from '../../src/import';
+import { Profile, RuleSet, FshCode, FshValueSet } from '../../src/fshtypes';
+import { InsertRule, CardRule, ConceptRule } from '../../src/fshtypes/rules';
+import { loggerSpy, assertCardRule, assertValueSetConceptComponent } from '../testhelpers';
+import { minimalConfig } from '../utils/minimalConfig';
+import { applyInsertRules } from '../../src/fhirtypes/common';
+
+describe('applyInsertRules', () => {
```
mint-thompson, line 8: Could you add tests to demonstrate the scenarios where a `RuleSet` is inserted multiple times? I'm thinking of two main cases: one where the `RuleSet` is simply repeated, and one where the it's less direct:
```
RuleSet: Cake
* ...

RuleSet: Pie
* insert Cake
* ...

Profile: BakeryProfile
* insert Cake
* insert Pie
```

src/fhirtypes/common.ts
```diff
@@ -346,14 +368,76 @@ export function applyMixinRules(
...
+          seenRuleSets.push(ruleSet.name);
+          applyInsertRules(ruleSet, tank, seenRuleSets);
+          ruleSet.rules.forEach(ruleSetRule => {
+            // On the import side, a rule that is intended to be a ValueSetConceptComponent can
+            // be imported as a ConceptRule because the syntax is identical. If this is the case,
+            // create a ValueSetConceptComponent that corresponds to the ConceptRule, and use that
+            if (
+              fshDefinition instanceof FshValueSet &&
+              ruleSetRule instanceof ConceptRule &&
+              ruleSetRule.definition == null
```
cmoesel, line 409: As it is now, if someone mistakenly puts a definition in a code in a `ValueSet`, then this `if` is not satisfied and it will end up being detected as an unsupported rule type in line 424 below -- and generate an error that `ConceptRule` cannot be applied to a `ValueSet`.  I think authors might be confused by this...

I think it would be better remove `ruleSetRule.definition == null` in the `if` statement and then do a special check inside the `if` statement and produce a more appropriate warning: `ValueSet concepts should not include a definition, only system, code, and display are supported.` (or something like that).  And still convert it to a `ValueSetConceptComponentRule` (but without the definition).

src/fshtypes/FshCodeSystem.ts
```diff
@@ -11,20 +10,26 @@ export class FshCodeSystem extends FshEntity {
...
     this.rules = [];
   }
 
-  addConcept(newConcept: FshConcept) {
-    if (this.concepts.find(existingConcept => existingConcept.code == newConcept.code)) {
+  addConcept(newConcept: ConceptRule) {
+    if (
+      this.rules
+        .filter(rule => rule instanceof ConceptRule)
+        .find((existingConcept: ConceptRule) => existingConcept.code == newConcept.code)
```
cmoesel, line 25: I think this implementation effectively results in two loops: one that loops over all rules and keeps only the ConceptRules, and then another that loops over all those rules trying to match on the code.

I think this would be more efficient as a single `find`:
```ts
this.rules.find(rule => rule instanceof ConceptRule && rule.code == newConcept.code)
```
(not sure, but you might need a cast in there).

src/fshtypes/RuleSet.ts
```diff
@@ -19,4 +20,8 @@ export class RuleSet extends FshEntity {
   get id() {
     return this.name;
   }
+
+  ruleIsAllowed(rule: Rule) {
+    return !(rule instanceof InsertRule);
+  }
```
cmoesel, line 26: I'm confused.  Don't we allow `InsertRule` in `RuleSet`s?

test/export/CodeSystemExporter.test.ts
```diff
@@ -247,15 +246,15 @@ describe('CodeSystemExporter', () => {
...
+    codeSystem.rules.push(...[new ConceptRule('myCode'), new ConceptRule('anotherCode')]);
     doc.codeSystems.set(codeSystem.name, codeSystem);
     const exported = exporter.export().codeSystems;
     expect(exported.length).toBe(1);
     expect(exported[0].count).toBe(5);
     expect(loggerSpy.getAllLogs('warn').length).toBe(0);
   });
 
-  it('should warn when ^count does not match number of concepts in #complete CodeSystem', () => {
+  it('should warn when ^count does not match number of rules in #complete CodeSystem', () => {
```
cmoesel, line 257: This description isn't exactly right -- because `^count` should correspond to the number of concepts, not rules.  Now that `CodeSystem` supports `include` rules, it's possible for a `CodeSystem` to have only one rule but have a `count` > 1.

Can you fix the description _and_ add a test to ensure the proper count when using `insert` rules?

src/export/InstanceExporter.ts
```diff
@@ -26,7 +28,7 @@ export class InstanceExporter implements Fishable {
     instanceDef: InstanceDefinition,
     instanceOfStructureDefinition: StructureDefinition
   ): InstanceDefinition {
-    let rules = fshInstanceDef.rules.map(r => cloneDeep(r));
+    let rules = fshInstanceDef.rules.map(r => cloneDeep(r)) as FixedValueRule[];
```
cmoesel, line 31: Same comment applies here as I left on `MappingExporter.ts` line 43: I'd prefer that we not _assume_ that all `insert` rules have already been filtered out.  See my comment on `MappingExporter.ts` for more discussion.

src/fshtypes/Profile.ts
```diff
@@ -1,18 +1,52 @@
...
 }
+
+export type SdRule =
```
cmoesel, line 43: I'm not sure this is the right place to export `SdRule`, since it is used by both `Profile` and `Extension`.  Perhaps it should be defined and exported somewhere in the `rules` folder?

src/import/FSHImporter.ts
```diff
@@ -416,40 +416,29 @@ export class FSHImporter extends FSHVisitor {
...
-        logger.error(
-          'Caret rule on ValueSet cannot contain path before ^, skipping rule.',
-          rule.sourceInfo
-        );
       } else {
-        valueSet.rules.push(this.visitCaretValueRule(caretValueRule));
+        valueSet.rules.push(rule);
       }
     });
+    valueSet.rules = valueSet.rules.filter(rule => rule);
```
cmoesel, line 441: What's the purpose of this filter?  Does `valueSet.rules` potentially have null entries and you're trying to filter them out?

src/import/FSHImporter.ts
```diff
@@ -498,25 +486,19 @@ export class FSHImporter extends FSHVisitor {
...
+          codeSystem.addConcept(rule);
+        } catch (e) {
+          logger.error(e.message, rule.sourceInfo);
+        }
       } else {
-        codeSystem.rules.push(this.visitCaretValueRule(caretValueRule));
+        codeSystem.rules.push(rule);
       }
     });
+    codeSystem.rules = codeSystem.rules.filter(rule => rule);
```
cmoesel, line 501: Another of those mysterious filters...

src/fshtypes/Extension.ts
cmoesel: I think I agree with @mint-thompson -- let's be explicit about what we allow.  I also think something like below might read more easily:
```ts
function ruleIsAllowed(rule: Rule) {
  switch (rule.constructor) {
    case InsertRule:
    case CardRule:
    case FlagRule:
      return true;
    default:
      return false;
  }
}
```
There is a limitation -- it wouldn't work for _subclasses_ of these things (while `instanceof` would) -- but I don't think that's really a problem for us.  We don't have subclasses of these, and even if we did, maybe we should explicitly test for them anyway.

cmoesel: OR... we could just have a single file where we manage the allowed and disallowed rules.  I tested that out and confirmed something like this would also work:

```ts
// This map is obviously incomplete, but proves the point...
const allowedRulesMap = new Map<any, any[]>([
  [Profile, [InsertRule, CardRule, FlagRule]],
  [FshValueSet, [InsertRule, ValueSetComponentRule]]
]);

export function isAllowedRule(fshDefinition: Profile | FshValueSet, rule: Rule): boolean {
  return allowedRulesMap.get(fshDefinition.constructor)?.find(r => r === rule.constructor) != null;
}
```

Of course the reverse map (where the rule is the key and entities are the values) would also work.  I'd have to type it all out to see which one would be preferred though.

ngfreiter: I'll try to do something like the map and be explicit. We do have subtypes though on `ValueSetComponentRule`, but there are only 2 subtypes so I will just list them as well.

src/fshtypes/RuleSet.ts
ngfreiter: Well we do, but when you are testing if a `ruleIsAllowed`, that test is after everything has been expanded. So I guess this function should have a better name than `ruleIsAllowed`, since it is really `ruleIsAllowedAfterExpandingInsertRules`. But that is too long. Maybe `ruleApplies`? That is still not totally clear, but maybe more clear?

test/export/CodeSystemExporter.test.ts
ngfreiter: Oh that looks like a find and replace typo, but it helped you think of a test case, so good job find and replace.

src/fshtypes/Extension.ts
cmoesel: I didn't test it, but if we want to support subtypes with the map, then something like this _might_ work:
```ts
export function isAllowedRule(fshDefinition: Profile | FshValueSet, rule: Rule): boolean {
  return allowedRulesMap.get(fshDefinition.constructor)?.find(r => rule instanceof r) != null;
}
```
(Darn -- deleted all my test code for this about 20 minutes ago!)

src/import/FSHImporter.ts
ngfreiter: Yeah sometimes `visitVsRule` will return undefined if the rule is something like a `CaretValueRule` with no a path in front. But...why not just not add the rule at all? I don't know. It made sense at the time I swear.

cmoesel: OK.  Then it might benefit from a comment.  And although what you have is technically correct, it might be more clear what the intent is if you changed it to:
```ts
    valueSet.rules = valueSet.rules.filter(rule => rule != null);
```

ngfreiter: I wasn't clear in my questioning of myself, but I meant that I am going to just not add undefined rules to the rules array. It seems simpler and more clear what is happening.

src/fhirtypes/common.ts
```diff
@@ -326,8 +338,18 @@ export function applyMixinRules(
   fshDefinition: Profile | Extension | Instance,
   tank: FSHTank
 ): void {
+  if (fshDefinition.mixins.length > 0) {
+    const insertString = '* insert ' + fshDefinition.mixins.join('\n* insert ') + '\n';
```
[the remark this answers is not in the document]
ngfreiter: Wait, but the code you have is the case where there is no extra `\n` at the end, since `join` doesn't put one at the end. If you move the `\n` to the map:
```
const insertString = fshDefinition.mixins.map(m => `* insert ${m}\n`).join('');
```
That is the case where you do have an extra `\n` at the end. Unless I'm misinterpreting something.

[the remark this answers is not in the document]
cmoesel: I think I just have not mastered the English language well enough yet.  That last bit was intended to mean this:
> ~~And~~ I don't think the final \n is necessary ~~, is it?~~. If ~~so~~ it is necessary, you could move the \n from the join to the map and just join on empty string.

So... My original suggested code does not put an `\n` at the very end, and I think that is OK -- but if it is not OK, then you can move that `\n` to the `map` in order to get the final `\n` back.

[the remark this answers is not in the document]
ngfreiter: Ahh now it all adds up. Gotcha. But you are correct, the final `\n` is really not necessary, so I will go with your original suggestion.

test/fhirtypes/applyInsertRules.test.ts
ngfreiter: Thanks for pointing this out.  The case you laid out was erroneously being logged as a circular dependency!

ngfreiter:
Ok @cmoesel and @mint-thompson, https://github.com/FHIR/sushi/pull/500/commits/d5c0596b008da5406094ea7f00cd23164db642bf should make the desired changes. I will update from master once the other pending PRs have been updated and merged.

src/fhirtypes/common.ts
```diff
@@ -388,51 +389,55 @@ export function applyInsertRules(
...
+    const ruleSet = tank.fish(rule.ruleSet, Type.RuleSet) as RuleSet;
+    if (ruleSet) {
+      if (seenRuleSets.includes(ruleSet.name)) {
+        logger.error(
+          `Inserting ${ruleSet.name} will cause a circular dependency, so the rule will be ignored`,
+          rule.sourceInfo
+        );
+        return;
+      }
+      // seenRuleSets.push(ruleSet.name);
```
cmoesel, line 405: Is this intentionally commented out?  If so, can we just delete it?

src/fshtypes/AllowedRules.ts
```diff
@@ -0,0 +1,79 @@
...
+ * it into its list of rules, and then export those rules.
+ * @param fshDefinition
+ * @param rule
+ * @returns - true if rule is allowed on fshDefinition, false otherwise
+ */
+export function isAllowedRule(
+  fshDefinition: Profile | Extension | Instance | FshValueSet | FshCodeSystem | Mapping | RuleSet,
+  rule: Rule
+): boolean {
+  return allowedRulesMap.get(fshDefinition.constructor)?.find(r => rule instanceof r) != null;
```
cmoesel, line 78: If you happen to be making changes anyway, it dawns on me that this would have been even better:
```ts
return allowedRulesMap.get(fshDefinition.constructor)?.some(r => rule instanceof r);
```

Issue #392, opened 2020-04-25
Sushi doesn't like strings that start with a # as the first character.


keithboone:
They work if I use """, but not "

ngfreiter:
The following seems to work for me:
```
Profile: MyPatient
Parent: Patient
* name.family = "#Hi"
```
Maybe it is something more specific to your use case? Do you have an example we could recreate this with?

ngfreiter:
@keithboone any updates on this and how we could recreate it?

keithboone:
Instance: states-NJ
InstanceOf: Location
Description : "Example Location for the State of New Jersey"
* name = "# Testing Hash Mark\n## Second test"
* id = "states-NJ"


This is the failing instance.

ngfreiter:
It seems like the issue is the `\n`, not the `#` at the start of the string. I haven't been able to cause an issue just using `#` with no `\n`, but the following:
```
Instance: states-NJ
InstanceOf: Location
Description: "nonsenese"
* name = "testing \n in this"
```
does cause an error.  Now it seems like you can accomplish the desired goal here using a multiline string:
```
Instance: states-NJ
InstanceOf: Location
Description: "nonsenese"
* name = """
  testing 
  in this"""
```
But, which gives the output in JSON as 
```
{
  "resourceType": "Location",
  "id": "states-NJ",
  "name": "testing \nin this"
}
```

But I'm not entirely sure if we are supposed to be allowing `\n` if someone wants to use `\n` instead of using a multiline string and using enter. I think @cmoesel and @mint-thompson know more about this and may have a better idea of what should be allowed and what is going on here.

cmoesel:
The FHIR Shorthand specification doesn't really define any special support for characters like `\n` in strings.  If we want to support something like this, I think we need to consider what special characters are supported and document them as such.  For example `\r`, `\n`, and `\t` might be good candidates -- this would align well with what FHIR allows in strings (although FHIR doesn't explicitly say to use those representations of the characters).

I think this is worth considering, but we also need to think through any possible implications.

keithboone:
Since Sushi seems mostly to be focused on JSON representations, I was thinking that it should support the same escape sequences in strings that JSON supports, or at least expect it to.

keithboone:
Added to ballot comments.

cmoesel:
Closing this as it was fixed a while back.

PR #503, opened 2020-06-25
Support escaped characters \n \r \t in strings

jafeltra:
This PR addresses https://standardhealthrecord.atlassian.net/browse/CIMPL-419 and #392 

In strings and multiline strings, you can now add the escaped characters that FHIR strings support, `\n`, `\t`, and `\r`. These characters in strings and multiline strings will be parsed correctly and replaced in the string similar to the other escaped characters that are allowed. You should be able to add a string in your FSH like: 

```
Description: "An important first and \n second line."
```

The tests that check this case have `\\n` to escape the `\` in the template string.


src/import/FSHImporter.ts
```diff
@@ -1499,6 +1499,9 @@ export class FSHImporter extends FSHVisitor {
     const str = stringCtx.getText();
     return str
       .slice(1, str.length - 1)
+      .replace(/\\n/g, '\n')
+      .replace(/\\r/g, '\r')
+      .replace(/\\t/g, '\t')
```
mint-thompson, line 1504: These replacements need to happen after the literal-backslash replacement. Otherwise, this FSH input would cause problems: `Some \\n text`

jafeltra: Good catch! I updated this and added a test to check that `Some \\n text` will get replaced properly and result in `Some \n text`.

jafeltra:
Rebased and resolved the merge conflicts. If Nick's PR is merged first, I can resolve the resulting conflicts.

mint-thompson:
Okay, the solution I offered wasn't quite right. The expected output for that input should be a literal backslash followed by the letter n. I think that rather than doing several `replace` calls, you'll have to do this by scanning through the string only once so that a literal backslash doesn't get reused as an escape-sequence backslash.

PR #509, opened 2020-06-30
Implied path search maintains ancestor indices

### LINKED ISSUE #502: Extra Extensions on Instances
See https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/extensions.20on.20repeating.20elements for more details. To recreate this bug, you can use the following full example:
```
Instance: PatInstance
InstanceOf: MyPatient
* generalPractitioner[0] = Reference(drwelby)       //the GP
* generalPractitioner[0].display = "Dr Marcus Welby"

* generalPractitioner[1] = Reference(gp-org1)       //the GP Practice
* generalPractitioner[1].display = "Good Health Medcial Centre"
* generalPractitioner[1].extension[edi-address].valueCode = #F
* generalPractitioner[1].extension[gp-practice-enrollment-date].valueBoolean = true

Profile: MyPatient
Parent: Patient
* generalPractitioner only Reference(Practitioner | Organization)
* generalPractitioner.extension contains
    http://hl7.org/fhir/us/core/StructureDefinition/us-core-birthsex named edi-address 0..1 and
    http://hl7.org/fhir/us/core/StructureDefinition/us-core-direct named gp-practice-enrollment-date 0..1
```

You should see that there are extraneous extension elements added to `generlPractitioner[0]`. This looks like it is happening in the `setImpliedPropertiesOnInstance` function, so that may be a good place to start investigating this.

mint-thompson:
Fixes #502 and completes task [CIMPL-429](https://standardhealthrecord.atlassian.net/browse/CIMPL-429).

When finding implied paths, the numeric index on the last part of the path should be removed. But, numeric indices on path parts earlier in the path should be maintained. This will prevent implied paths from including siblings of ancestor elements that have numeric indices and minimum cardinality of 0.


test/export/InstanceExporter.test.ts
```diff
@@ -1755,6 +1755,62 @@ describe('InstanceExporter', () => {
...
+          type: 'Organization',
+          isReference: true
+        }
+      );
+      // * generalPractitioner.extension contains
+      //   http://hl7.org/fhir/us/core/StructureDefinition/us-core-direct named direct-email 0..1
+      const containsRule = new ContainsRule('generalPractitioner.extension');
+      containsRule.items.push({
+        name: 'direct-email',
+        type: 'http://hl7.org/fhir/us/core/StructureDefinition/us-core-direct'
```
ngfreiter, line 1776: Instead of using this extension, we could just use on of the extensions that is already in the `testdefs/package` folder. That way we don't have to add another file, and we don't have to add a file from US Core. For example, the http://hl7.org/fhir/StructureDefinition/patient-mothersMaidenName extension is already in `testdefs/package`, and it should work pretty much analogously to the US Core direct email extension (just has a `valueString` instead of `valueBoolean`).

PR #510, opened 2020-06-30
Support BYO XML

ngfreiter:
This addresses https://standardhealthrecord.atlassian.net/browse/CIMPL-377 by adding support for BYO-XML resources. Now if someone has existing XML definitions that they want to incorporate into SUSHI by adding them to the `ig-data/input/` folder, they can do that.

The resources are converted to objects using the [fhir package](https://www.npmjs.com/package/fhir), and can then be used by FSH definitions in the same way that a BYO-JSON resource can be. When the resources are copied over, they are kept as XML files and not converted to JSON.


cmoesel:
This is definitely cool, but I do have one high-level concern.  It looks like they package the FHIR 4.0.0 StructureDefinitions (not 4.0.1, BTW) -- and together the three files they package are 40MB according to GitHub: https://github.com/lantanagroup/FHIR.js/tree/master/profiles/r4

Now, NPM says the unpacked size is 8.48MB, so they may be minimizing those files -- so that's good.  But that's still a lot of payload for files we might not need or that might be old anyway.  My primary concern is that for the playground, we don't want to add even 8 MB of JavaScript code that needs to be downloaded before SUSHI will run in the browser.

This could be a moot point.  If we don't actually use those files, Webpack might be smart enough to cut them out of the web-based JavaScript; but if it's not smart enough to do that, then we might want to consider other approaches.

ngfreiter:
I figured the 4.0.0 wouldn't be a problem, since the actual properties on resources and stuff haven't changed between 4.0.0 and 4.0.1, and as far as I can tell from my looking through the code, that is what the resource definitions are used for. I didn't realize it was that much space though.

For the playground, the packages that SUSHI depends on were already causing some problems. For one, `fs` was an issue. To solve that, this [null-loader](https://www.npmjs.com/package/null-loader) package was used to avoid loading `fs` at all. I think a similar approach could work here to avoid loading the `fhir` package? By the way, if you know of an easier way to handle that `fs`, please let @SamCitron know.

I can consider other approaches. The main other approach would seem to be writing our own converter. But let me know if you want me to consider those other approaches.

cmoesel:
There were some changes between 4.0.0 and 4.0.1, but hopefully they don't affect this library.  The one I remember most is the change to types in some StructureDefinitions (remember the crazy `System.*` types that @jafeltra had to add support for - where the real type is in an extension instead?  That was a change from 4.0.0 to 4.0.1).  It's probably harmless, but somehow it feels more painful to have to download them when they're not the latest version. ;-)

Yeah, I guess we could use `null-loader` to skip it in the playground --  I've used that in the past (IIRC, it was for `fs`, in fact!).  But I think if we do that with the `fhir` lib, you may need to re-write your logic to account for the case where the `fhir` package is not loaded.  The other thing is just to see what Webpack does with it.  It might be that if our code has no path to those large files, Webpack cuts it out.

I looked at the VS Code extension for FHIR Tools because it can convert FHIR XML to JSON, but it turns out it uses Lantana's library too!  So... let's stick with it and we'll figure out what to do w/ the playground after that.

cmoesel:
I've also used [resolve.alias](https://webpack.js.org/configuration/resolve/#resolvealias) to stub out specific files that were really large and I knew were not needed.  And reviewing the documentation now, it looks like Webpack says to use `resolve.alias` _instead_ of `null-loader` (which is apparently deprecated in Webpack 5).  So, @SamCitron -- take note!

ngfreiter:
They do have a section in their README about using different versions of FHIR, and I'm guessing 4.0.1 could work if we manually pointed the library to them. But then we have to have the definitions checked in to the sushi codebase, which is not so nice. 

I've just confirmed that the `null-loader` trick does seem to work. I did `npm run build` on `fsh-playground` without using `null-loader`, and the resulting build folder was about 12MB (ouch). Then I used `null-loader` on `fhir`, and the resulting build was about 2MB (oohh ahhh). So it doesn't seem like webpack automatically cuts it out, but it does seem like we can tell webpack to cut it out, and since we already have to cut out `fs` anyway, it isn't a big lift.

Ah you've followed the same set of clues (or I guess just one clue) I did to get to Lantana 🕵️. I looked a little closer at the Lantana code, and it actually is a little trickier than I thought it might be. I thought it wouldn't be a big deal to write our own converter, and I don't think it is a *big* deal still, but I guess it isn't so easy as xml2js and tack on a resourceType like I had hoped.

ngfreiter:
And may I add another "ooh ahh" for resolve.alias, that looks like a better, built in null-loader!

cmoesel:
I'm not so sure that their library would work easily with 4.0.1, otherwise I think they would have updated those files by now.  I suspect that those annoying `type` changes probably mess up some of their code too.  Not sure about that, but either way, I'm not interested in doubling up the code by checking in 4.0.1 to our source code anyway!

Also, the good news w/ `resolve.alias` is that if we do want to support the XML-to-JSON support in the playground at some future date, then we can just stub out those huge JSON files and leave everything else (as long as the library doesn't use them for XML-to-JSON).

cmoesel:
> And may I add another "ooh ahh" for resolve.alias, that looks like a better, built in null-loader!

Yeah, I had forgotten that it handles entire modules until I looked at it again.  Actually, it has one other upside, which is that if it's easier, we can replace the `fhir` module with a stubbed version that defines dummy functions for the functions we use so you don't have to guard against null modules.  If that makes sense.

ngfreiter:
Ok @cmoesel this should have all your comments addressed. I didn't merge updates from master though, since I will have to do that again after #513 gets merged.

PR #518, opened 2020-07-07
Reference contained resource

### LINKED ISSUE #508: References to contained resources
If you run SUSHI on the following snippet of FSH:
```
Instance: Foo
InstanceOf: ServiceRequest
* contained = annexSR2
* supportingInfo = Reference(annexSR2)

Instance: annexSR2
InstanceOf: Binary
Usage: #inline
* contentType = #zip
```
The output JSON has `supportingInfo.reference` set to `Binary/annexSR2`. The IG Publisher issues errors in the QA report, because `Binary/annexSR2` cannot be resolved, since it is an inline defined contained resource, so it does not exist on its own. Instead, the reference should be set as `#annexSR2` to keep the IG Publisher happy.

ngfreiter:
Fixes #508 by making it so that references to `#inline` resources are specified as `#id` instead of `ResourceType#id`. This can be tested using this example:
```
Instance: Foo
InstanceOf: ServiceRequest
* contained = annexSR2
* supportingInfo = Reference(annexSR2)

Instance: annexSR2
InstanceOf: Binary
Usage: #inline
* contentType = #zip
```
If you run SUSHI on this snippet on master, the `supportingInfo` reference will look like `Binary/annexSR2`. If you then run the IG Publisher on SUSHI's output, the build should succeed, but the QA report will contain an error like:
image

If you then run SUSHI on this snippet on this branch, the `supportingInfo` reference will be correctly specified as just `#annexSR2`. Then, when you run the IG Publisher on that output, the QA error shown above should go away.


ngfreiter:
@jafeltra I've seen that error report, but I don't know what SUSHI could do differently to make it go away. The previous error says to make the reference `#id`, and that is what we do. So in my opinion there is nothing different SUSHI could do. But I could definitely be wrong about that. The Publisher is mysterious. If you're unsure about it we can wait on this PR and try to figure out what that is about. I don't want to merge if you don't really approve.

jafeltra:
Okay I think that I'm okay with leaving this because I agree, you definitely changed this in the way the previous error specified. I think this the changes in this PR make sense, so I feel comfortable approving!

cmoesel:
Looks to be the same as the one reported here: https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/US.20Core.20QA.20Issue.20.234-.20bad.20link.20for.20rel.20ref.20in.20bundle

PR #519, opened 2020-07-07
Canonical keyword

jafeltra:
This PR addresses https://standardhealthrecord.atlassian.net/browse/CIMPL-420 and adds support for a `Canonical` keyword that works similarly to `Reference`.

I've only added support for using the Canonical keyword as part of a FixedValueRule, since that's the only place I could think of that it would be used, but I could easily be missing something. An example of its use:

```
CodeSystem: MyCodeSystem
* #onecode
```
```
Profile: MyObs
Parent: Observation
* code.coding.system = Canonical(MyCodeSystem) (exactly)
```


cmoesel:
> I am pretty sure I agree that the only place this needs to work is with FixedValueRules, but I think this will commonly be used on Instances, and while it seems like this should work fine for an Instance, could you add some Instance tests?

I haven't reviewed this PR (at all).  I only saw this comment come by.  But... I can think of two other places where people might *try* to use `Canonical`:
```
Alias: DT2 = Canonical(DiabetesTypeTwoValueSet)
```
and _maybe_:
```
* code from Canonical(SomeValueSet)
```

The first one I think has a use case -- if people want to define an alias that is _shorter_ than some long name (to make it easier to use).  Is it a _great_ use case?  Maybe not, but I am tempted to support it.

The second one seems silly to me since I think that anywhere that you say `code from Canonical(X)`, the simpler `code from X` also should work.  So... I don't feel compelled to support this.

jafeltra:
I talked to @cmoesel about supporting the other two use cases for `Canonical` that he mentioned, and we're not going to support them as part of this PR. I looked more into using `Canonical` in an `Alias`, and it will likely require refactoring how we do some of the import and alias resolving. Also, the FSH spec does not specifically mention that as a case to support, so we are planning to hold on implementing it for now. We are also going to wait on the use case for `code form Canonical(X)` until it seems it is needed or required.

I also added tests for Instances using `Canonical`, so I think this PR should be set for re-review.

src/fshtypes/FshCanonical.ts
```diff
@@ -0,0 +1,12 @@
+import { FshEntity } from './FshEntity';
+
+export class FshCanonical extends FshEntity {
+  public sdType: string;
```
cmoesel, line 4: What is `sdType` for?

src/import/FSHImporter.ts
```diff
@@ -1245,6 +1250,18 @@ export class FSHImporter extends FSHVisitor {
...
+  visitCanonical(ctx: pc.CanonicalContext): FshCanonical {
+    const canonicalText = this.parseCanonical(ctx.CANONICAL().getText());
+    const fshCanonical = new FshCanonical(canonicalText)
+      .withLocation(this.extractStartStop(ctx))
+      .withFile(this.currentFile);
+    return fshCanonical;
+  }
+
+  private parseCanonical(canonical: string): string {
+    return canonical.slice(canonical.indexOf('(') + 1, canonical.length - 1);
```
cmoesel, line 1262: The grammar allows spaces inside (like `Canonical(    Foo     )`).  Do we need to trim those?  It looks like we _don't_ trim them; I just can't tell if that's because we don't need to or if it's a mistake.

test/export/InstanceExporter.test.ts
```diff
@@ -1189,6 +1190,56 @@ describe('InstanceExporter', () => {
       );
     });
 
+    // Fixing using Canonical
+    it('should apply a FixedValue rule with an valid Canonical entity defined in FSH', () => {
```
cmoesel, line 1194: `an valid Canonical`, eh?

src/import/FSHImporter.ts
cmoesel: OK.  I checkout out the branch and updated the import test to have spaces inside the `(` `)`.  The result is that it created a `FshCanonical` whose entity name had spaces in it.  I then updated an exporter test to see what happens if a `FshCanonical` has spaces in the entity name, and it failed.  So... I think we need to trim the spaces!

test/import/FSHImporter.Profile.test.ts
```diff
@@ -1211,6 +1211,27 @@ describe('FSHImporter', () => {
...
+        const result = importSingleText(input);
+        const profile = result.profiles.get('ObservationProfile');
+        expect(profile.rules).toHaveLength(1);
+
+        const expectedCanonical = new FshCanonical('Example')
+          .withLocation([8, 32, 8, 49])
+          .withFile('');
+        assertFixedValueRule(profile.rules[0], 'code.coding.system', expectedCanonical);
+      });
+
```
cmoesel, line 1234: Can you add a similar test but use the optional spaces (e.g., `Canonical (   Example   )`)?  I think that the resulting `FshCanonical` should have the spaces trimmed out (else it doesn't export correctly).

src/fshtypes/FshCanonical.ts
cmoesel: I removed it and the tests all still passed, so... maybe it was just a copy/paste thing?

bkaney:
How does this work when adding `|version`?

cmoesel:
@bkaney -- good question!  We probably need to think about that, because here's the thing: you can only used `Canonical()` with an entity defined in your IG or with an entity from a dependency IG.  That being the case, it seems like the version is likely "locked in" to the version defined (or depended on) in your IG anyway, right?  For example, if you have a `MyPatient` profile at version 1.0.0 in your IG, does it make sense to say `Canonical(MyPatient|2.0.0)`?  It might, I just don't know...  These feature is mainly for cases where you want the canonical of something created in your IG, because the canonical isn't yet published for it (and may change?) -- but if you're referencing a version outside your IG, then the canonical is known (and stable) and maybe you should just use the canonical directly?

src/import/FSHImporter.ts
jafeltra: Ah I think I was using `Reference` as a guide when writing the grammar, which also allows spaces. You're definitely right that we should trim the spaces so we can identify the entity, but I think I also just want to double check that we do in fact want the spaces included in the grammar, and that I didn't make a bad assumption there.

src/fshtypes/FshCanonical.ts
jafeltra: Yeah I think it was a copy/paste issue. Thanks for catching it.

src/import/FSHImporter.ts
cmoesel: I think the spaces are good.  I'd say keep them in the grammar -- let's just make sure we handle them appropriately...

I wonder if maybe we don't handle `Reference`s with spaces appropriately then?  It may be worth checking.

jafeltra: I had the same thought and looked into it, and I don't think it's being handled correctly. If you have the following rule in your FSH

```
* subject = Reference(   mCODEPatientExample01 )
```
the output will be 
```
"reference": "   mCODEPatientExample01 "
```
which is probably not what was intended, but no errors are logged.

Additionally, if you have the following rule in  your FSH
```
* stage.assessment only Reference(   CancerStageParent )
``` 
SUSHI logs the following error and skips the rule:
```
error No definition for the type "   CancerStageParent " could be found.
```

bkaney:
> @bkaney -- good question! We probably need to think about that, because here's the thing: you can only used `Canonical()` with an entity defined in your IG or with an entity from a dependency IG. That being the case, it seems like the version is likely "locked in" to the version defined (or depended on) in your IG anyway, right? For example, if you have a `MyPatient` profile at version 1.0.0 in your IG, does it make sense to say `Canonical(MyPatient|2.0.0)`? It might, I just don't know... These feature is mainly for cases where you want the canonical of something created in your IG, because the canonical isn't yet published for it (and may change?) -- but if you're referencing a version outside your IG, then the canonical is known (and stable) and maybe you should just use the canonical directly?

My $0.02 -- 

For internal or external that have a FHIR package, `Canonical(MyCodeSystem)` might:
* prefer: `MyCodeSystem.url`
* if empty, construct `${config-yaml.canonical}/CodeSystem/MyCodeSystem|${myCodeSystem.version}`
* if `myCodeSystem.version` is empty, maybe there is an option to use IG version, e.g. `config-yaml.version`? Or option to leave version-less?

I have seen that some IGs (must have a process) that will update all the versions of resources to match the IG version (at publish time). But I know there are some IGs where the versions of individual resources aren't always equal to the IG version.


In the case you use an external reference and are not using a FHIR package dependency, I guess you wouldn't use the helper.

src/import/FSHImporter.ts
cmoesel: Yep.  Then that is broken for sure.  Can you maybe file a new bug for that and just address `Canonical` in this PR?

jafeltra: Sure! I'll log an issue for References.

jafeltra:
@cmoesel your comments should be addressed in 0ca5b90.

PR #521, opened 2020-07-07
0.14.0

ngfreiter:
SUSHI is truly a teenager now. Hopefully it won't start to rebel.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "fsh-sushi",
-  "version": "0.13.2",
+  "version": "0.14.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "fsh-sushi",
-  "version": "0.13.2",
+  "version": "0.14.0",
   "description": "Sushi Unshortens Short Hand Inputs (FSH Compiler)",
   "scripts": {
     "build": "rm -rf dist && tsc && cp -r src/ig/files dist/ig/files",


palfrey:
I'm now hitting https://github.com/yarnpkg/yarn/issues/5480 with 0.14.0, but wasn't with 0.13.2. Did anything change with the release process?

ngfreiter:
@palfrey Nothing about the release process should have changed, but we did update some scripts between this release and the last, and I (a Windows user) did this release, where @cmoesel (a Mac user) did 0.13.1, but I've done several other releases in the past. Do you have any info we could use to reproduce your error? I'm assuming you installed SUSHI with `yarn` instead of `npm`, but then what was the specific error that occurred?

palfrey:
On a Mac (10.14.6), doing `yarn run sushi` with 0.13.2 worked, but with 0.14.0 gets me
```
yarn run v1.22.4
warning package.json: No license field
$ <project path>/node_modules/.bin/sushi
env: node\r: No such file or directory
error Command failed with exit code 127.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

This is with yarn 1.22.4.

Notably, doing `file node_modules/.bin/sushi` on 0.13.2 gets me
```
node_modules/.bin/sushi: a /usr/bin/env node script text executable, UTF-8 Unicode text
```
where as for 0.14.0 it gets me
```
node_modules/.bin/sushi: a /usr/bin/env node script text executable, UTF-8 Unicode text, with CRLF line terminators
```

ngfreiter:
We've spent some time looking into this, and it seems like this has always been an issue for us, for example the same thing seems to happen with SUSHI 0.13.1 (released from a windows machine). It just hadn't come up yet. Thanks for letting us know about this, we will make sure it doesn't happen in future releases of SUSHI. For this release, would you be able to install from `npm` by doing `npm install -g fsh-sushi`? It seems npm does some handling of the line endings that yarn does not, so for this release, npm should work for you.

palfrey:
Using npm for install seems to work fine, thanks. I've also just been doing `node ./node_modules/.bin/sushi` which also works fine, even with yarn install, so I've got workarounds at least!

ngfreiter:
Thanks for being understanding for now! We will make sure future releases don't have this issue.

PR #522, opened 2020-07-08
Report FHIR Shorthand Version

### LINKED ISSUE #501: `sushi -v` should list SUSHI version AND the FSH version it supports
Users may want to know what version of FSH a given version of SUSHI supports...  We should support this.  E.g., something like:

```
$ sushi -v
SUSHI v0.14 (implements FHIR Shorthand specification v0.13.x)
```

cmoesel:
When `sushi -v` is called, report the SUSHI version and the FSH version it implements.

Fixes #501.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -22,6 +22,8 @@ import {
   getRawFSHes
 } from './utils/Processing';
 
+const FSH_VERSION = '0.13.x';
+
 app().catch(e => {
   logger.error(`SUSHI encountered the following unexpected error: ${e.message}`);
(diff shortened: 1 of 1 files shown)


ngfreiter:
You know how when you read a sentence and a word is in there multiple times your brain will just skip past it? Well I think that might have happened to both of us:
image
We are saying "SUSHI" twice in our version message. Sorry I missed this in my review, it just went right past me.

cmoesel:
Ah... I hadn't looked at the `Running...` log...  Good catch!

PR #523, opened 2020-07-08
No Hidden Files

### LINKED ISSUE #350: SUSHI shouldn't copy hidden files
When SUSHI copies files (like in `ig-data/input/pagecontent/`), it should skip hidden files like `.DS_Store`.

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/file.20copy.20from.20ig-data

ngfreiter:
Fixes #350.

SUSHI used to copy over hidden files like `.DS_Store` and `Thumbs.db`. Now those files will not be copied over. It is not *all* hidden files that are not copied, but rather just a few common hidden system files that get added by the OS. The list of system files to not copy is dictated by the `junk` [package](https://www.npmjs.com/package/junk), which I used to exclude these files.


ngfreiter:
Oh no. Case sensitive files have thwarted me!

ngfreiter:
> Oh no. Case sensitive files have thwarted me!

Alright everyone breathe easy, I've thwarted the case sensitive files back.

jafeltra:
I'm sorry I didn't notice this earlier, but I just realized that a warning is being logged when there are hidden files that we ignore in `ig-data/input/pagecontent/`. When I have a `.DS_Store` file in that folder, it is properly ignored and not copied over, but the warning that we log is now incorrect, and probably should only be logged on non-junk files:

```
warn  Files not in the supported file types (.md and .xml) were detected. These files will be copied over without any processing.
```

ngfreiter:
> I'm sorry I didn't notice this earlier, but I just realized that a warning is being logged when there are hidden files that we ignore in `ig-data/input/pagecontent/`. When I have a `.DS_Store` file in that folder, it is properly ignored and not copied over, but the warning that we log is now incorrect, and probably should only be logged on non-junk files:
> 
> ```
> warn  Files not in the supported file types (.md and .xml) were detected. These files will be copied over without any processing.
> ```

Better late than never, expect another PR

jafeltra:
> Better late than never, expect another PR

Better earlier than late though.😕 I will eagerly await the PR.

PR #524, opened 2020-07-09
Add type definition files for generated parser classes

mint-thompson:
Antlr generates a number of classes based on the FSH grammar definition. These classes are produced in javascript, using es5
function-style classes declarations. The typescript compiler will generate type definitions for these classes, but the generated
definitions contain errors that will appear when another project specifies fsh-sushi as a dependency. By manually including type
definitions that are error-free, fsh-sushi can safely be used as a dependency in other projects.

Because the classes most likely to be of use in other project are the FSH and FHIR types, they are added as exports to index.js. Note that the change to package.json which specifies index.js allows these classes to be included more easily, but does mean that fsh-sushi can no longer be run from its directory with a simple "node ." command.

To test including this in another project, `npm pack` can be used to build a `.tgz` file, which can then be used with `npm install` in the other project to add as a dependency. Including the current version of fsh-sushi will result in compilation errors, but including the refactored version will result in successful compilation.

Because the type definition files are now manually defined, this does require them to be updated when a grammar change results in a change to one of the interfaces in these files.


package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.14.0",
   "description": "Sushi Unshortens Short Hand Inputs (FSH Compiler)",
   "scripts": {
-    "build": "rm -rf dist && tsc && cp -r src/ig/files dist/ig/files",
+    "build": "rm -rf dist && tsc && cp -r src/ig/files dist/ig/files && cp src/import/generated/*.d.ts dist/import/generated/",
```
ngfreiter, line 6: Does this mean we still build the incorrect .d.ts files with `tsc` and then overwrite them with the `cp`?

package.json
```diff
@@ -28,11 +28,11 @@
     "type": "git",
     "url": "https://github.com/fhir/sushi.git"
   },
-  "main": "dist/app.js",
+  "main": "dist/index.js",
```
ngfreiter, line 31: > Because the classes most likely to be of use in other project are the FSH and FHIR types, they are added as exports to index.js. Note that the change to package.json which specifies index.js allows these classes to be included more easily, but does mean that fsh-sushi can no longer be run from its directory with a simple "node ." command.

This is just a change for developers you mean right? We can't run sushi by doing `node .` anymore, but this doesn't change anything from the perspective of a user. I'm pretty sure, but just want to double check.

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.14.0",
   "description": "Sushi Unshortens Short Hand Inputs (FSH Compiler)",
   "scripts": {
-    "build": "rm -rf dist && tsc && cp -r src/ig/files dist/ig/files",
+    "build": "rm -rf dist && tsc && cp -r src/ig/files dist/ig/files && cp src/import/generated/*.d.ts dist/import/generated/",
     "build:watch": "tsc -w",
     "build:grammar": "bash antlr/gradlew -p antlr generateGrammarSource",
```
ngfreiter, line 8: In my testing, a rebuild of the grammar wipes out the `generated/*.d.ts` files. Probably easiest just to move those elsewhere, and would make sense as well, since they aren't generated files anyway.

package.json
mint-thompson: That's correct: when sushi is installed, the `bin` section of `package.json` is what defines the command-line operation. So, `sushi` will still run `dist/app.js`.

package.json
mint-thompson: So far, the best I've been able to figure out with tsc is that if I want it to include the generated js files, I have to set `includeJs: true`, which means that type definition files get generated. I'd much prefer for tsc to notice that I already have my own type definition files, but I'm still trying to figure out how to set up `tsconfig.json` to make that happen.

ngfreiter: Ok, I have no problem with overwriting really, just wanted to make sure I was understanding correctly.

package.json
cmoesel: And I don't think it's a big deal for developers anyway because I think we usually should be doing `ts-node src/app.ts` anyway...

cmoesel:
@ngfreiter said:
> Should we instead be updating sushi to depend on @types/antlr4?

Typically you shouldn't add types to the runtime dependencies.  Did you try adding `@type/antlr4` to **ihsus**'s dev dependencies?

package.json
cmoesel: That does seem unfortunate... I wonder if we can keep these type files somewhere under the `antlr` folder and then update `build.gradle` to copy them over to `src/import/generated` after running the ANTLR generation?

Also... was fixing these file formulaic?  Like... would it be possible to write an algorithm that lets TSC generate the bad files but then automagically fixes them?  That would help us avoid the issue of the type files getting out of sync w/ the grammar...

ngfreiter:
> @ngfreiter said:
> 
> > Should we instead be updating sushi to depend on @types/antlr4?
> 
> Typically you shouldn't add types to the runtime dependencies. Did you try adding `@type/antlr4` to **ihsus**'s dev dependencies?

Yeah that is what I did, but I guess, should I have to do that? Shouldn't I just be able to `npm install fsh-sushi` and then use `fsh-sushi`? It seemed to me like it speaks to something that is not totally right about the packaging of SUSHI that we have to add another package.

src/index.ts
```diff
@@ -0,0 +1,2 @@
+export * as fhirtypes from './fhirtypes';
+export * as fshtypes from './fshtypes';
```
cmoesel, line 2: I expect we'll need more than this for the FSH Playground.  Can you talk to @SamCitron and @jafeltra to determine what they need exported to support the FSH Playground?

package.json
mint-thompson: I like this idea best, so I'll add something in the build directory to do this.

src/index.ts
cmoesel: I thought you used `Processing`, no?

cmoesel:
@ngfreiter said:
> Yeah that is what I did, but I guess, should I have to do that? Shouldn't I just be able to `npm install fsh-sushi` and then use `fsh-sushi`? It seemed to me like it speaks to something that is not totally right about the packaging of SUSHI that we have to add another package.

I was going based on what I had seen and what's said on the [DefinitelyTyped](https://definitelytyped.org/) page:

image

That said, it seems that [Microsoft has said otherwise](https://devblogs.microsoft.com/typescript/the-future-of-declaration-files-2/) and, more specifically, in [this thread](https://github.com/Microsoft/types-publisher/issues/81) said:

image

So... it sounds like moving `@types/antlr4` to `dependencies` is probably the right thing after all.  I should have known.  That @ngfreiter is a pretty smart guy!

src/index.ts
jafeltra: The `playgroundApp` function uses functions from `Processing`, but I had thought these exports were for things that are being directly imported in the separate project. If that's not the case, then I can get a more extensive list of all the functions that `playgroundApp` is using.

cmoesel: How does the playground use functions from `Processing` without importing them?

cmoesel: Also, we wouldn't need to export each function.  We'd probably just export the whole `utils` module.

jafeltra: Ah okay I see where the disconnect is here. Right now, the changes we made for the playground all live in functions that are in SUSHI. So the playground only imports that one function, `playgroundApp`. @cmoesel and I discussed this separately and maybe this is something we'd want to reconsider. If we were to export all the functions that the `playgroundApp` function is currently using if it were to be moved into the Playground repo we'd need the following modules also exported:

- `utils` (for various Processing functions, the logger, and the Type enum)
- `export` (for the `exportFHIR` function)
- `import` (for the `RawFsh` class)
- `fhirDefs` (for the `FHIRDefinitions` class)

This shift might be something larger we want to discuss, so I'm not sure if it's easier to hold off on exporting these or if it's easier to just export them all now even if they don't get used.

mint-thompson:
> This worked well for me, and I tested out merging in the changes from master locally, and it seemed the script approach continued to work even when the grammar changed, so that's cool! I think we should consider putting `build/fixGrammarTypes` into a folder not named `build`. That is a sensible name, but it is also the name of SUSHI's default output, so for example when testing this PR, I accidentally overwrote that folder with SUSHI's output.

Ah, so that's why `build` was in `.gitignore`! Somehow, I had managed to forget the name of the default output folder since I always specify an output folder. But yes, absolutely, that default output should stay the same, and I'll move the script.

dev/fixGrammarTypes.js
```diff
@@ -0,0 +1,40 @@
+const fs = require('fs-extra');
+const path = require('path');
+const readline = require('readline');
+const EOL = require('os').EOL;
```
cmoesel, line 4: These `require` statements are making lint fail.  I was going to say to change them to `import` statements -- but then I think you need to run it through babel for node to be able to process it.

So, I think your choices are to turn off this eslint rule for this file or to convert the file to `.ts` and use `ts-node` in the npm task instead of `node`.  Since this is a TypeScript project already, I kind of lean toward the latter.

dev/fixGrammarTypes.js
```diff
@@ -0,0 +1,40 @@
+const fs = require('fs-extra');
```
cmoesel, line 1: Can you add a comment at the top of the file indicating why we need this script and what it does?

PR #525, opened 2020-07-09
Fixing contained references

### LINKED ISSUE #508: References to contained resources

ngfreiter:
So #518 was supposed to fix #508, but in fixing it, it broke references to resources within a Bundle.

Now this fixes #508 without breaking references to any inline resource. The approach taken in this PR is specific to resources in the `contained` array. I couldn't figure out how to do it without post processing, and I don't really love post processing, so if you also don't love post processing and can think of a better way, please let me know.


test/export/InstanceExporter.test.ts
```diff
@@ -1043,16 +1043,18 @@ describe('InstanceExporter', () => {
       });
     });
 
-    it('should fix an inline reference while resolving the Instance being referred to', () => {
+    it('should fix a reference to a contained resource using a relative reference', () => {
```
mint-thompson, line 1046: A question to make sure I understand correctly: in this test case, `contained` gets a reference added to it, and it has the shortened `#id` form. This happens regardless of whether or not the instance being referenced is inline or not. But when a reference to an instance (inline or not) gets put into some element other than `contained`, should it always get the full `type/id` reference?

ngfreiter: Yes so `TestOrganization` is added to the `contained` array, and then the `managingOrganization` is set to a Reference to `TestOrganization`. Based on the documentation here: https://www.hl7.org/fhir/references.html#canonical-fragments and here: http://www.hl7.org/fhir/domainresource.html#invs, my understanding is that a resource in the `contained` array should always be referenced with `#id`. And yes, my understanding is that the `containd` array is special, these special reference rules do not apply in other contexts.

ngfreiter:
@jafeltra I took out the profile test because it didn't seem to make sense to me. If you fixed the value on a profile, you would be setting a `patternReference` to a specific resource. So any instance of that profile would have to have a `reference` to that same resource. But the resource being referenced is a contained resource, so it only exists in the context of the profile, so I just couldn't see how any instance of that profile would reference this resource that is contained within the profile.

PR #526, opened 2020-07-13
Stop logging SUSHI twice

ngfreiter:
We accidentally made it so that when SUSHI starts, it says "Running SUSHI SUSHI...". This fixes it to only say SUSHI once.


cmoesel:
I don't know... now when you call `sushi -v` it says:
```
$ ts-node src/app.ts -v
v0.14.0 (implements FHIR Shorthand specification v0.13.x)
```

I think Mark wanted it to specifically say:
```
SUSHI v0.14.0 (implements FHIR Shorthand specification v0.13.x)
```
to make it more clear regarding the SUSHI version vs the FSH version...

So... the real fix you may want to make is to _remove_ `SUSHI` from the logging statement that says `Running...`

ngfreiter:
I forgot to say but @cmoesel I made the change you requested.

PR #527, opened 2020-07-14
Add sushi --init|-i command

ngfreiter:
This adds an `--init|-i` command to SUSHI. It works by running `sushi --init`, which will start a short interactive session in the command prompt that looks like:
```
Name (Default: ExampleIG): NewIG
Id (Default: fhir.example): my.id
Canonical (Default: http://example.org): http://myid.org
Status (Default: draft): active
Version (Default: 1.0.0): 2.0.0
Initialize SUSHI project in C:\Users\nfreiter\dev\NewIG? [y/n]: y
```
These values are placed into a config.yaml file, which is combined with the other files necessary to create a simple IG with SUSHI and the publisher. The init project file structure is created in a folder named based on the user input name ("NewIG" above):
```
.
├── .gitignore
├── _genonce.bat     
├── _genonce.sh           
├── _updatePublisher.bat  
├── _updatePublisher.sh 
├── _gencontinuous.bat 
├── _gencontinuous.sh  
└── fsh
    ├── config.yaml   
    ├── ig-data
    │   └── input
    │       └── pagecontent
    │           └── index.md  
    └── patient.fsh    

```
Where the config.yaml file contains the user input configuration, plus some other fields which have default values. We could allow the user to interactively set more config values, but I tried to focus on what I felt were the most important without getting too confusing. The .gitignore file contains some basic things that should be ignored when making an IG. The patient.fsh file contains a basic profile of patient, and the index.md file contains some basic content for a home page. After running `sushi --init`, you should be able to go into the created directory, run `sushi`, and then run `_updatePublisher` and `_genonce`, and all builds should succeed without error, giving a basic IG.

I'd suggest actually running `npm pack` and then installing the `.tgz` file to test this, since this PR includes changes to package.json to allow some of the static init files to be packed.

On top of code review, please also don't be shy about saying if the overall `--init` workflow should be different!


src/utils/init-project/fsh/ig-data/input/pagecontent/index.md
```diff
@@ -0,0 +1,5 @@
+# Example SUSHI Project
+
+Welcome to the example SUSHI Project. This project is meant to help people get started using SUSHI to create an Implementation Guide.
+
+Feel free to modify this index page with your own awesome content!
```
cmoesel, line 5: Since we asked the user to tell us the name of the project, can we update this file to use it instead of "Example SUSHI Project"?  It's not really an _example_...

src/utils/init-project/fsh/patient.fsh
```diff
@@ -0,0 +1,6 @@
+// This is a simple example of a FSH file.
+// This file can be renamed, and additional FSH files can be added.
+// SUSHI will look for definitions in any file using the .fsh ending.
+Profile: MyPatient
+Parent: Patient
+* name.family = "Shorty"
```
cmoesel, line 6: I'm probably being too nit-picky, but this is actually a constraint we would never recommend!  I like our examples / placeholders to follow best practice as much as possible.  So... this could be as simple as `* active = true` or `name 1..* MS` -- but probably not a profile that requires the family name to be Shorty.  Besides, isn't Shorty his given name?

ngfreiter: Shorty is both Shorty's first and last name, Shorty only needs the one name.

src/utils/Processing.ts
```diff
@@ -168,6 +170,70 @@ export function writeFHIRResources(outDir: string, outPackage: Package, snapshot
...
 
+/**
+ * Initializes an empty sample FSH within a user specified subdirectory of the current working directory
+ */
+export function init(): void {
+  logger.info(
+    'This tool will initialize a SUSHI project by creating all necessary files to build a simple IG. ' +
+      'The content of this IG will be simple default values, and should be updated. ' +
+      'The tool will now interactively create a config.yaml file for your project.'
+  );
```
cmoesel, line 181: When running this, the "info" logger feels a little weird (since it's really a dialog).  I'd suggest we just straight-up `console.log` this.  Also, you will probably think I'm crazy but... I like it when the text pops out.  And you know how I like to pop text out, right?  Boxes!
```
╭──────────────────────────────────────────────────────────╮
│ This interactive tool will use your answers to create a  │
│ working SUSHI project configured with your project's     │
│ basic information. See the SUSHI documentation for more  │
│ guidance on project structure and configuration.         │
╰──────────────────────────────────────────────────────────╯
```

src/utils/Processing.ts
```diff
@@ -168,6 +170,70 @@ export function writeFHIRResources(outDir: string, outPackage: Package, snapshot
...
+
+  // Write init directory out, including user made config.yaml, files in utils/init-project, and build scripts from ig/files
+  const outputDir = path.resolve('.', configDoc.get('name'));
+  const initProjectDir = path.join(__dirname, 'init-project');
+  if (!readlineSync.keyInYN(`Initialize SUSHI project in ${outputDir}?`)) {
+    logger.info('Aborting initialization.');
+    return;
+  }
+
+  logger.info(`Adding files to ${outputDir}`);
```
cmoesel, line 206: Again... the logging message just seems out of place here.  I'd skip it but add something at the end to signal it's done.

src/utils/Processing.ts
```diff
@@ -168,6 +170,70 @@ export function writeFHIRResources(outDir: string, outPackage: Package, snapshot
...
+    path.join(outputDir, '.gitignore')
+  );
+  // Add the _updatePublisher, _genonce, and _gencontinuous scripts
+  const scriptsDir = path.join(__dirname, '..', 'ig', 'files');
+  const scriptsDirContents = fs.readdirSync(scriptsDir);
+  scriptsDirContents
+    .filter(file => file.startsWith('_'))
+    .forEach(file => {
+      fs.copyFileSync(path.join(scriptsDir, file), path.join(outputDir, file));
+    });
```
cmoesel, line 234: I think my affinity for boxes may be a sickness:
```
╭──────────────────────────────────────────────────────────╮
│          Project initialized at: ./MyProject.            │
├──────────────────────────────────────────────────────────┤
│ Now try this:                                            │
│                                                          │
│ > cd MyProject                                           │
│ > sushi .                                                │
│                                                          │
│ For guidance on project structure and configuration see  │
│ the SUSHI documentation:  https://fshschool.org/sushi    │
╰──────────────────────────────────────────────────────────╯
```

ngfreiter: I'm on it boxman.

ngfreiter:
@cmoesel boxes, coming to a PR near you in https://github.com/FHIR/sushi/pull/527/commits/49c2a623d5bee4cb025fd27db50fe5aae0d65996

src/utils/Processing.ts
```diff
@@ -168,6 +170,90 @@ export function writeFHIRResources(outDir: string, outPackage: Package, snapshot
...
+ * Initializes an empty sample FSH within a user specified subdirectory of the current working directory
+ */
+export function init(): void {
+  console.log(
+    '\n╭──────────────────────────────────────────────────────────╮\n' +
+      '│ This interactive tool will use your answers to create a  │\n' +
+      "│ working SUSHI project configured with your project's     │\n" +
+      '│ basic information. See the SUSHI documentation for more  │\n' +
+      '│ guidance on project structure and configuration.         │\n' +
+      '╰──────────────────────────────────────────────────────────╯\n'
```
cmoesel, line 183: I hate it when I find issues and realize that you were just doing _exactly_ as I told you.  Doh!

Well, anyway... I _meant_ to delete the last sentence in this box since I put a similar thing in the box when the process completes.  I don't think we need to tell them twice -- and I think it's more appropriate in the second box.  So I'd strike the last sentence from this box.

src/utils/init-project/config.yaml
```diff
@@ -0,0 +1,25 @@
+# ╭──────────────────────────────────────ImplementationGuide───────────────────────────────────────╮
+# │  The properties below are used to create the ImplementationGuide resource. For a list of       │
+# │  supported properties, see:                                                                    │
+# │  http://build.fhir.org/ig/HL7/fhir-shorthand/branches/beta/sushi.html#ig-development           │
```
cmoesel, line 4: And yet another thing you did exactly as I told you but is probably wrong now.  I think we need to change this to the new location: https://fshschool.org/sushi/configuration/

In fact, we probably need to change this in other parts of SUSHI too (because we produce a similar message to this when we generate config files from package.json.

ngfreiter:
Ok @cmoesel the boxes are fixed!

ngfreiter:
@mint-thompson, good point, I think `name`, `id`, and `status` do have constraints on them. I guess `canonical` is also constrained to be non-whitespace? And then `version` is the wild wild west. However, we don't check that these values are valid when someone just directly uses config.yaml, so as of right now any strings you plug in to `--init` would work if you just ran SUSHI afterward. The approach that makes sense to me would be to add a validation function for config.yaml that can be called when someone makes a config.yaml by hand, and when someone creates a config.yaml using `sushi --init`. Does that make sense? Or do you think I should just add validation here now, and if we go back and do validation for all of config.yaml later, we can refactor the validation I would make now?

mint-thompson:
@ngfreiter Hmm, maybe it's best to just have the validation happen when processing the config and not when creating it? I do think it should be added, but maybe it can be separate from project initialization.

ngfreiter:
Ok @mint-thompson and @cmoesel I resolved the conflicts but unfortunately it dismissed your approval, so this should be ready for a quick re-review!

PR #529, opened 2020-07-14
Fix vulnerabilities reported by npm audit

cmoesel:
NPM was complaining due to a vulnerability in lodash.  I ran `npm audit fix` and this is what it gave me.  Seems reasonable, removes the vulnerabilities, and all the tests still pass.


cmoesel:
I am making the executive decision that this only needs one reviewer.

PR #530, opened 2020-07-14
Allow profiles/extensions to inherit *some* extensions

cmoesel:
Instead of removing ALL extensions from the parent SD, allow some to be inherited.  This is implemented by keeping track of extensions NOT to inherit.
See: https://jira.hl7.org/browse/FHIR-27535

This is submitted as a DRAFT PR until I get confirmation if the [list of uninheritable extensions](https://jira.hl7.org/browse/FHIR-27535) is complete as-is or if others (like fmm and wg) should also be disallowed.

### DIFF
--- a/src/export/StructureDefinitionExporter.ts
+++ b/src/export/StructureDefinitionExporter.ts
@@ -34,6 +34,14 @@ import {
 } from '../fhirtypes/common';
 import { Package } from './Package';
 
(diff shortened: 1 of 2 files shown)


cmoesel:
For some reason GitHub isn't recognizing the second commit I just pushed.  I'm going to close this PR and open a new one.

PR #534, opened 2020-07-15
Allow profiles/extensions to inherit *some* extensions

cmoesel:
Instead of removing ALL extensions from the parent SD, allow some to be inherited.  This is implemented by keeping track of extensions NOT to inherit.
See: https://jira.hl7.org/browse/FHIR-27535


cmoesel:
Oh, yeah, sorry.  Here it is: https://chat.fhir.org/#narrow/stream/179239-tooling/topic/Forge.20added.20extension.20explicit-type-name/near/203906359

PR #536, opened 2020-07-17
Version on canonical

ngfreiter:
This adds the ability to put a version onto a `Canonical`. If the `Canonical` is specified as `Canonical(MyDef|version)`, we will try to get the version for `MyDef`, which either comes from a rule on `MyDef`, or the version in the config if `MyDef` is defined in FSH, or the actual definition of `MyDef` if it is defined externally. If a version is found, it is appended as `|<version>` to the canonical uri. If the `Canonical` is specified as `Canonical(MyDef|someString)`, then `|someString` is appended to the canonical uri for `MyDef`. If no `version` is specified, the canonical uri is unchanged.

It's possible `|version` is not the best syntax for indicating that the version should be automatically resolved. Or maybe we shouldn't automatically resolve version at all. So if you don't like this syntax, discuss below.


test/export/StructureDefinitionExporter.test.ts
```diff
@@ -1589,6 +1589,79 @@ describe('StructureDefinitionExporter', () => {
...
+    const sd = pkg.profiles[0];
+    const fixedSystem = sd.findElement('Observation.code.coding.system');
+    // still should set the value
+    expect(fixedSystem.patternUri).toEqual(
+      'http://hl7.org/fhir/us/minimal/CodeSystem/VeryRealCodeSystem|1.0.0'
+    );
+  });
+
+  it('should log an error when no version can be found for a Canonical entity', () => {
+    delete fisher.tank.config.version;
```
mint-thompson, line 1639: Could you restore this value after the test? Otherwise, other test cases that are added after this one that are expecting that value to be present wouldn't work as expected, and it could be hard to find the reason why.

ngfreiter: Good catch, done in https://github.com/FHIR/sushi/pull/536/commits/22a8a82ca7ced87af2cc888a00329617749309aa

bkaney:
Thanks for starting this dicussion!

I am not sure about typing in literal `|version` (where you would also use with a real version, such as `|1.0.1`).

Since the FHIR spec encourages the practice to include the version, what about adding a flag in cases you _don't_ want to automatically resolve:
```
* library = Canonical(MyLibrary) (omit-version)
```

To specific/force a specific version:
```
* library = Canonical(MyLibrary|1.0.3)
```

or, an alternative to the `omit-version` flag, maybe this way to force it to have a null version (e.g. a pipe with nothing after)
```
* library = Canonical(MyLibrary|)
```

ngfreiter:
@bkaney There is definitely something to the idea that version should just be included by default. We do have some concerns about using `Canonical(Foo|)` to mean don't get the version, since the `|` in FHIR always preceeds a version. We weren't totally sure what the best path forward is, so we've put it to Zulip to see if we can get a consensus. Here is the link https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Canonical.20with.20version.

ngfreiter:
So from the discussion in https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Canonical.20with.20version we decided to just keep it simple and remove any automatic resolution of the version. A version can be added explicitly by saying:
`Canonical(Foo|1.2.3)`, but that is the only way to get a version. https://github.com/FHIR/sushi/pull/536/commits/8739442d7a5eefa6b7d951b2d3d87244e4bf698b makes that change.

Issue #531, opened 2020-07-15
Extension names are conflicting


ngfreiter:
It seems like it can cause issues in SUSHI when two extensions start with the same string of characters. As an example, something like this:
```
Extension:      FastqPaired
Id:             FastqPaired
Title:          "Fastq Paired"
Description:    "Represents paired read data generated from sequencing performed on this specimen."

* extension contains
    forward 1..1 MS and
    forwardByteSize 0..1 and
    forwardChecksum 0..1 and
    reverse 1..1 MS and
    reverseByteSize 0..1 and
    reverseChecksum 0..1

* extension[forward] ^short = "Forward read file location"
* extension[forward].value[x] only string

* extension[forwardByteSize] ^short = "Forward file size in bytes"
* extension[forwardByteSize].value[x] only decimal

* extension[forwardChecksum] ^short = "Forward file MD5 checksum"
* extension[forwardChecksum].value[x] only string

* extension[reverse] ^short = "Reverse read file location"
* extension[reverse].value[x] only string

* extension[reverseByteSize] ^short = "Reverse file size in bytes"
* extension[reverseByteSize].value[x] only decimal

* extension[reverseChecksum] ^short = "Reverse file MD5 checksum"
* extension[reverseChecksum].value[x] only string
```
does not seem to work, because the `reverse` and `reverseByteSize` extensions conflict, and SUSHI gets them confused with each other. If you rename the byte size extensions as follows, things work:
```
Id:             FastqPaired
Title:          "Fastq Paired"
Description:    "Represents paired read data generated from sequencing performed on this specimen."

* extension contains
    forward 1..1 MS and
    byteSizeForward 0..1 and
    forwardChecksum 0..1 and
    reverse 1..1 MS and
    byteSizeReverse 0..1 and
    reverseChecksum 0..1

* extension[forward] ^short = "Forward read file location"
* extension[forward].value[x] only string

* extension[byteSizeForward] ^short = "Forward file size in bytes"
* extension[byteSizeForward].value[x] only decimal

* extension[forwardChecksum] ^short = "Forward file MD5 checksum"
* extension[forwardChecksum].value[x] only string

* extension[reverse] ^short = "Reverse read file location"
* extension[reverse].value[x] only string

* extension[byteSizeReverse] ^short = "Reverse file size in bytes"
* extension[byteSizeReverse].value[x] only decimal

* extension[reverseChecksum] ^short = "Reverse file MD5 checksum"
[...]

PR #537, opened 2020-07-20
Extension names 531

007vasy:
Test for catching [this bug](https://github.com/FHIR/sushi/issues/531)


007vasy:
@ngfreiter can you have a look?

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -343,6 +343,32 @@ describe('StructureDefinitionExporter', () => {
...
+    const FooFooCardRule = new CardRule('extension[FooFoo]');
+    FooFooCardRule.min = 1;
+    FooFooCardRule.max = '1'; // * extension[sliceB].extension 1..1
+
+    const FooBarCardRule = new CardRule('extension[FooBar]');
+    FooBarCardRule.min = 0;
+    FooBarCardRule.max = '1'; // * extension[sliceB].extension 0..0
+
+    const containsRule = new ContainsRule('extension');
+    containsRule.items = [{ name: 'FooFoo' }, { name: 'FooBar' }];
```
ngfreiter, line 362: Thanks for the test! I think I know why this test isn't catching the original bug. In the original report, the first extension name, `forward` is entirely contained within the second extension name, `forwardByteSize`. In this test, that is not the case, `FooFoo`  is not a prefix of `FooBar`. If you change this test so that the `FooFoo` extension is just named `Foo`, so it has a name that is a prefix of `FooBar`, then the test should correctly fail.

531.md
```diff
@@ -0,0 +1,27 @@
+ npm run test test/export/StructureDefinitionExporter.tests
```
ngfreiter, line 1: For this to be ready to merge, we will want to remove this file.

README.md
```diff
@@ -123,4 +123,4 @@ Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
-limitations under the License.
```
ngfreiter, line 126: We'd want to undo this change as well, looks like a whitespace change.

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -343,6 +343,32 @@ describe('StructureDefinitionExporter', () => {
     expect(exported.derivation).toBe('constraint');
   });
 
+  it('should export sub-extensions, with similar starting names and different types', () => {
```
ngfreiter, line 346: Once this test is correctly failing, can you skip it using `it.skip`? That way we can merge in the failing test to master before pushing up the fix. Then, when we have a bug fix ready, we can un-skip the test.

007vasy:
@cmoesel

PR #539, opened 2020-07-20
Local CodeSystem in CaretRules

### LINKED ISSUE #516: Using local CS in code definition doesn't always work
For example, this produces the error `Resolved value "MyCS" is not a valid URI.`:

```
CodeSystem: MyCS
Id: my-cs
* #mycode "My Display" "My Definition"

Profile: MyProfile
Parent: Patient
* active ^code[0] = MyCS#mycode
```

but if we change it to use an alias instead, it works:

```
CodeSystem: MyCS
Id: my-cs
* #mycode "My Display" "My Definition"

Alias: MyCSAlias = http://hl7.org/fhir/sushi-test/CodeSystem/my-cs

Profile: MyProfile
Parent: Patient
* active ^code[0] = MyCSAlias#mycode
```

Note that these _also_ work (using the CS name, _not_ the alias):

```
Profile: MyOtherProfile
Parent: Observation
* code = MyCS#mycode

Instance: MyInstance
InstanceOf: Observation
* code = MyCS#mycode
```

So this would suggest that this bug may be restricted to using local code systems in codes in caret rules.

ngfreiter:
This fixes #516.

CodeSystems defined in FSH were not being replaced with their canonical URLs when they were used in a CaretValueRule. It would work on a FixedValueRule, but not a CaretValueRule. So if the user did:
```
CodeSystem: MyCS
Id: my-cs
* #mycode "My Display" "My Definition"

Profile: MyProfile
Parent: Patient
* active ^code[0] = MyCS#mycode
```
The `MyCS` would not be converted to the URL for `MyCS`, resulting in an error. This was also actually an issue for References too. References to locally defined resources would not be resolved into the form `<resource.resourceType>/<resource.id>`. The code for replacing References and CodeSystems is in the same spot, so this PR fixes resolution for both of them.

To test this, you can use the example given above. On `master`, that will result in a `Resolved value "MyCS" is not a valid URI.` error. On this branch, it should build without error, and the value will be correctly replaced. Additionally, the tests on https://github.com/FHIR/sushi/commit/ffbe61805345e1ce141b2da4e88afb0cfc55fe3d fail on that commit, but then the next commit fixes them by adding the bug fix.

### DIFF
--- a/src/fhirtypes/common.ts
+++ b/src/fhirtypes/common.ts
@@ -226,12 +226,8 @@ export function getArrayIndex(pathPart: PathPart): number {
  * @param {Fishable} fisher - A fishable implementation for finding definitions and metadata
  * @returns {FixedValueRule} a clone of the rule if replacing is done, otherwise the original rule
  */
-export function replaceReferences(
-  rule: FixedValueRule | CaretValueRule,
-  tank: FSHTank,
-  fisher: Fishable
-): FixedValueRule | CaretValueRule {
-  let clone: FixedValueRule | CaretValueRule;
+export function replaceReferences<T>(rule: T, tank: FSHTank, fisher: Fishable): T {
+  let clone: T;
   const value = getRuleValue(rule);
   if (value instanceof FshReference) {
     const instance = tank.fish(value.reference, Type.Instance) as Instance;
@@ -271,7 +267,7 @@ export function replaceReferences(
  * @param rule - The rule to get a value from
  * @returns - The value on the rule
  */
-function getRuleValue(rule: FixedValueRule | CaretValueRule): FixedValueType {
+function getRuleValue(rule: any): FixedValueType {
   if (rule instanceof FixedValueRule) {
     return rule.fixedValue;
   } else if (rule instanceof CaretValueRule) {
--- a/src/fhirtypes/common.ts
+++ b/src/fhirtypes/common.ts
@@ -12,7 +12,9 @@ import {
   InsertRule,
   ConceptRule,
   ValueSetConceptComponentRule,
-  SdRule
+  SdRule,
+  CaretValueRule,
+  FixedValueType
 } from '../fshtypes/rules';
 import {
   FshReference,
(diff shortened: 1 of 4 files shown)


test/export/StructureDefinitionExporter.test.ts
```diff
@@ -2423,6 +2423,98 @@ describe('StructureDefinitionExporter', () => {
...
+    exporter.exportStructDef(profile);
+    const sd = pkg.profiles[0];
+
+    expect(sd.jurisdiction[0].coding[0]).toEqual({
+      code: 'foo',
+      system: 'http://hl7.org/fhir/us/minimal/CodeSystem/bar-id'
+    });
+  });
+
+  it('should apply a CodeSystem CaretValueRule on an ED and replace the Reference', () => {
```
cmoesel, line 2494: Typo:
> and replace the Reference

ngfreiter:
> Changing my "requested changes" to a normal comment so I don't hold up this being merged while I'm OOTO on Thursday. So... if @ngfreiter addresses my comments and someone else reviews and approves, feel free to merge.

I appreciate the confidence that I would get this done early enough for someone else to review, but it wasn't meant to be.

src/fhirtypes/common.ts
```diff
@@ -271,7 +267,7 @@ export function replaceReferences(
  * @param rule - The rule to get a value from
  * @returns - The value on the rule
  */
-function getRuleValue(rule: FixedValueRule | CaretValueRule): FixedValueType {
+function getRuleValue(rule: any): FixedValueType {
```
cmoesel, line 270: When I tested this locally, I did not have to change this to `any`.  I think I remember it working w/ `FixedValueRule | CaretValueRule` just fine.  Did you *have* to change this?

ngfreiter: I get:
```
Argument of type 'T' is not assignable to parameter of type 'FixedValueRule | CaretValueRule'.
  Type 'T' is not assignable to type 'CaretValueRule'
```
when I change it, which seems sensible right? Becuase TypeScript doesn't know what `T` is, and can't be sure it'll work for a FixedValueRule or CaretValueRule.

cmoesel: Oh, I didn't notice... you didn't constrain T at all.  It should be fine if you do this instead:
```ts
export function replaceReferences<T extends FixedValueRule | CaretValueRule>(rule: T, tank: FSHTank, fisher: Fishable): T {
```

ngfreiter: Ohhhh that makes more sense

PR #541, opened 2020-07-21
Translate some ANTLR errors to more user friendly messages

### LINKED ISSUE #318: Annoying white space dependency
THis is OK:   Alias:   $OrgType = http://terminology.hl7.org/CodeSystem/organization-type

This is NOT:  Alias:   $OrgType= http://terminology.hl7.org/CodeSystem/organization-type

cmoesel:
Provide user friendly messages if we detect the following problems:
* missing space around =
* missing space around ->
* missing space after *

This also addresses one issue that can arise when ANTLR hits a parsing error and tries to keep going -- which is that sometimes we'll process a rule that has missing required tokens.  In this case, we want to be a bit more defensive and due a null check before dereferencing the token.

This also adds a global.d.ts to import jest-extended to allow for jest-extended matchers to work properly in VS Code. This has no effect on the distributed module.

Fixes #318


cmoesel:
That caret-rule thing was pretty annoying, but it's fixed now. ;-)

PR #552, opened 2020-07-29
Allow quantities to be fixed to specializations

### LINKED ISSUE #515: Can't set age value in instance using quantity syntax
Attempting this (on instance of `FamilyMemberHistory`):
```
* condition.onsetAge = 30.0 'a'
```
Results in this:
```
error Cannot fix number value: 30. Value does not match element type: Age
```

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Instance.20age

cmoesel:
Allow quantities to be fixed to specializations like Age and Duration. Note that we do not do any extra validation (e.g., we don't ensure that Age uses valid units for Age, or that a Count uses the '1' unit) -- we leave that for the IG Publisher to validate.

To test this, you can first run the tests at commit 16a827e without the fix.  Four tests will fail.  Then run them again at commit ed136db (the fix) and they will succeed.

You can also test them via a profile like this:

```
Profile: ConditionWithFixedOnsetAge
Parent: Condition
* onset[x] only Age
* onset[x] = 42 'a'
```

Or an instance like this:
```
Instance: ConditionAcquiredAt42
InstanceOf: Condition
* subject = Reference(Patient/abc)
* onsetAge = 42 'a'
```

Fixes #515

### DIFF
--- a/test/fhirtypes/ElementDefinition.fixFshQuantity.test.ts
+++ b/test/fhirtypes/ElementDefinition.fixFshQuantity.test.ts
@@ -5,12 +5,15 @@ import { FHIRDefinitions } from '../../src/fhirdefs/FHIRDefinitions';
 import { StructureDefinition } from '../../src/fhirtypes/StructureDefinition';
 import { FshQuantity, FshCode } from '../../src/fshtypes';
 import { TestFisher } from '../testhelpers';
+import { ElementDefinitionType } from '../../src/fhirtypes';
 
 describe('ElementDefinition', () => {
   let defs: FHIRDefinitions;
+  let condition: StructureDefinition;
   let observation: StructureDefinition;
   let fshQuantity1: FshQuantity;
   let fshQuantity2: FshQuantity;
+  let fshQuantityAge: FshQuantity;
   let fisher: TestFisher;
 
   beforeAll(() => {
@@ -26,6 +29,8 @@ describe('ElementDefinition', () => {
     observation = fisher.fishForStructureDefinition('Observation');
     fshQuantity1 = new FshQuantity(1.23, new FshCode('mm', 'http://unitsofmeasure.org'));
     fshQuantity2 = new FshQuantity(1.24, new FshCode('mm', 'http://unitsofmeasure.org'));
+    condition = fisher.fishForStructureDefinition('Condition');
+    fshQuantityAge = new FshQuantity(42.0, new FshCode('a', 'http://unitsofmeasure.org'));
   });
 
   describe('#fixFshQuantity', () => {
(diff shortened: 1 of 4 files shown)


test/fhirtypes/ElementDefinition.fixFshQuantity.test.ts
```diff
@@ -55,6 +60,34 @@ describe('ElementDefinition', () => {
       expect(referenceRangeLow.patternQuantity).toBeUndefined();
     });
 
+    it('should fix a FshQuantity to a quantity specialization', () => {
+      const onsetX = condition.elements.find(e => e.id === 'Condition.onset[x]');
+      onsetX.type = [new ElementDefinitionType('Age')];
+      onsetX.fixValue(fshQuantityAge, false, fisher);
+      // @ts-ignore we don't elaborate over all possible pattern[x] in ElementDefinition
```
ngfreiter, line 67: I agree that we don't elaborate over all `pattern[x]`, but I think every other type that can be set by some syntax in FSH directly, so all primitive types, and then the select complex types with a special syntax like CodeableConcept, Quantity, or Ratio are explicitly on the ElementDefinition. So is there precedent that `patternAge` should be on the ElementDefinition?

cmoesel: I think the difference is that for those other ones we explicitly reference them in the production code (e.g., we have `src` code -- not just `test` code -- that uses `patternQuantity` directly). But we do not have `src` code that uses `patternAge` directly -- we only directly reference it in `test` code...

I could be convinced to add it... it just becomes unused when you package up the module without tests.

ngfreiter: I'm not sure that is generally the case. We do have `src` code that uses `patternQuantity`, but I can't find any that uses `patternCodeableConcept`, `patternCoding`, or most of the `pattern` elements it seems. From what I am seeing, `patternQuantity` is kind of special in that it does actually get used outside of tests.

That being said, I am still willing to be convinced we don't need to add `patternAge`, but I'm not convinced just yet.

cmoesel: Ah... Hmm... Maybe we used to reference them directly but it got refactored?  Or maybe we never did?  I don't know.  I'll just add `patternAge` and `fixedAge` so you stop bugging me. ;-)

cmoesel: OK.  Prepare to be overwhelmed by the glory of `fixedAge` and `patternAge` in 68f24d8.

Issue #532, opened 2020-07-15
Changing URLs break extension support


palfrey:
#407 fixed URL support for code systems and valuesets, and the code seems to indicate that it would do the same for extensions. However I'm still seeing errors.

Example FSH:
```
Extension: TestExtension
* ^url = "http://foo.bar.baz"

Profile: TestProfile
Parent: Communication
* payload.extension contains TestExtension named ContentConcept 1..
// * payload.extension contains http://foo.bar.baz named ContentConcept 1..
* payload.extension[ContentConcept].value[x] only CodeableConcept
```

gets me `error No element found at path payload.extension[ContentConcept].value[x] for TestProfile, skipping rule`. Switching it out for the commented line gets me `Cannot create ContentConcept extension; unable to locate extension definition for: http://foo.bar.baz.` instead.

Any thoughts? I need the URL to be a particular value for historical reasons with an existing data set.

cmoesel:
Hi @palfrey -- I agree, this _looks_ like it should work (both the uncommented and commented code should each work).  We'll take a look at it.

palfrey:
Awesome. Just tested the fixes locally, and fixes my original issue. Thanks!

PR #554, opened 2020-07-29
Fix custom extension urls

cmoesel:
If you override an extension's URL with a caret rule, the extension definition should fix the 'Extension.url' element value to the overridden URL. Prior to this fix, it fixed it to the canonical-based URL instead.  This caused problems if you tried to access and use that extension in a profile.

To test, you can run the test from commit f635fff without the fix.  It should fail.  If you run it with the fix (commit 43671f8) it should succeed.

You can also test it using the example FSH for #532:
```
Extension: TestExtension
* ^url = "http://foo.bar.baz"

Profile: TestProfile
Parent: Communication
* payload.extension contains TestExtension named ContentConcept 1..
// * payload.extension contains http://foo.bar.baz named ContentConcept 1..
* payload.extension[ContentConcept].value[x] only CodeableConcept
```

Fixs #532

### DIFF
--- a/src/export/StructureDefinitionExporter.ts
+++ b/src/export/StructureDefinitionExporter.ts
@@ -30,7 +30,8 @@ import {
   splitOnPathPeriods,
   applyMixinRules,
   cleanResource,
-  applyInsertRules
+  applyInsertRules,
+  getUrlFromFshDefinition
 } from '../fhirtypes/common';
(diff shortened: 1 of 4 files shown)


cmoesel:
Oh yeah.  I almost forgot.  This also fixes fishing by URL when the URL is overridden via a caret rule.

PR #556, opened 2020-07-30
id is not required in FSHOnly mode

mint-thompson:
The configured id is used when generating an IG, but is not used
otherwise. Do not require it in FSHOnly mode. The two remaining required
configuration values are:
- canonical: used to set the url on generated resources
- fhirVersion: used to download FHIR resources

The version property is no longer required. It is used to set the
version on generated resources, but this is an optional property.

### DIFF
--- a/src/import/importConfiguration.ts
+++ b/src/import/importConfiguration.ts
@@ -48,7 +48,7 @@ import { FshCode } from '../fshtypes';
 
 const MINIMAL_CONFIG_PROPERTIES = ['canonical', 'fhirVersion'];
 // Properties that are only relevant when an IG is going to be generated from output, and have no informational purpose
-const MINIMAL_IG_ONLY_PROPERTIES = ['id'];
+const MINIMAL_IG_ONLY_PROPERTIES = ['id', 'name', 'status', 'copyrightYear', 'releaseLabel'];
 const IG_ONLY_PROPERTIES = [
   'contained',
   'extension',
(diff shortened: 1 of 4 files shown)


src/import/importConfiguration.ts
```diff
@@ -99,6 +100,22 @@ export function importConfiguration(yaml: YAMLConfiguration | string, file: stri
...
+    )
+  ) {
+    logger.error(
+      'SUSHI minimally requires the following configuration properties to generate an IG: ' +
+        MINIMAL_IG_ONLY_PROPERTIES.join(', ') +
+        '.',
+      { file }
+    );
+    throw new Error('Minimal config not met');
+  }
```
cmoesel, line 118: I imagine it might be annoying for someone if they are missing a core required property and get an error that `canonical` and `fhirVersion` are required, then they add those and get another error that `id` is required.  It would probably be a better user experience to list _everything_ required based on their config one time.

You could probably reduce some duplicate code as well if you did something more like:
```ts
const minimalProperties = yam.FSHOnly ? MINIMAL_CONFIG_PROPERTIES : [...MINIMAL_CONFIG_PROPERTIES, ...MINIMAL_IG_PROPERTIES];
// ...
```
Or if you define `MINIMAL_IG_PROPERTIES` to include the `MINIMAL_CONFIG_PROPERTIES`, then the above can be even simpler.

mint-thompson: That sounds like a reasonable change to me. Right now, the issue I'm still twisting around is the fact that both `copyrightYear` and `copyrightyear` are allowed as spellings of the same property, which means the current way of checking for minimal properties will have to accommodate that.

cmoesel: Dang.  I guess you could make the minimal constants be lists of regular expressions instead of lists of strings... Then you can just include `/copyright[Yy]ear/`.  It seems a little silly but also seems like the least complex solution...

PR #561, opened 2020-08-03
Support fixing inline instances on profiles

### LINKED ISSUE #511: SUSHI does not support pattern[x]/fixed[x] w/ inline instances
The following should be supported, but currently is not:
```
Instance: USPostalAddress
InstanceOf: Address
Usage: #inline
* type = #postal
* country = "US"

Profile: USPatient
Parent: Patient
* address = USPostalAddress
```

As written, this should create the following attribute in the `USPatient` profile's `Patient.address` element:
```json
"patternAddress": {
  "type": "postal",
  "country": "US"
}
```

Unless I'm mistaken, supporting something like the FSH above is the _only_ way to get a `patternAddress` into an element without introducing a syntax that allows `{` `}` objects in FSH.

jafeltra:
This PR fixes #511 and supports fixing inline instances on profiles. The example in the issue is now supported, and the tests show a few other examples of what is not allowed.

I do have a couple questions about the implementation:
- In order to reach the `InstanceDefinition` case in `ElementDefinition.fixValue`, I had to make the change to the `applyMixins` function to not set the constructor name. I'm not entirely sure if this is the right approach, because it looks like the original implementation is the way that the [TypeScript documentation]( https://www.typescriptlang.org/docs/handbook/mixins.html#mixin-sample) describes. However, without this change, instances will be an instance of the `InstanceDefinition` class, but their constructor name is `HasId` because the `applyMixins` overwrites this. If others don't think this change is correct, I can always have `ElementDefinition.fixValue` go into the `InstanceDefinition` case whenever we see a `HasId` constructor name.
- I got rid of the error in the importer that we cannot fix instances to profiles since we are now allowing that. However, I'm not sure if there were other cases that we had in mind when writing that error that we still don't want to support. If so, I can add an error back in for those cases.

### DIFF
--- a/test/fhirtypes/ElementDefinition.fixInstanceDefinition.test.ts
+++ b/test/fhirtypes/ElementDefinition.fixInstanceDefinition.test.ts
@@ -1,5 +1,4 @@
 import path from 'path';
-import { cloneDeep } from 'lodash';
 import { loadFromPath } from '../../src/fhirdefs/load';
 import { FHIRDefinitions } from '../../src/fhirdefs/FHIRDefinitions';
 import { StructureDefinition, InstanceDefinition } from '../../src/fhirtypes';
@@ -9,7 +8,6 @@ describe('ElementDefinition', () => {
   let defs: FHIRDefinitions;
   let patient: StructureDefinition;
   let observation: StructureDefinition;
-  let location: StructureDefinition;
   let fisher: TestFisher;
 
   beforeAll(() => {
@@ -24,7 +22,6 @@ describe('ElementDefinition', () => {
   beforeEach(() => {
     patient = fisher.fishForStructureDefinition('Patient');
     observation = fisher.fishForStructureDefinition('Observation');
-    // location = fisher.fishForStructureDefinition('Location');
   });
 
   describe('#fixInstance', () => {
--- a/test/fhirtypes/ElementDefinition.fixInstanceDefinition.test.ts
+++ b/test/fhirtypes/ElementDefinition.fixInstanceDefinition.test.ts
@@ -1,0 +1,172 @@
+import path from 'path';
+import { cloneDeep } from 'lodash';
+import { loadFromPath } from '../../src/fhirdefs/load';
+import { FHIRDefinitions } from '../../src/fhirdefs/FHIRDefinitions';
+import { StructureDefinition, InstanceDefinition } from '../../src/fhirtypes';
+import { TestFisher } from '../testhelpers';
+
+describe('ElementDefinition', () => {
+  let defs: FHIRDefinitions;
+  let patient: StructureDefinition;
+  let observation: StructureDefinition;
+  let location: StructureDefinition;
+  let fisher: TestFisher;
+
+  beforeAll(() => {
+    defs = new FHIRDefinitions();
+    loadFromPath(
+      path.join(__dirname, '..', 'testhelpers', 'testdefs', 'package'),
+      'testPackage',
+      defs
+    );
+    fisher = new TestFisher().withFHIR(defs);
+  });
+  beforeEach(() => {
+    patient = fisher.fishForStructureDefinition('Patient');
+    observation = fisher.fishForStructureDefinition('Observation');
+    // location = fisher.fishForStructureDefinition('Location');
+  });
+
+  describe('#fixInstance', () => {
+    it('should fix an allowed type of an instance', () => {
+      const addressInstance = new InstanceDefinition();
+      addressInstance._instanceMeta.name = 'USPostalAddress';
+      addressInstance._instanceMeta.sdType = 'Address';
+      addressInstance._instanceMeta.usage = 'Inline';
+      addressInstance.country = 'US';
+      addressInstance.type = 'postal';
+
+      const address = patient.elements.find(e => e.id === 'Patient.address');
+      address.fixValue(addressInstance);
+      expect(address.patternAddress).toEqual({ country: 'US', type: 'postal' });
+      expect(address.fixedAddress).toBeUndefined();
+    });
+
+    it('should fix an allowed type of an instance (exactly)', () => {
+      const addressInstance = new InstanceDefinition();
+      addressInstance._instanceMeta.name = 'USPostalAddress';
+      addressInstance._instanceMeta.sdType = 'Address';
+      addressInstance._instanceMeta.usage = 'Inline';
+      addressInstance.country = 'US';
+      addressInstance.type = 'postal';
+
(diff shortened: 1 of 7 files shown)


test/fhirtypes/ElementDefinition.fixInstanceDefinition.test.ts
```diff
@@ -0,0 +1,169 @@
...
+      address.fixValue(workAddress, true);
+      expect(address.fixedAddress).toEqual({ country: 'US', type: 'postal', use: 'work' });
+      expect(() => {
+        address.fixValue(homeAddress, true);
+      }).toThrow(
+        'Cannot fix {"country":"US","type":"postal","use":"home"} to this element; a different Address is already fixed: {"country":"US","type":"postal","use":"work"}'
+      );
+    });
+
+    it('should throw FixedToPatternError when trying to change fixed[x] to patternp[x]', () => {
```
mint-thompson, line 135: small typo here: `patternp`

test/import/FSHImporter.Profile.test.ts
```diff
@@ -1352,7 +1352,7 @@ describe('FSHImporter', () => {
         assertFixedValueRule(profile.rules[0], 'identifier.system', 'http://example.org');
       });
 
-      it('should log an error and skip the rule when parsing fixed value Resource rule', () => {
+      it('should parsing fixed value Resource rule', () => {
```
mint-thompson, line 1355: slightly awkward grammar here, maybe `should parse a fixed value Resource rule`?

test/fhirtypes/ElementDefinition.fixInstanceDefinition.test.ts
jafeltra: That pesky "p" key is too close to that pesky "[" key.

test/import/FSHImporter.Profile.test.ts
jafeltra: Oh yes wow I should have read that more carefully after deleting.

jafeltra:
@mint-thompson thanks for reviewing! I think the `typeof value === 'object'` case is actually catching a few other cases, including classes that we don't support. One example I saw was the [tests for fixing a JavaScript date](https://github.com/FHIR/sushi/blob/master/test/fhirtypes/ElementDefinition.fixValue.test.ts#L29). In that case, we go into that `typeof value === 'object'` case, but we don't have a case for it in the switch statement below, so we get a MismatchedTypeError.

I had originally tried to add a case for `value instanceof InstanceDefinition` in addition to the generic object one thinking it would fall into the generic one if it wasn't an InstanceDefinition, but TypeScript was unhappy about that. I'm not exactly sure why it was upset (although I thought I did understand why at some point last week), so maybe I'll take another look at that case.

jafeltra:
I can't figure out exactly why TypeScript it upset about having both the `value instanceof InstanceDefinition` and `typeof value === 'object'` in the same if statement. My hunch was that it had to do with the `[key: string]: any;` that is on InstanceDefinition, but I'm not completely convinced of that now. If others think this is an issue, I may talk to you all about the best way forward, but otherwise, I think I'm going to pause my investigation until I hear more.

I also updated the few typos @mint-thompson pointed out!

cmoesel:
@jafeltra - I think it doesn't allow it because it knows that `value` should be one of the `FixedValueType` types; and it also knows that once you explicitly test for `InstanceDefinition`, there are no other types left that would have a constructor -- so _technically_ `value.constructor` _should_ be undefined every time at that point.  But... as you noted, we have some tests that force other values in there anyway -- so it is possible for `value.constructor` to still be defined, but we likely bent some rules somewhere to make that happen.

Anyway, I found an approach that I think is more or less the same, but keeps TypeScript happy (and the tests still pass):
image

What do you think of that?

cmoesel:
Oh wait.  That last line should probably be:
```ts
type = (type === 'object' && value.constructor?.name) ?? type;
```

jafeltra:
Ah! I had completely forgotten about the fact that it was taking `FixedValueType` into consideration. This makes me feel better. I like that approach, and it lets me get rid of the change in the `applyMixins` function, so I'll push up the change now.

test/fhirtypes/ElementDefinition.fixInstanceDefinition.test.ts
```diff
@@ -0,0 +1,169 @@
...
+  });
+
+  describe('#fixInstance', () => {
+    it('should fix an allowed type of an instance', () => {
+      const addressInstance = new InstanceDefinition();
+      addressInstance._instanceMeta.name = 'USPostalAddress';
+      addressInstance._instanceMeta.sdType = 'Address';
+      addressInstance._instanceMeta.usage = 'Inline';
+      addressInstance.country = 'US';
+      addressInstance.type = 'postal';
```
cmoesel, line 34: Technically, `type` is a `code`, so I think this should be:
```ts
addressInstance.type = new FshCode('postal'); // * type = #postal
```

test/fhirtypes/ElementDefinition.fixInstanceDefinition.test.ts
```diff
@@ -0,0 +1,169 @@
...
+      expect(address.fixedAddress).toBeUndefined();
+    });
+
+    it('should fix an allowed type of an instance (exactly)', () => {
+      const addressInstance = new InstanceDefinition();
+      addressInstance._instanceMeta.name = 'USPostalAddress';
+      addressInstance._instanceMeta.sdType = 'Address';
+      addressInstance._instanceMeta.usage = 'Inline';
+      addressInstance.country = 'US';
+      addressInstance.type = 'postal';
```
cmoesel, line 48: Same as above, I think this should be a `FshCode`, not a `string`.

test/fhirtypes/ElementDefinition.fixInstanceDefinition.test.ts
jafeltra: Since this is an `InstanceDefinition`, I think that the code would be represented just as the key/value pair with a string for 'postal'. I think if this was setting a rule on an Instance, then we'd use FshCode, but since this is the representation after it's exported, I think it might just be a string. I could be wrong though, so let me know if you still disagree.

cmoesel: Ah, yes.  You are right.  (Of course!)

test/fhirtypes/ElementDefinition.fixInstanceDefinition.test.ts
cmoesel: Nevermind!

jafeltra:
The button has been pressed!

PR #566, opened 2020-08-11
Don't change type code for id and url

### LINKED ISSUE #543: String type invokes http://hl7.org/fhir/StructureDefinition/structuredefinition-fhir-type extension
Running the following FSH
```
Profile: TestObservation
Parent: Observation
* code.id only string
* effective[x] only dateTime
```
gets the differential element section
```
      {
        "id": "Observation.code.id",
        "path": "Observation.code.id",
        "type": [
          {
            "code": "string",
            "extension": [
              {
                "url": "http://hl7.org/fhir/StructureDefinition/structuredefinition-fhir-type",
                "valueUrl": "string"
              }
            ]
          }
        ]
      },
      {
        "id": "Observation.effective[x]",
        "path": "Observation.effective[x]",
        "type": [
          {
            "code": "dateTime"
          }
        ]
      }
```
I was surprised by the extension section for `string`. Didn't think that would be needed for built-in types, and indeed it wasn't needed for `dateTime`. Had a quick dig through the sushi code, but couldn't seem to see anything that said why dateTime works with this but not string.

mint-thompson:
Fixes #543 and completes task [CIMPL-467](https://standardhealthrecord.atlassian.net/browse/CIMPL-467)

The id and url elements need to always have a type code of http://hl7.org/fhirpath/System.String. Do not let this code change when applying type constraints to these elements.

### DIFF
--- a/test/export/StructureDefinitionExporter.test.ts
+++ b/test/export/StructureDefinitionExporter.test.ts
@@ -1392,6 +1392,45 @@ describe('StructureDefinitionExporter', () => {
     expect(lastMessage).toMatch(/File: FooQuantity\.fsh.*Line: 6 - 11\D*/s);
   });
 
+  it('should apply an OnlyRule to constrain an id element', () => {
+    const specialString = new Profile('SpecialString');
+    specialString.parent = 'string';
+    doc.profiles.set(specialString.name, specialString);
+    const profile = new Profile('MyObservation');
+    profile.parent = 'Observation';
+    const onlyRule = new OnlyRule('code.id');
+    onlyRule.types = [{ type: 'SpecialString' }];
+    profile.rules.push(onlyRule);
+    exporter.exportStructDef(profile);
+    const sd = pkg.profiles[0];
+    const codeId = sd.findElement('Observation.code.id');
+    expect(codeId.type.length).toBe(1);
+    expect(codeId.type[0].getActualCode()).toBe('http://hl7.org/fhirpath/System.String');
+    expect(codeId.type[0].profile.length).toBe(1);
+    expect(codeId.type[0].profile[0]).toBe(
+      'http://hl7.org/fhir/us/minimal/StructureDefinition/SpecialString'
+    );
+  });
+
+  it('should apply an OnlyRule to constrain a url element', () => {
+    const specialUri = new Profile('SpecialUri');
+    specialUri.parent = 'uri';
+    doc.profiles.set(specialUri.name, specialUri);
+    const extension = new Extension('SpecialExtension');
+    const onlyRule = new OnlyRule('url');
+    onlyRule.types = [{ type: 'SpecialUri' }];
+    extension.rules.push(onlyRule);
+    exporter.exportStructDef(extension);
+    const sd = pkg.extensions[0];
+    const url = sd.findElement('Extension.url');
+    expect(url.type.length).toBe(1);
+    expect(url.type[0].getActualCode()).toBe('http://hl7.org/fhirpath/System.String');
+    expect(url.type[0].profile.length).toBe(1);
+    expect(url.type[0].profile[0]).toBe(
+      'http://hl7.org/fhir/us/minimal/StructureDefinition/SpecialUri'
+    );
+  });
+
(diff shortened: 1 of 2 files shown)


test/export/StructureDefinitionExporter.test.ts
```diff
@@ -1392,6 +1392,45 @@ describe('StructureDefinitionExporter', () => {
...
+    exporter.exportStructDef(profile);
+    const sd = pkg.profiles[0];
+    const codeId = sd.findElement('Observation.code.id');
+    expect(codeId.type.length).toBe(1);
+    expect(codeId.type[0].getActualCode()).toBe('http://hl7.org/fhirpath/System.String');
+    expect(codeId.type[0].profile.length).toBe(1);
+    expect(codeId.type[0].profile[0]).toBe(
+      'http://hl7.org/fhir/us/minimal/StructureDefinition/SpecialString'
+    );
+  });
```
ngfreiter, line 1413: I'm wondering if we should allow an only rule on one of these FHIRPath `System.<type>` types at all. The `SpecialString` profile is a profile of `string`, which though it is a FHIR primitive, has several elements, one of which is `value`, which has the FHIRPath `System.String` type. So I'm not sure it makes sense to say that `id` must conform to a profile on the FHIR `string`, when `id` does not have the elements on the `FHIR` string. I feel like in a sense the `System.<type>` types are being treated as the "true" primitives by FHIR, so it seems like we shouldn't be able to constrain their types any further.

I could be way off here, this is a confusing issue, but let me know what you think.

mint-thompson: That's a really good point. Is it even possible to Profile a FHIRPath primitive? I wasn't able to find StructureDefinitions for those types, so maybe not. So then to take that idea further, a profile on `integer` couldn't apply any kind of OnlyRule on `integer.value`, because there's no Profile or subtype that could be applied. I think you might be on to something with a more general rule of "If an element's type.code is a FHIRPath primitive, OnlyRule is not allowed". As a quick check, I looked at [the profile for Observation](http://hl7.org/fhir/observation.profile.json.html), and the only elements in there that have a `type.code` of a FHIRPath primitive are `id` elements.

I'm going to change the implementation to just give a hard "no" when trying to apply an OnlyRule to a FHIRPath primitive. The rule can safely be skipped. I think this can be logged as an error, as it is an invalid rule.

mint-thompson: So now, I'm not so sure. Maybe profiling an `id` is fine. Or, maybe SUSHI is allowing it when it shouldn't. Consider the following FSH input and FHIR output, which SUSHI processes without errors.
FSH:
```
Instance: UnusualObservation
InstanceOf: Observation
* status = #final
* code = http://loinc.org#10196-4
* valueString = "Value"
* valueString.id = "First ID"
* valueString.id.id = "Second ID"
```
FHIR:
```
{
  "id": "UnusualObservation",
  "status": "final",
  "code": {
    "coding": [
      {
        "code": "10196-4",
        "system": "http://loinc.org"
      }
    ]
  },
  "valueString": "Value",
  "_valueString": {
    "id": "First ID",
    "_id": {
      "id": "Second ID"
    }
  }
}
```
So either SUSHI is doing something wrong, and that inner `id` shouldn't be there, or SUSHI is behaving correctly, and an `id` element has an `id` element, indicating that it most likely is conforming to the FHIR `string` profile.

ngfreiter: To further give evidence to the idea that SUSHI is behaving correctly, it looks like we did do this intentionally: https://github.com/FHIR/sushi/pull/52. That PR contains the code that causes `id` to be unfolded even though it has the FHIRPath type, so it seems like we did this on purpose. So if that is the case, then I think what you have currently *is* the correct behavior, and I've unnecessarily taken us for a little loop here. 

Do you plan to make any changes relating to https://github.com/FHIR/sushi/pull/566#discussion_r469277917? It's up to you, but since I feel convinced that this behavior is correct now, I will approve if you don't plan to make any more changes.

PR #570, opened 2020-08-12
Add Reslicing

### LINKED ISSUE #542: SUSHI should handle reslicing better
Currently, when someone tries to reslice a slice:
```
Profile: TestProcedure
Parent: Procedure
* extension contains SlicedExtension named slicedExtension 0..*
* extension[slicedExtension] ^slicing.discriminator.type = #pattern
* extension[slicedExtension] ^slicing.discriminator.path = "extension[namepart].valueString"

* extension[slicedExtension] contains test1 0..1 and test2 0..1
* extension[slicedExtension][test1].extension[namepart].valueString = "test1" (exactly)
* extension[slicedExtension][test1].extension[valuepart].valueCodeableConcept from TransfusionProcedureCodeVS (extensible)

Extension: SlicedExtension
* extension contains namepart 1..1 valuepart 1..1
* extension[namepart].value[x] only string
* extension[valuepart].value[x] only CodeableConcept
```
SUSHI does not correctly handle the line doing the reslicing:
```
extension[slicedExtension] contains test1 0..1 and test2 0..1
```
It looks like SUSHI adds the element with an `id` of `Procedure.extension:slicedExtension:test1`, which is not correct, the `id` should be `Procedure.extension:slicedExtension/test1`. The `findElementByPath` logic then fails to find the element, since it has an incorrectly specified `id`. If we want to explicitly not support reslicing, that is also a possibility, but error messages would need to be more clear about that.

ngfreiter:
Fixes #542 by adding support for reslicing in `ContainsRule`s.

I tested this using the example given in #542, and it appears to work for me. As far as I can gather from the spec, there isn't really anything special about reslicing outside of the `sliceName` (and as a result the `id`) having the format `initialSlice/reslice1/reslice2`. As a result, this updates the `addSlice` function to handle the case in which the element we are adding to is itself a slice, in which case we have to reslice.

### DIFF
--- a/src/fhirtypes/StructureDefinition.ts
+++ b/src/fhirtypes/StructureDefinition.ts
@@ -1,5 +1,6 @@
 import upperFirst from 'lodash/upperFirst';
 import cloneDeep from 'lodash/cloneDeep';
+import escapeRegExp from 'lodash/escapeRegExp';
 import { ElementDefinition, ElementDefinitionType, LooseElementDefJSON } from './ElementDefinition';
 import { Meta } from './specialTypes';
 import { Identifier, CodeableConcept, Coding, Narrative, Resource, Extension } from './dataTypes';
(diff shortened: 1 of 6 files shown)


src/fhirtypes/StructureDefinition.ts
```diff
@@ -133,12 +134,13 @@ export class StructureDefinition {
...
-      if (element.id.startsWith(`${currentId}.`) || element.id.startsWith(`${currentId}:`)) {
+      if (new RegExp(`${escapeRegExp(currentId)}[.:/]`).test(element.id)) {
         lastMatchId = currentId;
       } else if (
-        (!currentId.startsWith(`${lastMatchId}.`) && !currentId.startsWith(`${lastMatchId}:`)) ||
+        !new RegExp(`${escapeRegExp(lastMatchId)}[./:]`).test(currentId) ||
         // If element is not a slice at this level, and the currentId is a slice, break to add children before slices
-        (element.id.startsWith(`${lastMatchId}.`) && currentId.startsWith(`${lastMatchId}:`))
+        (new RegExp(`${escapeRegExp(lastMatchId)}[.]`).test(element.id) &&
+          new RegExp(`${escapeRegExp(lastMatchId)}[:/]`).test(currentId))
```
mint-thompson, line 143: In order to get these regular expressions to behave the same as `startsWith`, they should have a `^` at the start.

ngfreiter: Good call, I will add that.

ngfreiter: @mint-thompson, this should be ready for re-review based on your comment.

PR #573, opened 2020-08-18
Fixing complex objects checks for fixed children

### LINKED ISSUE #560: Fixing the parent of an already fixed value is not working as expected
When fixing the parent of a fixed value in a conflicting way, SUSHI does not log an error and fixes both the parent and the child, even though the child is fixed to a different value by each rule. An example of this is below:

An example of a profile that works as expected:
```
Profile: MyObs
Parent: Observation
* valueQuantity = 10 'mm'
* valueQuantity.value = 20
```
With the above profile, SUSHI will log an error that it "cannot fix 20 to this element; a different decimal is already fixed: 10" and the differential only has the 10 mm value fixed.

However, the following profile does not work as expected (switching the order of the two rules above):
```
Profile: MyObs
Parent: Observation
* valueQuantity.value = 20
* valueQuantity = 10 'mm'
```
With this profile, SUSHI will not log any errors and the conflicting values are both set in the differential as follows: 
```
{
  "id": "Observation.valueQuantity",
  "path": "Observation.valueQuantity",
  ...
  "patternQuantity": {
    "value": 10,
    "code": "mm",
    "system": "http://unitsofmeasure.org"
  }
},
{
  "id": "Observation.valueQuantity.value",
  "path": "Observation.valueQuantity.value",
  "patternDecimal": 20
}
```

Another example of this can be seen when fixing values using inline instances. The following example works similarly to the second example and does not log any errors and produces a conflicting differential:

```
Instance: USPostalAddress
InstanceOf: Address
Usage: #inline
* type = #postal
* country = "US"

Profile: USPatient
Parent: Patient
* address.country = "not-US"
* address = USPostalAddress
```

mint-thompson:
Fixes #560 and completes task [CIMPL-496](https://standardhealthrecord.atlassian.net/browse/CIMPL-496).

When fixing a complex object to an element, children of that element or of that element's slicedElement may already have fixed values. When this happens, make sure that the two values do not conflict with one another.

The case of "an instance is assigned to a slice (such as `address[RecentAddress]`), and a later fixed value on a non-slice (such as `address.period.start`) would violate that slice" is out of scope for this task.

### DIFF
--- a/src/fhirtypes/ElementDefinition.ts
+++ b/src/fhirtypes/ElementDefinition.ts
@@ -1288,6 +1288,17 @@ export class ElementDefinition {
       throw new ValueAlreadyFixedError(fshValue, type, JSON.stringify(currentElementValue));
     }
 
+    // Children of elements with complex types such as Quantity may already have fixed values
+    this.children().forEach(child => this.checkChildFixedValue(child, fhirValue));
+
+    // if this is a slice, make sure that nothing on this.slicedElement() is being violated
+    const slicedElement = this.slicedElement();
+    if (slicedElement) {
+      slicedElement
+        .children()
+        .forEach(child => slicedElement.checkChildFixedValue(child, fhirValue));
+    }
+
     // If we made it this far, fix the value using fixed[x] or pattern[x] as appropriate
     if (exactly) {
       // @ts-ignore: Type 'any' is not assignable to type 'never'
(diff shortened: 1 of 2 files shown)


src/fhirtypes/ElementDefinition.ts
```diff
@@ -1300,6 +1311,33 @@ export class ElementDefinition {
...
+    const patternX = `pattern${upperFirst(childType)}` as keyof ElementDefinition;
+    const currentChildValue = child[fixedX] ?? child[patternX];
+    if (currentChildValue != null) {
+      // find the element on fhirValue that would get assigned to the child
+      const childPath = child.id.replace(`${this.id}.`, '').split('.');
+      let newChildValue = fhirValue;
+      for (const pathPart of childPath) {
+        if (newChildValue != null) {
+          newChildValue = newChildValue[pathPart];
+        }
```
ngfreiter, line 1326: I think this causes an issue in the case of an array. Consider this example:
```
Profile: MyObs
Parent: Observation
* valueCodeableConcept.coding.code = #bar
* valueCodeableConcept = #foo
```
This is pretty much the same situation as your test on `valueQuantity`, but because `valueCodeableConcept` contains an array, the test does not seem to work. You end up with `patternCode: "bar"` on `valueCodeableConcept.coding.code`, and `patternCodeableConcept: {coding: [{code: "foo"}]}` on `valueCodeableConcept`, and these two rules conflict.

The `fixedByAnyParent` function checks if a value is fixed by any parent of an element, and returns that element if so, and it does work with arrays, so my first thought was that you could use that function, but of course that is an issue because you can't use that function until you actually do fix the value on the parent element, but we are trying to decide if it is valid to fix the value. So one possibility would be to fix the value on the parent, determine if that fixing violated anything on the child elements, and un-fix the value on the parent if so. But maybe it is just simpler to add array handling logic here.

mint-thompson:
Thank you for providing the additional sample case! I've added a test case with that scenario, and I've updated `checkChildFixedValue` to traverse values that contain arrays.

PR #577, opened 2020-08-20
Allow SUSHI to run without config.yaml

ngfreiter:
This adds the ability for SUSHI to run without a config.yaml file. When no config.yaml is present, SUSHI looks for an `ImplementationGuide.json` file in the `../input` folder (assuming SUSHI is running in the IG Pub context), and if one is found, will extract the `canonical` and any `dependencies` from that file, and use that to process the FSH. SUSHI will run in `FSHOnly` mode if it has to extract the configuration in this way. If no `ImplementationGuide.json` is found, SUSHI will just fail for having no configuration as it used to.


src/utils/Processing.ts
```diff
@@ -59,8 +60,16 @@ export function ensureOutputDir(input: string, output: string, isIgPubContext: b
 export function readConfig(input: string): Configuration {
   const configPath = ensureConfiguration(input);
   if (configPath == null || !fs.existsSync(configPath)) {
-    logger.error('No config.yaml in FSH definition folder.');
-    throw Error;
+    const config = extractConfiguration(input);
```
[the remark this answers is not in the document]
ngfreiter: I thought this made sense, but looking a little deeper I'm less sure. It seems to me that `ensureConfiguration` ensures that a `config.yaml` file will exist. But when we extract the configuration, we don't actually make a `config.yaml` file, we are just directly creating the configuration from the ImplementationGuide JSON. I could still put this code in `ensureConfiguration`, but it feels a little off, since one method of ensuring the configuration, `generateConfiguration`, creates a file that is then read into memory, where the other method of ensuring the configuration `extractConfiguration`, generates a config directly in memory. Let me know what you think.

ngfreiter:
Ok @cmoesel, this should have all the changes you asked for, except for moving `extractConfiguration` into `ensureConfiguration`, see my comment https://github.com/FHIR/sushi/pull/577#discussion_r475620571 for why I did not yet do that. I can still do that, I just wasn't totally sure it was the right choice, and wanted to wait to hear what you thought of my comment before doing anything.

src/utils/Processing.ts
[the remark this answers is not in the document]
cmoesel: Ugh.  You're right -- it returns a config _path_ -- not a config itself.  Ugh.  What do you think of this?
* Rename `ensureConfiguration` to `ensureConfigurationFile`
* Rename `extractConfiguration` to `loadConfigurationFromIgJSON`

I think that clarifies things a bit better...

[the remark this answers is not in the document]
ngfreiter: That sounds good to me!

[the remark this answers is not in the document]
ngfreiter: Ok, https://github.com/FHIR/sushi/pull/577/commits/2282a68eab01c5a16b6e3957582c6260bcb68ca3 changes the names, I went with `loadConfigurationFromIgResource`, since the IG can now be either JSON or XML.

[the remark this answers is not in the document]
cmoesel: Perfect, especially since that avoids the whole "should it be 'JSON' or 'Json' debacle".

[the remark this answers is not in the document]
jafeltra: 'JSON' vs 'Json' and 'FHIR' vs 'Fhir' are two of life's unending debacles.

test/import/loadConfigurationFromIgResource.test.ts
```diff
@@ -0,0 +1,105 @@
...
+    expect(loggerSpy.getLastMessage('info')).toMatch(
+      new RegExp(
+        `from ${escapeRegExp(path.join(inputPath, '..', 'input', 'ImplementationGuide.json'))}`
+      )
+    );
+  });
+
+  it('should return null when there are two potentially valid IG files', () => {
+    const inputPath = path.join(__dirname, 'fixtures', 'ig-JSON-and-ig-XML', 'fsh');
+    expect(loadConfigurationFromIgResource(inputPath)).toBeNull();
```
jafeltra, line 88: I'm not actually sure that `null` will get returned if there is both a JSON and XML representation of the IG. I think the input path here should actually be `ig-XML-and-ig-JSON`. When I do switch that, the test fails. I think it might be because the `return null` in [loadConfigurationFromIgResource line 46](https://github.com/FHIR/sushi/pull/577/files#diff-41eb50e93ad9d6ea6e77945583b53084R46) just moves on to the next iteration of the forEach`.

ngfreiter: Good catch, in 359d325 I made the test actually test things, and fixed the code so that the case the test is testing for passes.

jafeltra: Looks good to me! The only other minor thought I had if you think it makes sense to log a message that multiple files were found and that's why the information can't be extracted. Right now I just get

```

File: src/fshtypes/Configuration.ts [hunk 3]
```diff
@@ -64,7 +64,7 @@ export type Configuration = {
   packageId?: string;
 
   license?: string;
-  fhirVersion: string[];
+  fhirVersion?: string[];
```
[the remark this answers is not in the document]
ngfreiter: The reason is...oops. I will fix it!

test/import/loadConfigurationFromIgResource.test.ts
jafeltra: I don't feel strongly about this if we don't want to have two messages and it's complicated to have the one message be conditional.

src/fshtypes/Configuration.ts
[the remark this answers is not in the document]
ngfreiter: Fixed in https://github.com/FHIR/sushi/pull/577/commits/7de8179e692de88f348a6cffef1bc282a3c7d1bd

test/import/loadConfigurationFromIgResource.test.ts
cmoesel: Ugh.  And it did it to me again.  I thought I submitted the above comment _yesterday_ -- but it just added it to my pending comments instead (even though it _looked_ like it submitted it).  Sorry!

ngfreiter: I hadn't seen Julia's further comments. I'll improve the error message.

ngfreiter: Error message added in https://github.com/FHIR/sushi/pull/577/commits/32d3efdd47a26e22ce6ae69a286c8c66f44c0fd0

src/utils/Processing.ts
```diff
@@ -57,13 +64,21 @@ export function ensureOutputDir(input: string, output: string, isIgPubContext: b
...
-    logger.error('No config.yaml in FSH definition folder.');
+    config = loadConfigurationFromIgResource(input);
+  } else {
+    const configYaml = fs.readFileSync(configPath, 'utf8');
+    config = importConfiguration(configYaml, configPath);
+  }
+  if (!config) {
+    logger.error(
+      'No config.yaml in FSH definition folder, and no configuration could' +
+        ' be extracted from an ImplementationGuide JSON file.'
```
jafeltra, line 78: I'm sorry to do this (I should have thought about it before), but should this actually say "ImplementationGuide resource" since it could be a JSON or XML?

ngfreiter: Yes it should!

ngfreiter: https://github.com/FHIR/sushi/pull/577/commits/f19af48d84428149513942705923060a5f5e10fc fixes this
.

ngfreiter:
> I did some additional manual testing on this and found the following results:
> 
> * with `config.yaml` -- each resource has the `version` from `config.yaml` populated in its definition
> * without `config.yaml` -- the `version` is _not_ populated in each definition (even if it is specified in IG JSON)
> 
> To get better parity w/ config, I think that if the IG JSON has a `version`, you should put it in the config so it gets populated into the resources.

I came across this difference when implementing, and found that, at least in my experience, the IG Publisher automatically adds `version`. When someone doesn't have a `config.yaml`, it seemed to me like that is a signal for SUSHI to be as hands off as possible, and since the Publisher adds version anyway, I thought it made sense to leave it out.

Does the Publisher adding version change your opinion? If not, let me know and I can change it to extract `version` as well.

ngfreiter:
PS, sorry @cmoesel I should've noted the `version` thing instead of making you have to dig to find it.

cmoesel:
> Does the Publisher adding version change your opinion? 

Not really, as I already knew it does that.  I just like having SUSHI generate profiles that are as complete as possible -- for people who want to view them pre-publisher or for other workflows where perhaps they're not using the publisher at all?

I see FSHOnly (and zero config) more as a signal that they don't want SUSHI to touch anything _except_ the definitional files -- but I think the definitional files should still be fully processed the same way.  If we want to consider no longer populating `version` in the definitions, we can -- but I think that if we do change it, we should change it across the board.

ngfreiter:
> Not really, as I already knew it does that. I just like having SUSHI generate profiles that are as complete as possible -- for people who want to view them pre-publisher or for other workflows where perhaps they're not using the publisher at all?

Somehow I had a feeling you already knew that 😉 . I will make the change.

ngfreiter:
Version, coming to a PR near you in https://github.com/FHIR/sushi/pull/577/commits/67daafac4119532250149f82cf690c141245ea58

PR #580, opened 2020-08-24
Add logic to skip hidden files when including non-FSH IG resources

masnick:
This prevents an error when macOS silently adds a `.DS_Store` file to one of the folders in `ig-data/input/{supported-resource-input-directory}/`.

I ran into this when trying to include some `.json` format instances in an IG under `ig-data/input/examples/`. When modifying the .json files in the `examples/` folder, macOS will sometimes silently add a `.DS_Store` file into `examples/`.

When this happens, the `sushi` command will fail with a generic "Invalid file detected in directory /path/to/examples/" message. Because the `.DS_Store` file is hidden by the OS in Finder, this issue is hard to debug.

This change will ignore any files that begin with `.` to avoid this problem.

### DIFF
--- a/src/fhirdefs/load.ts
+++ b/src/fhirdefs/load.ts
@@ -197,8 +197,7 @@ export function loadCustomResources(input: string, defs: FHIRDefinitions): void 
           if (file.startsWith('.')) {
             // Ignore hidden files on *nix, like `.DS_Store` on macOS
             continue;
-          }
-          else if (file.endsWith('.json')) {
+          } else if (file.endsWith('.json')) {
             resourceJSON = fs.readJSONSync(path.join(dirPath, file));
           } else if (file.endsWith('xml')) {
             resourceJSON = converter.xmlToObj(fs.readFileSync(path.join(dirPath, file)).toString());
--- a/src/fhirdefs/load.ts
+++ b/src/fhirdefs/load.ts
@@ -5,6 +5,7 @@ import path from 'path';
 import os from 'os';
 import tar from 'tar';
 import axios from 'axios';
+import junk from 'junk';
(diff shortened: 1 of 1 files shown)


src/fhirdefs/load.ts
```diff
@@ -194,7 +194,10 @@ export function loadCustomResources(input: string, defs: FHIRDefinitions): void
       for (const file of files) {
         let resourceJSON: any;
         try {
-          if (file.endsWith('.json')) {
+          if (file.startsWith('.')) {
```
ngfreiter, line 197: We have come across this situation in a different part of our code, and there we used a package called `junk` to filter out hidden files like `.DS_Store` but also certain Windows hidden files like `Thumbs.db`. I think it would be best to be consistent and use that here as well. To use `junk`, you'd want to add:
```
import junk from 'junk';
```
to the top of `load.ts`, and then line 197 would be something like:
```
if (junk.is(file)) {
```
The `junk.is` function just tests if the filename is on `junk`s list of system files.

cmoesel:
@masnick -- we're hoping to do a release of SUSHI today.  If you can fix up this PR today, we can get it in the release...

cmoesel:
If you can't we'll just hold it for the next release then.

masnick:
@ngfreiter thanks for the review, I've done as you suggested.

@cmoesel assuming the automated checks pass this should be ready to go into the release. I'm stepping away for a few hours now so I won't be able to fix problems immediately, so if this gets bumped to the next release because of that that's ok with me :)

masnick:
Ok, the checks passed and I tested this locally to ensure it works as intended. I think it's ready to merge in unless you see other issues. Thanks!

PR #581, opened 2020-08-25
Allow BYO resources to be in IG folder

ngfreiter:
This makes it so SUSHI will also search the IG folders for BYO-JSON/XML. This content only gets loaded in so that it can be used in the FSH definitions, it does not get copied over.

Since this is a part of the larger new IG Publisher integration, we can merge this into a feature branch that can then be merged into master at once.

I actually think that some of the concerns I had about loading definitions from the output folders of SUSHI weren't really a problem. If you run SUSHI and generate files into the IG Publisher folders, then on the next run of SUSHI you will load those in, since we don't have a `generated` folder yet, but this doesn't actually seem to be an issue. SUSHI prefers stuff defined in FSH to BYO-JSON/XML, and it doesn't do checks for duplicate definitions with BYO-JSON/XML content, so the generated definitions are just loaded in and then never used.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -83,7 +83,10 @@ async function app() {
   const dependencyDefs = loadExternalDependencies(defs, config);
 
   // Load custom resources specified in ig-data folder
-  loadCustomResources(input, defs);
+  loadCustomResources(path.join(input, 'ig-data', 'input'), defs);
+  if (isIgPubContext) {
+    loadCustomResources(path.join(input, '..', 'input'), defs);
+  }
 
   let tank: FSHTank;
   try {
--- a/src/app.ts
+++ b/src/app.ts
@@ -84,7 +84,7 @@ async function app() {
 
   // Load custom resources specified in ig-data folder
   loadCustomResources(path.join(input, 'ig-data', 'input'), defs);
-  if (isIgPubContext) {
+  if (isIgPubContext && !fs.existsSync(path.join(input, 'ig-data'))) {
     loadCustomResources(path.join(input, '..', 'input'), defs);
   }
 
--- a/src/fhirdefs/load.ts
+++ b/src/fhirdefs/load.ts
@@ -170,10 +170,10 @@ export function cleanCachedPackage(packageDirectory: string): void {
 
 /**
  * Loads custom resources defined in ig-data into FHIRDefs
- * @param {string} input - The input path to the cli
+ * @param {string} resourceDir - The path to the directory containing the resource subdirs
  * @param {FHIRDefinitions} defs - The FHIRDefinitions object to load definitions into
  */
-export function loadCustomResources(input: string, defs: FHIRDefinitions): void {
+export function loadCustomResources(resourceDir: string, defs: FHIRDefinitions): void {
   // Similar code for loading custom resources exists in IGExporter.ts addPredefinedResources()
   const pathEnds = [
     'capabilities',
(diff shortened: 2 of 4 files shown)


cmoesel:
> SUSHI prefers stuff defined in FSH to BYO-JSON/XML

Hmm... I don't think that's the way it was intended to work.  One of the reasons we introduced BYOJ was to allow people to put in _patched_ versions of JSON definitions if SUSHI generated them wrong.  The idea was supposedly that you could keep your FSH definition (waiting for SUSHI to fix whatever bug), but as long as your BYOJ was there, it would overwrite the bad SUSHI-generated file w/ the good BYOJ file...

src/app.ts
```diff
@@ -83,7 +83,10 @@ async function app() {
   const dependencyDefs = loadExternalDependencies(defs, config);
 
   // Load custom resources specified in ig-data folder
-  loadCustomResources(input, defs);
+  loadCustomResources(path.join(input, 'ig-data', 'input'), defs);
+  if (isIgPubContext) {
+    loadCustomResources(path.join(input, '..', 'input'), defs);
+  }
```
cmoesel, line 89: I feel like this could make for some really weird situations.  If you have definitions in `ig-data/input` then once you run SUSHI one time, those definitions will be in `../input`.  And then the next time you run SUSHI, it will load the definitions from `ig-data/input` and then load the same definitions from `../input`.

As a result, I could see the potential for two bad things (if I understand this correctly):

1. If you edit a file in `ig-data/input`, but the same file already exists in `../input` (due to a recent SUSHI run) -- then won't the old unedited file load over the new edited file (since `../input` gets loaded after `ig-data/input`)?

2. If you delete a file in `ig-data/input` but it still exists in `../input` (due to a recent SUSHI run), then SUSHI will load the old one from `../input` and then add it to the IG JSON `resources` list -- effectively making it as if it was not deleted at all.

To me, it seems like if `ig-data` exists, then we best just load stuff from there (and there only).  We should only load directly from `../input` when the author has intentionally chosen not to use `ig-data`.  Does that make sense?

ngfreiter: Yeah that does make sense, I'll make that change!

ngfreiter:
> > SUSHI prefers stuff defined in FSH to BYO-JSON/XML
> 
> Hmm... I don't think that's the way it was intended to work. One of the reasons we introduced BYOJ was to allow people to put in _patched_ versions of JSON definitions if SUSHI generated them wrong. The idea was supposedly that you could keep your FSH definition (waiting for SUSHI to fix whatever bug), but as long as your BYOJ was there, it would overwrite the bad SUSHI-generated file w/ the good BYOJ file...

Ah you are right, I phrased that poorly. When we have to fish for a definition in, so let's say a `Profile` declares its `Parent`, SUSHI will first look through things defined in FSH before going to externally loaded defs. But, when SUSHI writes files, it copies BYOJ files after writing the stuff defined in FSH, so the things defined in FSH do get copied over by BYOJ stuff. In this context though, the BYOJ stuff doesn't get copied over, it just gets loaded in, since it is already in the `input` folder.

cmoesel:
Ah.  I think maybe SUSHI isn't behaving consistently then.  It seems that if you want something in BYOJ to override FSH, then SUSHI should always use the BYOJ resource.  We may want to consider that (but not in this PR)...

src/app.ts
ngfreiter: Changed in https://github.com/FHIR/sushi/pull/581/commits/5b2a5223cd356a9aae1e0a1001aaa6ed2207efc4.

PR #584, opened 2020-08-27
Added puns

jafeltra:
This PR just adds a few puns provided by Nicole Ng and Ammu Irivinti!

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -191,7 +191,8 @@ const MESSAGE_MAP: { [key in ResultStatus]: string[] } = {
     'Everything is ship-shape!',
     'Ex-clam-ation point!',
     'Ac-clam-ations!',
-    'Fin-tastic job!'
+    'Fin-tastic job!',
+    'You dolphinitely have the hang of this!'
   ],
   warnings: [
     'Not bad, but you cod do batter!',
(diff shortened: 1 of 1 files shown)


cmoesel:
Ugh.

mint-thompson:
This is good and I like all of these. Thank you to our kind contributors.

cmoesel:
I think `Ugh` is exactly the confirmation our contributors were looking for.  I hope it pleases them.

src/app.ts
```diff
@@ -191,7 +191,8 @@ const MESSAGE_MAP: { [key in ResultStatus]: string[] } = {
     'Everything is ship-shape!',
     'Ex-clam-ation point!',
     'Ac-clam-ations!',
-    'Fin-tastic job!'
+    'Fin-tastic job!',
+    'You dolphinitely have the hang of this!'
```
ngfreiter, line 195: Unfor-tuna-tely, I think this line is slightly too long, so it ends up looking like this:
image

It looks like it is 3 characters too long.

ngfreiter: We could do "You're dolphinitely doing great", or "You dolphinitely got it!", or something like that. Or "You dolphinitely got the hang of it!" to stay close to the original pun.

jafeltra: Oh good call. Somehow I like "You're dolphinitely doing great!" so I think I'll go with that!

PR #585, opened 2020-08-28
Improve 'IG Publisher mode' messages

### LINKED ISSUE #555: IG Publisher context determined by directory name (undocumented)
I found this _very_ confusing in v0.15. I'd recommend at least documenting it if not removing it entirely.

I was getting a message of
> Current FSH tank conforms to an IG Publisher context. Output will be adjusted accordingly.

and no _gen... scripts were created.

I finally looked at the source:

```ts
// If a fsh subdirectory is used, we are in an IG Publisher context
const isIgPubContext = path.parse(input).base === 'fsh';
```

It appears that if you specify an ~~`--out`~~ (correction: input) directory named 'fsh' then the program behaves differently.

Please consider addressing this confusion. Thank you!

jafeltra:
This PR fixes #555 by improving the generic message that is logged when SUSHI encounters a `fsh/` directory and is in "IG Publisher mode."

I used the message that @cmoesel suggested on the issue, but made a small adjustment to include that the output folder can also be specified by the `--out` option.

One additional question I have that I wanted a second opinion on before changing is around the additional messages that are logged when we adjust the input and output paths. Right now, when the input path needs to be adjusted and we don't specify an output path, the following messages are logged in this order:

```
info  fsh/ subdirectory detected and added to input path
info  SUSHI detected a "fsh" directory in the input path. As a result, SUSHI will operate in "IG Publisher integration" mode. This means:
  - the "fsh" directory will be used as the input path
  - the parent of the "fsh" directory (e.g., "../fsh") will be used as the output path unless otherwise specified with --out option
  - generation of publisher-related scripts will be suppressed (i.e., assumed to be managed by you)
info  No output path specified. Output to ../sample-fsh-tanks/publisher-test
```
Do you think the first and third message are now unnecessary with the improved second message? Or do they seem helpful still? If so, it might make more sense to have the second, detailed message come first, and then have the first and third message come after if they are applicable. Before switching that around though, I wanted to know what other people thought.

### DIFF
--- a/src/utils/Processing.ts
+++ b/src/utils/Processing.ts
@@ -31,7 +31,7 @@ export function findInputDir(input: string): string {
   const fshSubdirectoryPath = path.join(input, 'fsh');
   if (fs.existsSync(fshSubdirectoryPath)) {
     input = path.join(input, 'fsh');
-    logger.info('fsh/ subdirectory detected and add to input path');
+    logger.info('fsh/ subdirectory detected and added to input path');
   }
   return input;
 }
@@ -39,7 +39,10 @@ export function findInputDir(input: string): string {
 export function ensureOutputDir(input: string, output: string, isIgPubContext: boolean): string {
   if (isIgPubContext) {
     logger.info(
-      'Current FSH tank conforms to an IG Publisher context. Output will be adjusted accordingly.'
+      'SUSHI detected a "fsh" directory in the input path. As a result, SUSHI will operate in "IG Publisher integration" mode. This means:\n' +
+        '  - the "fsh" directory will be used as the input path\n' +
+        '  - the parent of the "fsh" directory (e.g., "../fsh") will be used as the output path unless otherwise specified with --out option\n' +
+        '  - generation of publisher-related scripts will be suppressed (i.e., assumed to be managed by you)'
     );
   }
   let outDir = output;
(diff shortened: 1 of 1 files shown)


ngfreiter:
I think the second message captures the information provided by the first message, so the first message seems somewhat unnecessary to me now. But then the third message does seem to add something, since the second message says that the parent will be used unless they have specified a different directory, and the third message confirms whether or not they have specified an output directory. Does that make sense?

cmoesel:
I agree w/ @ngfreiter -- kill the first message; leave the second message.

jafeltra:
Okay! I had also had thoughts of having the first message include the current input path, something like:

```
info  fsh/ subdirectory detected and added to input path. Running SUSHI on FSH definitions in ../sample-fsh-tanks/publisher-test/fsh
```

Do you think that would still be redundant?

cmoesel:
Yeah, that still feels redundant to me.

jafeltra:
Alrighty! Works for me!

jafeltra:
First message is gone in cf45f23. I think this is set to be reviewed now!

cmoesel:
😢 
image
😢

cmoesel:
🥳 
image
🥳

jafeltra:
😕
image
😕

cmoesel:
🕺 
image
🕺

cmoesel:
👋 
image
👋

jafeltra:
🙏
image
🙏

PR #586, opened 2020-08-28
Rename config.yaml

ngfreiter:
This renames the `config.yaml` file to `sushi-config.yaml`. Using a `config.yaml` file is still allowed, but a warning message will be displayed to indicate that `config.yaml` is deprecated and should be changed to `sushi-config.yaml`.

This PR does not move the new configuration file to the root IG Publisher folder, because we already have the `config.yaml` file living in `fsh`, so it seemed like the path of least resistance for users to leave the `sushi-config.yaml` file in `fsh`, with the expectation that it will then live in `input/fsh` when we make that change.


test/import/ensureConfiguration.test.ts
```diff
@@ -17,20 +17,37 @@ describe('ensureConfiguration', () => {
...
+    expect(configPath).toBe(path.join(tank, 'config.yaml'));
+    expect(loggerSpy.getAllLogs('warn')).toHaveLength(1);
+    expect(loggerSpy.getLastMessage('warn')).toMatch(/Use of config\.yaml is deprecated/);
+  });
+
+  it('should return the path to a pre-existing config.yml and log a warning', () => {
     const tank = path.join(__dirname, 'fixtures', 'existing-config-yml');
     const configPath = ensureConfiguration(tank);
     expect(configPath).toBe(path.join(tank, 'config.yml'));
     // Generating a config causes a warning, so there should be NO warnings in this case
```
jafeltra, line 48: Very  minor, but I don't think we need this comment any more.

ngfreiter: We don't need that comment any more, so it is no more in 0deac15

jafeltra (dismissed):

src/ig/IGExporter.ts
```diff
@@ -197,7 +197,7 @@ export class IGExporter {
       // No need for the detailed log message since we already logged one in the package loader.
       logger.error(
         `Failed to add ${dependency.packageId} to ImplementationGuide instance because no ` +
-          'version was specified in your config.yaml.'
+          `version was specified in your ${path.basename(this.configPath)}.`
```
cmoesel, line 200: Thanks for changing all the `config.yaml` to `${path.basename(this.configPath)}`.  That must have been pretty annoying.  You do `${path.basename(this.configPath)}` often enough, however, that it seems you might have benefited by just declaring a new `configName` property in the class, e.g.:
```ts
this.configName = path.basename(this.configPath);
```
and then `${this.configName}` everywhere.

test/ig/IGExporter.SUSHI-GENERATED-FILES.test.ts
```diff
@@ -48,7 +48,7 @@ describe('IGExporter', () => {
       expect(content).toMatch(/\| _updatePublisher.bat .*\| generated \| \s* \|/);
       expect(content).toMatch(/\| _updatePublisher.sh .*\| generated \| \s* \|/);
       expect(content).toMatch(
-        /\| input[\/\\]ImplementationGuide-fhir\.us\.minimal\.json \s*\| generated \| .*[\/\\]customized-ig[\/\\]config\.yaml, \{all input resources and pages\} \s*\|/
+        /\| input[\/\\]ImplementationGuide-fhir\.us\.minimal\.json \s*\| generated \| .*[\/\\]customized-ig[\/\\]sushi-config\.yaml, \{all input resources and pages\} \s*\|/
```
cmoesel, line 51: This might not match what the user really has (if they have `config.yaml`) -- but I don't think I care.  We were already in that boat if they had `config.yml` anyway.

cmoesel (dismissed):
This looks to me like it was a fairly tedious and annoying activity.  Thanks for doing it.

I left a comment about how you could benefit from some re-use, but it's up to you if you want to do anything about it.  I approve either way.

src/ig/IGExporter.ts
ngfreiter: Yeah good point, added in https://github.com/FHIR/sushi/pull/586/commits/fedd542e656f51ffc3da5332ca16f1f8968e2660. Unfortunately that then dismissed your review and @jafeltra's review.

cmoesel (approved):
Looks good!

PR #589, opened 2020-08-31
Support FSH definitions in input/fsh

jafeltra:
This PR completes [CIMPL-508](https://standardhealthrecord.atlassian.net/browse/CIMPL-508). It adds support for reading FSH definitions from an `input/fsh` folder even if the subdirectory is not specified in the input path.

This also still maintains support for automatically reading in FSH definitions found in a `fsh/` subdirectory. I added comments to the places where we will need to remove the handling when we no longer want to support this folder. If we imagine that we just will continue to support it, I can either update or remove those comments.

There is one unintended side effect of maintaining support for both folders. Now, if we were to have a `myTank/input/fsh` folder with FSH definitions, you can run SUSHI pointing to `myTank` and we will automatically look in `myTank/input/fsh`, which is the intention of this task. However, you can also run SUSHI pointing to `myTank/input` and SUSHI will find the `fsh` folder now, but if we remove the support for that top level `fsh/` folder, that same SUSHI command will not work. I think this is okay because we aren't expecting someone with an `myTank/input/fsh` folder to be running SUSHI pointing at `input`, but I just wanted to raise it in case it was concerning to anyone else.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,10 +36,10 @@ Options:
 Additional information:
   [path-to-fsh-defs]
     Default: "."
-    If fsh/ subdirectory present, it is included in [path-to-fsh-defs]
+    If input/fsh/ subdirectory present, it is included in [path-to-fsh-defs]
   -o, --out <out>
     Default: "build"
-    If fsh/ subdirectory present, default output is one directory above fsh/
+    If input/fsh/ subdirectory present, default output is one directory above fsh/
 ```
 
 # IG Generation
--- a/src/app.ts
+++ b/src/app.ts
@@ -46,10 +46,12 @@ async function app() {
       console.log('Additional information:');
       console.log('  [path-to-fsh-defs]');
       console.log('    Default: "."');
-      console.log('    If fsh/ subdirectory present, it is included in [path-to-fsh-defs]');
+      console.log('    If input/fsh/ subdirectory present, it is included in [path-to-fsh-defs]');
       console.log('  -o, --out <out>');
       console.log('    Default: "build"');
-      console.log('    If fsh/ subdirectory present, default output is one directory above fsh/');
+      console.log(
+        '    If input/fsh/ subdirectory present, default output is one directory above fsh/'
+      );
     })
     .arguments('[path-to-fsh-defs]')
     .action(function (pathToFshDefs) {
@@ -67,8 +69,11 @@ async function app() {
 
   input = findInputDir(input);
 
-  // If a fsh subdirectory is used, we are in an IG Publisher context
-  const isIgPubContext = path.parse(input).base === 'fsh';
+  // If an input/fsh subdirectory is used, we are in an IG Publisher context
+  const fshFolder = path.parse(input).base === 'fsh';
+  const inputFshFolder = fshFolder && path.parse(input).dir.endsWith(`${path.sep}input`);
+  // TODO: Legacy support for top level fsh/ subdirectory. When no longer supporting, update to const isIgPubContext = inputFshFolder;
+  const isIgPubContext = fshFolder || inputFshFolder;
   const outDir = ensureOutputDir(input, program.out, isIgPubContext);
 
   let config: Configuration;
--- a/src/app.ts
+++ b/src/app.ts
@@ -49,9 +49,7 @@ async function app() {
       console.log('    If input/fsh/ subdirectory present, it is included in [path-to-fsh-defs]');
       console.log('  -o, --out <out>');
       console.log('    Default: "build"');
-      console.log(
-        '    If input/fsh/ subdirectory present, default output is one directory above fsh/'
-      );
+      console.log('    If input/fsh/ subdirectory present, default output is to input/generated/');
     })
     .arguments('[path-to-fsh-defs]')
     .action(function (pathToFshDefs) {
@@ -72,9 +70,10 @@ async function app() {
   // If an input/fsh subdirectory is used, we are in an IG Publisher context
   const fshFolder = path.parse(input).base === 'fsh';
   const inputFshFolder = fshFolder && path.parse(input).dir.endsWith(`${path.sep}input`);
-  // TODO: Legacy support for top level fsh/ subdirectory. When no longer supporting, update to const isIgPubContext = inputFshFolder;
-  const isIgPubContext = fshFolder || inputFshFolder;
-  const outDir = ensureOutputDir(input, program.out, isIgPubContext);
+  const isIgPubContext = inputFshFolder;
+  // TODO: Legacy support for top level fsh/ subdirectory. Remove when no longer supported.
+  const isLegacyIgPubContext = fshFolder && !inputFshFolder;
+  const outDir = ensureOutputDir(input, program.out, isIgPubContext, isLegacyIgPubContext);
 
   let config: Configuration;
   try {
@@ -116,7 +115,8 @@ async function app() {
 
   logger.info('Converting FSH to FHIR resources...');
   const outPackage = exportFHIR(tank, defs);
-  writeFHIRResources(outDir, outPackage, program.snapshot);
+  const useGeneratedFolder = !isLegacyIgPubContext;
+  writeFHIRResources(outDir, outPackage, program.snapshot, useGeneratedFolder);
 
   // If FSHOnly is true in the config, do not generate IG content, otherwise, generate IG content
   if (config.FSHOnly) {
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -36,10 +36,10 @@ Options:
 Additional information:
   [path-to-fsh-defs]
     Default: "."
-    If fsh/ subdirectory present, it is included in [path-to-fsh-defs]
+    If input/fsh/ subdirectory present, it is included in [path-to-fsh-defs]
   -o, --out <out>
     Default: "build"
-    If fsh/ subdirectory present, default output is one directory above fsh/
+    If input/fsh/ subdirectory present, default output is one directory above fsh/
```
ngfreiter, line 42: Maybe this isn't a huge concern, because we are planning on adding a `generated` folder, and you are planning on adding that `generated` folder, but the way this currently works, `input` is treated as the output directory, which means the true output of SUSHI goes into `input/input/...`. Is it worth changing that now just so you can change it again when you add a `generated` folder? Maybe not, since we are merging into a feature branch.

src/utils/Processing.ts
```diff
@@ -27,19 +27,30 @@ export function findInputDir(input: string): string {
     logger.info('path-to-fsh-defs defaulted to current working directory');
   }
 
+  // TODO: Legacy support. Remove when no longer supported.
   // Use fsh/ subdirectory if not already specified and present
   const fshSubdirectoryPath = path.join(input, 'fsh');
   if (fs.existsSync(fshSubdirectoryPath)) {
     input = path.join(input, 'fsh');
-    logger.info('fsh/ subdirectory detected and add to input path');
+    return input;
```
ngfreiter, line 35: I think this could be an appropriate place for a warning about how the `fsh` directory is deprecated, and they should switch to `input/fsh`.

src/utils/Processing.ts
```diff
@@ -27,19 +27,30 @@ export function findInputDir(input: string): string {
...
+  }
+
+  // Use input/fsh/ subdirectory if not already specified and present
+  const inputFshSubdirectoryPath = path.join(input, 'input', 'fsh');
+  if (fs.existsSync(inputFshSubdirectoryPath)) {
+    input = path.join(input, 'input', 'fsh');
```
ngfreiter, line 41: I think it makes sense to prefer the `input/fsh` dir over the `fsh` dir, if they happened to have both.

src/app.ts
```diff
@@ -46,10 +46,12 @@ async function app() {
...
       console.log('  [path-to-fsh-defs]');
       console.log('    Default: "."');
-      console.log('    If fsh/ subdirectory present, it is included in [path-to-fsh-defs]');
+      console.log('    If input/fsh/ subdirectory present, it is included in [path-to-fsh-defs]');
       console.log('  -o, --out <out>');
       console.log('    Default: "build"');
-      console.log('    If fsh/ subdirectory present, default output is one directory above fsh/');
+      console.log(
+        '    If input/fsh/ subdirectory present, default output is one directory above fsh/'
+      );
```
ngfreiter, line 54: Also here, this will change when `generated` is added.

README.md
jafeltra: Ah yeah I was trying to keep out the changes needed for output change, but I guess I hadn't really thought about how weird the output would be in this case. I had output it to a different directory when testing so hadn't run into this issue, so now I guess I am a little more uncomfortable with how this works as an intermediate solution. But it is still intermediate I guess.

The other possible solution to this that I had considered was putting the output and input changes into one PR. Do you think that would make more sense?

(I haven't actually come to a conclusion here... shhh.)

src/utils/Processing.ts
jafeltra: Ah good idea! I didn't even think about that.

src/app.ts
jafeltra: As I said above, I was planning to just change this when updating the output piece, but now I'm not sure if this is too weird to leave as is for now.

README.md
ngfreiter: I don't think it is too weird to leave for now, since we are merging this into a branch that is not `master`, but if you were planning on doing the `input/generated` task next, then I'm also fine with wrapping that PR into this one. But really doesn't matter to me, I'll leave it to your preference.

jafeltra:
a20a8e1 adds support for favoring the `input/fsh` directory over the `fsh/` directory and adds a warning when the `fsh/` directory is detected. I'm open to wording suggestions for the warning.

I'm going to look into how to handle the output changes (either adding them into this PR or keeping this one as is for now) and I'll let you know which way I decide in a bit!

jafeltra:
I think based on looking at the changes for output, it might be easiest to keep the input and output changes in separate PRs. So as long as you're okay with some intermediary weirdness, I think this is good to be re-reviewed.

ngfreiter (dismissed):
Looks good!

src/utils/Processing.ts
```diff
@@ -27,19 +27,34 @@ export function findInputDir(input: string): string {
...
   // Use fsh/ subdirectory if not already specified and present
   const fshSubdirectoryPath = path.join(input, 'fsh');
-  if (fs.existsSync(fshSubdirectoryPath)) {
+  if (!fs.existsSync(inputFshSubdirectoryPath) && fs.existsSync(fshSubdirectoryPath)) {
     input = path.join(input, 'fsh');
-    logger.info('fsh/ subdirectory detected and add to input path');
+    logger.warn(
+      'SUSHI detected a "fsh" directory that will be used in the input path. ' +
+        'The top level "fsh" directory is being deprecated. FSH definitions in ' +
+        '"fsh" should be moved to "input/fsh".'
```
cmoesel, line 44: Should we also say
* config.yaml moves to ${root}/sushi-config.yaml
* files in ig-data go in their respective folders starting from the root

??? (Only word it better?)

If so, maybe we could personalize the message w/ exact paths so it says something like:
* sushi-config.yaml moves to /Users/cmoesel/dev/MyIG/sushi-config.yaml
* ig-data/* files move to /Users/cmoesel/dev/MyIG/*
* .fsh files move to /Users/cmoesel/dev/MyIG/input/fsh/*

cmoesel (dismissed):
Looks good.  I had one suggestion for a more detailed logging message, but we can defer that if you'd like.  In short, migration from `./fsh` to `./input/fsh` is not a simple _mv fsh ../input/fsh_ because config goes somewhere else and ig-data goes somewhere else.  So we need to guide the user through that better.

src/utils/Processing.ts
jafeltra: I think this seems reasonable to update now. I like those last three bullets. Based off of #586 though, I think that sushi-config.yaml will go in `input/fsh` so I think the first bullet would be:
- sushi-config.yaml moves to /Users/cmoesel/dev/MyIG/input/fsh/sushi-config.yaml
I can update to this if you agree!

cmoesel: Hmmm... I was thinking that w/ the new approach, `sushi-config.yaml` lives at the root, even if you have `input/fsh`.  To me, config files at the root just make more sense than in nested subfolders.  I think I assumed that @ngfreiter's PR was to rename the config, but that we'd address the new config location in a subsequent PR.

I guess that sort of challenges the notion, however, that SUSHI _switches_ the input folder to `input/fsh` when you run in a project w/ an `input/fsh` folder.  To me, I think I kind of like challenging that notion though.  It makes more sense to me if I think of it as SUSHI running in the root folder but scoping its input for `.fsh` files to `input/fsh`.  But of course, that probably makes a bunch more work, doesn't it?

jafeltra: Well it turns out I have not waited for you to agree (sorry). But it is in df127fd (so there's still time to disagree). I'm not 100% sure about the wording, so let me know what you think of that and the location of sushi-config.

jafeltra: Oh no! Your reply wasn't loaded when I replied! :panic:

jafeltra: Wait, how does that make a bunch more work? I can reword the messages that say we "switch" the input folder to align with the idea of "scoping". I guess another question would be when do we switch support for the location of config.yaml? Was that part of this task?

cmoesel: I assume it is more work because I also assume that SUSHI expects things to be in certain places relative to its working directory.  Maybe that is one (or two) too many assumptions.  Is it?

jafeltra:
I rebased this and it now has the update for outputting to `input/generated`. I believe everything that was previously discussed on here has been resolved and we have a separate task for moving the `sushi-config.yaml` file to the root. So I think this should be set for re-review now!

cmoesel (approved):
I'm good with this!

PR #590, opened 2020-09-02
Include id property in user-defined groupings

masnick:
Currently SUSHI sets only `name` and (optionally) `description` for user-defined groups when generating `ImplementationGuide-*.json`.

The IG publisher expects there to also be an `id` property, like so:

```js
...
  "definition": {
    "grouping": [
      {
        "id": "Group name goes here", // <----- this is what this PR adds in
        "name": "Group name goes here",
        "description": "Some description.."
      }
    ],
...
```

Without this `id` property, the user-defined group will not show up in the `artifacts.html` file built by the IG. When this property is defined, the group and associated resources appear in `artifacts.html` as expected.

A few things that are known to be wrong/broken with this PR:

- There's not a way to actually set the `id` property separately from `name`. This would require monkeying around with `config.yaml`. It may make sense to make the following change:

    ```yaml
    # Current format
    groups:
      GroupA:
        description: The Alpha Group
        resources:
        - StructureDefinition/animal-patient
        - StructureDefinition/arm-procedure

    # Proposed change
    groups:
      GroupIdGoesHere:
        name: Group name goes here
        description: The Alpha Group
        resources:
        - StructureDefinition/animal-patient
        - StructureDefinition/arm-procedure
    ```

- There are two failing tests (at least locally).

I'm not experienced enough with the SUSHI codebase to fix these problems independently efficiently, but I'm happy to pair with someone to work on this (or feel free to make your own changes).

### DIFF
--- a/src/fhirtypes/ImplementationGuide.ts
+++ b/src/fhirtypes/ImplementationGuide.ts
@@ -47,6 +47,7 @@ export type ImplementationGuideDefinition = {
 };
 
 export type ImplementationGuideDefinitionGrouping = {
+  id: string;
   name: string;
   description?: string;
 };
--- a/src/ig/IGExporter.ts
+++ b/src/ig/IGExporter.ts
@@ -1024,6 +1024,7 @@ export class IGExporter {
       }
     } else {
       this.ig.definition.grouping.push({
+        id: name,
         name: name,
         ...(description && { description })
       });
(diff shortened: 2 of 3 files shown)


cmoesel (approved):
Thanks for the PR, @masnick.  The solution looks reasonable enough, although perhaps we can add support for independently setting the `id` sometime in the future.  But for now, I think this totally works!  I pushed up some fixes to the unit tests if you want to take a look.

ngfreiter (approved):
This looks good to me as well. Thanks @masnick!

masnick:
Thank you both!

PR #592, opened 2020-09-02
Support output to input/generated

jafeltra:
NOTE: This is currently has a base branch of `support-input-fsh`. If the base branch doesn't automatically update to `new-ig-publisher-integration`, I can update it once `support-input-fsh` gets merged.

This PR supports outputting to a `input/generated` folder when running SUSHI on a folder structure that supports our new IG Integration (i.e. it has an `input/fsh` folder) or when you just have a flat FSH tank (no publisher integration at all). When running SUSHI on a folder structure that supports the old IG Integration, output will be written to all the same subfolders that it does now (`input/profiles`, `input/extensions`, etc).

I've been testing this using a flat FSH tank (no `fsh` or `input/fsh` folders), the old IG publisher integration, and the new IG publisher integration. I've also tried to be sure to test each of those specifying an output directory and letting SUSHI decide the output, and having `FSHOnly` set to true and false. However, there's a very likely chance I missed something in one of those cases. I'm happy to walk through that with anyone reviewing it.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -49,9 +49,7 @@ async function app() {
       console.log('    If input/fsh/ subdirectory present, it is included in [path-to-fsh-defs]');
       console.log('  -o, --out <out>');
       console.log('    Default: "build"');
-      console.log(
-        '    If input/fsh/ subdirectory present, default output is one directory above fsh/'
-      );
+      console.log('    If input/fsh/ subdirectory present, default output is to input/generated/');
     })
     .arguments('[path-to-fsh-defs]')
     .action(function (pathToFshDefs) {
@@ -72,9 +70,10 @@ async function app() {
   // If an input/fsh subdirectory is used, we are in an IG Publisher context
   const fshFolder = path.parse(input).base === 'fsh';
   const inputFshFolder = fshFolder && path.parse(input).dir.endsWith(`${path.sep}input`);
-  // TODO: Legacy support for top level fsh/ subdirectory. When no longer supporting, update to const isIgPubContext = inputFshFolder;
-  const isIgPubContext = fshFolder || inputFshFolder;
-  const outDir = ensureOutputDir(input, program.out, isIgPubContext);
+  const isIgPubContext = inputFshFolder;
+  // TODO: Legacy support for top level fsh/ subdirectory. Remove when no longer supported.
+  const isLegacyIgPubContext = fshFolder && !inputFshFolder;
+  const outDir = ensureOutputDir(input, program.out, isIgPubContext, isLegacyIgPubContext);
 
   let config: Configuration;
   try {
@@ -116,7 +115,8 @@ async function app() {
 
   logger.info('Converting FSH to FHIR resources...');
   const outPackage = exportFHIR(tank, defs);
-  writeFHIRResources(outDir, outPackage, program.snapshot);
+  const useGeneratedFolder = !isLegacyIgPubContext;
+  writeFHIRResources(outDir, outPackage, program.snapshot, useGeneratedFolder);
 
   // If FSHOnly is true in the config, do not generate IG content, otherwise, generate IG content
   if (config.FSHOnly) {
@@ -124,7 +124,12 @@ async function app() {
   } else {
     const igDataPath = path.resolve(input, 'ig-data');
     logger.info('Assembling Implementation Guide sources...');
-    const igExporter = new IGExporter(outPackage, defs, igDataPath, isIgPubContext);
+    const igExporter = new IGExporter(
+      outPackage,
+      defs,
+      igDataPath,
+      isIgPubContext || isLegacyIgPubContext
+    );
     igExporter.export(outDir);
     logger.info('Assembled Implementation Guide sources; ready for IG Publisher.');
   }
--- a/src/utils/Processing.ts
+++ b/src/utils/Processing.ts
@@ -47,20 +47,31 @@ export function findInputDir(input: string): string {
   return input;
 }
 
-export function ensureOutputDir(input: string, output: string, isIgPubContext: boolean): string {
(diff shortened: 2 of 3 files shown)


src/app.ts
```diff
@@ -116,15 +115,21 @@ async function app() {
...
     logger.info('Exporting FSH definitions only. No IG related content will be exported.');
   } else {
     const igDataPath = path.resolve(input, 'ig-data');
     logger.info('Assembling Implementation Guide sources...');
-    const igExporter = new IGExporter(outPackage, defs, igDataPath, isIgPubContext);
+    const igExporter = new IGExporter(
+      outPackage,
+      defs,
+      igDataPath,
+      isIgPubContext || isLegacyIgPubContext
```
ngfreiter, line 131: Is the `||` here necessary? As far as I can tell, `isLegacyIgPubContext` is only ever true when `isIgPubContext` is already true. What do you think?

src/utils/Processing.ts
```diff
@@ -47,21 +47,32 @@ export function findInputDir(input: string): string {
...
+  isLegacyIgPubContext: boolean
+): string {
+  if (isIgPubContext || isLegacyIgPubContext) {
     // TODO: Message includes information about legacy support for top level fsh folder. Remove when not supported.
+    const directory = isIgPubContext ? 'input/fsh' : 'fsh';
     logger.info(
-      'SUSHI detected an "input/fsh" or "fsh" directory in the input path. As a result, SUSHI will operate in "IG Publisher integration" mode. This means:\n' +
-        '  - the "input/fsh" or "fsh" directory will be used as the input path\n' +
-        '  - the parent of the "fsh" directory (e.g., "../fsh") will be used as the output path unless otherwise specified with --out option\n' +
+      `SUSHI detected an "${directory}" directory in the input path. As a result, SUSHI will operate in "IG Publisher integration" mode. This means:\n` +
```
ngfreiter, line 60: If the directory is `fsh`, will this end up saying SUSHI detected `an fsh`?

src/utils/Processing.ts
```diff
@@ -47,21 +47,32 @@ export function findInputDir(input: string): string {
...
   let outDir = output;
-  if (isIgPubContext && !output) {
-    // When running in an IG Publisher context, default output is the parent folder of the tank
+  if (isLegacyIgPubContext && !output) {
+    // TODO: Legacy support for top level "fsh" directory. Remove when no longer supported.
+    // When running in a legacy IG Publisher context, default output is the parent folder of the tank
     outDir = path.join(input, '..');
     logger.info(`No output path specified. Output to ${outDir}`);
+  } else if (isIgPubContext && !output) {
+    // When running in a legacy IG Publisher context, default output is the parent folder of the input/fsh folder
```
ngfreiter, line 73: This is the non-legacy mode right?

ngfreiter (commented):
I had a couple of questions on this below.

src/app.ts
cmoesel: I think it is necessary.  My reading of it is that if `isIgPubContext` and `isLegacyIgPubContext` are mutually exclusive.  While they both can be `false` at the same time, they can never be `true` at the same time.

Line 73:
```
const isIgPubContext = inputFshFolder;
```

Line 75:
```
const isLegacyIgPubContext = fshFolder && !inputFshFolder;
```

One requires `inputFshFolder` and the other requires `!inputFshFolder`.

cmoesel (changes requested):
I tried this out and it works well.
* I tested it against a simple project using the new config and it worked as expected.
* I also tested against legacy config (using `fsh` folder) and confirmed no differences from master.
* I also tested against a flat project (no `fsh` or `input/fsh`).  In this case, it exported to `build/input/generated`.  This is a change from the way it was before -- but I think we discussed this previously and decided it was the right way forward...

So there are only minor things: one logging statement to fix and one comment to fix.

src/utils/Processing.ts
jafeltra: uh oh. It will. I'll update it.

src/utils/Processing.ts
jafeltra: Agreed!

src/utils/Processing.ts
```diff
@@ -47,21 +47,32 @@ export function findInputDir(input: string): string {
...
+  if (isIgPubContext || isLegacyIgPubContext) {
     // TODO: Message includes information about legacy support for top level fsh folder. Remove when not supported.
+    const directory = isIgPubContext ? 'input/fsh' : 'fsh';
     logger.info(
-      'SUSHI detected an "input/fsh" or "fsh" directory in the input path. As a result, SUSHI will operate in "IG Publisher integration" mode. This means:\n' +
-        '  - the "input/fsh" or "fsh" directory will be used as the input path\n' +
-        '  - the parent of the "fsh" directory (e.g., "../fsh") will be used as the output path unless otherwise specified with --out option\n' +
+      `SUSHI detected an "${directory}" directory in the input path. As a result, SUSHI will operate in "IG Publisher integration" mode. This means:\n` +
+        `  - the "${directory}" directory will be used as the input path\n` +
+        `  - the parent of the "${directory}" directory will be used as the output path unless otherwise specified with --out option\n` +
```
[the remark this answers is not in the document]
jafeltra: I think you're right. I'll try to clarify.

jafeltra:
I think I addressed the things @ngfreiter and @cmoesel pointed out. Let me know if the wording in the logged statement is still confusing or you notice anything else.

cmoesel (approved):
New changes look good -- and as noted before, it all works as expected.  Thanks, @jafeltra!

ngfreiter (approved):
Looks good to me. Thanks for fixing the `an`!

PR #593, opened 2020-09-02
Find slices of choice elements

### LINKED ISSUE #567: Repeated elements in StructureDefinition with nested slices
The following FSH can be used as input to demonstrate the behavior:
```
Profile: PizzaBusiness
Parent: Practitioner
* extension contains TheBusiness named business 1..1
* extension[business].valueString contains badSlice 0..1
* extension[business].valueString[badSlice] = "The Bad Slice"

Extension: TheBusiness
* valueString ^slicing.discriminator.type = #value
* valueString contains goodSlice 0..1
* valueString[goodSlice] = "The Good Slice"
```

In the resulting StructureDefinition for `PizzaBusiness`, element.differential contains the following:
```
...
{
  "id": "Practitioner.extension:business.valueString:badSlice",
  "path": "Practitioner.extension.valueString",
  "min": 0,
  "max": "1"
},
{
  "id": "Practitioner.extension:business.valueString:badSlice",
  "path": "Practitioner.extension.valueString",
  "min": 0,
  "max": "1"
},
{
  "id": "Practitioner.extension:business.valueString:badSlice",
  "path": "Practitioner.extension.valueString",
  "min": 0,
  "max": "1",
  "patternString": "The Bad Slice"
}
```

mint-thompson:
Fixes #567 and completes task [CIMPL-497](https://standardhealthrecord.atlassian.net/browse/CIMPL-497).

When a choice element is sliced, the new element is considered to be a
reslice of the choice slice. This does not line up perfectly with the
way PathParts are constructed, which would use the choice slice name
(such as `valueString`) as the PathPart's base. So, if one of the
elements to search through is a slice of a choice element, look for
other slices on that choice element when trying to find the matching
slice.

### DIFF
--- a/src/fhirtypes/StructureDefinition.ts
+++ b/src/fhirtypes/StructureDefinition.ts
@@ -228,7 +228,11 @@ export class StructureDefinition {
           ...unfoldedElements
         ]);
         if (matchingSlice) {
-          newMatchingElements.push(matchingSlice, ...matchingSlice.children());
+          newMatchingElements.push(
+            matchingSlice,
+            ...matchingSlice.children(),
+            ...matchingSlice.getSlices()
+          );
           fhirPathString = matchingSlice.path;
         }
       }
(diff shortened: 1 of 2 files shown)


ngfreiter (commented):
I think it's possible this bug can be addressed more directly. Instead of letting the element get filtered out of the matching elements, and then using the remaining elements to find it, it could be possible to just ensure the element does not get filtered out. On line 230, after the `sliceMAtchingValueX` function is called, we add the `matchingSlice` and its children:
```
newMatchingElements.push(matchingSlice, ...matchingSlice.children());
```
What if here we also added the slices of the newly creating `matchingSlice`:
```
newMatchingElements.push(
            matchingSlice,
            ...matchingSlice.children(),
            ...matchingSlice.getSlices()
          );
```
This way we wouldn't have to go back and search through connected elements, the code on line 701 should find this slice, since it will be in the `elements` array.

mint-thompson:
Huh, for some reason, I remember trying something like that. But I just made your change and all the tests still pass, so there you have it. Thank you for the recommendation!

ngfreiter (approved):
Looks good to me!

PR #597, opened 2020-09-11
Retain choice slice min for choice of exactly 1 type

### LINKED ISSUE #596: SUSHI changes min of `value[x]` to zero when type-slicing a choice narrowed to a single type
Given the following definition:
```
Alias: CitizenshipExtension = http://hl7.org/fhir/StructureDefinition/patient-citizenship

Profile: CitizenshipTest
Parent: Patient
* extension contains CitizenshipExtension named citizenship 0..1
* extension[citizenship].extension[code].valueCodeableConcept from http://foo.org
```

SUSHI creates a differential with this:
```
{
  "id": "Patient.extension:citizenship.extension:code.valueCodeableConcept",
  "path": "Patient.extension.extension.valueCodeableConcept",
  "min": 0,
  "max": "1",
  "binding": {
    "strength": "required",
    "valueSet": "http://foo.org"
  }
}
```

When the IG Publisher processes this, since `extension[citizenship].extension[code].valueCodeableConcept` is the same as `extension[citizenship].extension[code].value[x]` (since `value[x]` is constrained to `CodeableConcept` only), it applies `0..1` to `extension[citizenship].extension[code].value[x]` -- even though it was originally `1..1`.

I suspect SUSHI is making `valueCodeableConcept` `0..1` because when it does a type slice, it _assumes_ there is more than one choice, so it must make the slice `0..1`.  This is obviously an invalid assumption.

cmoesel:
Usually, we make the min of a new type slice 0 because in a type slice, if min isn't 0, then it's not really giving you much of a choice, is it?

But in the case of applying a type slice to a choice that has been constrained to just one element... we should keep the min since that type slice really is the same as the choice it slices (since it is no choice at all anymore).

Fixes #596

To test, you can try the example in the bug report on the master branch and on this branch.  If you compare output, you'll see `min: 0` (master) vs `min: 1` (this branch).  If you run IG Publisher on the results, QA will report errors on the master version and no errors on the PR version.

### DIFF
--- a/src/fhirtypes/ElementDefinition.ts
+++ b/src/fhirtypes/ElementDefinition.ts
@@ -1838,7 +1838,6 @@ export class ElementDefinition {
 
   /**
    * Creates a new slice on the element.
-   * TODO: Handle re-slicing?
    * @see {@link http://hl7.org/fhir/R4/profiling.html#slicing}
    * @param {string} name - the name of the new slice
    * @param { ElementDefinitionType } [type] - the type of the new slice; if undefined it copies over this element's types
@@ -1861,7 +1860,6 @@ export class ElementDefinition {
 
     // On a new slice, delete slice.min and slice.max and then reset them
     // so that they are always captured in diff
-    const originalMax = slice.max;
     delete slice.min;
     delete slice.max;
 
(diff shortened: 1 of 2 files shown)


ngfreiter (approved):
This works as promised and looks good to me!

jafeltra (approved):
Makes sense to me and works as described!

PR #599, opened 2020-09-15
Add function which runs SUSHI

ngfreiter:
This adds a `fshToFhir` function, which can be called to convert FSH definitions to FHIR. The function also tracks errors and warnings that occur during the running of SUSHI. The function runs a simplified version of SUSHI, no `IGExporter` specific processing is done, only the conversion from FSH to FHIR. The function takes a set of options that can be used to change the configuration that SUSHI runs with. The options allow you to set the logging level, set dependencies, set the version, and set a canonical. If no options are given, SUSHI runs with default values for each of these things.

### DIFF
--- a/src/utils/FSHLogger.ts
+++ b/src/utils/FSHLogger.ts
@@ -105,10 +105,16 @@ class LoggerStats {
 
 export const stats = new LoggerStats();
 
-class ErrorsAndWarnings {
+export class ErrorsAndWarnings {
   public errors: { message: string; location: string }[] = [];
   public warnings: { message: string; location: string }[] = [];
   public shouldTrack = false;
+
+  reset(): void {
+    this.errors = [];
+    this.warnings = [];
+    this.shouldTrack = false;
+  }
 }
 
 export const errorsAndWarnings = new ErrorsAndWarnings();
--- a/src/utils/FSHLogger.ts
+++ b/src/utils/FSHLogger.ts
@@ -50,6 +50,18 @@ const incrementCounts = format(info => {
   return info;
 });
 
+const trackErrorsAndWarnings = format(info => {
+  if (!errorsAndWarnings.shouldTrack) {
+    return info;
+  }
+  if (info.level === 'error') {
+    errorsAndWarnings.errors.push({ message: info.message, location: info.location });
+  } else if (info.level === 'warn') {
+    errorsAndWarnings.warnings.push({ message: info.message, location: info.location });
+  }
+  return info;
+});
+
 const printer = printf(info => {
   let level;
   switch (info.level) {
@@ -73,7 +85,7 @@ const printer = printf(info => {
 });
 
 export const logger = createLogger({
-  format: combine(incrementCounts(), withLocation(), printer),
+  format: combine(incrementCounts(), trackErrorsAndWarnings(), withLocation(), printer),
   transports: [new transports.Console()]
 });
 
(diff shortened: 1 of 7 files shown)


cmoesel (changes requested):
This is cool and pretty clean.  I like it.  In the future, we might consider extending options to allow a `FhirCache` object to be passed in that fulfills an API for storing and retrieving dependencies.  That way implementations don't _have_ to use the default `~/.fhir` file system storage.  And then FSHOnline can use this function as well.  But that's more of a rainy day project.

I've suggested a few minor changes to consider.

ngfreiter:
Alright @cmoesel, all your comments should be addressed in 8c340ac.

cmoesel (changes requested):
I tested this and it's pretty cool! I think we need to make one change, however.  Much like the CLI, we should _not_ export the `snapshot` by default.  I'm torn on whether or not we should make an option to allow others to turn it on.  SUSHI-produced snapshots are not really something I want people to rely on.  What do you think?

BTW -- for others testing this, here's a node script I wrote:
```js
const { sushiClient } = require('fsh-sushi');

const FSH0 = `
Alias: $USCoreLab = http://hl7.org/fhir/us/core/StructureDefinition/us-core-observation-lab
Alias: $USCorePatient = http://hl7.org/fhir/us/core/StructureDefinition/us-core-patient
Alias: $LOINC = http://loinc.org

Profile: MyObservationProfile
Parent: $USCoreLab
* value[x] only Quantity

Instance: MyPatient
InstanceOf: $USCorePatient
* identifier[0].system = "http://foo.org/id"
* identifier[0].value = "1"
* name[0].given[0] = "Homer"
* name.family = "Simpson"
* gender = #male


Instance: MyObservationInstance
InstanceOf: MyObservationProfile
* code = $LOINC#1234-5
* status = #final
* valueQuantity = 55 'mm'
`;

const FSH1 = `
Instance: MyOtherObservationInstance
InstanceOf: MyObservationProfile
* code = $LOINC#9876-5
* status = #final
* valueQuantity = 42 'cm'
`

const OPTIONS = {
  canonical: 'http://foo.org/bar',
  version: '0.0.1',
  dependencies: [{ packageId: 'hl7.fhir.us.core', version: '3.1.1' }],
  logLevel: 'debug'
}
void async function() {
  const results = await sushiClient.fshToFhir([FSH0, FSH1], OPTIONS);
  console.log(JSON.stringify(results, null, 2));
}();
```

ngfreiter:
In 2548f31 I removed the exporting of snapshots. Also, instead of returning a `Package`, the `fshToFhir` function now returns a flat array of FHIR artifacts (with no snapshots). I felt like this was simpler, and makes it so that someone doesn't have to understand our `Package` class to use this function.

cmoesel (approved):
Looks good, works well, I like it.  And I think the simplification is probably a good call.

src/utils/FSHLogger.ts
```diff
@@ -92,3 +112,17 @@ class LoggerStats {
 }
 
 export const stats = new LoggerStats();
+
+export class ErrorsAndWarnings {
```
jafeltra, line 116: Oooo I think this could help clean up how we access the errors and warnings in FSH Online ✨

jafeltra (approved):
This looks good to me! I used Chris's script to test out the function as well, and it looks good!

PR #600, opened 2020-09-15
Change sushi --init to work with new IG integration

ngfreiter:
This updates `sushi --init` to work with the new IG Publisher integration. The following changes were made:
* place `index.md` in `input/pagecontent` instead of `fsh/ig-data/input/pagecontent`.
* place `patient.fsh` in `input/fsh` instead of `fsh`.
* place `sushi-config.yaml` in root instead of `fsh`. 
* don't ignore `input` in the `.gitignore` file.

This can only be actually tested once #589 is merged to move the input folder, and once we merge a PR to move `sushi-config.yaml` to the root of the IG Publisher, instead of in `input/fsh`.


ngfreiter:
So I'm leaving it as a draft for now, since it can't totally be tested. But if we are fine with merging it without doing those tests, I can un-draft it.

ngfreiter:
This draft is just too far gone at this point, so I'm going to just close and re-create.

PR #605, opened 2020-09-16
Align SUSHI classes with FSH

ngfreiter:
This aligns two of the `Rule` classes in SUSHI with the spec. The changes are:
```
* FixedValueRule -> AssignmentRule
* ValueSetRule -> BindingRule
```
Changing the `FixedValueRule` class was relatively involved, and also involved my attempt to remove any instance of the word `fix` being used incorrectly. Anywhere `fix` was used, I've tried to change it to `assign`, to align with the name of the new rule. Changing the `ValueSetRule` was less involved.

If these changes are larger in scope than we want, we could just change the class names alone, and keep the other usage of the word `fix`, and just try to use `assign` in the future.


cmoesel (dismissed):
Oh man.  I had no idea how pervasive this change would be.  Wowsers.  Sorry for making you go through that tedious task, but look at the bright side:
1. I think it is better as a result.  I like the alignment w/ the terms in the spec.
2. You just got a ton of GitHub cred with all those lines of changed code!

jafeltra (commented):
I'm still looking through the file changes, but I noticed that we still use `fixedValueRule` in the grammar in FSH.g4. Is that something we'd want to update as well?

cmoesel:
> I'm still looking through the file changes, but I noticed that we still use `fixedValueRule` in the grammar in FSH.g4. Is that something we'd want to update as well?

@ngfreiter and I discussed this.  Technically, the grammar is part of the released spec (https://build.fhir.org/ig/HL7/fhir-shorthand/reference.html#appendix-formal-grammar) so we actually _don't_ want to change that right now.  We want the SUSHI grammar to match the published spec grammar.  But we might change the grammar for STU2.

jafeltra:
> Technically, the grammar is part of the released spec (https://build.fhir.org/ig/HL7/fhir-shorthand/reference.html#appendix-formal-grammar) so we actually _don't_ want to change that right now. We want the SUSHI grammar to match the published spec grammar. But we might change the grammar for STU2.

Ahhh I forgot about that wrinkle. That makes sense to me!

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1626,17 +1635,17 @@ export class ElementDefinition {
...
    */
-  getFixableDescendents(): ElementDefinition[] {
-    const fixableChildren = this.children(true).filter(e => e.min > 0);
-    let fixableDescendents: ElementDefinition[] = [];
-    fixableChildren.forEach(fc => {
-      fixableDescendents = fixableDescendents.concat(fc.getFixableDescendents());
+  getAssignableDescendents(): ElementDefinition[] {
+    const assignableChildren = this.children(true).filter(e => e.min > 0);
+    let assignableDescendents: ElementDefinition[] = [];
+    assignableChildren.forEach(fc => {
```
jafeltra, line 1645: This is incredibly minor, but the fc variable here should probably be ac instead now. But I'm thinking this might be too nit picky and it's just fine. I'm guessing you agree since that's what you have now. Is that a fair assumption?

test/fhirtypes/ElementDefinition.assignBoolean.test.ts
```diff
@@ -119,32 +129,34 @@ describe('ElementDefinition', () => {
...
-      }).toThrow('Cannot fix true to this element; a different boolean is already fixed: false.');
+        userSelected.assignValue(true, true);
+      }).toThrow(
+        'Cannot assign true to this element; a different boolean is already assigned: false.'
+      );
       expect(clone).toEqual(userSelected);
     });
 
-    it('should throw FixedToPatternError when trying to change fixed[x] to pattern[x]', () => {
+    it('should throw AssignedToPatternError when trying to change fixed[x] to pattern[x]', () => {
```
jafeltra, line 139: This is also super minor, but here and in a few other places, the test name references "AssignedToPatternError", which I think should still be "FixedToPatternError". Again, if this is too nit picky, that's fine. It is just in a test description so it doesn't effect anything.

test/fhirtypes/ElementDefinition.assignString.test.ts
```diff
@@ -0,0 +1,1095 @@
...
+        'Cannot assign "bar foo" to this element; a different string is already assigned: "foo bar".'
+      );
+      expect(() => {
+        batchLotNumber.assignValue('bar foo', true);
+      }).toThrow(
+        'Cannot assign "bar foo" to this element; a different string is already assigned: "foo bar".'
+      );
+    });
+
+    it('should throw ValueAlreadyAssignedError when assigning an already assigned string by assigned[x]', () => {
```
jafeltra, line 73: There's a handful of places in here that refer to `assigned[x]`. I think it should still be `fixed[x]` right?

jafeltra (commented):
These changes look fantastic! I got through all the diffs and just had a few very minor comments. Let me know if you think any of them are worth addressing.

src/fhirtypes/ElementDefinition.ts
ngfreiter: You're giving me a lot of credit to assume that I noticed the `fc` vs `ac`. I will change this, and do a quick look for some other sneaky `f`s.

test/fhirtypes/ElementDefinition.assignBoolean.test.ts
ngfreiter: This one I actually think is fine, since the `FixedToPatternError` is thrown when we try to convert a `fixed[x]` value to a `pattern[x]` value, which cannot be done, so the error is for going from "Fixed to Pattern".

test/fhirtypes/ElementDefinition.assignString.test.ts
ngfreiter: Correct!

ngfreiter:
Ok @jafeltra, I've made the changes from your comments.

src/fhirtypes/ElementDefinition.ts
jafeltra: I hadn't noticed any others, so I'd say the others are fair game to stay at this point.

jafeltra (approved):

PR #606, opened 2020-09-17
Remove generation of package-list from history

ngfreiter:
Fixes #603 by removing support for SUSHI generating or processing the `package-list.json` file. 

I wasn't sure if this made more sense to be merged into the `new-ig-publisher-integration`, or just `master`, but I put this up as `master` since I feel like this can stand alone without the rest of the integration changes.

PR #607, opened 2020-09-18
Update handling of non-FSH files

jafeltra:
This PR updates how we handle non-FSH files. The main differences are:

- pagecontent/page pages are no longer renamed in the new configuration, but continue to be renamed in the legacy configuration
- ig.ini does not merge values. This is a breaking change for the legacy configuration, but if we eventually remove all support for ig.ini, that should matter slightly less
- In the new configuration, if a user provided file and a sushi-config.yaml property both exist, we prefer the provided file. The legacy configuration still prefers the sushi-config property. The warning messages change for each case as appropriate.
- If a user provided a file, we do not copy the file with the SUSHI warning box at the top.

Some changes that have been discussed but I haven't implemented here:
- Output of all files to `fsh-generated/resources`, including the ImplementationGuide JSON. I believe this is part of @cmoesel's task, but if I'm mistaken, I can update this.
- Even when we do output to that separate directory, the code changes here assume that we do not copy over user provided files like `menu.xml`. I believe we discussed that, but I'm not positive.
- ig.ini follows the new changes even though I think it is being removed, but I don't think there's a task yet for it so I wasn't completely sure. However, since the removal of package-list.json is in a PR, I didn't make any changes to the handling of that since I assumed it would just go away.


src/ig/IGExporter.ts
```diff
@@ -46,16 +46,21 @@ export class IGExporter {
...
   private readonly config: Configuration;
   private readonly configName: string;
+  private shouldCopyFiles: boolean;
   constructor(
     private readonly pkg: Package,
     private readonly fhirDefs: FHIRDefinitions,
     private readonly igDataPath: string,
-    private readonly isIgPubContext: boolean = false
+    private readonly isIgPubContext: boolean = false,
+    private readonly isLegacyIgPubContext: boolean = false
```
cmoesel, line 55: If you assign a default variable, then I don't think you need to explicitly declare the type.  TS will infer it from the value.

src/ig/IGExporter.ts
```diff
@@ -267,7 +272,7 @@ export class IGExporter {
     // To work around this, we set the necessary permissions on executable
     // scripts after copying them to the IG path.
     try {
-      if (!this.isIgPubContext) {
+      if (!this.isIgPubContext || this.isLegacyIgPubContext) {
```
cmoesel, line 275: Shouldn't this be `if (!this.isgPubContext && !this.isLegacyIgPubContext) {`?

src/ig/IGExporter.ts
```diff
@@ -321,49 +326,76 @@ export class IGExporter {
...
       if (filePath) {
+        const filePathString = `${path.basename(this.igDataPath)}${path.sep}${path.relative(
+          this.igDataPath,
+          filePath
+        )}`;
+
+        let preferredFileMessage =
+          `Since a ${filePathString} file was found, the "indexPageContent" property in the ${this.configName} ` +
+          `will be ignored and an index.md file will not be generated. Remove the ${filePathString} ` +
+          `file to use the "indexPageContent" property in ${this.configName} to generate an index.md file instead.`;
```
cmoesel, line 339: We really do not want to encourage the use of `indexPageContent`.  So I think this message should actually just tell the user to remove the `indexPageContent` property from the config.

src/ig/IGExporter.ts
```diff
@@ -552,18 +586,19 @@ export class IGExporter {
     const pageData = pages.map(page => {
       const nameParts = page.match(/^(\d+)_(.*)/);
       let prefix: number = null;
-      let name: string;
+      let title: string;
```
cmoesel, line 589: I kind of find `title` confusing, as I expect it is the human-readable title.  But it's not.  It's just the name without the numeric prefix.  How about calling what you call `nameWithPrefix` just `name` (since that's the one that will survive past deprecation) and call what you currently call `title` `nameWithoutPrefix`.  What do you think?

src/ig/IGExporter.ts
```diff
@@ -659,22 +694,38 @@ export class IGExporter {
...
       return;
     }
 
     // If user provided file and config, log a warning but prefer the config.
     if (existsSync(menuXMLDefaultPath) && this.config.menu) {
-      logger.warn(
-        `Found both a "menu" property in ${this.configName} and a menu.xml file at ig-data${path.sep}input${path.sep}includes${path.sep}menu.xml. ` +
+      const filePathString = `${path.basename(this.igDataPath)}${path.sep}input${path.sep}includes${
+        path.sep
+      }menu.xml`;
```
cmoesel, line 706: I've noticed this pattern w/ `${path.sep}` a few times now.  You know about `path.join`, right?  E.g., the above could be:
```
const filePathString = path.join(path.basename(this.igDataPath), 'input', 'includes', 'menu.xml');
```

cmoesel (changes requested):
A few minor comments.

src/ig/IGExporter.ts
jafeltra: Oops this was definitely a tunnel vision issue, and I was too focused on what used to be there. I can go back through and use `path.join`.

jafeltra:
@cmoesel I believe all of your comments have been addressed in 066de98.

I may need to make some changes to support copying package-list.json, depending on where @ngfreiter's PR #606  is merged and when.

jafeltra:
2884875 adds the required flag so that when this is rebased, it will only copy the package-list.json file if it exists and we are in the legacy tank configuration. When #606 is merged and the `new-ig-publisher-integration` branch is rebased with master, we will need to be sure that the `if (this.config.history)` is removed and that only the `if (isIgDataPackageList && this.shouldCopyFiles)` is present. But that case should be the correct case now, and since I had changed the warning message in the `if (this.config.history)` case, there should be a conflict in the rebase, which I think will be a helpful reminder.  I can't take out that case now since that would required additional changes from #606 that I thought should stay separate from this PR.

With that, I think all the necessary changes have been made here and it's ready for re-review.

src/app.ts
```diff
@@ -88,8 +88,10 @@ async function app() {
 
   // Load custom resources specified in ig-data folder
   loadCustomResources(path.join(input, 'ig-data', 'input'), defs);
-  if (isIgPubContext && !fs.existsSync(path.join(input, 'ig-data'))) {
+  if (isLegacyIgPubContext && !fs.existsSync(path.join(input, 'ig-data'))) {
     loadCustomResources(path.join(input, '..', 'input'), defs);
+  } else if (isIgPubContext) {
+    loadCustomResources(path.join(input, '..'), defs);
   }
```
ngfreiter, line 95: I'm not sure it makes sense to have three possible ways to load here. The `if` statement on line 91 was originally added to load resources that were in the `<root>/input` folder for the new IG Publisher integration. But now I think that `if` statement is not necessary, since this case is handled by line 93-94.

If someone is using the "Legacy" context, then my expectation is that they would not anticipate that we would load in definitions from `<root>/input`, since that is not a pattern that has traditionally been a part of the "Legacy" context. So would this be better off as just two cases, If we are in "Legacy" mode, load from `ig-data/input`, and if we are in the new mode, load from `<root>/input`.

Let me know if you think I'm off base, I could just be confusing myself about something you've already thought through.

src/ig/IGExporter.ts
```diff
@@ -267,7 +272,7 @@ export class IGExporter {
     // To work around this, we set the necessary permissions on executable
     // scripts after copying them to the IG path.
     try {
-      if (!this.isIgPubContext) {
+      if (!this.isIgPubContext || !this.isLegacyIgPubContext) {
```
ngfreiter, line 275: Shouldn't this be `!this.isIgPubContext && !this.isLegacyIgPubContext`? The way this is right now, I think as long as you are in one of the contexts, this if statement will be true, and we will try to `chmod` on files that don't actually exist, I get an error like this:
image

src/ig/IGExporter.ts
```diff
@@ -976,14 +1031,15 @@ export class IGExporter {
...
-                pathEnd,
-                `${resourceJSON.resourceType}-${resourceJSON.id}${path.extname(file)}`
-              );
-              this.copyAsIs(inputPath, outputPath);
+
+              if (this.shouldCopyFiles) {
+                const inputPath = path.join(dirPath, file);
+                const outputFileName = this.isLegacyIgPubContext
+                  ? `${resourceJSON.resourceType}-${resourceJSON.id}${path.extname(file)}`
+                  : file;
```
ngfreiter, line 1039: I'm not sure I understand here. Under what conditions is this not a `LegacyIgPubContext`?

src/ig/IGExporter.ts
```diff
@@ -1125,7 +1195,9 @@ export class IGExporter {
     iniObj.template = this.config.template;
     const comment = [
       'This ig.ini was generated using the template property in sushi-config.yaml. To provide your own',
-      'ig.ini, create an ig.ini file in the ig-data folder with required properties: ig, template.',
+      `ig.ini, create an ig.ini file in the ${path.basename(
+        this.igDataPath
+      )} folder with required properties: ig, template.`,
```
ngfreiter, line 1200: Thanks for the help prettier 🙄

ngfreiter (changes requested):
I had a few small comments, and one thing that I think needs to be fixed. Thanks for doing this task, it definitely is a tricky one!

src/app.ts
cmoesel: Oh, yeah.  I missed this -- but agree w/ @ngfreiter.  In legacy mode, we read _only_ from `ig-data/input`.  In new (non-legacy) mode, we read _only_ from `$root/input`.

jafeltra: Ah no, that is a good point. I had lost track of the fact that the load from `<root>/input` case had been added as part of this new configuration, so I can consolidate.

src/ig/IGExporter.ts
jafeltra: Hm this is a good point. I had changed that line before we had gotten to the current configuration for output and hadn't thought to go back to it. I think that the check for `isLegacyIgPubContext` is wrong, but I actually think that it should have supported this name change in any legacy configuration, not just a legacy IG Publisher configuration. So I think getting rid of the ternary operator here and only supporting the rename is correct. And the surrounding `if (this.shouldCopyFiles)` will support the current configuration. This means some of the tests will need to change as well, so I'll make those changes  now.

jafeltra: I updated the test to specifically check that no files are copied, which I believe is the expected behavior. Any file that should be copied is from a legacy tank configuration and therefore should be renamed, so I got rid of the ternary operator on this line.

src/ig/IGExporter.ts
jafeltra: Ugh yes. I did a bad job fixing this line a few hours ago. Should be fixed now.

src/app.ts
jafeltra: I believe 7f06748 should have the simplified configuration, but let me know what you think.

src/ig/IGExporter.ts
```diff
@@ -1125,7 +1195,9 @@ export class IGExporter {
     iniObj.template = this.config.template;
     const comment = [
       'This ig.ini was generated using the template property in sushi-config.yaml. To provide your own',
```
cmoesel, line 1197: This is showing up in legacy projects where `config.yaml` is used too, so we should replace `sushi-config.yaml` with `${this.configName}`

src/app.ts
```diff
@@ -86,10 +86,13 @@ async function app() {
...
-  loadCustomResources(path.join(input, 'ig-data', 'input'), defs);
-  if (isIgPubContext && !fs.existsSync(path.join(input, 'ig-data'))) {
-    loadCustomResources(path.join(input, '..', 'input'), defs);
+  // Load custom resources
+  if (!isIgPubContext) {
+    // In legacy configuration (both IG publisher context and any other tank), resources are in ig-data/input/
+    loadCustomResources(path.join(input, 'ig-data', 'input'), defs);
+  } else if (isIgPubContext) {
+    // In current tank configuration (input/fsh), resources will be in input/
+    loadCustomResources(path.join(input, '..'), defs);
```
ngfreiter, line 95: Should this just be an `if`, `else`?

src/ig/IGExporter.ts
```diff
@@ -1313,7 +1360,8 @@ export class IGExporter {
       'files.',
       '',
       'NOTE: This file does not currently list the FHIR resources and examples generated from .fsh files. It only',
-      'lists those files generated from project configs or the contents in the ig-data folder.',
+      'lists those files generated from project configs or the contents in the',
+      `${path.basename(this.igDataPath)} folder.`,
```
cmoesel, line 1364: Is this intentionally adding a newline here?  This is an example of the message w/ master:
```
NOTE: This file does not currently list the FHIR resources and examples generated from .fsh files. It only
lists those files generated from project configs or the contents in the ig-data folder.
```
and this is in this PR:
```
NOTE: This file does not currently list the FHIR resources and examples generated from .fsh files. It only
lists those files generated from project configs or the contents in the
ig-data folder.
```

cmoesel (commented):
A couple more minor comments, otherwise looks good!

ngfreiter (commented):
I've noticed that I don't think this treats `ignoreWarnings.txt` quite right. If I am using SUSHI in the new mode, and I have a defined `ignoreWarnins.txt` in my `input` folder, SUSHI is overwriting it with the "generated" one. I don't think we want to do that. Which then leads me to wonder if we want to just forget about `ignoreWarnings.txt` all together, since we are going that way for other static files, and `ignoreWarnings.txt` seems the least useful of the static files we generate. So if we can make a decision now that `ignoreWarnings.txt` is on its way out, then I'm ok with leaving it as-is, otherwise, I think `ignoreWarnings.txt` needs a `shouldCopyFiles` flag.

cmoesel:
@ngfreiter -- I was also thinking that `ignoreWarnings.txt` should go the way of the buffalo.  As long as the bug you noted is only in the "new" mode, then I think we can address this in a separate "stop generating ignoreWarnings.txt" task.

ngfreiter:
> @ngfreiter -- I was also thinking that `ignoreWarnings.txt` should go the way of the buffalo. As long as the bug you noted is only in the "new" mode, then I think we can address this in a separate "stop generating ignoreWarnings.txt" task.

Aww man now I'm sad about the buffalo. Yeah the bug was only in the "new" mode, so we can leave it for now.

src/ig/IGExporter.ts
jafeltra: No. I think this was one of the places I tried to fight prettier and put it on a new line, but I'll revert it.

cmoesel (approved):
Looks good to me!

ngfreiter (approved):
Looks good!

PR #609, opened 2020-09-22
Exit with code 0 when there is no FSH

mint-thompson:
Completes task [CIMPL-519](https://standardhealthrecord.atlassian.net/browse/CIMPL-519).

The IG publisher may run SUSHI when no SUSHI project is present. In this case, show a message indicating that there are no FSH files to read. This should not be considered an error by the calling process. There may also be no configuration present when there are no FSH files. This should also not be considered an error by the calling process.

Note that this still logs an `error` message when configuration is missing and cannot be inferred. If folks think this is confusing, I can address that as well.


jafeltra:
Hm I think I see what you mean with the logged error message. When I have a tank that has no .fsh files and no config/nothing that can be used to infer configuration, I get a logged error message about a missing configuration and then I get the logged info message that there are no FSH files. I feel like it would be nice to not have the config error message when there are no .fsh files, because the lack of configuration isn't an issue in that case, right? If so, I don't feel super strongly about this, so if it's not easy or clean to add into the code or if you feel strongly in favor of how it is now, I'm good with how this works now!

jafeltra:
The order of the messages look good to me now. I think it makes more sense reading the messages to not have the config error when there are no `.fsh` files.

The one thing I can't quite decide about is if it feels weird to have the error message moved out of the `readConfig` function. It seems like anyone using that function would probably benefit from that error message and now they'd have to provide it themselves. Is there a reason we shouldn't just wait to call `readConfig` once we know there are `.fsh` files? That said though, the `readConfig` function is only used in that once place and so far only used by SUSHI. So maybe this isn't a big deal. What do you think?

jafeltra (approved):
This looks good to me! I like the current order of the error messages and I think I like this current approach with the missing config error message in the `readConfig` function. Plus the exit codes are still the appropriate exit codes!

cmoesel (dismissed):
This looks good and works well, but let's _not_ merge this until we merge all the new config stuff... because I don't want to do that massive rebase again!

cmoesel (approved):
Looks good!

PR #610, opened 2020-09-23
Revert "Remove generation of package-list from history (#606)"

cmoesel:
This reverts commit 6b17523161284176531f192f7bb1ac2275f82997.

To provide a smooth transition, we should deprecate package-list.json w/ warnings before outright removing it...


ngfreiter (approved):
image

PR #611, opened 2020-09-23
Remove static files

ngfreiter:
This removes the "static files" that SUSHI would add to IGs, namely the `_genonce`, `_gencontinuous`, and `_updatePublisher` scripts, along with `ignoreWarnings.txt`. The scripts are still added from https://github.com/FHIR/sample-ig, but they are only added once `sushi --init` is run. As far as I can tell, the IG Publisher runs without `ignoreWarnings.txt`, but @cmoesel, thought it may not, so a reviewer may want to double check me on that.


src/utils/Processing.ts
```diff
@@ -284,14 +285,16 @@ export function init(): void {
...
-    });
-
+  console.log('Downloading publisher scripts from https://github.com/FHIR/sample-ig');
+  for (const script of [
+    '_genonce.bat',
+    '_genonce.sh',
+    '_updatePublisher.bat',
+    '_updatePublisher.sh'
+  ]) {
+    const res = await axios.get(`http://raw.githubusercontent.com/FHIR/sample-ig/master/${script}`);
```
cmoesel, line 295: It looks to me like if the download fails, an error will be thrown and the function will bail here.  I think we want to wrap this in try/catch so if a download fails, we can inform the user (and maybe even direct them to the URL for manual download), and keep on processing the init.

It would be sad if someone is in a locked down environment, or has weird proxies or something, and can't run SUSHI init at all.

cmoesel (changes requested):
I think we need some error-handling on the download.

Also... I think it would be nice if SUSHI could detect that there are no scripts in the expected places and output an info message telling the user where they can download them.  Users who have relied on our script generation might appreciate that.

src/app.ts
```diff
@@ -138,6 +138,17 @@ async function app() {
     );
     igExporter.export(outDir);
     logger.info('Assembled Implementation Guide sources; ready for IG Publisher.');
+    if (
+      !isIgPubContext &&
+      !isLegacyIgPubContext &&
```
cmoesel, line 143: I actually think it was a mistake before for us not to include these scripts in publisher contexts.  It confused people -- and since most people actually _started_ with SUSHI they still needed the scripts.

Anyway, I think it doesn't matter what context we're in.  If we don't see the scripts, let's log that info message no matter what.

src/utils/Processing.ts
```diff
@@ -292,8 +292,14 @@ export async function init(): Promise<void> {
...
   ]) {
-    const res = await axios.get(`http://raw.githubusercontent.com/FHIR/sample-ig/master/${script}`);
-    fs.writeFileSync(path.join(outputDir, script), res.data);
+    try {
+      const res = await axios.get(
+        `http://raw.githubusercontent.com/FHIR/sample-ig/master/${script}`
+      );
+      fs.writeFileSync(path.join(outputDir, script), res.data);
+    } catch (e) {
+      logger.error(`Unable to download ${script}: ${e.message}`);
```
cmoesel, line 301: I think it would be good to tell the user the full URL that failed download -- that way they can download it themselves (or more easily investigate).

cmoesel (changes requested):
A couple more minor fixes.

jafeltra:
This looks pretty good to me! I did notice that when I tried to run the `_updatePublisher` (and probably `_genonce`, I just hadn't gotten to that yet), I got the following error: `zsh: permission denied: ./_updatePublisher.sh`.

I think we might need to set the permissions on the .sh files again like we were doing in the IGExporter. Something like this worked when I quickly tested it out: ```if (script.endsWith('.sh')) {
   fs.chmodSync(path.join(outputDir, script), 0o755);
}```

jafeltra (approved):
This looks good to me! It works as expected, I can run the scripts correctly now, and I didn't see any errors related to a missing `ignoreWarnings` file.

ngfreiter:
@cmoesel this should be ready for re-review from your comments.

cmoesel (approved):
I didn't get a chance to manually test it again, but I reviewed the code and the changes look good to me.  I trust @jafeltra's results from when she manually tested the code.  So I'm good with this.

PR #612, opened 2020-09-23
Warn when history is specified

ngfreiter:
This adds a warning when the `history` property is specified on the config and someone is in the legacy IG Publisher mode, and adds an error when someone is in the new IG Publisher mode, and in this second case the "history" property is ignored.

I felt icky about putting this warning into the reading of the configuration, because it would mean we have to pass in the information about which context SUSHI is running in to the `readConfig` function. It felt goofy to do that just for a warning that we plan on removing sometime soon anyway. But I figured that may have been an unpopular decision, so let me know if I should move this to the `importConfiguration` function.


cmoesel:
This is pretty clean, but I think we handled this type of thing in other cases in the IGExporter.  So that would be another approach (that would be a little more consistent w/ the rest).

ngfreiter:
@cmoesel moved to the IGExporter in 3bd323d.

jafeltra (approved):
This looks good to me!

However, looking at this, I realized that the `if (this.config.history)` case that has these changes in it might actually need to be updated. I think right now it will overwrite a provided package-list.json if there is also a history property. I hadn't made that change in my PR because I thought this case was going away, but since it's back, I think it'll need to be updated.

Definitely not part of this PR though, so - approved!

jafeltra:
I think I missed the `return` you have after the error, so I actually don't think the file will be overwritten.

ngfreiter:
Ok @jafeltra and @cmoesel, I've added the `package-list.json` that would be generated by `history` to the log message when in the new IG Publisher. After some discussion to @jafeltra, as we understand it, everything is working as expected in terms of overwriting files in the legacy mode and the new IG Publisher mode.

jafeltra (approved):
Makes sense to me! And it covers all the cases for copying and overwriting in legacy configuration and not overwriting in new configuration.

cmoesel (approved):
Looks good to me!

PR #613, opened 2020-09-24
Use fsh-generated (and more!)

cmoesel:
This PR makes the following changes to the _new_ publisher integration:
* resources are generated to `fsh-generated/resources`
* IG JSON is generated to `fsh-generated/resources`
* `menu.xml` is generated to `fsh-generated/includes`
* `"template"` config property is deprecated - if used, log error, do not generate
  * NOTE: in legacy mode, log warning and behave as in the past
* `sushi-config.yaml` should now be in the _root_ of the project
* updates log message when a legacy mode is used
* adds new log messages to output the sushi CLI arguments (which was helpful in debugging the publisher)
[NewIgStructure.zip](https://github.com/FHIR/sushi/files/5278146/NewIgStructure.zip)

* fixes some issues w/ actually using it w/ IG Publisher

To test the new mode manually, you can use the attached project -- which includes a special local `sushi-dev-template` that is configured to look in the new `fsh-generated` locations for resources and menu.  One tricky bit is that when you run `_genonce`, the IG Publisher will invoke your _globally installed_ SUSHI.  So you either need to `npm pack` from this PR and install that package globally, or you can run `_genonce -no-sushi` to prevent the IG Publisher from invoking SUSHI.


src/utils/Processing.ts
```diff
@@ -28,31 +28,62 @@ export function findInputDir(input: string): string {
...
+    }
+    if (!fs.existsSync(path.join(input, 'ig-data', 'ig.ini'))) {
+      msg += `  - if you used the "template" property in your config, remove it and manage .${path.sep}ig.ini directly\n`;
+    }
+    if (!fs.existsSync(path.join(input, 'ig-data', 'package-list.json'))) {
+      msg += `  - if you used the "history" property in your config, remove it and manage .${path.sep}package-list.json directly\n`;
+    }
+    msg +=
+      '  - ensure your .gitignore file is not configured to ignore the sources in their new locations\n' +
+      '  - add /fsh-generated your .gitignore file to prevent SUSHI output from being checked into source control\n\n' +
```
ngfreiter, line 62: "add /fsh-generated your" -> "add /fsh-generated to your"

src/utils/Processing.ts
```diff
@@ -28,31 +28,62 @@ export function findInputDir(input: string): string {
...
```
ngfreiter, line 84: "add /fsh-generated your" -> "add /fsh-generated to your"

src/utils/Processing.ts
```diff
@@ -199,8 +213,9 @@ export function writeFHIRResources(
     folder: string,
     resources: { getFileName: () => string; toJSON: (snapshot: boolean) => any }[]
   ) => {
-    folder = useGeneratedFolder ? 'generated' : folder;
-    const exportDir = path.join(outDir, 'input', folder);
+    const exportDir = useGeneratedFolder
```
ngfreiter, line 216: I think in this function, if the `useGeneratedFolder` flag is equivalent to the `isIgPubContext` flag in `app.ts`, can we just change the name here to match, so that within `writeFHIRResources`, the flag is called `isIgPubContext`?

src/ig/IGExporter.ts
```diff
@@ -1151,18 +1154,19 @@ export class IGExporter {
   addIgIni(igPath: string): void {
     const inputIniPath = path.join(this.igDataPath, 'ig.ini');
     if (this.config.template != null) {
+      const filePathString = path.join(path.basename(this.igDataPath), 'ig.ini');
       if (existsSync(inputIniPath)) {
-        const filePathString = path.join(path.basename(this.igDataPath), 'ig.ini');
-
```
ngfreiter, line 1156: I can't figure out why these lines had to switch. What am I missing?

src/ig/IGExporter.ts
```diff
@@ -1151,18 +1154,19 @@ export class IGExporter {
...
         let preferredFileMessage =
           `Since a ${filePathString} file was found, the "template" property in the ${this.configName} ` +
-          `will be ignored and an ig.ini file will not be generated. Remove the ${filePathString} ` +
-          `file to use the "template" property in ${this.configName} and generate an ig.ini file instead.`;
+          'will be ignored and an ig.ini file will not be generated. Support for the "template" property ' +
+          `in ${this.configName} will be removed in a future version of SUSHI. Please remove the "template" ` +
+          `property in ${this.configName} and manage the ig.ini file directly.`;
```
ngfreiter, line 1163: Something feels fishy about how these errors and warnings are lining up. What I'm seeing is:
* If I'm using an `input/fsh` folder and I have `template` in my `sushi-config.yaml` and no `ig.ini` file, I get an error telling me not to use `template`, with what would be the contents of the ini file.
* If I'm using an `input/fsh` folder and I have `template` in my `sushi-config.yaml` and I do have `ig.ini`, I get a warning telling me that my `template` property will be ignored.

I feel like the second case should be an error as well. If you disagree, then we will have to change the `package-list` PR, since the situation for `history` is analogous to the one for `template`, but the errors and warnings being logged are not currently analogous.

ngfreiter (commented):
I have a few comments/questions below.

src/utils/Processing.ts
cmoesel: That was there when I got here. ;-)  That said, it used to be used in the new IG Publisher context _and_ the legacy flat context.  Now that it's used only in the new IG Publisher context, your suggestion makes total sense.

src/ig/IGExporter.ts
cmoesel: Oh.  That's easy.  Because at one point in development (some commit in this multi-commit PR) I was using `filePathString` in the `else` clause.  Then I stopped -- and never rescoped `filePathString` back to the narrower scope.  Oops.  I can do that.

src/ig/IGExporter.ts
cmoesel: So... I was thinking that if you don't have `ig.ini` already, then the end result would be a project that _won't_ build w/ the publisher.  Clearly an error condition.  If you have an `ig.ini` and we're just not going to overwrite it, then you probably still have a valid IG overall -- so that's more of a warning condition.

Somewhat (but not totally) analogous: If your config has a `menu` property, but you also have an `input/includes/menu.xml` file, we won't overwrite the file and we log a warning.  So it's analogous in that we're warning when you have duplicate sources; it's not analogous in that `menu` property is not deprecated.

Re: template and ig.ini: I could probably be convinced either way.  What do you think?

ngfreiter: I could also go either way. I'm happy as long as package-list and this match in when it is an error and a warning. My logic was that we are only letting `template` hang around for people who have legacy IGs. If you are making an IG with the new structure, and you specify `template` or `history`, I liked having an error, because if we weren't supporting legacy modes we wouldn't let you have those config properties at all. So I didn't want to let people get away with something they shouldn't in the new structure, just because we need to allow it for the legacy structure.

cmoesel: OK.  Let's make it an error then!

src/utils/Processing.ts
cmoesel: I'm actually noticing there's a fair bit of redundant code here too.  I'm going to consolidate that as well.

cmoesel:
OK, @ngfreiter -- your comments have been addressed!

ngfreiter (approved):
Thanks for the changes, this looks good to me! I've tried to make it break but I have not succeeded.

cmoesel:
I've rebased this so it's up to date w/ the target branch.  One thing I noticed and wanted opinions on... If we detect a legacy style project, we output a warning w/ steps to convert it -- which includes things like renaming config.yaml and removing "history".  But... we also emit individual warnings for these as well.  See screenshot below:

image

Part of me thinks this is redundant and the individual messages should be removed.  Another part of me thinks they just drive home the point.  And yet another part of me has basically stopped caring at this point. ;-)

ngfreiter:
> Part of me thinks this is redundant and the individual messages should be removed. Another part of me thinks they just drive home the point. And yet another part of me has basically stopped caring at this point. ;-)

I think I feel myself similarly split. For me, as long as the messages are consistent, which they are as far as I can tell, I don't mind driving the point home.

ngfreiter (dismissed):
Making my approval fresh.

jafeltra (commented):
This looks really great! I tested (I think) all the scenarios you had in the project structure slides you created a couple weeks ago, and they all look pretty great. I just noticed a couple differences that I want to double check:

- I noticed that if I have an `ignoreWarnings.txt` file in my tank and I output to a different folder, that file gets copied over. Is that expected? I'm thinking we don't want that functionality.
- When I'm in the legacy flat project structure, if I have any `pagecontent` pages with the prefix numbers, such as `1_clients.md`, the files no longer get renamed. I think we wanted this case to work the same as it does on master, so I think we do want the files renamed. I think this one might have been my fault, and I think we might need to be using a different flag when we decide what name to give the page.

cmoesel:
Good finds, @jafeltra!  Thanks!  I'll take a crack at these!

cmoesel:
@jafeltra -- your comments have been addressed in 6106320 and 6155dc4, including a couple of other related things I discovered.  Thanks again for the review!

jafeltra (approved):
This looks good to me now! The two cases I had noticed are addressed in the latest commits and I don't see anything else!

ngfreiter:
> Yay! @ngfreiter -- my last two commits staled your approval. Were you planning to re-review or should we go ahead and merge?

Merge away!

ngfreiter:
> > Yay! @ngfreiter -- my last two commits staled your approval. Were you planning to re-review or should we go ahead and merge?
> 
> Merge away!

Why thank you I think I will

PR #615, opened 2020-09-29
sushi --init integration

ngfreiter:
This updates `sushi --init` to work with the new IG Publisher integration, and also makes it so that we don't make a SUSHI-GENERATED-FILES.md file when running `sushi` in the new IG Publisher mode.

I don't think this conflicts with #613 but it does depend on it for full ability to test. To test this, you need to:

1. `npm pack` the `fsh-generated` branch.
2. Globally install `sushi` from that packed version.
3. Run `ts-node src/app.ts --init` (or however you run in dev sushi) on this branch.
4. Run `sushi .` on the output of `--init` using the globally installed version. 
5. Install the IG Publisher to `input-cache` (`_updatePublisher` is still broken for me, so I do this manually).
6. Get `sushi-dev-template` from this zip file on #613: https://github.com/FHIR/sushi/files/5278146/NewIgStructure.zip.
7. Change the `ig.ini` file generated by `--init` to have a template of `#sushi-dev-template`.
8. Run `_genonce`.

This should allow you to test out how `sushi --init` works in the environment that will exist once we release the new integration, and once the IG Publisher releases the template on their end.


cmoesel:
So it appears there is a problem w/ the downloaded `_updatePublisher.sh` script:
image

Sure enough, the shebang is _not_ a shebang:
```
#/bin/bash
```
Fixing it to `#!/bin/bash` resolved the problem.  I'll do a PR against the sample-ig we download the script from.

cmoesel (changes requested):
When I run `_genonce` I get a few errors.  I think we should aim for as clean a first run as possible.  One error is because `ignoreWanrings.txt` is not there:
image

I know we decided to to generate this in sushi anymore, but for `--init` only, I think it might still make sense to just stub out the file.  I think the contents could be:
```
== Suppressed Messages ==

# Add warning and/or information messages here after you've confirmed that they aren't really a problem
# (And include comments like this justifying why)
# See https://github.com/FHIR/sample-ig/blob/master/input/ignoreWarnings.txt for examples
```

It also complains about the `package-list.json` not being there.  Do you think that `--init` should just lay down a stubbed out version of that?  Something like the [package-list.json in Sample IG](https://github.com/FHIR/sample-ig/blob/master/package-list.json), but w/ appropriate values (like package-id and canonical) replaced?  I'm not sure about this one -- so it's an honest question, looking for your opinion.

cmoesel:
Actually, I think the errors I saw and assumed were due to package-list.json are the same as reported here: https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/The.20.3Cscript.3E.20'http.3A.2F.2Fhl7.2Eorg.2Ffhir.2Fhistory-cm.2Ejs'.20is.20llegal/near/211752918

So perhaps we do not need to do anything about package-list.json.

ngfreiter:
Ok @cmoesel https://github.com/FHIR/sushi/pull/615/commits/27c3e480ef515c36e40b7169209023ec2888caa0 adds an `ignoreWarnings.txt` with the message you recommended.

cmoesel (approved):
Changes look good!

We should not merge this, however, until the `fsh-generated` is merged -- since it depends on it.

jafeltra (approved):
I approve also!

cmoesel:
Ruh-roh.  Looks like there are conflicts @ngfreiter.  Sorry!

ngfreiter:
Conflicts should be fixed in https://github.com/FHIR/sushi/pull/615/commits/504a5a9a544530958d570ab6c2934da6e2d06a34!

PR #618, opened 2020-09-30
Clean fsh-generated

ngfreiter:
This makes it so that the `fsh-generated` folder will be wiped clean before SUSHI runs. SUSHI will only do this if the output folder contains a folder called `fsh-generated`, so it is not a universal cleaning of output folders. I don't think we can really safely universally clean the output folders, because in the legacy modes input to the publisher that is not managed by SUSHI mixes with output that is managed by SUSHI.

I feel safe about wiping out a folder recursively, because we only ever wipe out a folder named `fsh-generated`, so I think a very unfortunate chain of events would have to occur for something very bad to happen.


jafeltra (approved):
Looks good to me!

cmoesel (commented):
I was curious what would happen if SUSHI failed to delete the files.  I forced this scenario by "locking" a file in `fsh-generated` so it cannot be deleted.  When I did that, I got this error:
> error SUSHI encountered the following unexpected error: EPERM: operation not permitted, unlink '../../fsh/SimpleNewIG/fsh-generated/includes/menu.xml'

After that, SUSHI immediately quit.  We should decide how we want to handle this:
* leave as it is
* catch the error and provide a more user-friendly message (maybe still including the underlying OS error)
* catch the error, log it as a warning, but still try to muddle through somehow.

I think I'm in favor of the 2nd bullet, but welcome your opinion.

ngfreiter:
@cmoesel, if I understand correctly, the difference between your second and third bullet is that if we log as an error, the publisher won't run, but if we log as a warning it will?

Assuming that is the case, I feel like I support doing a warning, because maybe for some reason someone really wants a file in `fsh-generated`, and that's why they locked it in there, and maybe we would be doing a disservice by stopping their build process for that. I don't think I've ever unintentionally made a file impossible to delete, so I'd lean towards trusting the authors intent.

cmoesel:
@ngfreiter -- I feel like there have been times when I couldn't delete files for whatever reason (and it wasn't intentional).  That was mainly back in my Windows day; so maybe that is a thing of the past now.  My only worry is that it might result in inconsistent results (like an old file never being overwritten w/ the newer contents).  But I guess if we warn about it then the user should know to be on guard for that.  So... I can go either way.

ngfreiter:
You make a good point @cmoesel, Windows can still be a bully about deleting files that are open in other programs. I wasn't able to make it happen in some testing, but it has happened to me in other contexts, and I'd rather be safe than sorry. So I've elevated this to the level of an error in 5ae1f41.

jafeltra (approved):
The update makes sense to me. When I locked a file that SUSHI tries to overwrite, I did get a second error saying

```
error SUSHI encountered the following unexpected error: EPERM: operation not permitted, open '../publisher-fsh-tanks/NewIgStructure/fsh-generated/resources/CodeSystem-MyCodeSystem.json'
```
but I feel like that is fine since it is explained more clearly earlier in the logs and that log shows the full error, so it clearly is related (the same).

cmoesel (changes requested):
The test message needs to be updated to indicate it logs an error now.  I tried GitHub's "suggestion" mode.  Let me know how it works out.

ngfreiter:
> The test message needs to be updated to indicate it logs an error now. I tried GitHub's "suggestion" mode. Let me know how it works out.

Ohh fancy, seems to have worked well, see https://github.com/FHIR/sushi/pull/618/commits/7efd362a3416ee0984efb1d59cc0c7a70d96464f

cmoesel (approved):
Almost there!

PR #619, opened 2020-09-30
Update README for new publisher integration

ngfreiter:
This updates the README to be accurate in reference to the new publisher integration. This accuracy is mainly achieved by just giving FSHSchool the responsibility for being accurate, that way we can keep all our docs in one spot.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 ![GitHub Workflow Status](https://img.shields.io/github/workflow/status/fhir/sushi/Lint%20and%20Test)
 ![npm](https://img.shields.io/npm/v/fsh-sushi)
 
-# SUSHI (WORK IN PROGRESS)
+# SUSHI
 
 SUSHI (aka "SUSHI Unshortens Short Hand Inputs") is a reference implementation command-line interpreter/compiler for FHIR Shorthand (FSH).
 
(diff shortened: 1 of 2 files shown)


jafeltra (approved):
I like this. I think it'll be easier to keep our documentation up to date in one place, so this change seems reasonable to me.

jafeltra:
Actually, this makes me wonder. Should we point to FSH School even earlier in the README? If you're using SUSHI and you download it from NPM, the first thing you read for documentation is the README. Maybe it makes sense for those users to be pointed to https://fshschool.org/docs/sushi/ at the end of the `Installation for SUSHI Users` section for instructions for building a FSH tank and running SUSHI. Or maybe even we could have a `Initializing` section with brief `sushi --init` instructions and then the link to https://fshschool.org/docs/sushi/.

Either way, these are just broad thoughts on the README and I don't think they're necessarily part of this PR.

cmoesel:
As Julia suggests, I don't think it would hurt to put a link to the SUSHI FSH School doc earlier.

Also, I think it's time to remove the big `(IN PROGRESS)` from the headline of the README.

cmoesel:
Also, it's not exactly the README, but can you go ahead and change [app.ts line 26](https://github.com/FHIR/sushi/blob/master/src/app.ts#L26) so it reads:
```ts
const FSH_VERSION = '1.0.0';
```

ngfreiter:
Ok, moved a link up, changed the `FSH_VERSION`, and removed the `IN PROGRESS`, should be ready for re-review.

jafeltra (approved):
Looks good!

cmoesel (approved):

Issue #621, opened 2020-10-01
Nested Ad-Hoc extensions fail


ngfreiter:
See this thread for full details: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Questionnaire.20item.20extensions.

In the following example:
```
Instance: MyQ
InstanceOf: Questionnaire
* extension[us-core-race].id = "string"
* item[0].extension[us-core-race].id = "string"
* item[0].item[0].extension[us-core-race].id = "string"
```
SUSHI is correctly able to add an ad-hoc extension to the instance in the first line, but fails to do so in the second and third line. It appears to be an issue with how SUSHI handles nested paths in this case.

PR #622, opened 2020-10-01
Fix nested ad hoc extensions

ngfreiter:
Fixes https://github.com/FHIR/sushi/issues/621.

See the issue for an example of the bug. To test this, you can run that example on `master` and observe the errors, and note that no `extensions` are created within `item` objects on the instance. Then, run this example on this branch, and notice that those errors go away (still expect to see other errors and warnings related to elements not having correct cardinality, please excuse those), and the extensions are added to the `item` objects.


cmoesel (approved):
That's a wacky bug.  What was that `'.'` doing there in the first place?

Good find.  Fix works.  The world rejoices.

cmoesel:
This was a super simple fix.  I'm merging w/ one review.

PR #628, opened 2020-10-12
Reset errors and warnings between fshToFhir calls

### LINKED ISSUE #625: fshToFHIR is not clearing the errors array between invocations

ngfreiter:
Fixes #625 by clearing out the `errors` and `warnings` arrays at the start of the `fshToFhir` function, so that the reported errors and warnings only relate to that call.

To test, I've been using the following script:
```
const { sushiClient } = require('fsh-sushi');

const FSH0 = `
Alias: $USCoreLab = http://hl7.org/fhir/us/core/StructureDefinition/us-core-observation-lab
Alias: $USCorePatient = http://hl7.org/fhir/us/core/StructureDefinition/us-core-patient
Alias: $LOINC = http://loinc.org

Profile: MyObservationProfile
Parent: $USCoreLab
* value[x] only Quantity

Instance: MyPatient
InstanceOf: $USCorePatient
* identifier[0].system = "http://foo.org/id"
* identifier[0].value = "1"
* name[0].given[0] = "Homer"
* name.family = "Simpson"
* gender = #male


Instance: MyObservationInstance
InstanceOf: MyObservationProfile
* code = $LOINC#1234-5
* status = #final
* valueQuantity = 55 'mm'
`;

const FSH1 = `
Instance: MyOtherObservationInstance
InstanceOf: MyObservationProfile
* code = $LOINC#9876-5
* status = #final
* valueQuantity = 42 cm'
`

const OPTIONS = {
  canonical: 'http://foo.org/bar',
  version: '0.0.1',
  dependencies: [{ packageId: 'hl7.fhir.us.core', version: '3.1.1' }],
  logLevel: 'debug'
}
void async function() {
  const results = await sushiClient.fshToFhir([FSH0, FSH1], OPTIONS);
  console.log(results.errors)

  const results2 = await sushiClient.fshToFhir([FSH0, FSH1], OPTIONS);
  console.log(results2.errors)


  const results3 = await sushiClient.fshToFhir([FSH0, FSH1], OPTIONS);
  console.log(results3.errors)
}();
```
which has an error on the last line, the `42 cm'` is missing a leading `'`. You can run this script with node, just make sure it can find your locally built version of `fsh-sushi`. If you run this script with a dependency on the built version of `master`, you will see 2 errors, then 4, then 6, as the errors accumulate over the three calls. If you run this script with a dependency on the built version of this branch, you will see 2 errors, then 2, then 2, as the errors no longer accumulate. I also removed the manual resetting that was happening in the tests, as that is no longer necessary.


cmoesel:
So this reveals a potentially deeper issue with our implementation: we're using a global error collector which we _reset_ at the start of each call to an _asynchronous_ function.  In other words, there is opportunity for this function to be invoked multiple times in an asynchronous manner such that invocations happen in parallel.  Which means that these situations are theoretically possible:
* invocation _a_ collects an error, but invocation _b_ resets it before invocation _a_ has returned the error to the caller
* or invocation _b_ starts before _a_ has returned, invocation _a_ collects errors _after_ _b_ has reset them, so invocation _b_ still ends up returning invocation _a_'s errors
* or invocation _b_ starts before _a_ has returned, invocation _b_ collects an error, and then invocation _a_ returns it

So... I think we should be ensuring that loggers are unique to a given function call, but dang, that might be hard given the way we currently do loggers.

ngfreiter:
> So... I think we should be ensuring that loggers are unique to a given function call, but dang, that might be hard given the way we currently do loggers.

Do you have any ideas for this? Because I'm struggling to see how we could do this (without major changes).

cmoesel:
I was hoping maybe I didn't remember the logger quite right and there would be an easy solution.  Here are a few thoughts:
* We could experiment w/ ways of forcing an invocation to wait until a previous invocation has finished (maybe using a global variable, a while loop, and `process.tick()`?)
* We could have a synchronous version of FHIR definition loading and use that instead (?)
* We could refactor so that we're managing the errors separately in each synchronous section of the function execution -- in other words, not trusting any errors between async boundaries.
* We could refactor the whole logging mechanism to have more context of what "run" it is in -- but this might be fairly large I think, and I wonder if it would force us to pass a context around more (probably).
* We could just document it: wait until this function has returned before calling it again, otherwise reported errors may be inconsistent.

ngfreiter:
Right, I like bullet number 2, but the issue with that is we did have that working synchronously, and then had to change to make it async because the synchronous library we were using to make requests was deprecated, so we were forced to go async. I expect I would hit the same issues now as I did then.

Bullet number 4 feels like the way to do this "right", but I think it will be a big effort, and I think we would have to consider if it is worth that effort. Let me know what you think.

cmoesel:
Yeah, I suppose that bullet 2 would be bad news for running this SUSHI function browser-side (like in FSHOnline).  Before we jump into bullet 4, however, I'd like to have a better understanding of the impact.  Perhaps you could look into it and determine what actual changes would be needed?

One thing I found that seems promising is [async hooks](https://nodejs.org/api/async_hooks.html#async_hooks_async_hooks).  My understanding is that they can help you to identify and maintain context between asynchronous calls.  [This guide](https://medium.com/the-node-js-collection/async-hooks-in-node-js-illustrated-b7ce1344111f) and [this guide](https://stackabuse.com/using-async-hooks-for-request-context-handling-in-node-js/) provide some decent examples.  I'm not 100% sure if it's the right fit, but it's worth looking into.  It's a node feature though, so I don't know what that would mean for our webpacked module.

cmoesel:
BTW -- I noted async hooks because it doesn't require you to pass context around through function calls, which is my main concern about bullet 4.

ngfreiter:
So I've been looking into async hooks...and honestly I'm a little confused/in over my head. But if I am understanding, then the hooks track when async resources are used, but I don't seem to see that they would allow us to tell which invocation of a function is causing an error to be logged.

I would be fine with going with bullet number 5. I don't think there is an easy way to make this work, and I'd want to be certain it is really something people are feeling they want before we go deeper on it.

cmoesel:
So... I was thinking that `ErrorsAndWarnings` as well as `LoggerStats` would no longer be single global instances -- but rather, callers would get them via function calls (e.g., `FSHLogger.getErrorsAndWarnings()`).  Then that function might be defined something like this:
```ts
import { executionAsyncId } from 'async_hooks';

// ...

const errorsAndWarningsMap: Map<number, ErrorsAndWarnings> = new Map();

export function getErrorsAndWarnings() {
  const asyncId = executionAsyncId();
  if (!errorsAndWarningsMap.has(asyncId)) {
    errorsAndWarningsMap.set(asyncId, new ErrorsAndWarnings());
  }
  return errorsAndWarningsMap.get(asyncId);
}
```

I could totally be misunderstanding how this works, but I *think* `executionAsyncId` should return the same number throughout an async "thread" -- but different numbers between async "threads" ("thread" isn't the technically correct term, but it's how I think about it).  That said, I'm not sure *when*/*how* Node decides what should get different ids.  Clearly, a for-loop that calls an async function in each iteration _should_ result in different ids for each invocation.  I think?

I don't know.  Obviously I'm not sure about this stuff either.  This is just what I was thinking of when I mentioned async hooks.

cmoesel:
I think we can just document it for now, but thought I would put that reply there in case we want to come back to it or in case it triggers any other ideas in your head.

ngfreiter:
I'm not entirely sure. I was trying to call `executionAsyncId` several times within the single call of `fshToFhir`, and it was different each time I called it within the function, which if I'm understanding you correctly, we would need it to be the same within that function call.

cmoesel:
```
¯\_(ツ)_/¯
```

cmoesel:
OK... can you document the limitation using JSDoc on the `fshToFhir` function?  As an API function, it should probably be documented w/ JSDoc anyway.

ngfreiter:
Ok @cmoesel, documented in https://github.com/FHIR/sushi/pull/628/commits/203c8d5187d88239bdfb8f645c50108c2a51e389

cmoesel (approved):
Thanks, @ngfreiter!  Looks good, but needs to be synced with master!

jafeltra (approved):
This looks fantastic!

PR #631, opened 2020-10-13
Do not copy IG files in new configuration

### LINKED ISSUE #626: SUSHI appending comments to files in input
Reported by Eric Haas:

Eric Haas3:13 PM
Sushi is prepending text into markdown files in my input/includes directory each time I run it.
```
<!-- example-list-generator.md {% comment %}
*****************************************************************************************
*                            WARNING: DO NOT EDIT THIS FILE                             *
*                                                                                       *
* This file is generated by SUSHI. Any edits you make to this file will be overwritten. *
*                                                                                       *
* To change the contents of this file, edit the original source file at:                *
* Davinci-CDEX/input/includes/example-list-generator.md                                 *
*****************************************************************************************
{% endcomment %} -->
<!-- example-list-generator.md {% comment %}
*****************************************************************************************
*                            WARNING: DO NOT EDIT THIS FILE                             *
*                                                                                       *
* This file is generated by SUSHI. Any edits you make to this file will be overwritten. *
*                                                                                       *
* To change the contents of this file, edit the original source file at:                *
* Davinci-CDEX/input/includes/example-list-generator.md                                 *
*****************************************************************************************
{% endcomment %} -->
<!-- example-list-generator.md {% comment %}
*****************************************************************************************
*                            WARNING: DO NOT EDIT THIS FILE                             *
*                                                                                       *
* This file is generated by SUSHI. Any edits you make to this file will be overwritten. *
*                                                                                       *
* To change the contents of this file, edit the original source file at:                *
* Davinci-CDEX/input/includes/example-list-generator.md                                 *
*****************************************************************************************
{% endcomment %} -->
```

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Announcing.20SUSHI.201.2E0.2E0.20Beta.201/near/212471917

jafeltra:
This PR fixes #626, along with an additional couple cases.

There were a few cases were we would incorrectly copy and write IG publisher related files in the new tank configuration. One was reported in #626 where the `input/includes` content was being copied with the warning text in the new configuration. We still want to include this warning for anyone on the legacy project configuration, but we should not be copying over anything in the new configuration. The other case that Mark discovered was files in `input/pagecontent` were being copied over with the warning text. Similarly, we still want this warning in the legacy project configuration, but not in the new configuration.

I added the same check we use in other functions to decide if we should be copying the files. I did another scan through of the functions and the only other place that we were copying files was in the `input/images` folder. Technically I don't think keeping this copying in this folder made a difference since we didn't add any warnings, but I still decided to add the check there as well.


ngfreiter (commented):
This is looking good, and I am not getting those extra warnings anymore. I am wondering though if we need all of those files in the fixtures? I guess it doesn't really matter, but wouldn't just one file be sufficient for testing? Or at most one xml and one md file?

jafeltra:
I had been using the `example-config.yaml` file in the test, so I created a fixture file for each of the ones specified in the `pages` in that config file. But, I could probably simply it to just set my own `pages` field on the `minimalConfig` and remove some of the fixtures. Do you think that would be simpler and better?

ngfreiter:
> I had been using the `example-config.yaml` file in the test, so I created a fixture file for each of the ones specified in the `pages` in that config file. But, I could probably simply it to just set my own `pages` field on the `minimalConfig` and remove some of the fixtures. Do you think that would be simpler and better?

I don't much mind either way. I just didn't understand where the inspiration for those files was coming from. So I'll leave it to you now that it makes sense.

jafeltra:
Okay I took another look and I think I'm going to change it slightly for clarity. I had originally wanted to use `example-config.yaml` just because it was already an existing config file that had configured pages. But since we use `minimalConfig` as a base in other places and update it as necessary, I think it's actually probably clearer when reading the test if I take that approach. So I've updated the test to add configured pages to the minimal config and I took out a few of the pages as well. Let me know what you think.

cmoesel (approved):
This looks good to me!  Thanks, @jafeltra!

ngfreiter (approved):
Looks good!

PR #633, opened 2020-10-14
Avoid Duplicate Resources in fsh-generated

### LINKED ISSUE #623: SUSHI beta does not check for duplicate resources between input/* and input/fsh
If there is a FSH definition that has the same canonical URL as a JSON/XML resource already in one of the other `input/*`, SUSHI should issue a warning and ignore the FSH definition. Currently it still generates the FSH definition, causing duplicates in the source to the IG Publisher.

Reported here: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/SUSHI.201.2E0.2E0.20Beta.20discussion/near/212163679

ngfreiter:
Fixes #623 by avoiding writing a duplicate resource to the `fsh-generated` folder. Before any resource is written, the list of predefined resources (those coming from `input`) is checked to see if any of those resources match the `url` of the resource about to be written. If they do match, a warning is logged, and the FSH resource is not generated. This means that the pre-defined resource is preferred, and this matches how things work in previous versions of SUSHI, pre-defined resources have been preferred to FSH defined resources.


cmoesel (changes requested):
While this suppresses the FSH definition from being written to a file, it does not suppress the FSH definition from being processed.  Unfortunately, this means that if it is declared as a `Parent`, `InstanceOf`, or used in a `Reference`, SUSHI will use its _FSH_ definition rather than the external definition.  This causes inconsistencies in the outputs.

I've attached a simple project that demonstrates this:
* `input/profiles/StructureDefinition-MyPatient.json`
  * `MyPatient` constrains `name` to `0..0`
* `input/fsh/patient.fsg`
  * `MyPatient` adds a 3rd extension and constrains `name` to `1..*`
  * `MyPatientChild` declares `MyPatient` as its parent and constrains `name` to `1..1`

If the external definition were preferred, SUSHI would log an error for `MyPatientChild` since you cannot narrow `name` from `0..0` to `1..1`.  Instead, SUSHI produces `MyPatientChild` with a diff showing only `name.max` changed to `1` and also inheriting the 3rd extension added by the parent FSH definition (which _should_ be ignored).

[ExampleIG.zip](https://github.com/FHIR/sushi/files/5380222/ExampleIG.zip)

ngfreiter:
I can do that, because I know we've taken that sort of approach in the past (although I think this would've worked the same in the past, use FSH during processing, but export the external definition). But I still don't totally understand why we allow it to be a non-error if someone has an external file that overrides their FSH.

cmoesel:
Yeah -- agreed that in the past we probably suffered from this issue too.  And I've thought of it before today -- but I guess never was in a position to really try it out and see what happened.

I think the idea is that if someone wants to temporarily override a FSH definition, they need only drop in the file -- and then to remove the override, they need only remove the file.  That said -- how about we meet in the middle and call it a warning when we encounter this?

The question is -- should we:
- change the fishers to look at external defs first (before the tank), or
- leave fishers as-is and remove the FSH definition from the tank

If the latter, do we need to do a presweep of the tank?

ngfreiter:
> * change the fishers to look at external defs first (before the tank), or
> * leave fishers as-is and remove the FSH definition from the tank
> 
> If the latter, do we need to do a presweep of the tank?

I think there is something to be said for looking locally before looking at defs. For example, maybe I happen to collide the name of some profile I am making in FSH with some existing FHIR extension. I think this would be allowed, since the Profile and the Extension would be different types, and therefore not conflict.

So I am thinking option 2. But option 2 is tricky, since we figure out URLs as we go. Although maybe we could run through all artifacts and set all their URLs before exporting. And then remove the ones that conflict. Is that what you mean by presweep?

cmoesel:
I was thinking something like that for the "pre-sweep", but I just thought of something that might work...  What if, before we start exporting the tank, we go through all the predefined resources and fish the tank for their URLs?  If the fisher returns something, then we know it's a duplicate and we issue a warning and remove it from the tank?  I think that would work because the tank fisher already knows how to handle the URL (whether it be constructed via the canonical or overridden via ^url).

ngfreiter:
@cmoesel, a5e93c3 makes it so that predefined resources are found first when `fish`ing with the `MasterFisher`. It should work as you'd expect on the example you posted now. I did it this way instead of by pre-sweeping the definitions because:
1. It's actually not so easy to remove a definition
2. I don't like the idea of removing actual FSH definitions but only logging a warning. It just feels wrong in my heart. The approach I took here will at least still find a FSH definition if it is referenced by name or something, if that name differs from the external name.

I know I just submitted the code that this suggestion would make obsolete, but I'd like to give one more plea to consider that this could be an error. I know some user(s) want the ability to override resources they've already defined, but I think that is very easily solved by a comment. On the other hand, we are now only giving a warning when external resources cause the actual FSH definitions to be ignored. So someone could be autobuilding with the IG Publisher, seeing that SUSHI succeeds (since no errors) and being very confused as to why the changes they are making to their FSH code are not being reflected in their IG, while not realizing that it is because they have an extra sneaky definition in `input/profiles` which is blocking their FSH.

cmoesel:
Jose feels pretty passionately about this topic.  Let me see what he thinks on it...  He's the primary one pushing for this.

cmoesel:
OK, good call @ngfreiter -- looks like **error** is the way to go: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Same.20profile.20defined.20in.20FSH.20*and*.20JSON.3A.20Error.20or.20Warning.3F

ngfreiter:
Converted the warning to an error in https://github.com/FHIR/sushi/pull/633/commits/bdfac33fb5144a1b765a3957e280377846317bc2

src/fhirdefs/FHIRDefinitions.ts
```diff
@@ -103,6 +107,43 @@ export class FHIRDefinitions implements Fishable {
...
+    this.predefinedResources = new Map();
+  }
+
+  fishForPredefinedResource(item: string, ...types: Type[]): any | undefined {
+    const resource = this.fishForFHIR(item, ...types);
+    if (
+      resource &&
+      this.allPredefinedResources().find(
+        predefResource =>
+          predefResource.id === resource.id && predefResource.resourceType === resource.resourceType
```
cmoesel, line 120: Why test on `id` and `resourceType` rather than `url`?  I think `url` is really the only _globally unique_ identifier.  It's technically possible (and I think valid) to have two different resources w/ the same `id` and `resourceType` (as long as they have different URLs).

src/fhirdefs/FHIRDefinitions.ts
```diff
@@ -103,6 +107,43 @@ export class FHIRDefinitions implements Fishable {
...
+
+  fishForPredefinedResourceMetadata(item: string, ...types: Type[]): any | undefined {
+    const resource = this.fishForFHIR(item, ...types);
+    if (
+      resource &&
+      this.allPredefinedResources().find(
+        predefResource =>
+          predefResource.id === resource.id && predefResource.resourceType === resource.resourceType
+      )
+    ) {
```
cmoesel, line 135: If you used `this.fishForPredefinedResource` on line 128, then couldn't this just be `if (resource) {`?

src/fhirdefs/FHIRDefinitions.ts
cmoesel: Also, you're using `url` as the matcher in `Processing.ts` so it seems like it would be more consistent.

cmoesel (changes requested):
The code worked well in my tests, but I would recommend a few small changes.

src/fhirdefs/FHIRDefinitions.ts
ngfreiter: Right, I agree they should match, so I need to fix that either way, but I went with `resourceType` and `id` because I don't think matching on `url` works for Instances right? Because an Instance may not have a `url`. But we do require that an Instance have a `resourceType` and `id`, since those are used to generate the file names.

cmoesel: Oh...  I wasn't thinking of instances...  Yeah, I guess you're right.  Can we match on all three then?  In the case of instances, it should still match since `url` would be `undefined` for both.  And in my edge case it would properly detect things that have matching `id` and `resourceType` but different `url`.

ngfreiter:
Alright @cmoesel, I used all three to check for duplicates, and made the other improvements you suggested, thanks!

cmoesel (approved):
Alright.  Works well, looks good.  I accept!

jafeltra (approved):
This makes sense to me! I tested various cases with the example IG Chris attached, and everything worked as expected, and the code makes sense to me!

PR #634, opened 2020-10-14
Don't use non-conformance type descriptions when building IG resource

### LINKED ISSUE #629: IG resource generating "description" incorrectly
Reported by Eric Haas on Zulip ([link](https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/bug.20in.20ig.20.20resource.20generation.20for.20v.201.2E0.2E0)):

The IG Resource has a incorrect structure when add in a Goal example resource to the definitions list.
briefly is creating description object instead of string using the content from Goal.description which is type codeable concept. This may happen with other non conformance resource with a `description` element too, so I would check that out...

Here is the example file: https://www.hl7.org/fhir/us/core/Goal-goal-1.html

here is the offending bit in the generated IG resource:

```yaml
    - reference:
        reference: Goal/goal-1
      description:  #####should be string not object####
        text: Patient is targeting a pulse oximetry of 92% and a weight of 195 lbs
      name: goal-1
      exampleCanonical: 'http://hl7.org/fhir/us/core/StructureDefinition/us-core-goal'
```

should be:

```yaml
    - reference:
        reference: Goal/goal-1
      description:  Patient is targeting a pulse oximetry of 92% and a weight of 195 lbs
      name: goal-1
      exampleCanonical: 'http://hl7.org/fhir/us/core/StructureDefinition/us-core-goal'
```

cmoesel:
Fixes #629

### DIFF
--- a/src/ig/IGExporter.ts
+++ b/src/ig/IGExporter.ts
@@ -32,6 +32,25 @@ import { logger, Type } from '../utils';
 import { FHIRDefinitions } from '../fhirdefs';
 import { Configuration } from '../fshtypes';
 
+// List of Conformance and Terminology resources from http://hl7.org/fhir/R4/resourcelist.html
+const CONFORMANCE_AND_TERMINOLOGY_RESOURCES = new Set([
+  'CapabilityStatement',
+  'StructureDefinition',
+  'ImplementationGuide',
+  'SearchParameter',
+  'MessageDefinition',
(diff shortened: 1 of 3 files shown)


mint-thompson (approved):

jafeltra (approved):
This makes sense to me and works as expected using both the new and legacy tank configurations.

PR #636, opened 2020-10-15
Set input directory default

### LINKED ISSUE #627: Can no longer just run `sushi`
We used to be able to do `sushi` to run SUSHI on the current working directory, but now we have to do `sushi .`. I don't think there is a good reason why we can't just assume `.` when there is no input directory, so we should bring back `sushi` with no dot as an allowed command.

jafeltra:
This PR fixes #627. There was a bug where you could no longer run `sushi` (without the input directory specified). This was because we were trying to use the input directory in a `path.resolve` before we had set a default if it was not specified. I moved that functionality to it's own function and called it earlier to ensure we always have a default value when we try to use it.


ngfreiter (approved):
Looks good, worked well, easy to review, a lovely PR.

cmoesel (approved):
I'd buy that for a dollar!

PR #637, opened 2020-10-15
Fix extraction of config info from IG resource in new publisher config

### LINKED ISSUE #635: SUSHI does not correctly extract config from IG XML/JSON
If you run SUSHI without a `sushi-config.yaml`, it *should* extract information from the IG XML or JSON if it can find it.  This is not working correctly.

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/How.20I.20use.20Sushi.201.2E0.2E0.20to.20manage.20an.20IG/near/213361487

cmoesel:
Also logs the extracted config so the user can see what values are being used.

Fixes #635


src/import/loadConfigurationFromIgResource.ts
```diff
@@ -11,19 +11,30 @@ import { logger } from '../utils';
...
+export function loadConfigurationFromIgResource(
+  input: string,
+  isLegacyIgPubContext: boolean
+): Configuration | null {
   // First, look in the ig.ini file for a path to the IG resource
   let igPath: string;
-  const igIniPath = path.join(input, '..', 'ig.ini');
+  const igIniPath = isLegacyIgPubContext
+    ? path.join(input, '..', 'ig.ini')
+    : path.join(input, 'ig.ini');
```
ngfreiter, line 22: It seems like you have this pattern a few times:
```
isLegacyIgPubContext ? path.join(input, '..', 'ig.ini') : path.join(input, 'ig.ini');
```
Would it work to just do something like that once to find the root directory (the one containing ig.ini), and use that everywhere?

ngfreiter (commented):
Looks good, one minor suggestion.

src/import/loadConfigurationFromIgResource.ts
cmoesel: Yeah, OK.  I thought about that at first but thought (at that point) there were only two places where this happened.  Then in testing I found I missed one and did not revisit the decision.  I'll refactor.

cmoesel:
@ngfreiter -- In the midst of this, I realized it would be better to actually just refactor the legacy flag out of the `loadConfigurationFromIgResource` function -- because the only thing that function cares about is the IG root -- which is the same in either configuration.  This also allowed me to delete some of the "legacy" tests I had previously introduced.

ngfreiter (approved):
Oh cool! Looking good 😎

src/import/loadConfigurationFromIgResource.ts
```diff
@@ -56,14 +58,16 @@ export function loadConfigurationFromIgResource(input: string): Configuration |
...
-      FSHOnly: true,
-      dependencies: igResource.dependsOn?.filter(dep => dep.packageId && dep.version),
+      version: igResource.version,
       fhirVersion: igResource.fhirVersion ?? [],
-      version: igResource.version
+      dependencies: igResource.dependsOn?.filter(dep => dep.packageId && dep.version),
+      FSHOnly: true
     };
+    logger.info(`Extracting FSHOnly configuration from ${igPath}:`);
+    Object.entries(config).forEach(e => logger.info(`  ${e[0]}: ${JSON.stringify(e[1])}`));
```
jafeltra, line 69: Do you think it would be worthwhile to format the JSON string that's logged here? With how it works now, I got a pretty dense block of text logged: 

image

I did try using `JSON.stringify(e[1], null, 2)`, and I'm not entirely sure it's better because the "info" boxes, but figured I'd ask your opinion.

image

jafeltra (commented):
These changes make sense to me! I just had one small styling type question.

src/import/loadConfigurationFromIgResource.ts
cmoesel: I also tried `JSON.stringify` when I was developing this and decided I liked it better _without_ it.  It just takes up a lot of space with it and looks kind of funky with the labels.  In reality, these log messages are mainly helpful for debugging, so they don't need to be super readable every time.  And also, usually people will have at most one IG dependency (at least now that seems to be the case) -- so the block won't be as large as yours above.

All that said, I'm willing to go the other way if you all think it's better to format it.

cmoesel: Although, if we're going to use JSON stringify with formatting, then maybe it would be better not to use a separate log message per property but to put the whole thing in a single log message (e.g., "Extracting FSHOnly configuration..." with the formatted JSON appended to the same message).

jafeltra: Yeah I agree that the "info" labels make it look kind of funky, and given your reasoning, I'm good leaving it as it is!

jafeltra (approved):
These changes make sense to me! I just had one small styling type question.

PR #639, opened 2020-10-15
Dont process spreadsheets

### LINKED ISSUE #638: SUSHI tries to read spreadsheets as XML resources
The `input/resources`, `input/profiles`, etc, folders can include spreadsheet XML files (e.g., `resources-spreadsheet.xml`).  SUSHI tries to read these as FHIR resources and then emits an error:
```
error Loading resources-spreadsheet.xml failed with the following error:
Unknown resource type: Workbook
```

cmoesel:
If an XML file name ends with "-spreadsheet.xml" or it contains the mso-application declaration for excel, ignore it.

Fixes #638


jafeltra (approved):
This makes sense to me!

ngfreiter (commented):
Could you add a test for the case where the file is ignored because it comes from Excel? Also might it make sense to use smaller files? It's not really a big deal, but those are some pretty beefy xml files for tests where we aren't actually testing the content. On the other hand, I can't begrudge you those line change counts.

cmoesel:
> Could you add a test for the case where the file is ignored because it comes from Excel?

@ngfreiter -- I don't understand what you're asking...  I have tests that show it is ignored (no error logged).  Previous behavior was to log an error but keep going.  I don't know what else / how else you want me to test that it was "ignored".

ngfreiter:
> > Could you add a test for the case where the file is ignored because it comes from Excel?
> 
> @ngfreiter -- I don't understand what you're asking... I have tests that show it is ignored (no error logged). Previous behavior was to log an error but keep going. I don't know what else / how else you want me to test that it was "ignored".

Oh I see, I misunderstood. I thought there were two distinct types of spreadsheets we were looking for, one where the spreadsheet ends with `-spreadsheet.xml`, and one where the spreadsheet matches this regex:
```
/<\?mso-application progid="Excel\.Sheet"\?>/
```
but now I see that the spreadsheets that end with `-spreadsheet.xml` are just a subset, and all of the spreadsheets we want to ignore should match that regex. I'm guessing you're just ignoring the ones that end with `-spreadsheet.xml` so that you can avoid having to read in big spreadsheets if we can? If that's the case, then this makes sense.

cmoesel:
Yes -- looking for `-spreadsheet.xml` is an efficiency to avoid having to load in and grep a potentially very large XML file.  So in the tests, `resources-spreadsheet.xml` tests the first case (although I guess it's not possible to be sure since it also covers the second case) and `sneaky-spread-like-bread-sheet.xml` definitely covers only the second case.

As for the files themselves, sure, I can replace them with smaller files.  Although I do have to say, based on past experience, checking in large text files is very good for the GitHub street cred.  I'm still riding high from checking in all the FHIR R4 definitions early in this project (before you implemented the package loader).  You may have more commits, but I have oh-so-many-more lines!
image

ngfreiter:
That's it, no more squash and merge for me. I've gotta keep my commit numbers up.

cmoesel:
Ha.  Yes, @ngfreiter, now that you're 50% like the rest of us, at least it levels the playing field.  I should have guessed you'd try to cheat to keep your advantage.

BTW -- I made those files smaller for you, since I'm past my cheating days now.

jafeltra (approved):
My review got dismissed, but I approve even more strongly now that these XML files are so teeny.

ngfreiter (approved):
I don't know what you're talking about...I've just heard that new best practice is one commit per line change 🤷‍♂️ . It's the cutting edge...trust me...don't google it or anything.

Thanks for the changes, looks great!

PR #644, opened 2020-10-20
Default to resourceType and id when reporting duplicate resources

### LINKED ISSUE #642: SUSHI reports duplicate "undefined" resource
A user has reported receiving the following message:
> **error** Ignoring FSH definition for undefined since it duplicates existing pre-defined resource. To use the FSH definition, remove the conflicting file from "input". If you do want the FSH definition to be ignored, please comment the definition out to remove this error.

ngfreiter:
Fixes #642.

We were logging a message that used `resource.url`, but it was possible that `resource` was an instance, and did not have a `url` on it. We can guarantee that every resource, instances included, will have a `resourceType` and `id`, so if the `url` is undefined, we now use the `resourceType` and `id` when logging the error message.

### DIFF
--- a/test/utils/Processing.test.ts
+++ b/test/utils/Processing.test.ts
@@ -468,7 +468,7 @@ describe('Processing', () => {
               tempIGPubRoot,
               'fsh-generated',
               'resources',
-              'Patient-my-duplicate-profile.json'
+              'StructureDefinition-my-duplicate-profile.json'
             )
           )
         ).toBeFalsy();
--- a/test/utils/Processing.test.ts
+++ b/test/utils/Processing.test.ts
@@ -403,6 +403,14 @@ describe('Processing', () => {
       myFSHDefinedProfile.id = 'my-duplicate-profile';
       myFSHDefinedProfile.url = 'http://example.com/StructureDefinition/my-duplicate-profile';
 
+      const myPredefinedInstance = new InstanceDefinition();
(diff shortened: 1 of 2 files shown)


test/utils/Processing.test.ts
```diff
@@ -459,13 +468,30 @@ describe('Processing', () => {
               tempIGPubRoot,
               'fsh-generated',
               'resources',
-              'StructureDefinition-my-duplicate-profile.json'
+              'Patient-my-duplicate-profile.json'
```
cmoesel, line 471: Shouldn't this _still_ be `StructureDefinition-my-duplicate-profile.json`?  You didn't make any changes to profile-related code or test fixtures -- and there is nothing in the predefined resources or FSH resources that _would_ make a file called `Patient-my-duplicate-profile.json` anyway -- so you're not really testing anything as it currently is.  If it's `StructureDefinition-my-duplicate-profile.json` though, then you're at least testing that the FSH profile was not created.

cmoesel (changes requested):
I think maybe you just had a find-and-replace mistake in the tests?

test/utils/Processing.test.ts
ngfreiter: Fixed in 5917ef7

cmoesel (approved):
Looks good to me!

jafeltra (approved):
Makes lots of sense to me!

PR #645, opened 2020-10-20
Quantity fix

### LINKED ISSUE #608: Overwriting value on a Quantity
See this issue for full details: https://jira.hl7.org/browse/FHIR-28552.

When an author sets `value` on a Quantity, and then sets the Codeable properties on that Quantity as in the following:
```
* dosage.doseAndRate.doseQuantity.value = 250.0
* dosage.doseAndRate.doseQuantity = UCUM#mg "mg"
```
SUSHI overwrites the setting of `value` when setting the Codeable properties in the second line.

julianxcarter:
Fixes #608 in which setting a quantity value, then setting the rest of the quantity fields in a succeeding line would overwrite the previously set value.

### DIFF
--- a/src/fhirtypes/common.ts
+++ b/src/fhirtypes/common.ts
@@ -199,16 +199,13 @@ export function setPropertyOnInstance(
           current = current[key];
         } else {
           // Check if the instance already has the element being defined
-          if (current[pathPart.base] != null && typeof current[pathPart.base] === 'object') {
+          if (current[key] != null && typeof current[key] === 'object') {
             // Check if the instance already has a quantity element
             // Quantity elements are the only FHIR types with both a code and value
-            if (
-              current[pathPart.base].hasOwnProperty('value') &&
-              assignedValue.hasOwnProperty('code')
-            ) {
+            if (current[key].hasOwnProperty('value') && assignedValue.hasOwnProperty('code')) {
               // Ensure that the existing value is not being overwritten
               assignedValue = {
-                value: current[pathPart.base].value,
+                value: current[key].value,
                 ...assignedValue
               };
             }
--- a/src/fhirtypes/common.ts
+++ b/src/fhirtypes/common.ts
@@ -123,7 +123,7 @@ export function setImpliedPropertiesOnInstance(
 
 export function setPropertyOnInstance(
   instance: StructureDefinition | ElementDefinition | InstanceDefinition | ValueSet | CodeSystem,
-  pathParts: PathPart[],
+  pathParts: PathPart[], 
   assignedValue: any
 ): void {
   if (assignedValue != null) {
@@ -198,6 +198,19 @@ export function setPropertyOnInstance(
           if (current[key] == null) current[key] = {};
           current = current[key];
         } else {
+          // Check if the instance  already has the element being defined
+          if (current.hasOwnProperty([pathPart.base])) {
+            // Check if the instance already has a quantity element
+            // Quantity elements are the only FHIR types with both a code and value
+            if (current[pathPart.base].hasOwnProperty('value')) {
+              if (assignedValue.hasOwnProperty('code')) {
+                // Ensure that the existing value is not being overwritten
+                assignedValue = { 
+                  value: current[pathPart.base].value, ...assignedValue, 
+                };
+              }
+            }
+          }
           current[key] = assignedValue;
         }
       }
--- a/src/fhirtypes/common.ts
+++ b/src/fhirtypes/common.ts
@@ -198,18 +198,19 @@ export function setPropertyOnInstance(
           if (current[key] == null) current[key] = {};
           current = current[key];
         } else {
-          // Check if the instance  already has the element being defined
-          if (current.hasOwnProperty([pathPart.base])) {
+          // Check if the instance already has the element being defined
+          if (current[pathPart.base] != null && typeof current[pathPart.base] === 'object') {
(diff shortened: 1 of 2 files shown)


src/fhirtypes/common.ts
```diff
@@ -198,6 +198,20 @@ export function setPropertyOnInstance(
           if (current[key] == null) current[key] = {};
           current = current[key];
         } else {
+          // Check if the instance  already has the element being defined
+          if (current.hasOwnProperty([pathPart.base])) {
```
mint-thompson, line 202: I'm surprised this works, but it clearly does. The [MDN page for hasOwnProperty](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty) doesn't mention taking an array parameter, though, so I would suggest passing the argument as a string rather than an array.

src/fhirtypes/common.ts
```diff
@@ -198,6 +198,20 @@ export function setPropertyOnInstance(
...
+            // Check if the instance already has a quantity element
+            // Quantity elements are the only FHIR types with both a code and value
+            if (current[pathPart.base].hasOwnProperty('value')) {
+              if (assignedValue.hasOwnProperty('code')) {
```
mint-thompson, line 206: The conditions in the three if statements can safely be combined into a single if statement, since the only code that modifies data is nested inside all three.

src/fhirtypes/common.ts
cmoesel, line 202: I'm confused by the array syntax here.  Why `current.hasOwnProperty([pathPart.base])` and not `current.hasOwnProperty(pathPart.base)`?  Is it just a left-over of a previous version where maybe you previously had `current[pathpart.base]`?  Or is there some other new fancy syntax thing I don't know about?  After all, it seems to work -- so I'm inclined to think maybe it was intentional!

It's not a big deal, but we could be a little more specific here and check that it's actually an object:
```ts
if (current[pathPart.base] != null && typeof current[pathPart.base] === 'object') {
```

mint-thompson (changes requested):
Great work on your first PR! I have a few minor stylistic questions, but also, the test case you've written doesn't actually make it to your third if statement. So, I think you'll need to modify the test case to satisfy all of your conditions.

cmoesel (commented):
First PR!  Awesome!  I had one question about a syntax I'm not familiar with...

src/fhirtypes/common.ts
cmoesel: Oh.  I see @mint-thompson already noted this (in addition to some things I missed).  Thanks, @mint-thompson!

julianxcarter: Looking at this now, I'm actually surprised that it works as well haha. The brackets were a holdover from a previous implementation that I somehow glanced over a few times before I pushed this PR. I'll remove this in an edit, and then combine the three conditionals as well.

julianxcarter: It's not intentional! It was a holdover from a previous implementation where I was checking if `current[pathPart.base]` was truthy rather than using the `hasOwnProperty()` function. Looking at it now, I'm actually somewhat surprised it still worked! I also think that checking that it's actually an object would be useful, I'll add that app in another commit.

mint-thompson (dismissed):
Hooray!

src/fhirtypes/common.ts
```diff
@@ -198,6 +198,21 @@ export function setPropertyOnInstance(
           if (current[key] == null) current[key] = {};
           current = current[key];
         } else {
+          // Check if the instance already has the element being defined
+          if (current[pathPart.base] != null && typeof current[pathPart.base] === 'object') {
```
cmoesel, line 202: I just noticed this now, but I'm wondering... why do you use `current[pathPart.base]` here (and after) rather than `current[key]`?  I only noticed this because on line 211, we essentially assign `assignedValue.value` to `current[pathPart.base].value`, then on line 216, we assign `current[key] = assignedValue`.  So if you follow along, we say we're preserving the `value` but we're basically doing `current[key].value = current[pathPart.base].value`; which really should only work if `key === pathPart.base`.

And sure enough, if you look up at lines 136-137 we have:
```ts
const key =
  pathPart.primitive && i < pathParts.length - 1 ? `_${pathPart.base}` : pathPart.base;
```
And we only get to your new code if `i === pathParts.length` so I think we do in fact know that `key === pathPart.base` 100% of the time we are in this block (lines 201-215).  So... to me, it would be more clear if we just used `current[key]` instead of `current[pathPart.base]` since the code directly above and below this uses `current[key]` already.

cmoesel (commented):
This looks really good.  I just have one minor question/suggestion I noticed as I reviewed it a second time.

src/fhirtypes/common.ts
julianxcarter: You're right! Using `pathPart.base` isn't necessary when the `key` value is set to be the same thing on lines 136-137. There's no particular reason for my not using the `key` variable here, I had just blanked on the fact that they should always be equal in this block of code. I'll update the conditionals on lines 202 and below to use the `key` variable.

cmoesel (approved):
Looks great!  Thanks, @julianxcarter!

cmoesel:
Given that the changes are trivial since @mint-thompson last approved, I am going to go ahead and merge this now.

PR #646, opened 2020-10-20
Change configuration format to support separate group names and IDs

masnick:
Previously, custom groups defined in `sushi-config.yaml` used the same string for both the `id` and `name` properties for a given group.

For example, given the following YAML:

    groups:
      GroupA:
        name: Group A
        description: The Alpha Group
        resources:
        - StructureDefinition/animal-patient
        - StructureDefinition/arm-procedure

Previously, both the `name` and `id` properties in the generated IG JSON would have been `Group A`. This caused errors like the following when building the IG with the FHIR publisher:

> ImplementationGuide.definition.resource[0].groupingId
> id value 'Group A' is not valid

This commit changes SUSHI to set `id` to `GroupA` and `name` to `Group A`. This avoids the error when building the IG.

Groups where `name` is undefined are still supported:

    groups:
      GroupB:
        description: The Beta Group
        resources:
        ...

In this case, both `id` and `name` will be `GroupB`.

This is related to #590.

### DIFF
--- a/src/fshtypes/Configuration.ts
+++ b/src/fshtypes/Configuration.ts
@@ -138,6 +138,7 @@ export type Configuration = {
 };
 
 export type ConfigurationGroup = {
+  id: string;
   name: string;
   description?: string;
   resources: string[];
--- a/src/ig/IGExporter.ts
+++ b/src/ig/IGExporter.ts
@@ -1080,21 +1080,28 @@ export class IGExporter {
   /**
    * Adds or updates a group in the implementation guide.
    *
-   * @param {string} name - name of the group, used as unique identifier
+   * @param {string} id - unique identifier for group
+   * @param {string} name - name of the group
    * @param {string} description - optional description of the group
    */
-  private addGroup(name: string, description?: string): void {
+  private addGroup(id: string, name?: string, description?: string): void {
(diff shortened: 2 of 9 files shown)


cmoesel:
Thanks, @masnick.  I'll have to review this more closely, but I understand the approach.  In some ways, it makes sense since `ImplementationGuide.definition.resource.groupingId` refers to the `id` (not the `name`) -- so it makes sense that `id` would be the key for our groups.

That said, `id` is a more restrictive data type than `string` (which is data type of `name`).  If people have existing configs and currently use a group name that has invalid characters for `id` (like spaces or underscores) -- then this would present a problem (although, I guess that is already the case since we were assigning `name` to `id` before).  Alright...  We might want to consider logging an `error` when people use an invalid character for an `id`.

So I'll review this more closely later.  Thanks for the PR!

masnick:
Thanks!

I don't feel strongly about this, but another option might be allowing people to manually override `id` like:

	someKeyHere:
		id: some_other_id
 		name: Some name

cmoesel (approved):
Looks good to me.  Thank you, @masnick!  I'll update the documentation for FSHSchool to reflect this change!

PR #651, opened 2020-10-26
Added support for coded values on Age types

### LINKED ISSUE #617: Handle "Age" when assigned FSHCode values
The line:
```
* condition.onsetAge = UCUM#a "years"
```
fails to set these values on `onsetAge` with the error:
```
error Cannot bind value set to Age; must be coded (code, Coding, CodeableConcept, Quantity), or the data types (string, uri).
  File: C:\Users\mkramer\Documents\GitHub\fhir-mCODE-ig\fsh\ExtendedExample1.fsh
  Line: 522
```
This is because `Age` is a specialization of `Quantity`, and we do not correctly handle that when a Coding type value is being assigned. We do have code to handle these specializations, since we correctly handle `Age` for the `64.0 'a'` syntax that can be used on a `Quantity`.

julianxcarter:
fixes #617

### DIFF
--- a/src/fhirtypes/ElementDefinition.ts
+++ b/src/fhirtypes/ElementDefinition.ts
@@ -1537,7 +1537,7 @@ export class ElementDefinition {
       );
     } else if (type === 'Coding') {
       this.assignFHIRValue(code.toString(), code.toFHIRCoding(), exactly, 'Coding');
-    } else if (type === 'Quantity') {
+    } else if (type === 'Quantity' || type === 'Age') {
       // Since code only maps to part of Quantity, we want to ensure that if there are other (non-code) parts
       // already assigned, we take them on too -- as we don't want to overwrite them with blanks.
       const existing = this.fixedQuantity ?? this.patternQuantity ?? this.assignedByAnyParent();
@@ -1548,7 +1548,7 @@ export class ElementDefinition {
       if (existing?.comparator != null) {
         quantity.comparator = existing.comparator;
       }
-      this.assignFHIRValue(code.toString(), quantity, exactly, 'Quantity');
+      this.assignFHIRValue(code.toString(), quantity, exactly, type);
     } else {
       throw new CodedTypeNotFoundError([type]);
     }
--- a/src/fhirtypes/ElementDefinition.ts
+++ b/src/fhirtypes/ElementDefinition.ts
@@ -1171,7 +1171,7 @@ export class ElementDefinition {
         this.assignString(value as string, exactly);
         break;
       case 'Code':
-        this.assignFshCode(value as FshCode, exactly);
+        this.assignFshCode(value as FshCode, exactly, fisher);
         break;
       case 'Quantity':
         value = value as FshQuantity;
@@ -1520,12 +1520,22 @@ export class ElementDefinition {
    * @throws {ValueAlreadyAssignedError} when the code is already assigned to a different code
    * @throws {InvalidUriError} when the system being assigned is not a valid uri
    */
-  private assignFshCode(code: FshCode, exactly = false): void {
+  private assignFshCode(code: FshCode, exactly = false, fisher?: Fishable): void {
     if (code.system && !isUri(code.system.split('|')[0])) {
       throw new InvalidUriError(code.system);
     }
 
     const type = this.type[0].code;
+
+    // Check if the code is a specialization of the Quantity type
+    let quantityType = type === 'Quantity';
+    if (!quantityType) {
+      const actualTypeSD = fisher?.fishForFHIR(type, Type.Type);
+      if (actualTypeSD?.baseDefinition === 'http://hl7.org/fhir/StructureDefinition/Quantity') {
+        quantityType = true;
+      }
+    }
+
     if (type === 'code' || type === 'string' || type === 'uri') {
       this.assignFHIRValue(code.toString(), code.code, exactly, type);
     } else if (type === 'CodeableConcept') {
@@ -1537,7 +1547,7 @@ export class ElementDefinition {
       );
     } else if (type === 'Coding') {
       this.assignFHIRValue(code.toString(), code.toFHIRCoding(), exactly, 'Coding');
-    } else if (type === 'Quantity' || type === 'Age') {
+    } else if (quantityType) {
       // Since code only maps to part of Quantity, we want to ensure that if there are other (non-code) parts
       // already assigned, we take them on too -- as we don't want to overwrite them with blanks.
       const existing = this.fixedQuantity ?? this.patternQuantity ?? this.assignedByAnyParent();
--- a/src/fhirtypes/ElementDefinition.ts
+++ b/src/fhirtypes/ElementDefinition.ts
@@ -1171,7 +1171,7 @@ export class ElementDefinition {
         this.assignString(value as string, exactly);
         break;
       case 'Code':
-        this.assignFshCode(value as FshCode, exactly, fisher);
+        this.assignFshCode(value as FshCode, fisher, exactly);
         break;
       case 'Quantity':
         value = value as FshQuantity;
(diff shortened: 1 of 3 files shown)


src/fhirtypes/ElementDefinition.ts
```diff
@@ -1537,7 +1537,7 @@ export class ElementDefinition {
       );
     } else if (type === 'Coding') {
       this.assignFHIRValue(code.toString(), code.toFHIRCoding(), exactly, 'Coding');
-    } else if (type === 'Quantity') {
+    } else if (type === 'Quantity' || type === 'Age') {
       // Since code only maps to part of Quantity, we want to ensure that if there are other (non-code) parts
```
julianxcarter, line 1541: Just as a note: This is what I was talking about when I said there was existing code to fix issue #608. Lines 1550-1547 appear to be designed to do exactly what's being done in PR #645, only it occurs as rules are being validated rather than later on when the rules are being applied to the instance definition. The `fixedQuantity` property of this class is never set before its referenced on line 1543 though, which is why values were still being overwritten.

ngfreiter: Good find! It looks like this bit of code prevents overwriting from happening when we are assigning a value on a Profile or an Extension, which sets the `fixedQuantity` or `patternQuantity` field. The bug you fixed only seemed to occur when we were assigning a value on an Instance. So it looks to me like this little bit of code is working for the purpose it was meant to, Profiles and Extensions, but we overlooked making things work similarly on Instances.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1537,7 +1537,7 @@ export class ElementDefinition {
       );
     } else if (type === 'Coding') {
       this.assignFHIRValue(code.toString(), code.toFHIRCoding(), exactly, 'Coding');
-    } else if (type === 'Quantity') {
+    } else if (type === 'Quantity' || type === 'Age') {
```
ngfreiter, line 1540: This does work to address the task as it was described, but there may be a way to do this more generally. `Age` is not the only specialization of `Quantity`, there are several others. If you check out the graph of General-Purpose Data types on [this](https://www.hl7.org/fhir/datatypes.html#2.24.0) page, the types that are shown to inherit from `Quantity` with blue arrows will all act similarly to `Age`.

For any value that is a `Quantity` or specialization of `Quantity`, we can assign values using two different syntaxes:
```
* some.quantity.element = 64.0 'a'
```
to set the `value`, `code` and implicitly the `system` (which automatically gets set to http://unitsofmeasure.org, since this is a very common `system` to use), or
```
* some.quantity.element = UCUM#a "years"
```
which sets the `system`, `code`, and `units`. It's a little weird, but each syntax is useful for its own purposes.

Anyway, if you test out the first syntax on an element of type `Age` or `Duration` (another specialization of `Quantity`), you will notice it already works. This is because of code on lines 1180-1191 in ElementDefinition.ts. You will notice that this code does not hard code in the name of the specializations, but rather dynamically checks if a type inherits from `Quantity`, and if it does, we let the `64.0 'a'` syntax be used on that element.

I think a similar dynamic check would be useful on this PR. Instead of checking if the type is `Age`, we can look up the StructureDefinition for the given type, and if that StructureDefinition inherits from `Quantity`, then we can assign the FHIR value to the `Quantity` or specialization of `Quantity` element. The code on line 1180-1191 should be a useful example of how to do this, and also let me know if you want more of an explanation.

ngfreiter (commented):
Good work, and nice test cases. I think we can make this PR more general, see my more detailed comment below, and let me know what you think or if you have any questions.

src/fhirtypes/ElementDefinition.ts
julianxcarter: I was wondering if this was a problem specific to the `Age` type or all of the other `Quantity` specializations, it'd definitely be useful to make this fix more general. I'll implement this next, thanks for the heads up and for the example!

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1520,12 +1520,22 @@ export class ElementDefinition {
    * @throws {ValueAlreadyAssignedError} when the code is already assigned to a different code
    * @throws {InvalidUriError} when the system being assigned is not a valid uri
    */
-  private assignFshCode(code: FshCode, exactly = false): void {
+  private assignFshCode(code: FshCode, exactly = false, fisher?: Fishable): void {
```
ngfreiter, line 1523: Since `exactly` is an optional argument, could we keep it last? If we put `fisher` last, we will then always have to specify `exactly`.

ngfreiter (commented):
Nice work! Just had one small additional comment. Besides that, it worked for me with both `Age` and `Duration`. Thanks for making it more general.

ngfreiter (approved):
Looks good!

src/fhirtypes/ElementDefinition.ts
cmoesel: Hmmm...  So I think I might disagree with @ngfreiter (!!).  While I agree that optional arguments should be last, I'd also note that `fisher` is an optional argument, so...

I actually think `fisher` should go last in order to maintain consistency between function signatures.  For example, the primary (public) `assignValue` signature takes these same three arguments and has `fisher` _last_:
```ts
assignValue(value: AssignmentValueType, exactly = false, fisher?: Fishable): void
```

The other (private) assign functions take the `exactly` argument after the `value`:
```ts
assignFHIRValue(fshValue: string, fhirValue: any, exactly: boolean, type: string)
assignNumber(value: number, exactly = false): void
assignString(value: string, exactly = false): void
```

So, to maintain consistency with `assignValue` (and to an extent, the others), I'd suggest reverting back to @julianxcarter's original signature:
```ts
private assignFshCode(code: FshCode, exactly = false, fisher?: Fishable): void
```

ngfreiter: Ohhh that's my bad. @cmoesel is definitely right here. Sorry for making you change that @julianxcarter!

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1520,12 +1520,22 @@ export class ElementDefinition {
...
     const type = this.type[0].code;
+
+    // Check if the code is a specialization of the Quantity type
+    let quantityType = type === 'Quantity';
+    if (!quantityType) {
+      const actualTypeSD = fisher?.fishForFHIR(type, Type.Type);
+      if (actualTypeSD?.baseDefinition === 'http://hl7.org/fhir/StructureDefinition/Quantity') {
+        quantityType = true;
+      }
+    }
```
cmoesel, line 1537: Since this is executed _before_ the `if`/`else if` checks below, it's being executed even in the case that the type is `code`, `string`, `uri`, `Coding`, or `CodeableConcept`.  Since _fishing_ can be an expensive operation, and since the quantity specialization use case is the minority use case (meaning this would happen needlessly in most cases), I think it would be best to delay this until we know it's not one of those other cases.

One potential approach would be to create a tiny helper function based on your code above:
```ts
private isQuantityType(type: string, fisher?: Fishable): boolean {
    if (type === 'Quantity') {
      return true;
    }
    const actualTypeSD = fisher?.fishForFHIR(type, Type.Type);
    return actualTypeSD?.baseDefinition === 'http://hl7.org/fhir/StructureDefinition/Quantity';
  }
}
```
Then in the `else if` (line 1550), you would do:
```ts
} else if (isQuantityType(type, fisher)) {
```
this would delay execution of the moderately expensive code until we _know_ we need to check for a quantity type.

cmoesel (changes requested):
The code look good and works well.  I separately commented (outside of this review) that I think you should revert the signature back to what it was (w/ `exactly` second and `fisher` last).  I've also included a suggestion that would improve performance.

src/fhirtypes/ElementDefinition.ts
julianxcarter: I really like this solution, and it'd be much more efficient as you were saying. I'll implement this approach and switch the parameters back in a new commit.

ngfreiter (approved):
Looks even better!

cmoesel (approved):
Yep. A++.  Would buy again.

PR #653, opened 2020-10-27
Must Support does not automatically transfer from sliced elements to slices

mint-thompson:
Completes task [CIMPL-570](https://standardhealthrecord.atlassian.net/browse/CIMPL-570).

When setting `mustSupport = true` on an element with slices, do not automatically set `mustSupport = true` on the slices.
When `mustSupport = true` is set on a slice, include this in the differential even if the sliced element also has `mustSupport = true`.

### DIFF
--- a/src/fhirtypes/ElementDefinition.ts
+++ b/src/fhirtypes/ElementDefinition.ts
@@ -1872,10 +1872,11 @@ export class ElementDefinition {
       throw new DuplicateSliceError(this.structDef.name, this.id, name);
     }
 
-    // On a new slice, delete slice.min and slice.max and then reset them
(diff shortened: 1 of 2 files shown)


cmoesel (approved):
This works splendidly and I'm relieved at how small the changes were to accomplish this!

FYI @markkramerus -- I ran this against mCODE and it _did_ result in changes.  Previous `MS` flags were being suppressed in slices in the `CancerGeneticVariant`, `CancerGenomicsReport`, and `MCODEPatientBundle` profiles -- but based on the new guidance from FHIR, these should be surfaced (and _are_ when you use this PR branch).

@mint-thompson -- can you sync this w/ `master` so it is ready to be merged after its second approval?

ngfreiter (approved):
Looks good!

PR #656, opened 2020-10-29
Group properties when exporting Instances

### LINKED ISSUE #632: Instance generation should place primitive extensions adjacent to their primitive values
See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/A.20question.20of.20style.2E.2E.2E

mint-thompson:
Completes task [CIMPL-571](https://standardhealthrecord.atlassian.net/browse/CIMPL-571). Fixes #632 .

A primitive property on an Instance may have extensions or other nested attributes on it. This is represented on the Instance with a property whose key begins with an underscore. For easier reading, when a property and its underscored version both exist on an Instance, those properties should be grouped together on the serialized Instance.

Object keys are technically unordered, but this implementation seems to exhibit the desired behavior. A potentially more robust solution would be building a `Map`, which has ordered keys, in `toJSON`. But, this would then remove the "round-trip" nature of `toJSON` and `fromJSON`, and it would also require a custom serializer, because `JSON.stringify` [will only return a Map's enumerable properties](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#Description), not its key-value pairs.

### DIFF
--- a/src/fhirtypes/InstanceDefinition.ts
+++ b/src/fhirtypes/InstanceDefinition.ts
@@ -1,6 +1,6 @@
-import { cloneDeep, remove, pull } from 'lodash';
+import { difference } from 'lodash';
 import { Meta } from './specialTypes';
-import { HasId } from './common';
+import { HasId, orderedCloneDeep } from './common';
 import { applyMixins } from '../utils';
 import { InstanceUsage } from '../fshtypes';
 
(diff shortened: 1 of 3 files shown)


src/fhirtypes/InstanceDefinition.ts
```diff
@@ -26,8 +26,51 @@ export class InstanceDefinition {
...
+    }
+    if (this._id) {
+      clone._id = cloneDeep(this._id);
+    }
+    if (this.meta) {
+      clone.meta = cloneDeep(this.meta);
+    }
+    if (this._meta) {
+      clone._meta = cloneDeep(this._meta);
+    }
```
cmoesel, line 54: I feel like this can be made more succinct in one of two ways.

1. You could just loop through a set of these keys:
```ts
const clone: any = {};
['resourceType', '_resourceType', 'id', '_id', 'meta', '_meta'].forEach(key => {
  if (this[key]) {
    clone[key] = cloneDeep(this[key]);
  }
});
```

2. Couldn't you just add them to `orderedKeys` first and then let the logic in lines 70-72 take care of it?
```ts
const orderedKeys = ['resourceType', '_resourceType', 'id', '_id', 'meta', '_meta']
  .filter(key => this[key] != null);
// _instanceMeta is only needed for lookup and IG config - not a FHIR property 
const keys = difference(Object.keys(this), [...orderedKeys, '_instanceMeta']);
const underscoreKeys = remove(keys, key => key.startsWith('_'));
```
I think something like the above 5 lines could replace the current lines 29 - 58.

src/fhirtypes/InstanceDefinition.ts
```diff
@@ -26,8 +26,51 @@ export class InstanceDefinition {
...
+        orderedKeys.push(`_${key}`);
+        pull(underscoreKeys, `_${key}`);
+      }
+    });
+    underscoreKeys.forEach(key => {
+      orderedKeys.push(key);
+    });
+    orderedKeys.forEach(key => {
+      clone[key] = cloneDeep(this[key]);
+    });
```
cmoesel, line 72: Another thought -- instead of doing a clone operation on every single key, might it be better just to clone `this` at the start and use that to populate a `result` object?  E.g.,
```ts
const clone = cloneDeep(this);
const result: any = {}; // this assumes you're using my 2nd proposed optimization above
orderedKeys.forEach(key => {
  result[key] = clone[key];
});
```

cmoesel (changes requested):
I've proposed a few optimizations for your consideration, but the reason I'm requesting changes is that I think this only handles the top level keys, but it really needs to recursively handle all values, e.g. this should work too:
```json
{
  "resourceType": "Patient",
  "name": [{
    "given": ["Michael"],
    "family": "Bolton",
    "_family": {
      "extension": {
        "url": "http://question.org/that/michael/bolton?",
        "valueBoolean": false
      }
    }
  }]
}

cmoesel (approved):
This looks good to me.  The only question I have is why `orderedCloneDeep` should be in `common.ts` when the only place it is used is `InstanceDefinition.ts`?  Do you anticipate using it in other places?

mint-thompson:
> This looks good to me. The only question I have is why `orderedCloneDeep` should be in `common.ts` when the only place it is used is `InstanceDefinition.ts`? Do you anticipate using it in other places?

The function is written to be adaptable to other resource types, but since it's only being called in `InstanceDefinition.ts` for now, it's fine for it to live in there until something else actually needs it.

cmoesel (approved):
Lovin' it.

ngfreiter (approved):
This looks good to me!

PR #657, opened 2020-10-30
Merge existing Mappings

jafeltra:
This PR adds support for adding rules to existing `Mapping`s on StructureDefinitions ([CIMPL-583](https://standardhealthrecord.atlassian.net/browse/CIMPL-583)).

On master, if you were to make a `Mapping` that has the same `id` as one that is already on the Parent of the Profile you are adding the mapping to, SUSHI will add a duplicate Mapping at the top level of the StructureDefinition and will also add any rules. This PR makes it so that you can add a `Mapping` that has the same `id` as one already on the SD and add rules. As long as any specified metadata matches what is on the existing `Mapping`, no metadata will be added to the SD but any rules on the `Mapping` will be added to the SD. If there is any metadata specified on the `Mapping` that does not match what is on the Mapping on the SD, an error will be logged and no rules will be added.

I had also intended to include fixes for the two things @cmoesel pointed out [here](https://github.com/FHIR/GoFSH/pull/28#pullrequestreview-519770578), but after looking at the original US Core SDs, these changes seem to be caused by the caret value rules GoFSH was creating since we were accessing indexes in arrays in the differential. I believe all the mapping information should be the same as the original US Core SDs.

### DIFF
--- a/src/export/MappingExporter.ts
+++ b/src/export/MappingExporter.ts
@@ -2,7 +2,7 @@ import { FSHTank } from '../import';
 import { Package } from '.';
 import { logger, Type, MasterFisher } from '../utils';
 import { Mapping } from '../fshtypes';
-import { StructureDefinition, idRegex } from '../fhirtypes';
+import { StructureDefinition, StructureDefinitionMapping, idRegex } from '../fhirtypes';
 import { InvalidFHIRIdError } from '../errors';
 import { MappingRule } from '../fshtypes/rules';
 import { applyInsertRules } from '../fhirtypes/common';
@@ -72,7 +72,38 @@ export class MappingExporter {
       Type.Extension
     ) as StructureDefinition;
     if (sourceStructDef) {
-      this.setMetadata(sourceStructDef, fshDefinition);
+      const parent = this.fisher.fishForFHIR(
+        sourceStructDef.baseDefinition,
+        Type.Resource,
+        Type.Type,
+        Type.Profile,
+        Type.Extension
+      );
+      const matchingParentMapping = parent?.mapping.find(
+        (m: StructureDefinitionMapping) => m.identity === fshDefinition.id
+      );
+      if (matchingParentMapping != null) {
+        const isMatchingTitle = fshDefinition.title
+          ? fshDefinition.title === matchingParentMapping.name
+          : true;
+        const isMatchingTarget = fshDefinition.target
+          ? fshDefinition.target === matchingParentMapping.uri
+          : true;
+        const isMatchingDescription = fshDefinition.description
+          ? fshDefinition.description === matchingParentMapping.comment
+          : true;
+        if (!isMatchingTitle || !isMatchingTarget || !isMatchingDescription) {
+          // If the mapping identity matches one on the parent, all other metadata must also match in order to merge MappingRules
+          logger.error(
+            `Unable to add Mapping ${fshDefinition.name} because it conflicts with one already on the parent of ${fshDefinition.source}.`,
+            fshDefinition.sourceInfo
+          );
+          return;
+        }
+      } else {
+        // Only add metadata if it does not already exist on the parent
+        this.setMetadata(sourceStructDef, fshDefinition);
+      }
       this.setMappingRules(sourceStructDef, fshDefinition);
     } else {
       logger.error(`Unable to find source "${fshDefinition.source}".`, fshDefinition.sourceInfo);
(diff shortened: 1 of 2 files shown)


src/export/MappingExporter.ts
```diff
@@ -72,7 +72,38 @@ export class MappingExporter {
...
+          ? fshDefinition.title === matchingParentMapping.name
+          : true;
+        const isMatchingTarget = fshDefinition.target
+          ? fshDefinition.target === matchingParentMapping.uri
+          : true;
+        const isMatchingDescription = fshDefinition.description
+          ? fshDefinition.description === matchingParentMapping.comment
+          : true;
+        if (!isMatchingTitle || !isMatchingTarget || !isMatchingDescription) {
+          // If the mapping identity matches one on the parent, all other metadata must also match in order to merge MappingRules
```
cmoesel, line 96: I know this is _exactly_ what I suggested when we discussed this, but now I'm having second thoughts...  If the `mapping.comment` is the only thing that changes, is that so bad?  Is it possible that in a profile you might want to provide additional comments about the mapping (in general) that maybe didn't apply to the mapping in the parent?  I think maybe that should be allowed.

So I guess I'm suggesting we still check for matching `name` (`title`) and `uri` (`target`) but perhaps we allow for changes in `comment` (`description`).  What do you think?

cmoesel: BTW -- at first I thought maybe even requiring `name` (`title`) to match was too strict, but when I looked at the definition in FHIR, it says "A name for the specification that is being mapped to" and says you _must_ have either a `uri` _or_ `name` -- so I think that probably _shouldn't_ change (but `comment` still seems changeable to me).

cmoesel (commented):
I'm changing my mind on one thing we discussed.  Let me know what you think!

src/export/MappingExporter.ts
jafeltra: Second thoughts are always allowed! I think I'm fine with this change! But out of curiosity, do you have a link to that documentation you're referencing? I've been struggling to find the right FHIR documentation for Mappings.

cmoesel: This is what I was looking at: http://hl7.org/fhir/R4/structuredefinition-definitions.html#StructureDefinition.mapping

jafeltra: Ah okay. I've been there so I at least feel better about that! And I agree with your second comment. It feels like if you should have a `uri` or `name`, we should match both of those. But a `comment` seems reasonable for a profile to change. So I'll make that update!

test/export/MappingExporter.test.ts
```diff
@@ -194,6 +194,116 @@ describe('MappingExporter', () => {
...
+      const status = observation.elements.find(e => e.id === 'Observation.status');
+      const originalStatusMappingLength = status.mapping.length;
+
+      exporter.export();
+      expect(loggerSpy.getAllMessages('error')).toHaveLength(1);
+      expect(observation.mapping.length).toBe(originalMappingLength); // No metadata added
+      expect(status.mapping.length).toBe(originalStatusMappingLength); // No rule added to status element
+    });
+
+    it('should not log an error and not add metadata but add rules for a Mapping that is inherited from the parent and has additional metadata not on the parent', () => {
```
cmoesel, line 282: I feel like this test description is a little confusing.  This test is about testing that changing the description doesn't log an error, right?  And it should probably also verify that the new description _does_ get set in the FHIR mapping?

jafeltra:
@cmoesel I've updated this to reflect the changes we talked about this morning. Hopefully the test description is a bit more clear, but all the new test descriptions are long, so I can try to reword them if they're still confusing.

27184ab makes it so that any inherited mapping that has additional or changed comments will add that field to the `mapping` on the structure definition. It also updates the test to check that the value on the `mapping` has been updated. I also added that check for the equality of the `mapping` for the first test to ensure nothing was changed when we don't update the metadata because there was no changed metadata. I can add that check to the other tests if we feel it is necessary.

29b5a6b just re-orders the tests so the test that checks for an error comes last in this group. The diff got confusing when I reordered it, so I just spit it into it's own commit so I could keep better track of the changes.

cmoesel (approved):
It works! (No, I'm not actually surprised).

I ran GoFSH on US Core and then ran SUSHI `master` on the GoFSH results and this PR's branch on the GoFSH results.  When I compared, there _were_ differences: `master` created duplicate mappings in the SDs, but this PR's branch did _not_.  So, that's GOOD!

ngfreiter (approved):
Yep, this is looking good to me.

Issue #659, opened 2020-11-09
SUSHI incorrectly counts minimum cardinality on sliced elements


ngfreiter:
If you create the following simple profile
```
Profile: Foo
Parent: Questionnaire
* item ^slicing.discriminator[0].type = #value
* item ^slicing.discriminator[0].path = "type"
* item contains group 0..1
```
Then create an instance of that profile like:
```
Instance: Q
InstanceOf: Foo
* status = #active
* item[0].linkId = "foo"
* item[0].type = #group
* item[0].answerOption[0].valueString = "1"
```
SUSHI will run with no errors. On the other hand, if you do:
```
Instance: Q
InstanceOf: Foo
* status = #active
* item[group].linkId = "foo"
* item[group].type = #group
* item[group].answerOption[0].valueString = "1"
```
You should get the error: "Element Questionnaire.item.answerOption.value[x] has minimum cardinality 1 but occurs 0 time(s)."

In the second case, that error message is extraneous, since the instance is valid, and does define a `valueString` element. This does seem to be unique to choice elements.

PR #660, opened 2020-11-09
Correctly count minimum cardinality on sliced elements

ngfreiter:
Fixes https://github.com/FHIR/sushi/issues/659.

The issue description gives a full example of this bug. When a choice element was the child of a slice, SUSHI would not count this choice element when checking the cardinality of the corresponding child on the sliced element. I would recommend looking at the example, because I think that illustrates it best.

To test, you can test it out on that example, and you can run the tests on https://github.com/FHIR/sushi/commit/882e942ba0c5792f555e0a20cd6b5d740d01f0dc, and the test added in that commit should fail, but then pass in https://github.com/FHIR/sushi/commit/b0a12332e0127deea640aac826c96aeb23bc6983.


src/export/InstanceExporter.ts
```diff
@@ -142,7 +142,11 @@ export class InstanceExporter implements Fishable {
       let instanceChild = instance[`_${childPathEnd}`] ?? instance[childPathEnd];
       // If the element is a choice, we will fail to find it, we need to use the choice name
       if (instanceChild == null && childPathEnd.endsWith('[x]')) {
-        const choiceSlices = children.filter(c => c.path === child.path && c.sliceName);
+        const possibleChoiceSlices = children;
```
jafeltra, line 145: This is pretty small, but when I was using the debugger trying to understand what was going on, I noticed that `children` was changing as new items were pushed to `possibleChoiceSlices`. It doesn't seem like there are any unintended consequences to this, but would it potentially be safer to use a `cloneDeep` here? What do you think?

jafeltra (commented):
This makes sense to me. I had one really minor comment, but besides that it looks good to me!

src/export/InstanceExporter.ts
cmoesel: I agree w/ Julia... if the intent is that `children` _should_ change, then I'd rather just remove line 145 and refer to `children` in the logic that follows.  If the intent is that `children` should not change, then `possibleChoiceSlices` should be a different array.

If we're not mutating the _elements_ themselves (but rather, just the _array_), then `children.slice()` would probably be more efficient than `cloneDeep(children)`.

cmoesel (commented):
It seems to work, but I think @jafeltra may be on to something, so I'd recommend addressing that.  Since I will be OOTO on Thursday, I'm _not_ marking this `request_changes`.  As long as Julia's comment is addressed, you can assume that I _approve_.

ngfreiter:
Good point @jafeltra, since as @cmoesel mentioned, only the array and not the objects inside are being copied, I only did a shallow copy using the spread operator. I used the spread operator, because who am I to disagree with [Dan](https://stackoverflow.com/a/20547803)

jafeltra (approved):
Who am I to disagree with you not disagreeing with Dan?

PR #662, opened 2020-11-10
Left-align contents of ignoreWarnings.txt

cmoesel:
In running regression tests, I found that in SUSHI 1.0.0 we don't properly left align the comments in the ignoreWarnings.txt file.  This may be harmless, but it's probably not intended -- and it does represent a change.  So this PR fixes it so the comments are left-aligned again.


jafeltra (approved):
I have vague memories of talking about this (or at least seeing it), so I'm glad it's fixed!

PR #663, opened 2020-11-10
Regression Tests

cmoesel:
_EDIT: This regression script currently only works on Mac._

See the README for details about the regression tests and how to run them.  This also includes a `regression/find-repos.ts` script that finds all pre-1.0.0 FSH projects at github.com/HL7 and in the auto-build history and uses it to populate `regression/all-repos.txt`.

By default, the regression tests the local SUSHI code vs the `master` SUSHI branch (pulling it from GitHub).  This is the best approach for regression-testing any PR against master -- _but_ it is _extremely_ slow.  If you want to get the full effect of testing all repos, I recommend running it like this:
```sh
$ bash regression/run-regression.sh regression/all-repos.txt 0.16.0 1.0.0-beta.4
```

If you want to try testing local vs master (or any other branch), I recommend slimming down the list of repos to 1 or 2.  You can do this by editing all-repos.txt or by creating another text file and passing it in as an argument.

If you want to test `find-repos.ts` you can delete `all-repos.txt` and then run it.  You can also try things like commenting out lines (they should stay commented after running `find-repos.txt` again) or deleting lines (which would result in those repos being re-added after running `find-repos.ts` again).

I've only testing this on Mac.  I have no idea if it works in Windows.  I expect that the only chance there is of it working would be in the Windows 10 Bash shell.


cmoesel:
Screenshot of summary report w/ links to diff files: 
image

cmoesel:
Screenshot of diff report for one repo: 
image

jafeltra (approved):
This PR is really cool! I tried running it and almost everything worked as expected. It seems like I get an error when trying to use the `diff2html` piece, so I don't get the nice HTML diffs linked in the summary page. I get `command not found: diff2html` when I remove the `-q` flag from that command. However, I looked at the `.diff` files, and the differences are all what I would expect, so I think the script itself is running as expected and this might be something on my end. So I don't think this should block my approval! But I might ask you about it again tomorrow.

cmoesel:
Doh.  It was relying on a locally installed version of `diff2html-cli` that I had.  I've fixed it now.  Can you try it again, @jafeltra?

jafeltra:
Ah! I have some beautiful diffs now!

jafeltra:
Ah wait one more quick question. I think this should be fine and is what we'd expect, but I do get a diff from the missing "external" logo now that that's been removed, for example: 

image

I think that's expected and should be fine.

cmoesel:
> I think that's expected and should be fine.

Agreed!

jafeltra (approved):
Oops I didn't realize my approval was dismissed! Re-approving!

PR #664, opened 2020-11-10
Only exit early if no FSH files AND no config file

cmoesel:
Previously, we exited early if no FSH files were found.  This caused a regression for a project that only used SUSHI for IG assembly (no FSH!).  Now we check for FSH files or a config file.

If you want to test this w/ the repo that caused the problem, clone [smart-web-messaging](https://github.com/HL7/smart-web-messaging) and run `master` against it and then this branch against it.  You'll see that `master` exits early, but this branch does not.


jafeltra (approved):
This is a cool edge case you were able to discover! It seems addressed appropriately!

src/app.ts
```diff
@@ -107,8 +107,14 @@ async function app() {
...
+      rawFSH.length === 0 &&
+      !fs.existsSync(path.join(originalInput, 'config.yaml')) &&
+      !fs.existsSync(path.join(originalInput, 'sushi-config.yaml')) &&
+      !fs.existsSync(path.join(originalInput, 'fsh', 'config.yaml')) &&
+      !fs.existsSync(path.join(originalInput, 'fsh', 'sushi-config.yaml'))
+    ) {
+      logger.info('No FSH files or sushi-config.yaml present.');
       process.exit(0);
     }
     config = readConfig(isIgPubContext ? originalInput : input, isLegacyIgPubContext);
```
ngfreiter, line 120: Would it be simpler to move the check you've added above below line 114, where the config is read in? That way couldn't you just check if the config existed, and when we deprecate `config.yaml`, we don't have it being checked for in two places.

That would mean that SUSHI would still run if it extracts the configuration from an IG resource, or if it creates one from a package.json, so if we don't want that, then we can just keep it as is.

cmoesel: I had considered that, but the problem is that if no config is found, `readConfig` logs an error and throws.  The behavior we want is that if there are no FSH files and no config, we exit gracefully (because it's not a FSH project) -- so logging an error and throwing would make for an ungraceful exit.

ngfreiter (approved):

PR #665, opened 2020-11-12
Inline extensions are not allowed on profiles

### LINKED ISSUE #591: SUSHI should emit error if "inline" extension is declared on a profile
Inline extensions should only be supported within complex extensions.  If an author attempts to create an inline extension in a profile, SUSHI should emit an error.

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Extension.2Eurl.20must.20be.20an.20absolute.20URL

mint-thompson:
Completes task [CIMPL-592](https://standardhealthrecord.atlassian.net/browse/CIMPL-592) and fixes #591 .

Log an error when an inline extension would be used on a profile. Many existing test cases were using inline extensions on profiles, so those tests are updated such that they still test the same code. Inline extensions are still allowed on extensions.

### DIFF
--- a/src/export/StructureDefinitionExporter.ts
+++ b/src/export/StructureDefinitionExporter.ts
@@ -322,6 +322,10 @@ export class StructureDefinitionExporter implements Fishable {
           this
         );
         urlElement.assignValue(slice.sliceName, true);
+        // Inline extensions should not be used on profiles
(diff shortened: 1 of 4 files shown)


ngfreiter (approved):
This looks good to me!

jafeltra (approved):
This makes sense to me!

Issue #654, opened 2020-10-27
adding description and name extensions to non conformacen resources.


Healthedata1:
see this Zulip chat for background: https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/Adding.20Description.20for.20Examples.20to.20ImplementationGuide.20-.20how

I would like to see sushi implement this ASAP.

1. create a sushi namespace extension for name and definition for now until can figure out what the final base will be to make the validator happy.

1.  for all non- conformance instance  in the input/resources, examples etc folder look for this extension when building the IG Resource to use for name and definitions.

1. Add neme and description keyword to FSH to create this extension on resources.

PR #666, opened 2020-11-13
Use name and description extensions to set values in generated IG

ngfreiter:
Fixes https://github.com/FHIR/sushi/issues/654 and addresses [CIMPL-591](https://standardhealthrecord.atlassian.net/browse/CIMPL-591).

If a BYO resource has an extension on `meta` with a URL of "http://hl7.org/fhir/StructureDefinition/instance-name", then the value of that extension is used to populate `ImplementationGuide.definition.resource.name`. If this extension is found on a resource that has a `title` field, then it is ignored. If the resource has an extension on `meta` with a URL of "http://hl7.org/fhir/StructureDefinition/instance-description", then the value of that extension is used to populate `ImplementationGuide.definition.resource.description` for that resource. If the extension is found on a resource that has a `description` field, then it is ignored. This is per the description given here: https://jira.hl7.org/browse/FHIR-29406.

If a resource has a configured name and/or description, that value is preferred, since I think it is less likely someone adds those values to their configuration by accident.

To test this, you can make a sample IG with resources like this:
```
{
  "resourceType": "Patient",
  "meta": {
    "extension": [
      {
        "url": "http://hl7.org/fhir/StructureDefinition/instance-description",
        "valueString": "Patient description"
      },
      {
        "url": "http://hl7.org/fhir/StructureDefinition/instance-name",
        "valueString": "Patient name"
      }
    ]
  },
  "id": "MyPatient",
  "name": {
    "family": "Test"
  }
}
```
in the non-fsh `input/*` folders.


src/ig/IGExporter.ts
```diff
@@ -1143,6 +1150,32 @@ export class IGExporter {
...
   }
 
+  private getMetaExtensionDescription(resource: InstanceDefinition): string {
+    const description = resource.meta?.extension?.find(
+      e => e.url === 'http://hl7.org/fhir/StructureDefinition/instance-description'
+    )?.valueString;
+    const definition = this.fhirDefs.fishForFHIR(resource.resourceType, Type.Resource);
+    if (
+      definition &&
+      !definition.snapshot.element.find((e: any) => e.id === `${resource.resourceType}.description`)
```
cmoesel, line 1160: I feel like this is more aggressive than it should be.  It may capture some non-conformance resources that happen to have fields named description or title.  For example [Goal](http://hl7.org/fhir/R4/goal.html) has a description field, but authors may still want to put a description for the example (notably because `Goal.description` is a `CodeableConcept`.  I think you should only exclude the "conformance" and "terminology" resource types, as described in the Jira issue.

I know Grahame said "they won't be allowed on resources that have title and description elements", but I think he just meant "conformance" and "terminology" resources.  But if you really want to implement what he said literally then you should check for both "title" and "description" being present at the same time.

cmoesel (changes requested):
I didn't have time for a full review, but found one issue that I think needs to be addressed.

src/ig/IGExporter.ts
ngfreiter: Yeah I misread the Jira issue and used the constraints that they voted out. Fixed in https://github.com/FHIR/sushi/pull/666/commits/8eeb5f2771d24aec630231b19767d237744ba9f9.

cmoesel: Actually, I missed that part in the Jira issue, so now I'm less sure.  It *does* kind of look like they intended it to be the way you originally coded it, but as I noted, that has flaws.  For now, I like the way you fixed it better anyway, so let's keep it like that until we hear otherwise.

ngfreiter: Unless I'm misreading the issue again they added this:
```
`%resource.title.exists().not()` for the name one and `%resource.description.exists().not()` for the description one
```
as a description on the 2020-11-02, but then changed that to this:
```
Will also define appearance constraints to ensure that these extensions aren't used in resources where there is a "proper" (non-extension-based) slot for name/description (i.e., conformance resources cannot use these extensions). (FHIRPath reflection allows for this kind of constraint, saying the resource type does not have an element where name = 'title', see http://hl7.org/fhirpath/#reflection)
```
Three days ago. So I think the way it is now is more consistent with the current description.

jafeltra (dismissed):
Based on what I understand from the discussions on Zulip, I think this makes sense!

src/ig/IGExporter.ts
```diff
@@ -1143,6 +1150,24 @@ export class IGExporter {
     }
   }
 
+  private getMetaExtensionDescription(resource: InstanceDefinition): string {
+    const description = resource.meta?.extension?.find(
+      e => e.url === 'http://hl7.org/fhir/StructureDefinition/instance-description'
+    )?.valueString;
```
cmoesel, line 1156: Based on the Zulip conversation, I think this is supposed to be `valueMarkdown`. See [zulip comment](https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/Adding.20Description.20for.20Examples.20to.20ImplementationGuide.20-.20how/near/215415480) and related [jira issue](https://jira.hl7.org/browse/FHIR-29405).

cmoesel (changes requested):
Looks good except I think they were planning to define the _description_ extension using `markdown` type.  So, once you adjust for that we should be good to go.

cmoesel (approved):
BINGO!

jafeltra (approved):

PR #669, opened 2020-11-16
Log dependencies individually when extracted from IG

mint-thompson:
When a configuration is extracted from an ImplementationGuide resource, info messages are logged for the elements of the configuration. The dependencies array may have many elements. To avoid an overly-long message, each element of the array is logged separately.


ngfreiter (approved):
This looks good to me!

cmoesel (approved):
Perfect!  And I confirmed that it now plays nicely w/ the IG Publisher, even w/ the project that previously caused issues in the publisher.  Thank you, @mint-thompson!

PR #671, opened 2020-11-17
Fix content reference bug

ngfreiter:
This fixes the bug that first arose here: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Possible.20bug.20with.20dependencies.20and.20nesting.

The issue here actually was not about dependencies, but about the way we find `contentReference` elements, `PlanDefinition.action.action` in this case. We had assumed that a `contentReference` would only be a relative reference, like `#PlanDefinition.action.action`, as this is how it is in the core FHIR spec. However, in the package being generated by the IG, a full reference is used, `http://hl7.org/fhir/StructureDefinition/PlanDefinition#PlanDefinition.action`. The logic we had failed in that case, and this PR fixes that.

The easiest way to test is to use the definition given below, and make a FSH tank with that definition as BYO-JSON, and then an instance defined in FSH which inherits from the `ReproPlanDefinition` profile. On `master`, you should see SUSHI spit out bugs like the ones shown in the linked zulip conversation. On this branch, you should instead see SUSHI run without error.

[StructureDefinition-ReproPlanDefinition.zip](https://github.com/FHIR/sushi/files/5556130/StructureDefinition-ReproPlanDefinition.zip)

### DIFF
--- a/src/fhirtypes/ElementDefinition.ts
+++ b/src/fhirtypes/ElementDefinition.ts
@@ -1700,7 +1700,9 @@ export class ElementDefinition {
         if (json) {
           const def = StructureDefinition.fromJSON(json);
           // Content references start with #, slice that off to id of referenced element
-          const referencedElement = def.findElement(this.contentReference.slice(1));
+          const referencedElement = def.findElement(
+            this.contentReference.slice(this.contentReference.indexOf('#') + 1)
+          );
           newElements = referencedElement?.children().map(e => {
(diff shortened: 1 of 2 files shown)


jafeltra (approved):
This makes sense to me and works as described!

cmoesel (approved):
Looks good.  I tested w/ your attached profile and the instance from the zulip thread.  Confirmed it is broken on master but works in this branch.

PR #675, opened 2020-11-25
Shrinkwrap

cmoesel:
This PR switches `package-lock.json` to `npm-shrinkwrap.json`.  The only difference between npm-shrinkwrap and package-lock is that npm-shrinkwrap will be published, but package-lock will not.  Apparently `package-lock.json` is only enforced in development environments, but not when installed as a global package from NPM.  Node discourages use of npm-shrinkwrap in most cases, but specifically identifies CLI tools that will be globally installed as one VALID use case.

We need this because we just got bit by an unintended dependency upgrade.  ANTLR 4.9 introduced a dependency on Node 14.

This PR also takes the extra precaution of changing dependencies on antlr4 to use `~` instead of `^`.  `~` will allow patch upgrades (e.g., `4.8.x`) but does not allow minor upgrades (e.g. `4.9.0`).

It's easiest to test this if you are running Node 12 or below.  In that case, do this on master:
```
$ npm pack
$ npm install -g ./fsh-sushi-1.0.1.tgz
```
You will see a message like this:
> npm WARN notsup Unsupported engine for antlr4@4.9.0: wanted: {"node":">=14"} (current: {"node":"12.19.0","npm":"6.14.8"})
> npm WARN notsup Not compatible with your version of node/npm: antlr4@4.9.0

Now do the same on the PR branch (pack and install) and you will no longer get that message.


cmoesel:
OK.  Given that the autobuilder is currently breaking on every SUSHI build, and that many SUSHI users are also struggling with this, and that I can't get in contact w/ any other team members, I am using my super powers to merge this _now_.

ngfreiter:
> This PR also takes the extra precaution of changing dependencies on antlr4 to use ~ instead of ^. ~ will allow patch upgrades (e.g., 4.8.x) but does not allow minor upgrades (e.g. 4.9.0).

Is the `~` something I should be seeing in the PR? Or am I misunderstanding?

cmoesel:
> Is the `~` something I should be seeing in the PR? Or am I misunderstanding?

Uh... yeah.  So I did:
```
$ npm install --save antlr4@~4.8.0
$ npm install --save-dev @types/antlr4@~3.7.2
```
and assumed that it put the right things into `package.json`.  But obviously did not confirm that.  Doh.  As you've seen, I already noticed this and fixed it in another PR.

PR #677, opened 2020-11-25
Switch antlr4 deps from ^ to ~

cmoesel:
We only want patch upgrades on antlr4.  I thought I did this before, but apparenlty:
```
$ npm install --save antlr4@~4.8.0
```
results in `^4.8.0` in the `package.json`.  Seems like a bug to me.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
   ],
   "license": "Apache-2.0",
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


ngfreiter (approved):
Looks good!

mint-thompson (approved):

PR #679, opened 2020-11-25
Switch package-lock to npm-shrinkwrap

cmoesel:
Same as #675 but for 0.16.x.  This PR switches `package-lock.json` to `npm-shrinkwrap.json`.  This PR also takes the extra precaution of changing dependencies on antlr4 to use `~` instead of `^`.  `~` will allow patch upgrades (e.g., `4.8.x`) but does not allow minor upgrades (e.g. `4.9.0`).

It's easiest to test this if you are running Node 12 or below.  In that case, do this on the branch `v0.x`:
```
$ npm pack
$ npm install -g ./fsh-sushi-1.0.1.tgz
```

You will see a message like this:
> npm WARN notsup Unsupported engine for antlr4@4.9.0: wanted: {"node":">=14"} (current: {"node":"12.19.0","npm":"6.14.8"})
> npm WARN notsup Not compatible with your version of node/npm: antlr4@4.9.0

Now do the same on this PR branch (pack and install) and you will no longer get that message.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1216,9 +1216,10 @@
       }
     },
     "@types/antlr4": {
-      "version": "4.7.1",
-      "resolved": "https://registry.npmjs.org/@types/antlr4/-/antlr4-4.7.1.tgz",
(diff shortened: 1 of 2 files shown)


ngfreiter (approved):
Looks good, warnings on 0.16.0, no warnings here.

PR #683, opened 2020-11-30
Update FshCode toString to support spaces in codes

jafeltra:
This PR updates the `toString` method of `FshCode` to support codes that have spaces in them. Codes with spaces should be wrapped in quotation marks. I also added tests to test the `toString` function, which includes a test that checks for codes with spaces. That particular test would not pass without the update to the `toString` function. Since there were not tests for FshCode to start with, I haven't tested every function in the class, but I can add those if we think that's something we need and I should include now.

Note that using these changes with GoFSH will also allow us to correctly generate FSH from codes that had spaces in them, which is where we originally noticed this issue.

### DIFF
--- a/src/fshtypes/FshCode.ts
+++ b/src/fshtypes/FshCode.ts
@@ -7,7 +7,7 @@ export class FshCode extends FshEntity {
   }
 
   toString(): string {
-    const str = `${this.system ?? ''}#${this.code}`;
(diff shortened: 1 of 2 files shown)


cmoesel:
Woah. Sorry about all that weird assign/unassign stuff.

julianxcarter (approved):
This looks good to me!

cmoesel (approved):
💯

Issue #672, opened 2020-11-18
Feature to have sushi to run without any fsh file  ( sushi-config.yaml only option)


Healthedata1:
A feature to have sushi to run without any fsh file  ( sushi-config.yaml only option) would be handy for developers (like me) who only want to use the Sushi-config.yaml to create a ig resource and menu and not have worry about adding fsh files to the input folder and not have the ig-publisher automatically rerun sushi.

cmoesel:
Currently, SUSHI can run without FSH _files_, but it seems it does require the `input/fsh` folder to at least _exist_.  So... a workaround would be to include an empty `input/fsh` folder.  But... if you're trying to avoid triggering the IG Publisher to run SUSHI (by _not_ having an `input/fsh` folder), then adding `input/fsh` would obviously defeat the purpose.

Healthedata1:
I thought it would be nice feature but maybe others don't just use it
solely to manage the implementation guides.

Eric

Eric M Haas, DVM, MS
Health eData Inc
35 Crescent Avenue, Sausalito, CA 94965
707.227.2608|Skype: haas.eric1
ehaas@healthedatainc.com

cmoesel:
I think it is something we might be able to support.  I was just providing some additional information.

The tricky bit for us is that older versions of SUSHI supported projects without `input/fsh` or `fsh` folders -- so we need to be able to disambiguate between that legacy use case and a use case like yours (that should look in `input` for additional resources rather than `ig-data`).

Healthedata1:
go it


Eric M Haas, DVM, MS
Health eData Inc
35 Crescent Avenue, Sausalito, CA 94965
707.227.2608|Skype: haas.eric1
ehaas@healthedatainc.com

PR #685, opened 2020-12-02
Support a sushi-config.yaml-only option (no FSH files in a current tank configuration)

jafeltra:
This PR addresses #672. This supports a sushi-config.yaml-only option and have SUSHI run on a current tank configuration without any FSH files.

On master, if we run SUSHI on a current tank configuration but do not have a `fsh` subdirectory with FSH files, SUSHI will incorrectly think we are in a legacy tank configuration. In order to appropriately detect that we are in a current tank configuration but without the `input/fsh` path, we need to check a few different conditions:
- there must be no `.fsh` files anywhere in the tank (if they are in any arbitrary folder, we are in a legacy "flat" tank)
- there must be no `fsh/` directory (if there is, we are in a legacy IG Publisher tank)
- there must be no `ig-data` directory (if there is, we are in a legacy "flat" tank)

Once we detect that we are indeed in the current tank configuration without FSH files (the sushi-config.yaml-only option), we need SUSHI to properly look in the correct places for BYO-resources, configuration, etc. Since we have been changing the "input" path based on the tank configuration we are in, we do a lot of path manipulation (going up a directory, up two directories, etc) in a bunch of different places. The easiest way I could figure out to have this sushi-config.yaml-only option be supported without changing all those places we manipulate paths was to set the input path as if there was an `input/fsh` folder. This lets the rest of the code operate as it would if we were in a typical current tank configuration. The one place that needs to be handled separately was when we process the FSH files. If we are in this new case where our path is to an `input/fsh` directory that does not exist, we can safely say there are no FSH files.

I had also attempted to change how we change and use the "input" path, but it proved to be fairly convoluted and didn't feel like a huge improvement, so I decided to go with this approach due to the relative simplicity of it, especially given that we might be able to simplify a lot of this when we stop supporting legacy tank configurations.

### DIFF
--- a/src/utils/Processing.ts
+++ b/src/utils/Processing.ts
@@ -47,6 +47,7 @@ export function findInputDir(input: string): string {
   const fshSubdirectoryPath = path.join(originalInput, 'fsh');
   const rootIgDataPath = path.join(originalInput, 'ig-data');
   const currentTankWithNoFsh =
+    !fs.existsSync(inputFshSubdirectoryPath) &&
     !fs.existsSync(fshSubdirectoryPath) &&
     !fs.existsSync(rootIgDataPath) &&
     !hasFshFiles(originalInput);
--- a/src/utils/Processing.ts
+++ b/src/utils/Processing.ts
@@ -37,13 +37,16 @@ export function findInputDir(input: string): string {
   // TODO: Legacy support. Remove when no longer supported.
   // Use input/fsh/ subdirectory if not already specified and present
   const inputFshSubdirectoryPath = path.join(originalInput, 'input', 'fsh');
+  const inputSubdirectoryPath = path.join(originalInput, 'input');
   if (fs.existsSync(inputFshSubdirectoryPath)) {
+    input = path.join(originalInput, 'input', 'fsh');
+  } else if (fs.existsSync(inputSubdirectoryPath)) {
     input = path.join(originalInput, 'input', 'fsh');
   }
 
   // Use fsh/ subdirectory if not already specified and present
   const fshSubdirectoryPath = path.join(originalInput, 'fsh');
-  if (!fs.existsSync(inputFshSubdirectoryPath)) {
+  if (!fs.existsSync(inputFshSubdirectoryPath) && !fs.existsSync(inputSubdirectoryPath)) {
     let msg =
       '\n\n!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! IMPORTANT !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!\n';
     if (fs.existsSync(fshSubdirectoryPath)) {
@@ -187,6 +190,13 @@ export function loadExternalDependencies(
 export function getRawFSHes(input: string): RawFSH[] {
   let files: string[];
   try {
+    if (
+      path.basename(path.dirname(input)) === 'input' &&
+      path.basename(input) === 'fsh' &&
+      !fs.existsSync(input)
+    ) {
+      return [];
+    }
     files = getFilesRecursive(input);
   } catch {
     logger.error('Invalid path to FSH definition folder.');
--- a/src/utils/Processing.ts
+++ b/src/utils/Processing.ts
@@ -31,22 +31,36 @@ export function ensureInputDir(input: string): string {
   return input;
 }
 
+function checkForFshFiles(path: string): boolean {
+  try {
+    const files = getFilesRecursive(path).filter(file => file.endsWith('.fsh'));
+    return files.length > 0;
+  } catch (error) {
+    return false;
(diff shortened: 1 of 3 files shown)


cmoesel (changes requested):
This looks good and the implementation doesn't add too much complexity -- so that's even better!  I manually tested this against:
- SUSHI 1.0.0 IG w/ FSH files
- SUSHI 1.0.0 IG w/out FSH files
- Legacy IG w/ FSH files
- Legacy IG w/out FSH files
- Flat IG w/ FSH files
- Flay IG w/out FSH files

SUSHI behaved the same between master and this PR for all of the test cases except SUSHI 1.0.0 w/out FSH files.  The change on SUSHI 1.0.0 w/out FSH files was exactly as we wanted, so that's good too!

I noted one thing in the code that I'd like to change for efficiency...

src/utils/Processing.ts
```diff
@@ -31,19 +31,36 @@ export function ensureInputDir(input: string): string {
...
+  const fshFilesPresent = checkForFshFiles(originalInput);
+  const fshSubdirectoryPath = path.join(originalInput, 'fsh');
+  const rootIgDataPath = path.join(originalInput, 'ig-data');
+  const currentTankWithNoFsh =
+    !fshFilesPresent && !fs.existsSync(fshSubdirectoryPath) && !fs.existsSync(rootIgDataPath);
+
   if (fs.existsSync(inputFshSubdirectoryPath)) {
+    // Use input/fsh/ subdirectory if not already specified and present
+    input = path.join(originalInput, 'input', 'fsh');
+  } else if (currentTankWithNoFsh) {
```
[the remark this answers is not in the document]
jafeltra: I didn't intentionally split them; I think this is just a small window into how my brain was separating the two. That, and the fact that I went back and forth with what the input path should be. I'll combine them!

jafeltra:
@cmoesel your comments should be addressed. I also realized I didn't add any tests for `hasFshFiles` so I added two simple tests for that function.

And thank you for testing this out on so many combinations of tanks! I'm always doubtful that I checked every case, so I appreciate it! The only additional (I think) case that I had tested that you didn't explicitly list was SUSHI 1.0.0 IG w/ fsh/ with no files. That should work the same on this and on master.

mint-thompson (approved):
Tested with the same scenarios mentioned by @cmoesel and observed the same results. No additional findings beyond what has already been mentioned and addressed.

src/utils/Processing.ts
```diff
@@ -44,17 +44,16 @@ export function findInputDir(input: string): string {
...
+    !hasFshFiles(originalInput);
 
-  if (fs.existsSync(inputFshSubdirectoryPath)) {
-    // Use input/fsh/ subdirectory if not already specified and present
-    input = path.join(originalInput, 'input', 'fsh');
-  } else if (currentTankWithNoFsh) {
-    // Use input/fsh/ subdirectory when in the current tank configuration without FSH files
+  // Use input/fsh/ subdirectory if not already specified and present
+  // or when in the current tank configuration without FSH files
+  if (fs.existsSync(inputFshSubdirectoryPath) || currentTankWithNoFsh) {
```
cmoesel, line 56: I hate to be annoying, but I'm realizing now that `currentTankWithNoFsh` does not check `inputFshSubdirectoryPath`.  So in our most common use case (a project w/ `input/fsh`), the assignment to `currentTankWithNoFsh` will still cause `hasFshFiles(...)` to be invoked even though `currentTankWithNoFsh` isn't even looked at if `input/fsh` exists.

So, to really avoid that expensive call, you'd need to either include `!fs.existsSync(inputFshSubdirectoryPath)` in the assignment of `currentTankWithNoFsh` or you'd need to make an `isCurrentTankWithNoFsh` _function_ and call that on line 56 above (so it isn't called if `input/fsh` exists).

cmoesel: Sorry I missed that case when I reviewed the first time.  I actually had thought of it, but then for some reason thought it was fine after reviewing the code.  I think maybe I saw `fshSubdirectoryPath` and got it mixed up with `inputFshSubdirectoryPath` (or something... who knows).

cmoesel (changes requested):
Oops.  I realized that you did as I suggested (thanks!) but it still causes `hasFshFiles` to be called in most cases (no thanks to me!)...

src/utils/Processing.ts
jafeltra: No worries I should have thought of that case as well. I went with the first approach you described and added the check for `!fs.existsSync(inputFshSubdirectoryPath)` to the assignment of `currentTankWithNoFsh` because it seemed to make more sense to me mentally, but I'm happy to update it if you prefer the other approach. Fixed in 4d8f692.

Also, if the variable names are getting confusing, I can try renaming them.

cmoesel (approved):
I'll take it!

PR #687, opened 2020-12-04
Import Parameterized RuleSet

mint-thompson:
Completes task [CIMPL-573](https://standardhealthrecord.atlassian.net/browse/CIMPL-573).

This PR includes the grammar and code changes necessary for importing the new FSH type `ParamRuleSet`. These objects store the name, parameter list, and text content of a parameterized RuleSet. When a parameterized RuleSet is inserted, text substitution on content is used to generated a `RuleSet`. This way, existing code for applying an `InsertRule` can be mostly reused.

Because a `ParamRuleSet` may be used multiple times with different parameters, the parameters used for a given `InsertRule` are tracked in the `InsertRule` and the `FSHDocument`. To explain a bit about the new `appliedRuleSets` map in `FSHDocument`: the map keys are string lists, where the first element is the ruleset name, and successive elements are the provided parameters. The `Immutable` classes are used in order to enable this, as they provide comparison by list contents rather than by object reference.

A side effect of these changes is that an empty RuleSet will no longer be added to a `FSHDocument`. See `FSHImporter.RuleSet.test.ts` for the relevant test case.

The main points I am wondering about:
- Is it necessary to have two separate maps in `FSHDocument` for parameterized and non-parameterized `RuleSet`s? Based on the rules for the keys in `appliedRuleSets`, a non-parameterized `RuleSet` would get a key that is a list of one string. On the other hand, it may be handy during export steps to definitively search one map or the other.
- I made the implementation decision that parameters would be applied sequentially. I provided a test case in `ParamRuleSet.test.ts` to demonstrate this behavior. However, I do not have strong feelings about this behavior as opposed to simultaneously applied parameters, so I would like feedback on this behavior.
- I am particularly interested if you can think of any additional test cases that would help clarify the behavior of this new feature.


antlr/src/main/antlr/FSH.g4
```diff
@@ -27,6 +29,72 @@ csRule:             concept | caretValueRule | insertRule;
 ruleSet:            KW_RULESET SEQUENCE ruleSetRule+;
 ruleSetRule:        sdRule | concept | vsComponent;
 
+paramRuleSet:       KW_RULESET SEQUENCE PARAMETER_DEF_LIST paramRuleSetRule+;
+paramRuleSetRule:   STAR
```
ngfreiter, line 33: I'm having trouble understanding why this definition has to have the long list of tokens. Would we have to make sure any new tokens that we add are added to this list.

mint-thompson: in short: yes. The way that the grammar knows that the contents of the `paramRuleSet` are complete is by running into a token that defines the start of a new `entity`. Therefore, this list has to be all tokens that do not start a new `entity`. I tried doing this by negating an alternative of tokens, similar to this:
```
paramRuleSetRule: STAR ~(KW_PROFILE | KW_EXTENSION | KW_INVARIANT | ...)+;
```
But, the resulting parser did not behave as expected.

ngfreiter: I tried it with a rule like this:
```
paramRuleSetRule:   STAR ~(KW_PROFILE | KW_ALIAS | KW_EXTENSION | KW_INSTANCE | KW_INVARIANT | KW_VALUESET | KW_CODESYSTEM | KW_RULESET | KW_MAPPING)*;
```
And it seemed to work kinda alright. The tests in `FSHImporter.ParamRuleSet.test.ts` fail, but they only fail because for some reason the white space is parsed in a bit differently, and I don't think that would be a huge barrier to overcome. I'm not sure why the `*` works where the `+` doesn't, the difference in logic there isn't totally clicking for me.

If we can make the negation approach work, I think that be a lot less risky going forward if we end up adding new tokens.

mint-thompson: dang. You are absolutely right. I've updated the grammar accordingly. This is much better. I owe you an ice-cream

antlr/src/main/antlr/FSH.g4
```diff
@@ -27,6 +29,18 @@ csRule:             concept | caretValueRule | insertRule;
...
```
ngfreiter, line 33: As I understand it, a `paramRuleSetRule` isn't really a single rule, but rather a block of rules. Like if the `RuleSet` was:
```
RuleSet: Foo (one, two)
* rule1
* rule2
* rule3
```
 then in this case the `paramRuleSetRule` would be all of `* rule1\n*rule 2\n*rule 3`, since the grammar only stops when it hits another keyword for an entity. Am I wrong about that and that text gets split up into three separate `paramRuleSetRule`s? If I'm not wrong, I think it would make sense to rename `paramRuleSetRule` to something like `paramRuleSetContent`, and update the grammar to indicate that there is only one block of content, not several rules.

src/import/FSHImporter.ts
```diff
@@ -1380,8 +1421,110 @@ export class FSHImporter extends FSHVisitor {
...
+    const [rulesetName, ruleParams] = this.parseRulesetReference(ctx.RULESET_REFERENCE().getText());
+    insertRule.ruleSet = rulesetName;
+    if (ruleParams) {
+      insertRule.params = this.parseInsertRuleParams(ruleParams);
+      const ruleSet = this.paramRuleSets.get(insertRule.ruleSet);
+      if (ruleSet) {
+        const ruleSetIdentifier = Immutable.List<string>([ruleSet.name, ...insertRule.params]);
+        if (
+          ruleSet.parameters.length === insertRule.params.length &&
+          !this.currentDoc.appliedRuleSets.has(ruleSetIdentifier)
```
ngfreiter, line 1433: If the `currentDoc` does have the given `ruleSetIdentifier` already, that drops into the `Incorrect number of parameters` error condition (unless I'm getting mixed up in my nesting here), and that doesn't seem to make sense.

Should this be an error condition in the first place? I'm not sure why a certain rule set couldn't be inserted twice into the same document with the same params.

ngfreiter (commented):
I noticed a couple of things that should maybe change, and a couple of questions I had. But overall this is looking pretty dang cool to me.

antlr/src/main/antlr/FSHLexer.g4
```diff
@@ -0,0 +1,129 @@
...
+fragment NONWS: ~[ \t\r\n\f\u00A0];
+fragment NONWS_STR: ~[ \t\r\n\f\u00A0\\"];
+
+// IGNORED TOKENS
+WHITESPACE:         WS -> channel(HIDDEN);
+BLOCK_COMMENT:      '/*' .*? '*/' -> skip;
+LINE_COMMENT:       '//' .*? [\r\n] -> skip;
+
+mode INSERT_RULE;
+RULESET_REFERENCE:      WS* NONWS+ (WS* ('(' ('\\)' | '\\\\' | ~[)])+ ')'))? -> popMode;
```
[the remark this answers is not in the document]
mint-thompson: Tokens from other modes aren't available in a given mode, but fragments are.

src/import/FSHImporter.ts
```diff
@@ -130,9 +133,11 @@ export class FSHImporter extends FSHVisitor {
   private currentFile: string;
   private currentDoc: FSHDocument;
   private allAliases: Map<string, string>;
+  paramRuleSets: Map<string, ParamRuleSet>;
```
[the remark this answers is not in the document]
mint-thompson: Right now, I only have it visible so that I can check it in the tests. But, you're right that it makes sense for it to be private. Do you think it would be better to add a public method for test support rather than having the field itself be public?

antlr/src/main/antlr/FSH.g4
mint-thompson: That's a good point. I think at one point I had `STAR` as one of the prohibited tokens, so it would actually get more than one match, but as it is written, `paramRuleSetContent` is a much more accurate name.

src/import/FSHImporter.ts
mint-thompson: Ah, I shouldn't have combined both of those boolean expressions. It is, of course, completely fine to insert a rule set twice with the same parameters. There's no need to generate the applied rule set again, so that step can be skipped, but nothing wrong has occurred.

antlr/src/main/antlr/FSHLexer.g4
```diff
@@ -0,0 +1,129 @@
...
+fragment WS: [ \t\r\n\f\u00A0];
+fragment NONWS: ~[ \t\r\n\f\u00A0];
+fragment NONWS_STR: ~[ \t\r\n\f\u00A0\\"];
+
+// IGNORED TOKENS
+WHITESPACE:         WS -> channel(HIDDEN);
+BLOCK_COMMENT:      '/*' .*? '*/' -> skip;
+LINE_COMMENT:       '//' .*? [\r\n] -> skip;
+
+mode INSERT_RULE;
```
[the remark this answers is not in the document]
mint-thompson: When we have an insert rule for a ruleset with parameters, the list of parameters can have all kinds of contents. For example, consider these insert rules:
```
* insert RuleSetA (#final)
* insert RuleSetB("something")
* insert RuleSetC (example)
```
If we use the tokens defined in the default mode, we will get some awkward results from the lexer. `(#final)` in RuleSetA will be a `CODE` token, `(example)` in RuleSetC will be `KW_EXAMPLE`, and `RuleSetB("something")` is just a big `SEQUENCE`. So, changing modes lets us temporarily ignore all of those lexer rules, and instead grab the right chunk of text for us to process later.

src/import/FSHImporter.ts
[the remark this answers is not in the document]
ngfreiter: I'm not too worried about it, since I don't think we will start using `FSHImporter` other places outside of testing. So I'm fine with it as is, just wanted to understand why.

ngfreiter (dismissed):
Testing on mCODE the grammar changes didn't cause any issues. I am interested in seeing the results of @cmoesel's regression tests, but barring anything too unexpected there, this looks good to me!

cmoesel:
I'll be glad to run my new regression script against it if @mint-thompson can resolve these new conflicts.

mint-thompson:
@cmoesel merge conflicts are resolved! I'd run the regression script myself, but, well, you know. Windows and all that.

cmoesel:
I've run the regression and I'm glad to report that no IGs were harmed in the making of this PR.

I'll do a fuller review of the actual PR tomorrow!

src/import/FSHImporter.ts
```diff
@@ -1384,8 +1422,110 @@ export class FSHImporter extends FSHVisitor {
...
+  private parseGeneratedRuleSet(input: string, name: string) {
+    // define a temporary document that will contain this RuleSet
+    const tempDocument = new FSHDocument(this.currentFile);
+    // save the currentDoc so it can be restored after parsing this RuleSet
+    const parentDocument = this.currentDoc;
+    this.currentDoc = tempDocument;
+    const subContext = this.parseDoc(input, this.currentFile);
+    this.visitDoc(subContext);
+    // restore parentDocument
+    this.currentDoc = parentDocument;
```
cmoesel, line 1520: If `this.parseDoc` or `this.visitDoc` throws, then `this.currentDoc` will never get set back.  I'm guessing that if they throw maybe we bomb out of everything and it doesn't matter anyway, but... in case we _do_ gracefully recover (now or in the future), it might be safer to wrap lines 1517-1518 in `try` and put line 1520 in a `finally` -- just to guarantee we set it back no matter what happens.

test/fshtypes/ParamRuleSet.test.ts
```diff
@@ -0,0 +1,42 @@
...
+      expect(appliedContents).toBe(expectedContents);
+    });
+
+    it('should replace parameters sequentially', () => {
+      const ruleSet = new ParamRuleSet('MyParamRuleSet');
+      ruleSet.parameters = ['first', 'second'];
+      ruleSet.contents = ['* code from {first}', '* category from {second}'].join(EOL);
+
+      const appliedContents = ruleSet.applyParameters(['{second}Deluxe', 'MySystem']);
+      const expectedContents = ['* code from MySystemDeluxe', '* category from MySystem'].join(EOL);
```
cmoesel, line 38: Oh... I see now.  So `['{second}Deluxe', 'MySystem']` works, but `['MySystem', '{first}Deluxe']` would not...

Hmm...  I'm not sure I like that.  I feel like either _no_ 2nd-level substitutions should happen or _all_ 2nd-level substitutions should happen.  This inbetween state is likely to confuse users.

test/import/FSHImporter.Profile.test.ts
```diff
@@ -1838,6 +1895,260 @@ describe('FSHImporter', () => {
...
+        Profile: ObservationProfile
+        Parent: Observation
+        * insert MysteriousRuleSet ("mystery")
+        `;
+        importer.import([new RawFSH(input, 'Insert.fsh')]);
+        expect(loggerSpy.getLastMessage('error')).toMatch(
+          /Could not find RuleSet with parameters named MysteriousRuleSet/s
+        );
+        expect(loggerSpy.getLastMessage('error')).toMatch(/File: Insert\.fsh.*Line: 4/s);
+      });
```
cmoesel, line 2151: I think it would be good to have negative tests for when there are parser errors when applying the ruleset.  For example:
```
CardRuleSet(path, min, max)
* {path} {minId}..{max}

Profile: MyObservation
Parent: Observation
* insert CardRuleSet(path with spaces, 1, *) // <-- what happens here?
* insert CardRuleSet(code, *, 1) // <-- oops, got min and max backwards and '*' isn't valid min
* insert CardRuleSet(code, 1, infinityAndBeyond) // <-- Buzz Lightyear
```

cmoesel: Similarly, I think we want to test to ensure that if a line in a ruleset is valid in the parser, but invalid in the exporter (for example, a path that doesn't exist), that the reported error is appropriately tracked to the right locations (e.g., the insert rule where it was applied and the underlying ruleset rule that originated the error).

cmoesel (changes requested):
Very cool stuff!  I have a few questions, comments, and suggestions...

src/fshtypes/ParamRuleSet.ts
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  applyParameters(values: string[]): string {
+    let appliedContents = this.contents;
+    this.parameters.forEach((parameter, index) => {
+      appliedContents = appliedContents.replace(
+        new RegExp(`{${escapeRegExp(parameter)}}`, 'g'),
+        values[index] ?? ''
+      );
+    });
```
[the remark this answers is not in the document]
cmoesel: I saw in the tests that actually the current behavior is somewhere halfway between... more comments in that part of the tests. ;-)

antlr/src/main/antlr/FSH.g4
```diff
@@ -27,6 +29,18 @@ csRule:             concept | caretValueRule | insertRule;
 ruleSet:            KW_RULESET SEQUENCE ruleSetRule+;
 ruleSetRule:        sdRule | concept | vsComponent;
 
+paramRuleSet:       KW_RULESET SEQUENCE PARAMETER_DEF_LIST paramRuleSetContent;
```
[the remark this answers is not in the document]
mint-thompson: The `paramRuleSetContent` rule just gets a big blob of text. It doesn't match any individual rules the way that `ruleSetRule` does. So, there would need to be a second parse on the content of the non-parameterized `ruleSet` the way there is for rule sets generated by applying parameters to a `paramRuleSet`.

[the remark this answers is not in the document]
cmoesel: Right, but if we already have the code to do the 2nd parse, it seems to me like it's more maintainable to have one way to handle all RuleSets.  Then we don't have two RuleSet classes, two ruleSet maps, and two sets of ruleset import code.  And as a dev, you don't need to think so much about -- is this a _normal_ RuleSet or a _parameterized_ RuleSet?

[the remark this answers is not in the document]
mint-thompson: The catch is, the second parse is using the current grammar for `ruleSet` and `ruleSetRule`. Applying parameters to a parameterized RuleSet is what allows it to be parsed as a non-parameterized RuleSet using those rules.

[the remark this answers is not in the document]
cmoesel: Oh dang.  That is a catch, isn't it?  Well, shoot.  Nevermind.  Nothing to see here. ;-)

test/import/FSHImporter.Profile.test.ts
mint-thompson: I'm adding in a negative test as you described, but I think that the case of "a line in a ruleset is valid in the parser, but invalid in the exporter" falls under the scope of Export Parameterized RuleSet.

The tricky part I'm running into is how to usefully describe the error. For example, in the first insert rule you've described above, we get a good old-fashioned "no viable alternative" error from the parser. This happens when parsing the generated RuleSet, which means we've called `parseDoc` on a manually constructed block of text. The error is logged by a `FSHListener`, which happens before `parseDoc` returns. Once we're back in `parseGeneratedRuleSet`, it's too late to change what the error is. I think the best that could be managed here is if the listener kept an internal list of any calls to its `syntaxError` method, we could check that list once we're back in `parseGeneratedRuleSet` and emit an additional message along the lines of "at least one bad thing happened when I tried to insert a parameterized rule set on line N", which at least lets the user know what line the offending insert rule is on. And if the error that is logged while parsing the generated document is not handled by our `FSHListener`, then we'd have to do something else. I guess if we ask the logger before we start how many errors it has, and it has more of them after running `parseDoc`, we can assume something went wrong with the insert rule?

test/fshtypes/ParamRuleSet.test.ts
mint-thompson: That's reasonable. I think that between those two options, I would want no extra substitutions to take place. I'll make that change.

src/import/FSHImporter.ts
```diff
@@ -578,6 +590,32 @@ export class FSHImporter extends FSHVisitor {
...
+        .split(',')
+        .map(param => param.trim());
+      paramRuleSet.contents = this.visitParamRuleSetContent(ctx.paramRuleSetContent());
+      this.paramRuleSets.set(paramRuleSet.name, paramRuleSet);
+    }
+  }
+
+  visitParamRuleSetContent(ctx: pc.ParamRuleSetContentContext): string {
+    const result = ctx.start.getInputStream().getText(ctx.start.start, ctx.stop.stop);
+    return result;
```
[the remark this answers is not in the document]
mint-thompson: `ctx.getText()` returns the matched tokens, but without the spaces that separated them, since those were piped to the `HIDDEN` channel. Accessing the input stream allows us to get the original contents.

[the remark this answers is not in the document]
cmoesel: Ah right.  I forgot about that.  OK.  Makes sense!

src/fshtypes/ParamRuleSet.ts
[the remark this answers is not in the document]
cmoesel: OK.  Based on [this comment](https://github.com/FHIR/sushi/pull/687#discussion_r542609004) I think you're planning to allow no recursive substitutions.

test/fshtypes/ParamRuleSet.test.ts
cmoesel: Yeah.  I think that makes sense-- because one big issue w/ recursive substitutions is that suddenly the var names are _meaningful_ outside of the RuleSet definition. So if you simply renamed a var in your RuleSet, you could break code that uses it (if we allowed recursive substitution).

test/import/FSHImporter.Profile.test.ts
cmoesel: I think I'm OK w/ starting w/ a reasonable solution (not too many hoops to jump through) and seeing what it looks like (i.e., if it's "good enough").  If making it perfect adds a lot of complication, then I think we need to weigh the tradeoffs.

cmoesel: After reviewing the code and tests, I think we need to do _something_ special here.  My main concern at this point is that the error message indicates its source as the file w/ the insert rule, but the line number is relative to the in-memory doc going through the second parse.  I think it's OK to return incomplete information if we need to, but we shouldn't return incorrect information.

One potentially simple (?) solution might just be to change the doc filename that gets set when we do the 2nd parse.  If we can change the name to something that clearly indicates its the in-memory application of a specific ruleset rule, that might be useful.

src/fshtypes/ParamRuleSet.ts
```diff
@@ -0,0 +1,31 @@
...
+  contents: string;
+
+  constructor(public name: string) {
+    super();
+    this.parameters = [];
+    this.contents = '';
+  }
+
+  applyParameters(values: string[]): string {
+    const paramUsage = new RegExp(`{(${this.parameters.map(escapeRegExp).join('|')})}`, 'g');
```
cmoesel, line 15: How would you feel about allowing for optional whitespace between the `{`/`}` and parameter names?  E.g., right now, you support `{cheese}` but not `{ cheese }`.  I think it might be nice to allow for that whitespace (as I think I personally would use it).

src/fshtypes/ParamRuleSet.ts
```diff
@@ -0,0 +1,31 @@
...
+        return values[matchIndex];
+      } else {
+        return '';
+      }
+    });
+  }
+
+  getUnusedParameters() {
+    return this.parameters.filter(param => {
+      return this.contents.indexOf(`{${param}}`) === -1;
```
cmoesel, line 28: If you change it to support whitespace around the variable name above, then don't forget to do it here too!

cmoesel (commented):
Thanks for detecting unused variables.  Your code made me think of one change I think would be helpful to keep things a bit more flexible.  See individual comments.

mint-thompson:
Whitespace in substitution tokens is now supported! Also, I added the parser error collection strategy that we discussed yesterday. However, while working on this, I realized that the error collector will only collect errors generated by the parser, and not any of the other errors that are directly logged by `FSHImporter`, such as a cardinality with neither side specified. Those errors will still be logged directly, with the source info from the generated text. So, I think it may be necessary to modify how those errors are emitted in order to consistently collect all errors from parsing generated text.

test/import/FSHImporter.Profile.test.ts
```diff
@@ -2145,7 +2177,88 @@ describe('FSHImporter', () => {
...
+          /Warnings parsing insert rule with parameterized RuleSet WarningRuleSet/s
+        );
+        expect(loggerSpy.getLastMessage('warn')).toMatch(/File: Insert\.fsh.*Line: 4/s);
+      });
+
+      it('should log one error when an insert rule with parameters results in non-parser errors', () => {
+        const input = `
+        Profile: MyObservation
+        Parent: Observation
+        * insert CardRuleSet(nonExistentPath, , )
```
ngfreiter, line 2255: I tried some manual testing with this test case, and one issue I've noticed is that when an `InsertRule` references a `RuleSet` that is not defined, we end up adding `undefined` to the list of rules for that profile, so we end up with an error like this 
image

ngfreiter: This FSH alone:
```
Profile: MyObservation
Parent: Observation
* insert CardRuleSet(nonExistentPath, , )
```
should recreate this.

ngfreiter (commented):
I noticed an error case we may want to improve, but it's looking good besides that.

test/import/FSHImporter.Profile.test.ts
mint-thompson: Ah, I understand, since this is now a scenario where a rule visitor might not return an instance of a rule. Do you think it would make more sense to return the `InsertRule` anyway and instead emit the error when applying the rule? This lines up with the timing for errors for non-parameterized rule sets, but it does mean that we would need to be keeping parameterized rule sets in their original forms on the documents in our tank so that we can correctly discern between "wrong number of parameters" and "undefined ruleset" errors for parameterized rule sets. The alternative to this is to modify some of the code in the importer so that `undefined` doesn't end up in a list of rules.

ngfreiter: I think if it's easier to implement, it's fine to just not add the rule to the list of rules at all. It's a bit of a break from what we have done with other rules, but at the same time this is a unique case.

cmoesel (changes requested):
I've tried to manually test this, but it's not working.  When I try to process this FSH:
```
// Example from: https://fshschool.org/downloads/sept-2020-fsh-tutorial-part-3.pdf

Alias: $ExpExt = http://hl7.org/fhir/StructureDefinition/capabilitystatement-expectation

RuleSet: SupportResource (i, resource, expectation)
* rest.resource[{i}].type = {resource}
* rest.resource[{i}].extension[0].url = $ExpExt
* rest.resource[{i}].extension[0].valueCode = {expectation}

RuleSet: SupportProfile (i, j, profile, expectation)
* rest.resource[{i}].supportedProfile[{j}] = {profile}
* rest.resource[{i}].supportedProfile[{j}].extension[0].url = $ExpExt
* rest.resource[{i}].supportedProfile[{j}].extension[0].valueCode = {expectation}

RuleSet: SupportInteraction (i, j, interaction, expectation)
* rest.resource[{i}].interaction[{j}] = {interaction}
* rest.resource[{i}].interaction[{j}].extension[0].url = $ExpExt
* rest.resource[{i}].interaction[{j}].extension[0].valueCode = {expectation}

Instance: TestCapabilityStatement
InstanceOf: CapabilityStatement
* status = #active
* date = "2020-12-18"
* kind = #requirements
* fhirVersion = #4.0.1
* format[0] = #json
// MeasureReport requirements
* insert SupportResource(0, #MeasureReport, #SHALL)
* insert SupportProfile(0, 0, "http://hl7.org/fhir/us/saner/StructureDefinition/PublicHealthMeasureReport", #SHALL)
* insert SupportInteraction(0, 0, #create, #SHALL)
* insert SupportInteraction(0, 1, #create, #SHALL)
// Measure requirements
* insert SupportResource(1, #Measure, #SHALL)
* insert SupportProfile(1, 0, "http://hl7.org/fhir/us/saner/StructureDefinition/PublicHealthMeasure", #SHOULD)
* insert SupportProfile(1, 1, "http://hl7.org/fhir/us/saner/StructureDefinition/PublicHealthMeasureStratifier", #SHOULD)
* insert SupportInteraction(1, 0, #create, #SHOULD)
* insert SupportInteraction(1, 1, #create, #SHOULD)
// Location requirements
* insert SupportResource(2, #Location, #SHALL)
* insert SupportProfile(2, 0, "http://hl7.org/fhir/us/saner/StructureDefinition/saner-resource-location", #SHOULD)
* insert SupportInteraction(2, 0, #create, #SHOULD)
* insert SupportInteraction(2, 1, #create, #SHOULD)
// QuestionnaireResponse requirements
* insert SupportResource(3, #QuestionnaireResponse, #SHALL)
* insert SupportProfile(3, 0, "http://hl7.org/fhir/us/saner/StructureDefinition/SanerQuestionnaireResponse", #SHALL)
* insert SupportInteraction(3, 0, #create, #SHALL)
* insert SupportInteraction(3, 1, #create, #SHALL)
```

I get errors like this:
> error Unable to find definition for RuleSet SupportResource.
>   File: /Users/cmoesel/dev/tmp/TestParamRS/input/fsh/patient.fsh
>   Line: 28
> error Unable to find definition for RuleSet SupportProfile.
>   File: /Users/cmoesel/dev/tmp/TestParamRS/input/fsh/patient.fsh
>   Line: 29
> error Unable to find definition for RuleSet SupportInteraction.
>   File: /Users/cmoesel/dev/tmp/TestParamRS/input/fsh/patient.fsh
>   Line: 30
> error Unable to find definition for RuleSet SupportInteraction.
>   File: /Users/cmoesel/dev/tmp/TestParamRS/input/fsh/patient.fsh
>   Line: 31

I think this should work, right?  Also I noticed that if I defined a rule set like this:
```
RuleSet: SupportResource (i, resource, expectation)
```
then it failed the grammar.  Ideally the space between the RuleSet name and `(` should be optional.

mint-thompson:
@cmoesel The `Unable to find definition for RuleSet` are from the export side. In hindsight, I should have opened the PR for the export side first.

I'll see what I can do about about being flexible about whitespace between the RuleSet name and the start of the parameter list.

mint-thompson:
@cmoesel The `Unable to find definition for RuleSet` are from the export side. In hindsight, I should have opened the PR for the export side first.

I'll see what I can do about about being flexible about whitespace between the RuleSet name and the start of the parameter list.

cmoesel:
Oh right!  I forgot about export.  I think in the past, most features didn't cause errors if we only implemented `import` without `export` -- but since this is an enhancement to an existing feature, I can see why it would cause problems.  Sorry for the false alarm!

cmoesel (approved):
Awesome.  Can't wait until people start using it!

PR #689, opened 2020-12-04
Allow block comments without whitespace

### LINKED ISSUE #681: /* */ require spaces
A user has reported that the comment delimiters `/*` and `*/` seem to require spaces to be recognized -- and that SUSHI may hang when it encounters these without spaces (for example, `/*========*/`).  We need to investigate and see (a) why it hangs, and (b) what we can do to be more flexible about this (if feasible).

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Stuck.20at.20Importing.20FSH.20text

mint-thompson:
Fixes #681 and completes task [CIMPL-604](https://standardhealthrecord.atlassian.net/browse/CIMPL-604).

By placing the lexer rule for block comments ahead of the lexer rule for sequences, a whitespace-free block comment is no longer interpreted as a sequence.


julianxcarter (dismissed):
This fixed the problem for me, nice work!

cmoesel (approved):
Looks good!  I wanted to be sure it didn't unintentionally break some other use case, so I used my fancy new regression script to test it against 52 FSH projects and there were no regressions.  Nice!

PR #690, opened 2020-12-04
Specify "unit" field on Quantities

### LINKED ISSUE #652: Cannot easily specify "unit" on Quantity
We support an easy syntax for specifying a Quantity when the units are taken from UCUM:
```
* valueQuantity = 155.0 '[lb_av]'
```
But there is no way to specify the `unit` field when doing this. It might make sense to support a syntax like:
```
* valueQuantity = 155.0 '[lb_av]' "lb"
```
where the `valueQuantity.unit` would then get set to "lb". The `unit` is meant to be a human readable value, and there is plenty of precedent for setting human readable values as strings at the end of assignments, so I don't think users would find this a confusing addition.

This came out of a conversation with @markkramerus on creating an instance of the [bodyweight](https://www.hl7.org/fhir/bodyweight.html) profile, which does require the `unit` keyword. Wanted to put this out here for discussion, and if we think this is a good idea, I'll add it to the backlog for SUSHI and to JIRA for an update to the FSH spec.

jafeltra:
This PR fixes #652. It adds the ability to specify a human-readable "unit" field on Quantities by specifying a string after the provided unit. For example:

```
* valueQuantity = 155.0 '[lb_av]' "lb"
```

sets `valueQuantity.unit`. This is similar to the display on Codings.


mint-thompson (approved):
This looks just fine to me! I've added https://standardhealthrecord.atlassian.net/browse/CIMPL-606 to the backlog in order to capture these changes in GoFSH.

cmoesel:
I don't remember.  Did we ever run this by @markkramerus? Mark -- I assume you're OK w/ this addition to the grammar?

jafeltra:
I just talked to @markkramerus and he said he was good with making these changes!

julianxcarter (approved):
Looks good to me!

markkramerus:
I also updated the FHIR Shorthand doc, just so you know.

PR #700, opened 2020-12-18
Soft indexing

julianxcarter:
Addresses [CIMPL-572](https://standardhealthrecord.atlassian.net/browse/CIMPL-572?atlOrigin=eyJpIjoiZmU0MWJlNzVhMzNkNDkxMWFjOGU2MTRjN2IxOWRlOTQiLCJwIjoiaiJ9), introducing soft indexing as a post-1.0 feature. For example:

This...
```
* rest[serverSlice].operation[0].name = "mcode-patient-bundle"
* rest[serverSlice].operation[0].definition = "http://hl7.org/fhir/us/mcode/OperationDefinition/mcode-patient-everything"
* rest[serverSlice].operation[0].extension.url = "http://hl7.org/fhir/StructureDefinition/capabilitystatement-expectation"
* rest[serverSlice].operation[0].extension.valueCode = #SHALL
```

Can now be rewritten as...

```
* rest[serverSlice].operation[+].name = "mcode-patient-bundle"
* rest[serverSlice].operation[=].definition = "http://hl7.org/fhir/us/mcode/OperationDefinition/mcode-patient-everything"
* rest[serverSlice].operation[=].extension.url = "http://hl7.org/fhir/StructureDefinition/capabilitystatement-expectation"
* rest[serverSlice].operation[=].extension.valueCode = #SHALL
```
In the current implementation, a `common.ts` file has been created within the `export` folder. This new file contains the `resolveSoftIndexing` helper function, which takes an array of `Rule`s as an argument and converts the soft indexing characters within the path to the corresponding index number. The Exporter classes import this `resolveSoftIndexing` helper function, and call it immediately prior to applying rules. Test cases for soft indexing have been added to `InstanceExporter.test.ts`, although more can be added to the other Exporter test files if we feel that's appropriate.


cmoesel (changes requested):
Exciting stuff.  I know this is a draft PR, but here is some early feedback to consider.

src/fhirtypes/common.ts
```diff
@@ -428,6 +429,7 @@ export function applyInsertRules(
   seenRuleSets: string[] = []
 ): void {
   const expandedRules: Rule[] = [];
+  resolveSoftIndexing(fshDefinition.rules);
```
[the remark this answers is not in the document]
cmoesel: I think that this `resolveSoftIndexing` call still needs to be removed -- and ideally we should have a test case like my snippet above to confirm that it's working as expected w/ RuleSets.

src/utils/PathUtils.ts
```diff
@@ -0,0 +1,127 @@
...
+ * Replaces soft indexs in rule paths with corresponding numbers
+ * @param {Rule[]} rules - An array of Rules
+ */
+export function resolveSoftIndexing(rules: Rule[]): void {
+  const pathMap: Map<string, number> = new Map();
+  const caretPathMap: Map<string, Map<string, number>> = new Map();
+
+  // Parsing and separating rules by base name and bracket indexes
+  const parsedRules = rules.map(rule => {
+    const parsedPath: any = { path: parseFSHPath(rule.path) };
```
cmoesel, line 61: I think you can actually type this more specifically than `any`:
```ts
    const parsedPath: { path: PathPart[], caretPath?: PathPart[] } = { path: parseFSHPath(rule.path) };
```
Then you'll get some of the type-safety features of having a strongly typed object.

src/utils/PathUtils.ts
```diff
@@ -0,0 +1,127 @@
...
+      }
+    });
+    rule.path = assembleFSHPath(rule.path); // Assembling the separated rule path back into a normal string
+
+    rule.caretPath?.forEach((element: PathPart) => {
+      // Caret path indexes should only be resolved in the context of a specific path
+      // Each normal path has a separate map to keep track of the caret path indexes
+      if (!caretPathMap.has(rule.path)) {
+        const caretMap = new Map();
+        caretPathMap.set(rule.path, caretMap);
```
cmoesel, line 97: Not a big deal, but why not just simplify to:
```ts
        caretPathMap.set(rule.path, new Map());
```

src/utils/PathUtils.ts
```diff
@@ -0,0 +1,127 @@
...
+      }
+
+      const workingPathMap = caretPathMap.get(rule.path);
```
cmoesel, line 100: Can we come up w/ a better name than `workingPathMap`?  Maybe `elementCaretPathMap`?

src/utils/PathUtils.ts
```diff
@@ -0,0 +1,127 @@
...
+            const newIndex = workingPathMap.get(element.base) + 1;
+            element.brackets[index] = newIndex.toString();
+            workingPathMap.set(element.base, newIndex);
+          } else if (bracket === '=') {
+            const currentIndex = workingPathMap.get(element.base);
+            element.brackets[index] = currentIndex.toString();
+          }
+        });
+      }
+    });
```
cmoesel, line 118: Is this code exactly the same as lines 72-89, except it's working on `workingPathMap` instead of `pathMap`?  If so, can we extract it into a separate private function and just call it from both places (passing in the respective map as appropriate)?

src/utils/PathUtils.ts
```diff
@@ -0,0 +1,127 @@
...
+    const parsedPath: any = { path: parseFSHPath(rule.path) };
+    // If we have a CaretValueRule, we'll need a second round of parsing for the caret path
+    if (rule instanceof CaretValueRule) {
+      parsedPath.caretPath = parseFSHPath(rule.caretPath);
+    }
+    return parsedPath;
+  });
+
+  // Replacing Soft indexes with numbers
+  parsedRules.map(rule => {
```
cmoesel, line 70: Do you mean `parsedRules.forEach`?  Because this doesn't seem to be mapping to anything?

src/utils/PathUtils.ts
cmoesel: Oh, I see.  You're actually re-assigning these as strings later on.... In that case you could still add `| string` to the type info for each.  But I'm thinking maybe we can avoid that re-assignment so that this can stay as a single static type.  More in a later comment...

src/utils/PathUtils.ts
```diff
@@ -0,0 +1,127 @@
...
+            element.brackets[index] = newIndex.toString();
+            pathMap.set(element.base, newIndex);
+          } else if (bracket === '=') {
+            const currentIndex = pathMap.get(element.base);
+            element.brackets[index] = currentIndex.toString();
+          }
+        });
+      }
+    });
+    rule.path = assembleFSHPath(rule.path); // Assembling the separated rule path back into a normal string
```
cmoesel, line 90: `rule.path` _was_ a `PathPart[]` and now you're changing it to a `string`.  Then later on (lines 122-126) you assign this to the original rule.  I'm wondering if you can avoid this by adding an `index` to the `forEach` on line 71, then you could just do `rules[index] = assembleFSHPath(rule.path)` here.

src/utils/PathUtils.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    rule.caretPath?.forEach((element: PathPart) => {
+      // Caret path indexes should only be resolved in the context of a specific path
+      // Each normal path has a separate map to keep track of the caret path indexes
+      if (!caretPathMap.has(rule.path)) {
```
cmoesel, line 95: If you switch to using `rules[index]` to hold the assembled path (as suggested above), then you'll need to change this (and a few other spots) to also refer to that (instead of `rule.path`).

src/utils/PathUtils.ts
```diff
@@ -0,0 +1,127 @@
...
+            element.brackets[index] = newIndex.toString();
+            workingPathMap.set(element.base, newIndex);
+          } else if (bracket === '=') {
+            const currentIndex = workingPathMap.get(element.base);
+            element.brackets[index] = currentIndex.toString();
+          }
+        });
+      }
+    });
+    if (rule.caretPath) rule.caretPath = assembleFSHPath(rule.caretPath);
```
cmoesel, line 119: Similar to `rule.path`, I'd consider setting `rules[index].caretPath` here instead.

src/utils/PathUtils.ts
```diff
@@ -0,0 +1,127 @@
...
+  });
+
+  // Converting individual elements into whole rule paths
+  rules.forEach((rule, index) => {
+    rule.path = parsedRules[index].path;
+    if (rule instanceof CaretValueRule) rule.caretPath = parsedRules[index].caretPath;
+  });
```
cmoesel, line 126: If you assign the `rules` individually in the loops above, then you shouldn't need this anymore.

test/export/InstanceExporter.test.ts
```diff
@@ -916,6 +916,51 @@ describe('InstanceExporter', () => {
...
+      patientInstance.rules.push(assignedValRule3);
+      const exported = exportInstance(patientInstance);
+      expect(exported.name).toEqual([
+        {
+          given: ['John', 'Johnny'],
+          family: 'Johnson'
+        }
+      ]);
+    });
+
```
cmoesel, line 963: I think it would be good to have some more tests.  Specifically:
* a test w/ rulesets (as previously noted)
* a test (or two) w/ caret rules (since there is some significant logic there)
* tests to ensure soft indexing gets applied in each required case: Profiles, Extensions, ValueSets, CodeSystems, Instance

cmoesel (changes requested):
It's looking really good. I like the refactor and I'm glad you found the `assembleFSHPaths` function.  I forgot about that one!

Please see below for some additional feedback!

src/fhirtypes/common.ts
```diff
@@ -448,7 +449,7 @@ export function applyInsertRules(
       ruleSet.rules.forEach(ruleSetRule => {
         // On the import side, a rule that is intended to be a ValueSetConceptComponent can
         // be imported as a ConceptRule because the syntax is identical. If this is the case,
-        // create a ValueSetConceptComponent that corresponds to the ConceptRule, and use that
+        // create a ValueSetConceptComponent that corr esponds to the ConceptRule, and use that
```
mint-thompson, line 452: how did this extra space sneak in here. such mischief from the space character.

test/export/InstanceExporter.test.ts
mint-thompson: In addition to these, I would like to see a test that includes both hard and soft indexing, assuming such a thing is allowed.

mint-thompson (changes requested):
Most of my suggestions were already pointed out by @cmoesel , so I only have minor comments to add.

mint-thompson (changes requested):
Things are looking good! I think that once you add in test cases for applying RuleSets that have soft-indexed paths in their rules, this will be ready to go.

src/utils/PathUtils.ts
```diff
@@ -0,0 +1,132 @@
...
+    }
+    return parsedPath;
+  });
+
+  // Replacing Soft indexes with numbers
+  parsedRules.forEach((parsedRule, index) => {
+    const originalRule = rules[index];
+    parsedRule.path.forEach((element: PathPart) => {
+      convertSoftIndexes(element, pathMap);
+    });
```
cmoesel, line 113: I don't think this approach is quite right, because it's calling `convertSoftIndexes` with only the `PathPart` (any portion of the path between `.`).  So, for example, for the path `foo[+].bar[+]`, `convertSoftIndexes` will first get called w/ `foo[+]` and then `bar[+]` -- and each will be the key in the shared map.  In this example, it happens to work, but consider a path that has nested items w/ the same name: `item[+].item[+]`.  That should convert to `item[0].item[0]` (since `item` and `item[0].item` are distinct arrays), but instead gets converted to `item[0].item[1]`.

I think the map needs the prefixes of the pathParts too (and the prefixes need to be already resolved).  So in my first example case, it should be tracking `foo` and `foo[0].bar` in the map.

cmoesel (changes requested):
I found a slight issue in how the indexes are being stored and referenced in the map.  See my individual comment for details.

I also think it would be easier (and perhaps better) to test the intricacies of `resolveSoftIndexing` in a `PathUtils.test.ts` file that just tests that actual function rather than testing all the intricacies in the separate exporters.  Then the tests in the exporters would be pretty simple: they'd just need to be enough to ensure that the soft indexing resolver is getting called.

In order to confirm my suspicion about the path map problem, I created a unit test for it.  So here is a head start on a `PathUtils.test.ts` file (for which the 2nd test _fails_).

```ts
import { Rule } from '../../src/fshtypes/rules';
import { resolveSoftIndexing } from '../../src/utils';

describe('PathUtils', () => {
  describe('#resolveSoftIndexing', () => {
    it('should resolve simple (non-nested) soft indexing', () => {
      const rules = ['name[+]', 'name[=]', 'name[=]', 'name[+]', 'name[=]'].map(r => new Rule(r));
      resolveSoftIndexing(rules);
      expect(rules.map(r => r.path)).toEqual([
        'name[0]',
        'name[0]',
        'name[0]',
        'name[1]',
        'name[1]'
      ]);
    });

    it('should resolve nested soft indexing w/ recursive paths', () => {
      // These types of paths are possible in instances of Questionnaire
      const rules = [
        'item[+].item[+].item[+]',
        'item[=].item[+].item[+]',
        'item[=].item[=].item[+]',
        'item[=].item[=].item[=]'
      ].map(r => new Rule(r));
      resolveSoftIndexing(rules);
      expect(rules.map(r => r.path)).toEqual([
        'item[0].item[0].item[0]',
        'item[0].item[1].item[1]',
        'item[0].item[1].item[2]',
        'item[0].item[1].item[2]'
      ]);
    });
  });
});
```

src/utils/PathUtils.ts
julianxcarter: You're totally right. This is something that occurred to me as a potential edge case, but I wasn't sure if recursive paths where allowed in the FHIR spec. I'll update this function to support edge cases such as these. Also, I think I separate `PathUtils` test file makes a lot of sense!

src/utils/PathUtils.ts
```diff
@@ -0,0 +1,141 @@
...
+/**
+ *
+ * @param {PathPart} element - A single element in a rules path
+ * @param {Map<string, number} pathMap - A map containing an element's name as the key and that element's updated index as the value
+ */
+function convertSoftIndexes(element: PathPart, pathMap: Map<string, number>) {
+  // Must account for a pathPart's base name, it's position in the path, as well as any slices it's contained in.
+  const mapName = element.slices
+    ? `${element.base}${element.slices.join('|')}${element.pathPosition}`
+    : `${element.base}${element.pathPosition}`;
```
cmoesel, line 77: Using the position in the path is clever, but unfortunately is still not unique enough.  For example, consider `ValueSet` which can have the following two possible paths:
```
compose.include[+].concept[+]
compose.exclude[+].concept[+]
```
The depth of `concept[+]` is `1` in both paths, but they reference unique arrays, so need to be tracked separately.  I think you're going to need to index on the resolved path prefix (e.g., `compose.include[0]` and `compose.exclude[0]`) instead of the depth.  If you did that, then the above could be something like:
```ts
const mapName = `${element.prefix ?? ''}.${element.base}|${(element.slices ?? []).join('|')}`
```

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -1989,6 +1989,46 @@ describe('StructureDefinitionExporter', () => {
...
+      slicingPathLine,
+      slicingRulesLine,
+      containsAddressLine,
+      apartmentNameCard
+    );
+
+    exporter.exportStructDef(testPatient);
+    const sd = pkg.profiles[0];
+    const apartmentNameElement = sd.findElement('Patient.address.line:ApartmentName');
+    expect(apartmentNameElement).toBeDefined();
```
cmoesel, line 2029: I think it would be more clear that you're testing the soft indexing if you checked the values of `apartmentNameElement.slicing.discriminator[0].type` and `apartmentNameElement.slicing.discriminator[0].path`.

cmoesel (changes requested):
I hate to be the bearer of bad news, but this still won't work exactly right in all cases.  I've commented individually below, but here is a test case that demonstrates the remaining issue:

```ts
    it('should resolve nested soft indexing w/ similar nested path items', () => {
      // These types of paths are possible in instances of ValueSet
      const rules = [
        'compose.include[+].concept[+]',
        'compose.include[=].concept[+]',
        'compose.exclude[+].concept[+]',
        'compose.exclude[=].concept[+]'
      ].map(r => new Rule(r));
      resolveSoftIndexing(rules);
      expect(rules.map(r => r.path)).toEqual([
        'compose.include[0].concept[0]',
        'compose.include[0].concept[1]',
        'compose.exclude[0].concept[0]',
        'compose.exclude[0].concept[1]'
      ]);
    });
```

test/export/StructureDefinitionExporter.test.ts
julianxcarter: Addressed in the most recent commit! The `slicing` property was held in `Patent.address.line` element, rather than in the sliced element itself.

src/utils/PathUtils.ts
julianxcarter: This solution worked well, and I believe it's pretty air tight in regards to edge cases. Thank you for the suggestion!

cmoesel (changes requested):
Cool!  I made a few comments with minor suggestions.

I also tested this manually, in combination with rulesets and ran into two issues.

**Normal RuleSets and Soft Indexing**

If you run SUSHI on this FSH:
```
RuleSet: NameHimBob
* name[+].given[0] = "Bob"

Instance: Bob
InstanceOf: Patient
* insert NameHimBob
```
everything is fine.  But if you run it on this instead:
```
RuleSet: NameHimBob
* name[+].given[0] = "Bob"

Instance: Bob
InstanceOf: Patient
* insert NameHimBob
* insert NameHimBob
```
you _should_ get an instance w/ two items in the `name` array (each having given name "Bob"), but you actually get an unexpected error instead:  
> error SUSHI encountered the following unexpected error: Cannot convert undefined or null to object

**Parameterized RuleSets and Soft Indexing**

If you run SUSHI on this FSH:
```
RuleSet: Name(first, last)
* name[+].given[0] = {first}
* name[=].family = {last}

Instance: ManyNamedPatient
InstanceOf: Patient
* insert Name("Robert", "Smith")
* insert Name("Bob", "Smith")
* insert Name("Robert", "Smith")
```
you get this result:
```json
{
  "resourceType": "Patient",
  "id": "ManyNamedPatient",
  "name": [
    {
      "given": ["Robert"],
      "family": "Smith"
    },
    {
      "given": ["Bob"],
      "family": "Smith"
    }
  ]
}
```
Note that there are only two names in the output, but there should be *three*.  Here is my guess as to what is going on:
* Expanded parameterized rulesets (e.g., w/ arguments applied) are stored in a map that uses the ruleset name _and argument values_ as the key.
* As such, the second `insert Name("Robert", "Smith")` references the _same_ expanded ruleset instance as the first `insert Name("Robert", "Smith")`.
* So what I'm guessing is happening is that when the first one is processed, the rule paths get rewritten with the actual indexes (e.g., `[0]` instead of `[+]`/`[=]`).  Then when the second one gets processed, it sees the rewritten rule paths (not the originals) -- so ends up basically re-applying the same exact thing (a no-op).

I have a hunch that maybe it will be as easy as doing a `deepClone` on the RuleSet rules when you expand the `insert` statements.  At least I hope so!  Since @mint-thompson is very familiar w/ RuleSets, I recommend you work with them on resolving the issue.

src/utils/PathUtils.ts
```diff
@@ -89,6 +86,8 @@ function convertSoftIndexes(element: PathPart, pathMap: Map<string, number>) {
       } else if (bracket === '=') {
         const currentIndex = pathMap.get(mapName);
         element.brackets[index] = currentIndex.toString();
+      } else if (indexRegex.test(bracket)) {
+        pathMap.set(mapName, parseInt(bracket));
```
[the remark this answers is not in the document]
julianxcarter: That's exactly why this was added! I'll add a comment for clarity.

test/export/CodeSystemExporter.test.ts
```diff
@@ -236,6 +236,37 @@ describe('CodeSystemExporter', () => {
...
 
+  it('should resolve soft indexing when applying Caret Value rules', () => {
+    const codeSystem = new FshCodeSystem('CaretCodeSystem');
+    const contactRule1 = new CaretValueRule('');
+    contactRule1.caretPath = 'contact[+].name';
+    contactRule1.value = 'Example Name';
+    codeSystem.rules.push(contactRule1);
+
+    const contactRule2 = new CaretValueRule('');
+    contactRule2.caretPath = 'contact[=].telecom[+].rank';
```
mint-thompson, line 247: What happens here if the caretPath is instead `contact[=].telecom[=].rank`? That is to say, the `telecom` part of the path is given a soft `=` before any are added with a `+` or a specific number. I would expect that I would get the same result, with the soft `=` being turned into a 0, since even though there aren't any defined elements, this is the first one. Do you think that a warning should be emitted when this is done? Or, do you think that this should be handled as an error?

src/utils/PathUtils.ts
```diff
@@ -0,0 +1,145 @@
...
+}
+
+/**
+ *
+ * @param {PathPart} element - A single element in a rules path
+ * @param {Map<string, number} pathMap - A map containing an element's name as the key and that element's updated index as the value
+ */
+function convertSoftIndexes(element: PathPart, pathMap: Map<string, number>) {
+  // Must account for a pathPart's base name, prior portions of the path, as well as any slices it's contained in.
+  const mapName = `${element.prefix ?? ''}.${element.base}|${(element.slices ?? []).join('|')}`;
```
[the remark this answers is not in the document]
julianxcarter: That's exactly what would happen. A simple base of `coding` would evaluate to `.coding|` in this expression, which is a little weird due to the extra characters but is generated consistently. This `mapName` variable is also only used as a unique key for that element within the map that keeps track of separate elements on a path.

test/export/CodeSystemExporter.test.ts
julianxcarter: Actually, the way things are set up now, the `=` wouldn't be converted to a number at all without a preceding `+`. I think that's probably something that we'd want to handle as a warning, while still converting the `=` to a 0 without a preceding `+`. I'll ask Chris and/or Mark about it in standup!

src/utils/PathUtils.ts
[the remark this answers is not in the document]
cmoesel: Right -- it doesn't matter if the string looks weird, as long as it is built in a consistent and repeatable way that will be unique for a given set of inputs (but reliably the same when the inputs are the same).  So I think we're OK.

test/export/CodeSystemExporter.test.ts
cmoesel: I agree -- if we run into a `[=]` and there was no preceding `[+]` or `[n]`, then we should convert to `[0]` but issue a warning (since it may be a mistake and should be fixed anyway).

src/fhirtypes/common.ts
```diff
@@ -486,7 +486,7 @@ export function applyInsertRules(
       logger.error(`Unable to find definition for RuleSet ${rule.ruleSet}.`, rule.sourceInfo);
     }
   });
-  fshDefinition.rules = expandedRules;
+  fshDefinition.rules = cloneDeep(expandedRules);
```
cmoesel, line 489: I don't think we need to clone the whole `rules` array, as many of those rules don't come from the RuleSet.  I think it is only the rules from the RuleSet that need to be cloned (probably on line 477 above).

julianxcarter: You're right! I removed the unnecessary copying in the most recent commit, wouldn't want to waste memory.

cmoesel:
So... it turns out my hunch was partially correct, but when I tested this again, my second use case above (with Robert Smith) was still not working correctly.  It turns out there were multiple things going on.  In this case, the third `insert` rule (which was exactly the same as the first) was getting _dropped_.  My commit 533bafc fixes this.  Can @julianxcarter or @mint-thompson review that before we merge?

cmoesel (approved):
I think we're ready for prime time!

mint-thompson (approved):

mint-thompson:
thank you @cmoesel for fixing my mistake in the importer. hooray for soft indexing!!

julianxcarter:
Nice catch, @cmoesel! I can both confirm the issue and also confirm the fix in your commit, thank you for the fix!

PR #701, opened 2020-12-21
Export Parameterized RuleSet

mint-thompson:
Completes task [CIMPL-574](https://standardhealthrecord.atlassian.net/browse/CIMPL-574).

These changes accompany #687 so that insert rules with parameters are applied correctly during the export process. So, the key function being updated is `applyInsertRules` in `fhirtypes/common.ts`. The logic for fishing up a non-parameterized RuleSet  and an application of a parameterized RuleSet are different: a non-parameterized RuleSet can be found by the `name` attribute on the RuleSet, but an application of a parameterized RuleSet must be found by using the name and parameters to form a valid key for the `FSHDocument`'s `appliedRuleSets` map. Thus, a separate fishing function is used when fishing for an applied parameterized RuleSet. Similarly, the logic for detecting circular RuleSet references is updated to track by a combination of RuleSet name and parameter values.

Once the RuleSet is fished up, the logic for applying its contents is unchanged.


cmoesel (changes requested):
I really like that this barely required any new code.  Very nice.

The code looks good, and when I manually tested this using the example from our Wicked FSH slides (modified a bit since we don't have soft indexing yet), it worked well.  The one issue I ran into was when there was an error in the RuleSet, the logged location wasn't quite right.

For example, if you use the following FSH file:
```
Profile: MyPatientProfile
Parent: Patient
Description: "I exist just to show the line offset issue"
* name 1..*

RuleSet: Name(i, given, family)
* name[{i}].given[0] = {given}
* name[{i}].family = {family}

Instance: MyPatient
InstanceOf: MyPatientProfile
* insert Name(0, "William", "Jones")
* insert Name(1, "Billybob", 456)
```
You will get back the following error:
> error Cannot assign number value: 456. Value does not match element type: string
>  File: /Users/cmoesel/dev/tmp/TestParamRS/input/fsh/testParamRS.fsh
>  Line: 3
>  Applied in File: /Users/cmoesel/dev/tmp/TestParamRS/input/fsh/testParamRS.fsh
>  Applied on Line: 13

Notice that the first `File` and `Line` point to the wrong location.  I think we discussed this before and determined that we would need to adjust.  Valid options are:
* Have `File`/`Line` point to the _range_ of the original (unexpanded) `RuleSet` definition in the source file _(my preferred approach, if possible)_
* Remove the `File` and `Line` from the log statement entirely
* Something else?

One potential idea if option one above is not possible and option 2 is not enough, would be to update `Location` to allow a `RuleSet` _name_ to be supplied.  Then, in this case, we could say something like:
> error Cannot assign number value: 456. Value does not match element type: string
>  RuleSet: Name
>  Applied in File: /Users/cmoesel/dev/tmp/TestParamRS/input/fsh/testParamRS.fsh
>  Applied on Line: 13

cmoesel (approved):
Yes!  I dig it.  And the error line sourcing is perfect... so helpful!  I also tested it by splitting the ruleset to a different file and all the file names and line numbers reported were still correct.  Thanks, @mint-thompson!

julianxcarter (approved):
This all looks great to me! It works well, and the test cases all make a lot of sense to me, so I'm approving. Great work Mint!

PR #704, opened 2020-12-21
Allow any configured page

### LINKED ISSUE #696: SUSHI should allow artifacts.html to be listed in pages
SUSHI currently checks that every page listed in the `pages` entry in `sushi-config.yaml` actually exists.  But authors may want to list some pages that the IG Publisher generates (such as artifacts.html).  We should either update SUSHI to allow references to _any_ pages (without checking for existence) or at least allow for certain known-to-be-generated pages.

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/pages

jafeltra:
This PR fixes #696. I took the approach of letting SUSHI reference _any_ page in the configuration because I thought that was the latest approach we talked about. However, if we'd prefer this implementation to only allow certain known-to-be-generated pages and check the existence for others, I can update it.


cmoesel (approved):
This looks good!  I tested it manually and it works as expected!

mint-thompson (approved):

PR #707, opened 2020-12-30
Relax required slice validation

ngfreiter:
This addresses [CIMPL-618](https://standardhealthrecord.atlassian.net/browse/CIMPL-621), and also https://github.com/FHIR/GoFSH/issues/57 by relaxing the validation we do on required slices. Previously, when SUSHI was validating the cardinality of slices, it would not count things that had a numerical index in the FSH. For example:
```
Instance: ctc-adverse-event-example-1
InstanceOf: CTCAdverseEvent
Usage: #example
* extension[0].url = "http://hl7.org/fhir/us/mcode/StructureDefinition/ctcae-grade"
* extension[0].valueCodeableConcept = $ctcae-grade-code-system#2 "Moderate Adverse Event"
```
the extension at `extension[0]` would not satisfy the cardinality of a `ctae-grade` extension, which was `1..1`. If instead the definition looked like this:
```
Instance: ctc-adverse-event-example-1
InstanceOf: CTCAdverseEvent
Usage: #example
* extension[grade].url = "http://hl7.org/fhir/us/mcode/StructureDefinition/ctcae-grade"
* extension[grade].valueCodeableConcept = $ctcae-grade-code-system#2 "Moderate Adverse Event"
```
SUSHI would be satisfied, since it knows which slice is being defined.

This check was overly strict, but we are likely not interested in the business of deciding which slice is being defined, so this PR takes the simple approach of just not validating cardinality of a slice if that slice _could_ be referred to by a numerically indexed path. So in the example above, `extension[0]` could be referring to any extension, so we don't validate cardinality for any required extensions.

### DIFF
--- a/src/export/InstanceExporter.ts
+++ b/src/export/InstanceExporter.ts
@@ -156,6 +156,16 @@ export class InstanceExporter implements Fishable {
       }
       // Recursively validate children of the current element
       if (Array.isArray(instanceChild)) {
+        // If the child is a slice, and there are no array elements with matching slice names
+        // but there are array elements that could match (since they use numerical indices)
+        // we can go no further in validation, since we can't know which slice is represented
+        if (
+          child.sliceName &&
+          !instanceChild.find((arrayEl: any) => arrayEl?._sliceName === child.sliceName) &&
+          instanceChild.find((arrayEl: any) => !arrayEl?._sliceName)
+        ) {
+          return;
+        }
         // Filter so that if the child is a slice, we only count relevant slices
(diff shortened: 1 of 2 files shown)


mint-thompson (approved):
This looks fine to me! 🎊

cmoesel (approved):
Looks good to me.  The only thing I'll note is that for the specific use case you mentioned (_extensions_), we may want to put in a special case check since it's fairly easy to identify what slice an extension belongs to (by its URL).  But that's not the responsibility of this PR.

Issue #688, opened 2020-12-04
Canonical keyword does not work w/ inline contained resources having no `url`


cmoesel:
`Canonical(MyInstance)` does not work when `MyInstance` is an inline contained resource and does not have a `url`.  For example, the following fails:

```
Instance: MyValueSetInstance
InstanceOf: ValueSet
* status = #active
* compose.include.system = "http://foo.org/"
* compose.include.concept[0].code = #bar
* compose.include.concept[1].code = #baz

Instance: MyQuestionnaireInstance
InstanceOf: Questionnaire
* contained[0] = MyValueSetInstance
* status = #active
* item.linkId = "A"
* item.type = #choice
* item.answerValueSet = Canonical(MyValueSetInstance)
```

> error Cannot use canonical URL of MyValueSetInstance because it does not exist. Be sure that MyValueSetInstance exists and it has a URL.

Contained canonical fragments should be supported, as the FHIR spec uses an example just like this:
https://www.hl7.org/fhir/references.html#canonical-fragments

See also: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/example.20of.20a.20FSH.20Instance.20with.20contained.20resources/near/218773957

PR #709, opened 2020-12-31
Support canonical fragments

ngfreiter:
Fixes https://github.com/FHIR/sushi/issues/688.

This supports using fragments of the form `#{id}` when the `Canonical` keyword is used for an inline instance that does not have an absolute URL. The issue description provides a good reference to test this on, but the first instance in that example must be set to `Usage: #inline` for this to work. I wasn't sure if this should only work on inline instances, or all instances, so let me know if you think I approached that wrong here.

### DIFF
--- a/src/export/Package.ts
+++ b/src/export/Package.ts
@@ -59,6 +59,8 @@ export class Package implements Fishable {
       const metadata: Metadata = {
         id: result.id,
         name: result instanceof InstanceDefinition ? result._instanceMeta.name : result.name,
+        instanceUsage:
+          result instanceof InstanceDefinition ? result._instanceMeta.usage : undefined,
         url: result.url
       };
       if (result instanceof StructureDefinition) {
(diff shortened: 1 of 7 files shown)


jafeltra (commented):
This looks good! From what I understand in the documentation, it seems like this makes sense to have just for inline instances.

The only thing that I noticed is that there seem to be similar tests for applying assignment rules with Canonical in `StructureDefinitionExporter.test.ts`. Do you think it makes sense to add an equivalent test for an inline instance there as well?

ngfreiter:
@jafeltra, I've added that test here: https://github.com/FHIR/sushi/pull/709/commits/00bf538b776e65f279f574e064414eb3543bc3ec.

jafeltra (approved):
Looks good to me! Thanks for adding that test!

mint-thompson (approved):

PR #710, opened 2021-01-04
Fix multiline string minimum indent detection

### LINKED ISSUE #708: Multiline strings losing letters
I've noticed that it seems like the leading letters in multiline strings can still get cut off in an unexpected way. I thought we had addressed this before, but maybe I am remembering incorrectly. For example, with the following definition:
```
Profile: CancerRelatedMedicationRequest
Parent: USCoreMedicationRequestProfile
Id: mcode-cancer-related-medication-request
Title: "Cancer-Related Medication Request"
Description: """A record of a medication prescription, administration, or consumption used to treat cancer. The medication may reported by the prescriber, prescribing organization, or patient. It does not have to be directly observed.

Conformance statement:

 MedicationRequest resources associated with an mCODE patient and associated with cancer treatment MUST conform to this profile. Beyond this requirement, a producer of resources SHOULD ensure that any resource instance associated with an mCODE patient that would reasonably be expected to conform to this profile (i.e., a prescription, administration, or medication exposure) SHOULD be published in this form."""
```
The resulting string is:
```
" record of a medication prescription, administration, or consumption used to treat cancer. The medication may reported by the prescriber, prescribing organization, or patient. It does not have to be directly observed.\n\nonformance statement:\n\nMedicationRequest resources associated with an mCODE patient and associated with cancer treatment MUST conform to this profile. Beyond this requirement, a producer of resources SHOULD ensure that any resource instance associated with an mCODE patient that would reasonably be expected to conform to this profile (i.e., a prescription, administration, or medication exposure) SHOULD be published in this form."
```
Notice that the "A" to start the description and the "C" in "Conformance are both cut off. I know we do some processing to remove leading spaces, and the "MedicationRequest resources..." line is unnecessarily indented one space, so I'm guessing that is what causes this, but still even if that space is likely a typo this behavior seems unexpected.

mint-thompson:
Fixes #708 and completes task [CIMPL-626](https://standardhealthrecord.atlassian.net/browse/CIMPL-626).

When the last line of a multiline string was the only one with leading spaces, it was detected as having the minimum indentation, causing loss of data on previous lines. Use a min function instead of comparing with zero in order to determine the minimum indentation.

### DIFF
--- a/src/import/FSHImporter.ts
+++ b/src/import/FSHImporter.ts
@@ -59,6 +59,7 @@ import {
 import isEqual from 'lodash/isEqual';
 import sortBy from 'lodash/sortBy';
 import upperFirst from 'lodash/upperFirst';
+import _min from 'lodash/min';
 import { parseCodeLexeme } from './parseCodeLexeme';
 import { EOL } from 'os';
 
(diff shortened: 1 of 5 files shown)


ngfreiter (approved):
Yep, this looks good to me!

jafeltra (approved):
These changes make sense to me!

PR #711, opened 2021-01-04
copyrightYear changed to 2021

julianxcarter:
Simple PR with a fix for any tests involving a `config` object. Now that we've crossed over into the new year, config files are created with a `copyrightYear` value of `2021+` rather than `2020+`. This PR updates all of the config fixtures to reflect this change.


mint-thompson (approved):

jafeltra (approved):
Makes sense!

Issue #703, opened 2020-12-21
Confusing error message on incorrect assignment rule


cmoesel:
If someone tries to assign `{URL}` or `{URL}|{version}` to an element representing a `CodeableConcept`, SUSHI gives an error.

For example:
```
* event = http://terminology.hl7.org/CodeSystem/MDRAE|v777
```
gives the error:
> error Cannot find definition for Instance: http://terminology.hl7.org/CodeSystem/MDRAE|v777. Skipping rule.

and:
```
* event = http://terminology.hl7.org/CodeSystem/MDRAE
```
gives the error:
> error Cannot find definition for Instance: http://terminology.hl7.org/CodeSystem/MDRAE. Skipping rule.

Oddly, if you use an alias, then it gives a more reasonable answer:
```
Alias: MDR = http://terminology.hl7.org/CodeSystem/MDRAE
...
* event = MDR
```
produces:
> error Cannot assign string value: http://terminology.hl7.org/CodeSystem/MDRAE. Value does not match element type: CodeableConcept

Ideally the other errors should also be more helpful.

Reported by @markkramerus

PR #714, opened 2021-01-06
Log less confusing error on inline instance assignment

ngfreiter:
Fixes https://github.com/FHIR/sushi/issues/703 by slightly clarifying the error message we log when someone assigns an inline instance. I don't think there is a clear way here to guess the intent of the user. Even if the element is of type CodeableConcept, the user could be trying to assign an inline instance of a CodeableConcept to that element. So all this PR does is log the type of the element being assigned onto, if it is possible. I think this should potentially help people understand what is going wrong when they assign a value.

If this doesn't seem sufficient, and someone sees a better way to log this message and/or guess the intent of the user, let me know.

### DIFF
--- a/src/export/StructureDefinitionExporter.ts
+++ b/src/export/StructureDefinitionExporter.ts
@@ -162,7 +162,15 @@ export class StructureDefinitionExporter implements Fishable {
               const instanceExporter = new InstanceExporter(this.tank, this.pkg, this.fisher);
               const instance = instanceExporter.fishForFHIR(rule.value as string);
               if (instance == null) {
-                logger.error(`Cannot find definition for Instance: ${rule.value}. Skipping rule.`);
+                if (element.type?.length === 1) {
+                  logger.error(
+                    `Cannot assign Instance at ${rule.value} to element of type ${element.type[0].code}. Instance definition not found.`
+                  );
+                } else {
+                  logger.error(
+                    `Cannot find definition for Instance: ${rule.value}. Skipping rule.`
+                  );
+                }
                 continue;
               }
               rule.value = instance;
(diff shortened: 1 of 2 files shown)


ngfreiter:
Idk why but as of right now I can't add @jafeltra as a reviewer, so @jafeltra I'm not purposefully excluding you, please also review if you please!

src/export/StructureDefinitionExporter.ts
```diff
@@ -162,7 +162,15 @@ export class StructureDefinitionExporter implements Fishable {
               const instanceExporter = new InstanceExporter(this.tank, this.pkg, this.fisher);
               const instance = instanceExporter.fishForFHIR(rule.value as string);
               if (instance == null) {
-                logger.error(`Cannot find definition for Instance: ${rule.value}. Skipping rule.`);
+                if (element.type?.length === 1) {
+                  logger.error(
+                    `Cannot assign Instance at ${rule.value} to element of type ${element.type[0].code}. Instance definition not found.`
```
jafeltra, line 167: I feel like I find the phrase "Instance at ${rule.value}" a little confusing. I think "at" made me thing it would be referring to a path. Maybe we could say 

```
`Cannot assign Instance at path ${rule.path} to element of type ${element.type[0].code}. Definition not found for Instance: ${rule.value}.`
```

Although, now I can't tell if that's too wordy or if it even is clearer.

jafeltra (commented):
This looks good. I'm not 100% I know what the intention of this error message was going to be, but I was slightly confused by the update (but also confused by the old one too), so I added a potential suggestion. But feel free to disagree though.

src/export/StructureDefinitionExporter.ts
ngfreiter: I tried to clarify the error message in https://github.com/FHIR/sushi/pull/714/commits/61139c3c727b7b1bf5905c511a6d7cdb0ec64aa7. The idea of this message is just that having the type of the element is helpful for people who may be trying to assign a Codeable, but are accidentally using the syntax to assign an instance.

jafeltra (approved):
This makes sense to me. I think this wording is clearer and will hopefully help people out!

julianxcarter (approved):
This looks good to me!

Issue #702, opened 2020-12-21
SUSHI should allow ALIAS|Version


cmoesel:
In FHS, we use `|version` to indicate a `CodeSystem` version (much like in FHIR `canonical` datatype).  As such, the following should probably work:
```
Alias: MDR = http://terminology.hl7.org/CodeSystem/MDRAE
* event = MDR|v777#10002198 "Anaphylatic reaction"
```
but it does not:
> error Resolved value "MDR|v777" is not a valid URI.

Reported by @markkramerus

PR #715, opened 2021-01-06
Allow ALIAS|VERSION

ngfreiter:
This fixes https://github.com/FHIR/sushi/issues/702 by allowing an alias to have a version appended to it. For example, the following:
```
Alias: MDR = http://terminology.hl7.org/CodeSystem/MDRAE
* event = MDR|v777#10002198 "Anaphylatic reaction"
```
should now work. The `MDR` alias will resolve to `http://terminology.hl7.org/CodeSystem/MDRAE`, and the final `system` set on `event` will be `http://terminology.hl7.org/CodeSystem/MDRAE|v777`.

This PR also makes it so that an alias cannot include a `|` character, since that character is now reserved for indicating a version.

### DIFF
--- a/src/import/FSHImporter.ts
+++ b/src/import/FSHImporter.ts
@@ -1868,8 +1868,13 @@ export class FSHImporter extends FSHVisitor {
   }
 
   private aliasAwareValue(parentCtx: ParserRuleContext, value = parentCtx.getText()): string {
-    this.validateAliasResolves(parentCtx, value);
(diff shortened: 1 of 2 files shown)


ngfreiter:
@cmoesel, also happy to have you review this if you want, but this time I am unable to add you as a reviewer.

mint-thompson (approved):

jafeltra (approved):
Top notch!

PR #716, opened 2021-01-06
FshQuantity toString updates

jafeltra:
This PR updates `FshQuantity`'s `toString` method to include the unit display text. This was missing when the display text was added to the quantity syntax.

By adding tests for this method, I discovered that the `toString` method will not return an accurate string if the `system` on the `unit` FshCode is not UCUM. This isn't a problem for now for SUSHI because the only FshQuantities created by SUSHI use UCUM as the system. I talked with @cmoesel about this, and we realized that we will need to ensure that GoFSH is not creating FshQuantities with non-UCUM systems, and in the future, we may consider supporting regular codes on Quantities if the unit is not a UCUM unit. However, those changes are separate from this PR.

### DIFF
--- a/src/fshtypes/FshQuantity.ts
+++ b/src/fshtypes/FshQuantity.ts
@@ -9,7 +9,10 @@ export class FshQuantity extends FshEntity {
 
   toString(): string {
     let str = this.value.toString();
-    if (this.unit?.code != null) str += ` '${this.unit.code}'`;
+    if (this.unit?.code != null) {
+      str += ` '${this.unit.code}'`;
+      if (this.unit?.display != null) str += ` "${this.unit.display}"`;
+    }
     return str;
   }
 
--- a/src/fshtypes/FshQuantity.ts
+++ b/src/fshtypes/FshQuantity.ts
@@ -11,7 +11,7 @@ export class FshQuantity extends FshEntity {
     let str = this.value.toString();
     if (this.unit?.code != null) {
       str += ` '${this.unit.code}'`;
-      if (this.unit?.display != null) str += ` "${this.unit.display}"`;
+      if (this.unit.display != null) str += ` "${this.unit.display}"`;
     }
     return str;
   }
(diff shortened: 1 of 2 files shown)


src/fshtypes/FshQuantity.ts
```diff
@@ -9,7 +9,10 @@ export class FshQuantity extends FshEntity {
 
   toString(): string {
     let str = this.value.toString();
-    if (this.unit?.code != null) str += ` '${this.unit.code}'`;
+    if (this.unit?.code != null) {
+      str += ` '${this.unit.code}'`;
+      if (this.unit?.display != null) str += ` "${this.unit.display}"`;
```
mint-thompson, line 14: Very minor, but I think at this point, `this.unit` is guaranteed to be non-null, so the `?` isn't necessary.

mint-thompson (commented):
Just to make sure I understand what's going on regarding non-UCUM quantity units: SUSHI currently will always make a `FSHQuantity` with the unit's system set to UCUM when it encounters a rule in the format `* valueQuantity = 25 'mm'`. But, there are cases in goFSH where it might output a rule with that syntax even though the input to goFSH did not use UCUM as the unit's system. Is that correct?

src/fshtypes/FshQuantity.ts
jafeltra: Ah good point! I moved that around after I wrote it and didn't look very closely at it. I'll update!

jafeltra:
@mint-thompson your description above is correct. I found at least one spot in GoFSH where we create a `FSHQuantity` for a property that `endsWith('Quantity')` without checking if the system on that quantity is UCUM. However, in SUSHI, we only create a `FSHQuantity` in two places (as far as I can tell) and they either specify UCUM or don't include any unit. I believe @cmoesel said he was going to log an issue for GoFSH.

src/fshtypes/FshQuantity.ts
jafeltra: Should be addressed in 1c6738f.

ngfreiter (approved):
Looks good to me!

mint-thompson (approved):
With the issue logged in goFSH, this change set is good to go.

PR #719, opened 2021-01-11
Add Extension Description/Title to root element

ngfreiter:
This addresses [CIMPL-608](https://standardhealthrecord.atlassian.net/browse/CIMPL-608) by adding a configuration option, `applyExtensionMetadataToRoot`, that can be used to control if the `Description` and `Title` of an `Extension` are used to set the `definition` and `short` on the root element of the `Extension`. The flag defaults to true, so if users don't want this behavior, they will have to explicitly set the flag to false.

To test this, you can try running SUSHI with this new flag set to `true`, `false`, and leaving the flag unset. When the flag is `true` or unset, metadata from the `Extension` will get copied onto the root element (the element with path `Extension`). Setting the `short` or `definition` of that element directly with a `^` rule should override the automatic setting. When the flag is explicitly `false`, no metadata should be copied over. It may also be useful to test this all the way through with the IG Publisher, so you can see how the `short` and `definition` elements are used as @cmoesel describes in this [thread](https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Extension.20short.2C.20definition.2C.20and.20default.20values).

### DIFF
--- a/src/export/StructureDefinitionExporter.ts
+++ b/src/export/StructureDefinitionExporter.ts
@@ -98,6 +98,9 @@ export class StructureDefinitionExporter implements Fishable {
     structDef.setName(fshDefinition.name, fshDefinition.sourceInfo);
     if (fshDefinition.title) {
       structDef.title = fshDefinition.title;
+      if (fshDefinition instanceof Extension) {
+        structDef.elements[0].short = fshDefinition.title;
+      }
     } else {
       delete structDef.title;
     }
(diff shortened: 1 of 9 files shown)


ngfreiter:
Also, if you take a look at this PR, please take a look at the corresponding documentation update: https://github.com/FSHSchool/site/pull/36.

mint-thompson (approved):
A fine new configuration option. Hooray for configuration!

julianxcarter (approved):
This looks quite good to me!

Issue #718, opened 2021-01-07
SUSHI init should download scripts from new ig-publisher-scripts repo


cmoesel:
IG Publisher scripts will now be hosted at https://github.com/HL7/ig-publisher-scripts instead of the sample IG.

PR #720, opened 2021-01-11
Update the publisher scripts location

ngfreiter:
Fixes https://github.com/FHIR/sushi/issues/718 by updating `sushi --init` to look in the new location for the IG Publisher scripts, which is this URL: https://github.com/HL7/ig-publisher-scripts.

For testing, I think just running `sushi --init`, making sure the scripts download and can then be run to generate the simple example IG is sufficient.

### DIFF
--- a/src/utils/Processing.ts
+++ b/src/utils/Processing.ts
@@ -371,14 +371,14 @@ export async function init(): Promise<void> {
     path.join(outputDir, 'input', 'ignoreWarnings.txt')
   );
   // Add the _updatePublisher, _genonce, and _gencontinuous scripts
-  console.log('Downloading publisher scripts from https://github.com/FHIR/sample-ig');
+  console.log('Downloading publisher scripts from https://github.com/HL7/ig-publisher-scripts');
   for (const script of [
     '_genonce.bat',
     '_genonce.sh',
     '_updatePublisher.bat',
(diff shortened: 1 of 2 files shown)


cmoesel (changes requested):
GitHub must employ some magic to make this work, because that HL7 project does not have a `master` branch -- it uses the new `main` branch naming scheme: https://github.com/HL7/ig-publisher-scripts/branches

Let's avoid magic and go w/ `main`.

julianxcarter (dismissed):
Apart from Chris's comment about changing the branch name within the url, I'm gonna move forward and approve this since it's working well for me now (and I assume it will after the change). Nice work!

ngfreiter:
> GitHub must employ some magic to make this work, because that HL7 project does not have a `master` branch -- it uses the new `main` branch naming scheme: https://github.com/HL7/ig-publisher-scripts/branches
> 
> Let's avoid magic and go w/ `main`.

Fixed in https://github.com/FHIR/sushi/pull/720/commits/f35bee27611806d0f0425876df65f0359c2c27a2.

cmoesel (approved):
I initially found a bug in the scripts themselves, so I was waiting for HL7/ig-publisher-scripts#3 to get merged before approving and merging this.  It's merged so all looks good here too!

PR #721, opened 2021-01-11
Carry over extensions when expanding content references

### LINKED ISSUE #699: Content References should recursively inherit extensions
See https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Questionnaire.20item.20extensions/near/216143049
and https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/Clarification.20on.20contentReference/near/204963141

cmoesel:
This is one of those times that I over-complicated things, wrote a bunch of code that didn't work, then stepped back and found a simpler solution, which also didn't work, and then finally fixed that.  I have to admit, the regression checker also saved me from pushing up a solution that was majorly broken; so put this one in the win column for regression tests.

To manually test this, try this FSH (inspired by the Zulip thread):
```
Profile: AdvancedQuestionnaire
Parent: Questionnaire
* item.extension contains  minValue named minIntValue 0..1 MS

Instance: MyQuestionnaire
InstanceOf: AdvancedQuestionnaire
* status = #active
* item.linkId = "1"
* item.type = #group
* item.item.linkId = "1.1"
* item.item.type = #integer
* item.item.extension[minIntValue].valueInteger = 130
```

On the main branch, it will result in an error (path does not exist), but on this PR branch, it will succeed and the extension will be correctly generated in the JSON.

Fixes #699.


test/utils/PathUtils.test.ts
```diff
@@ -1,3 +1,4 @@
+import '../testhelpers/loggerSpy'; // side-effect: silences logs
```
jafeltra, line 1: It looks like git got confused when merging these changes. I don't think we need this line since it's also included at the bottom of the import list. I'll let you decide which place you like it better though!

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1699,27 +1699,59 @@ export class ElementDefinition {
...
+            // Now carry over the extension elements (and their children) from the profile
+            const elementsToCarryOver = this.structDef
+              ?.findElement(contentRefId)
+              ?.children()
+              ?.filter(e => {
+                return (
+                  // an extension of child of extension under the unrolled path
+                  /(^|\.)(modifierE|e)xtension([.:]|$)/.test(e.id.slice(contentRefId.length + 1)) &&
+                  // but not under a recursive content reference because that would be a real mess!
+                  !recursiveContentReferencePaths.some(p => e.path.startsWith(`${p}.`))
```
jafeltra, line 1740: I was trying to understand the case where we would filter out these recursive content reference paths, which led me to try taking out this line, which made me notice that it doesn't look like either test includes a check to ensure these are filtered out properly. Does it make sense to add that to one of the tests or add a new test? Or am is this too much of an edge case?

jafeltra (commented):
This looks good to me! I tested it out with the example you provided and I ran the regression script on 58 known FSH projects and only one had meaningful changes, and they all were what I would expect. I just had one minor suggestion and a question below.

ngfreiter (commented):
This looks good to me, all I noticed was what @jafeltra already said, so I will hold off on approving until those comments are addressed.

test/utils/PathUtils.test.ts
cmoesel: Hmmm... I kind of like it on line 1.  Plus that gives me more git cred.  But at the same time, @ngfreiter got his in first, so it would be kind of silly to have this PR do nothing in that file except move the import line so... I guess I'll let Nick's stay. :-(

src/fhirtypes/ElementDefinition.ts
cmoesel: I'll try to cover it in a test.

cmoesel:
OK, @jafeltra and @ngfreiter -- 36a5ab0 should address the comments.  @jafeltra -- you'll notice that if you uncomment that line, the test now fails.  I had a hard time wording the comment in the test, but hopefully it makes at least 51% sense?

ngfreiter (dismissed):
This made my head hurt, but I think the test makes sense and looks good to me!

jafeltra (approved):
Agreed, my head hurts, but I think (?) this makes sense. Thanks for adding it.

cmoesel:
I just reviewed the regression that @jafeltra found with `HL7NZ/cca#master`.  I don't think that's an intentional regression; I think something is wrong.  So... putting this back into _DRAFT_ until I figure it out.

cmoesel:
OK @jafeltra and @ngfreiter -- this is ready for re-review.  Since your last review, I made the following changes:
* If a contains rule has multiple items, and there is an error on one of them (e.g., duplicate slice name), it will continue to process the others.  Before this it threw an exception that was caught to high up the chain and skipped all the other items.
* If a non-inline extension contains rules redeclares a slicename with the same extension URL, just treat it as a warning since it's basically a no-op.  It's possible the contains rule has a different cardinality, but the normal cardinality rules will still get applied, so if that ends up being a problem, an error will be produced anyway.
* Regression script now reports the number of errors returned by each SUSHI invocation so you can notice if the error count changes.  This also required a change to `app.ts` so the exit code is now the number of errors encountered (instead of always `1` if there are errors, no matter how many errors).

cmoesel:
Oh, I forgot.  Also:
* If an extension is carried over from the source of a contentReference, it will now be reflected in the differential as well.  This is because the IG Publisher doesn't correctly produce the snapshot indicating carried over extensions unless we do so.

jafeltra (approved):
I'm still not 100% sure I understand everything that's happening here or what was originally going wrong with the HL7NZ/cca repo, but I think all the individual changes and their tests make sense. I also ran the regression on all the repos and confirmed there were no differences and all the logs looked the same for the few that had errors. So I'm approving!

ngfreiter (approved):
Looks good to me!

cmoesel:
NOTE: Based on a new understanding of the behavior of content references, much of this PR was _backed out_ in #753.

Issue #705, opened 2020-12-22
SUSHI --init references ig-data in generated config


cmoesel:
When you run `sushi --init`, the generated config contains this comment:
```yaml
# ╭────────────────────────────────────────────menu.xml────────────────────────────────────────────╮
# │  To use a provided ig-data/input/includes/menu.xml file, delete the "menu" property below.     │
# ╰────────────────────────────────────────────────────────────────────────────────────────────────╯
```

Since the generate project is a SUSHI 1.0 project, it should not reference `ig-data`.

cmoesel:
Fixed in SUSHI 1.1.0

PR #722, opened 2021-01-12
Update relative paths in file comments

### LINKED ISSUE #706: Wrong relative path in generated menu.xml
When SUSHI generates the `menu.xml` from `sushi-config.yaml`, it references the path to the `input/includes` folder in its comment.  But it references the path relative to the wrong root.  For example, this is the comment from the generated mCODE menu:
 
```xml
<!-- menu.xml {% comment %}
*****************************************************************************************************
*                                  WARNING: DO NOT EDIT THIS FILE                                   *
*                                                                                                   *
* This file is generated by SUSHI. Any edits you make to this file will be overwritten.             *
*                                                                                                   *
* To change the contents of this file, edit the "menu" attribute in the tank sushi-config.yaml file *
* or provide your own menu.xml in the HL7-fhir-mCODE-ig/input/includes folder                         *
*****************************************************************************************************
{% endcomment %} -->
```

Note that it should be `input/includes`, _not_ `HL7-fhir-mCODE-ig/input/includes`.  This is relatively minor, but it does cause unnecessary diffs when it happens in regression tests (where the root of the two folders we are comparing is different).

jafeltra:
This PR fixes #706 and #705.

There were a few places where output files were referencing the wrong relative path in comments:

- The `menu.xml` file referenced the root folder name when in the current IG publisher configuration (`myTank/input/includes`), but now it just references `input/includes`, as described in #706.
- When looking into this, I also noticed that something similar could happen in a generated `index.md` file, so I made that change as well and added a test for it.
- The generated `sushi-config.yaml` file created when running `--init` referenced the old `ig-data/input/includes/menu.xml` file, so the file path is updated, as described in #705.


cmoesel (approved):
Looks good.  I also ran regression against 52 known FSH repos and all results were as expected!

ngfreiter (approved):
Looks good to me!

PR #723, opened 2021-01-12
Regression should use npx diff2html-cli

cmoesel:
The npm module name is actually diff2html-cli (although, once installed, the binary command is diff2html).

This corresponds to FHIR/GoFSH#65 which makes a similar change.


cmoesel:
Switching to `draft` as I ended up encountering problems w/ `npx diff2html-cli` over in `GoFSH` and want to figure out what's going on there before pressing forward with this one.

cmoesel:
Nevermind.  This is not a good idea.

PR #724, opened 2021-01-12
Suppress logs from test

ngfreiter:
I noticed that there were some logs coming from some tests, this suppresses those logs.


cmoesel (approved):
Oh weird.  I made these fixes too -- and I *thought* I pushed them up to the soft-indexing PR, but... it looks like I did not?  Anyway, yep.  Looks good to me.

jafeltra (approved):
Looks good to me and I don't get any more logs in the tests!

PR #725, opened 2021-01-12
Puns and punting prettiness

cmoesel:
This PR has two main components:

1. It exports the puns so they can be re-used in other apps (GoFSH and FSH Online)
2. It attemps to detect when SUSHI is called from Java and use less pretty characters in the summary

To test the less pretty characters, you'll need to install this version of SUSHI globally and then run it from the publisher.  Step by step instructions:
1. `npm remove -g fsh-sushi`
2. `npm pack`
3. `npm install -g ./fsh-sushi-1.0.2.tgz` (adjust as necessary for Windows)
4. `sushi --init` (accept all defaults)
5. `cd ExampleIG`
6. `sushi .` 
7. Confirm that summary uses pretty boxes
8. `./_updatePublisher.sh` (adjust as necessary for Windows)
9. `./_genonce.sh` (adjust as necessary for Windows)
10. Confirm that summary uses ugly boxes

@mint-thompson -- I'd really like to confirm this works on Windows, so can you give it a try?


cmoesel:
Now that I'm working on GoFSH, I can see that many of these puns actually wouldn't work there.  For example, when there is an error in GoFSH, it doesn't make sense to shame the user or tell them to check the docs -- because most likely it is not the user's fault (and FSH docs won't help anyway).  So... these will still be useful to FSH Online, but probably not GoFSH.

jafeltra (approved):
On my Mac, this looks good. I ran locally and got cool boxes and ran within the publisher and got boring boxes. I also tried out using the puns in the playground and it looks good there too (although I didn't implement that).

ngfreiter (approved):
Tested on Windows, ugly boxes became pretty boxes!

PR #732, opened 2021-01-17
remove smart quotes from puns

costateixeira:
Reduce mojibake when running sushi in an ide


costateixeira:
I don't do typescript, but I understand that strings can use either single or double quotes, so I just went and made this minor change. Please check if it is ok.

cmoesel (approved):
Good call, @costateixeira!  Thank you for the fix!  I hate smart quotes -- so I can't believe I missed them!

PR #735, opened 2021-01-19
Fix array extensions

### LINKED ISSUE #733: Extension use in arrays: Cannot convert undefined or null to object
This works fine:
```
* name[0].given[0] = "John"
* name[0].given[1] = "Bob"
* name[0].given[0].extension[nameQualifier].valueCode = #CL 
```
This gives "SUSHI encountered the following unexpected error: Cannot convert undefined or null to object":
```
* name[0].given[0] = "John"
* name[0].given[1] = "Bob"
* name[0].given[1].extension[nameQualifier].valueCode = #CL
```
and this is the extension:
```
Profile: SEBaseHumanName
Parent: http://hl7.org/fhir/StructureDefinition/HumanName
Description: "This profile provides the use of Swedish middlename as extensions and is used in the base Patient profile"
* insert SEStandardRuleSet
* extension contains SEBaseMiddleNameExtension named middleName 0..1
* extension contains SEBaseOwnFamilyExtension named ownFamily 0..1
* given.extension contains http://hl7.org/fhir/StructureDefinition/iso21090-EN-qualifier named nameQualifier 0..1
* given.extension[nameQualifier].valueCode from SEBaseNamePartQualifierVS (required)
```
Any thoughts? If the extension is used on both given:s it also works but that is not what is intended.

cmoesel:
Fixes #733 (problem w/ extensions on non-zero index of primitive arrays).

To reproduce, isolate commit b65e302 and run the test.  It will fail.  Then run the test after applying 95b531e.  It will pass.

OR test manually w/ the following FSH:
```
Profile: MyPatient
Parent: Patient
* address.line.extension contains
    http://hl7.org/fhir/StructureDefinition/iso21090-ADXP-additionalLocator named locator 0..1

 Instance: SomePatient
 InstanceOf: MyPatient
 * address.line[1].extension[locator].valueString = "3rd Floor"
```


src/fhirtypes/InstanceDefinition.ts
```diff
@@ -53,17 +53,20 @@ export class InstanceDefinition {
  * Otherwise, use all properties from the input.
  *
  * @param input - the value to clone
- * @param keys - optionally, the properties of the value to include in the clone
+ * @param keys - optionally, the properties of the value to include in the clone, default's to input keys if not specified
```
cmoesel, line 56:
```suggestion
 * @param keys - optionally, the properties of the value to include in the clone, defaults to input keys if not specified
```

test/export/InstanceExporter.test.ts
```diff
@@ -1998,6 +1998,51 @@ describe('InstanceExporter', () => {
       });
     });
 
+    it('should something', () => {
```
cmoesel, line 2001: Oops.  Forgot to describe the test.
```suggestion
    it('should handle extensions on non-zero element of primitive arrays', () => {
```

ngfreiter (approved):
This looks good to me!

julianxcarter (approved):
This looks good!

PR #739, opened 2021-01-25
Differential only includes new constraints and mappings

### LINKED ISSUE #731: SUSHI repeats mappings/invariants in differential
Some arrays in FHIR differentials are intended to be _additive_ -- meaning that they don't _replace_ the array in the snapshot, but rather append to it.  I believe this is the case for invariants and mappings (although we should confirm by testing the behavior of the IG publisher).

If this is the case, SUSHI should be fixed such that it only generates the _additional_ items for these arrays in the differential (rather than repeating all the existing items).

mint-thompson:
Fixes #731 and completes task [CIMPL-644](https://standardhealthrecord.atlassian.net/browse/CIMPL-644).

The constraint and mapping properties on an ElementDefinition are defined by FHIR to be additive. Therefore, the differential only needs to contain newly added items.

From http://hl7.org/fhir/elementdefinition.html#interpretation: "Additional constraints and mappings can be defined, but they do not replace the ones in the base definition"


cmoesel:
This is a really smart approach, @mint-thompson.  Well done.  There is only one suggestion I have.

When I ran the regression (`npm run regression`), the `HL7/livd#master` results seemed a little odd at first.  For example:
image
I didn't understand why all those elements were in the differential in the first place if there were no _different_ elements to begin with.  But it actually makes sense if you look at their FSH:
```
* . ^mapping[0].identity = "v2"
* . ^mapping[0].map = "N/A"
* . ^mapping[1].identity = "rim"
* . ^mapping[1].map = "N/A"
```
They are basically duplicating the existing mappings (for some reason) and now we just wipe them all out because they're all duplicates.  Which is fine.

The one thing I'd suggest is that in the case of the additive elements, if we are left with _nothing_ after removing duplicates, then that property shouldn't be in the diff at all (e.g., delete the property rather than setting to `[]`).

jafeltra (commented):
The changes here make sense to me. I was looking through the diffs when I ran the regression on a few repos and I just wanted to make sure one of the diffs I got makes sense to you as well. This happened to be on `HL7/livd`.

The original `mapping` looked like this: 
```
[
  {
    identity: "rim",
    map: "Entity. Role, or Act",
  },
  {
    identity: "v2",
    map: "N/A",
  },
  {
    identity: "rim",
    map: "N/A",
  },
  {
    identity: "cda",
    map: "ClinicalDocument, if bundle.type = document",
  },
]
```
And the generated one that we're building looked like this:

```
[
  {
    identity: "v2",
    map: "N/A",
  },
  {
    identity: "rim",
    map: "N/A",
  },
  {
    identity: "rim",
    map: "N/A",
  },
  {
    identity: "cda",
    map: "ClinicalDocument, if bundle.type = document",
  },
]
```
It looks like the only difference is on the `rim` `map` but when they're passed into `differenceWith`, they end up being the same, so the `mapping` isn't included in the output. Is this expected?

cmoesel:
> It looks like the only difference is on the `rim` `map` but when they're passed into `differenceWith`, they end up being the same, so the `mapping` isn't included in the output. Is this expected?

@jafeltra - I assume that the second array you are showing is _before_ the code removes mappings that are already in the parent, right?  Because when I run regression on `livd` I'm seeing _no mapping_ there, so those all get removed.  If I understand your comment correctly, then this is what is expected.  We should only output mappings that are new -- so even though the `"rim" ... "N/A"` mapping is twice in our output, it's still not new since it exists in the original mappings.  And even though the original mapping has `"rim" ... ""Entity. Role, or Act"` and our generated mapping does not, that's still OK because we wouldn't want to repeat that mapping in the differential anyway.  I think it ended up like that in the generated mapping because the `livd` FSH overwrote some mapping array entries using `^` rules and indexers.

jafeltra:
Okay that makes sense, thanks for the additional explanation! I think I was probably getting confused with the mapping in general and the difference that was supposed to be captured in the differential. I had been looking at this [line](https://github.com/FHIR/sushi/pull/739/files#diff-59689dd9a9bc06ce76ca74dc9c92316970979146f28894eab5f6ef9640359312R375), where it does

```
diff[prop] = differenceWith(this[prop], original[prop], isEqual)
```
Those two mapping arrays are compared and considered equal, which seemed confusing to me. But I think since the outcome is what we would expect, I'm good.

jafeltra (approved):

cmoesel (approved):
Looks good to me!

PR #740, opened 2021-01-27
Simple syntax for non-UCUM systems on Quantity assignment rules

### LINKED ISSUE #730: Support non-UCUM codes in Quantity syntax
Currently, FSH only allows a single line syntax for quantities when the unit is UCUM.  E.g., this:
```
* valueQuantity = 55 'mm' "millimeters"
```
Anything inside `'` `'` is expected to be UCUM.  If the unit is not ucum, you need to do it more like this:
```
* valueQuantity = http://mysystem.org/codes#foo "Foos"
* valueQuantity.value = 42
```

I think for STU2 we should support this syntax for the latter:
```
* valueQuantity = 42 http://mysystem.org/codes#foo "Foos"
```
Of course, aliases would be supported too:
```
* valueQuantity = 42 MYSYS#foo "Foos"
```

julianxcarter:
Fixes #730 and addresses [CIMPL-643](https://standardhealthrecord.atlassian.net/browse/CIMPL-643), allowing quantities using a non-UCUM system to be assigned in a single line rather than necessitating two lines of FSH. 

To test this, place an assignment rule with this schema into a test FSH file and ensure that the output looks correct: 
```
* valueQuantity = 42 http://mysystem.org/codes#foo "Foos"
```


antlr/src/main/antlr/FSH.g4
```diff
@@ -96,7 +96,7 @@ value:              SEQUENCE | STRING | MULTILINE_STRING | NUMBER | DATETIME | T
 item:               SEQUENCE (KW_NAMED SEQUENCE)? CARD flag*;
 code:               CODE STRING?;
 concept:            STAR code (STRING | MULTILINE_STRING)?;
-quantity:           NUMBER UNIT STRING?;
+quantity:           NUMBER (UNIT | CODE)? STRING?;
```
ngfreiter, line 99: I think `(UNIT | CODE)?` should be `(UNIT | CODE)`, no `?`, since we want one or the other. And we assume that one or the other is present in later processing.

test/fshtypes/FshQuantity.test.ts
```diff
@@ -64,7 +64,7 @@ describe('FshQuantity', () => {
       const code = new FshCode('bar', 'http://foo.com', 'Display');
       const quantity = new FshQuantity(100, code);
       const result = quantity.toString();
-      expect(result).toEqual('100 \'bar\' "Display"');
+      expect(result).toEqual('100 http://foo.com#bar "Display"');
```
ngfreiter, line 67: I can't comment on the line I want, but on 61-62 there is a comment that looks like it is no longer true, so we can take that out.

ngfreiter: Oh there is also the same comment on 38-39 that can go.

src/fshtypes/FshQuantity.ts
```diff
@@ -10,7 +10,10 @@ export class FshQuantity extends FshEntity {
   toString(): string {
     let str = this.value.toString();
     if (this.unit?.code != null) {
-      str += ` '${this.unit.code}'`;
+      str +=
+        this.unit.system != null && this.unit.system != 'http://unitsofmeasure.org'
+          ? ` ${this.unit.system}#${this.unit.code}`
+          : ` '${this.unit.code}'`;
```
ngfreiter, line 16: We now allow someone to do something like this:
```
Extension: Example
* valueQuantity = 12 #foo "hello"
```
To set the `value`, `code`, and `display`, but in this case I think the `toString` function would give something like this:
```
12 'foo' "hello"
```
Which seems off since that is the syntax for assuming UCUM as the system, but in this case there is no system. I think we would prefer it to be `12 #foo "hello"`. 

Or we disallow what I am doing altogether and require a `system` when setting a Quantity in this way. I think @cmoesel would know better than me if that is what we want to do.

ngfreiter (changes requested):
A couple of small things, and a question that I'd like to hear @cmoesel's thoughts on.

antlr/src/main/antlr/FSH.g4
cmoesel: Hmmm... on one hand, it might be nice to support `1.0 "hot fudge sundae"` for a quantity that uses only the human-readable unit and no code.

On the other hand, since everything but `NUMBER` is optional in the grammar above, now `1.0` alone would match on `quantity` -- which isn't really our desired behavior.

Since the prior behavior required `UNIT` and this PR is just about also allowing `CODE`, I think I agree w/ Nick -- it should not be optional.  If we want to make it optional, that should be a different task.

src/fshtypes/FshQuantity.ts
cmoesel: The spec says that if you have a `code`, you must have a `system` too:
image

So technically `12 #foo "hello"` is invalid.  That being the case, I'd say that if we encounter this (a quantity that uses a system-less code), we should log an error.  But for the purposes of `toString()`, we should still represent whatever we have in memory, so `12 #foo "hello"` is the correct stringified representation (even if it is invalid).  Does that make sense?

antlr/src/main/antlr/FSH.g4
julianxcarter: Understood! This makes a lot fo sense, I'll change this so that either `UNIT` or `CODE` is required.

jafeltra (commented):
This looks good to me! The one thing I noticed is that I think you need to regenerate the grammar since you made your last changes. But besides that, I think this looks good!

test/fshtypes/FshQuantity.test.ts
```diff
@@ -58,20 +56,18 @@ describe('FshQuantity', () => {
...
     });
 
     it('should return string for unit code with code with spaces', () => {
       const code = new FshCode('milli meters');
       const quantity = new FshQuantity(100, code);
       const result = quantity.toString();
-      expect(result).toEqual("100 'milli meters'");
+      expect(result).toEqual('100 #milli meters');
```
jafeltra, line 70: I think that since this code is 'milli meters' with a space, we actually need to have quotes around the text after the `#`, so I think the expected output should be `100 #"milli meters"`. This made me realize that it might be better to use the `FshCode`'s ``toString` method if the system isn't UCUM. I think the `FshCode.toString` method should account for the quotes around space and should handle the case where there is no system.

src/fshtypes/FshQuantity.ts
```diff
@@ -10,7 +10,10 @@ export class FshQuantity extends FshEntity {
   toString(): string {
     let str = this.value.toString();
     if (this.unit?.code != null) {
-      str += ` '${this.unit.code}'`;
+      if (this.unit.system) {
+        if (this.unit.system == 'http://unitsofmeasure.org') str += ` '${this.unit.code}'`;
+        else str += ` ${this.unit.system}#${this.unit.code}`;
+      } else str += ` #${this.unit.code}`;
```
jafeltra, line 16: I mentioned this in the other comment, but I think it might be safer to use the `FshCode.toString` method here if we are not generating a string for the special UCUM units case. If we are, we can keep the handling here since it's a special syntax for a quantity. Otherwise, we're just using regular `FshCode` syntax, so we can take advantage of the logic there.

jafeltra (commented):
I realized one more thing below (it's basically the same idea in two comments).

src/fshtypes/FshQuantity.ts
julianxcarter: That's a great suggestion! I'm making this change now!

jafeltra (dismissed):
This looks good to me! Thanks for making the `FshCode` change!

ngfreiter (approved):
Looks good to me! Just needs an update branch

jafeltra (approved):

ngfreiter (approved):

PR #745, opened 2021-02-02
Fix Duplicate meta.profile Values

### LINKED ISSUE #737: Duplicate meta.profile values when profiles assign meta.profile value
When a profile assigns a meta.profile value (meaning all instances must have it), SUSHI will generate instances that have the profile value twice.  This is likely because the instance exporter generates it based on the fixed value rule, but also generates it automatically based on the `InstanceOf:` metadata.

This was seen in the carin-bb IG, but can be replicated with [this FSH code](https://fshschool.org/FSHOnline/#/share/2LQcQms):
```
Profile: MyPatientAbstract
Parent: Patient
Id: MyPatientAbstract
Title: "My Patient Abstract"
Description: "Simple example test Abstract"
* meta 1..1 MS
* meta.profile 1..* MS

* meta.profile ^slicing.discriminator.type = #pattern
* meta.profile ^slicing.discriminator.path = "$this"
* meta.profile contains supportedPatientProfile 1..1 MS



Profile: MyPatient
Parent: MyPatientAbstract
Id: MyPatient
* meta.profile[supportedPatientProfile] = Canonical(MyPatient)


Instance: MyPatientInstance
InstanceOf: MyPatient
Usage: #example
* name[0].family = "LastName"
```

Resulting in this instance:
```json
{
  "resourceType": "Patient",
  "id": "MyPatientInstance",
  "meta": {
    "profile": [
      "http://example.org/StructureDefinition/MyPatient",
      "http://example.org/StructureDefinition/MyPatient"
    ]
  },
  "name": [
    {
      "family": "LastName"
    }
  ]
}
```

jafeltra:
This PR fixes #737 and [CIMPL-652](https://standardhealthrecord.atlassian.net/browse/CIMPL-652).

The issue has an example of how to replicate this bug on `master`. The test that I added also replicates this example.

There were duplicate values in `meta.profile` because the `InstanceExporter` set the `meta.profile` value based on the `InstanceOf` profile at the start of the extractor. Then later in the extractor, it also sets values based on implied values on the parent. In the example in the issue, the `InstanceOf` profile has `meta.profile` set to be `1..1` and it assigns the value. The InstanceExporter sees this as a rule we should imply and set on the Instance, and so it does.

I considered a few different ways to solve this, but settled on de-duplicating `meta.profile` values in the InstanceExporter after we finish assigning all the values. This problem is pretty specific to `meta.profile` on instances, so I thought it made sense to keep it within the exporter, and I don't think there should ever be a case when we want a duplicate.


src/export/InstanceExporter.ts
```diff
@@ -300,6 +300,9 @@ export class InstanceExporter implements Fishable {
     );
     cleanResource(instanceDef);
 
+    // Once all rules are set, we should ensure that we did not add a duplicate profile URL anywhere
+    if (instanceDef.meta?.profile) instanceDef.meta.profile = uniq(instanceDef.meta.profile);
```
cmoesel, line 304: I wonder if this should be moved into `cleanResource`?  I know the comments for the `cleanResource` function seem to indicate it only removes temporary properties and such, but... it seems like it has evolved since those comments were written.  It looks like it is doing other types of cleanup now too.  What do you think?

cmoesel (commented):
Looks good and I am willing to approve as-is; but before I do, I'd like your thoughts on my question...

src/export/InstanceExporter.ts
jafeltra: I'm not entirely against this idea, but my current reasoning for keeping it out of that function is that `cleanResource` is used in other places (okay I think only SD Exporter) but this issue really arises because of how we implement our `InstanceExporter`. The case that this line is trying to fix is really only caused by the addition of `meta.profile` in this exporter. If someone in their Structure Definition had added two of the same profiles, I don't see why that would be right, but I felt like maybe we should avoid un-doing that for them. But then again, there's also an edge case that they could have done that in an Instance as well, which is why I'm don't feel too strongly, I'm just slightly on the `InstanceExporter` side. I could be convinced otherwise.

cmoesel: Yeah, OK.  I don't have strong feelings about it, so we'll keep it as-is!

cmoesel (approved):

ngfreiter (approved):
Looks good!

PR #746, opened 2021-02-03
Calculate cardinality correctly when reslicing elements and exporting instances

### LINKED ISSUE #742: Issues w/ cardinality counting and re-slices
SUSHI seems to struggle with cardinality counting on instances of profiles with re-slicing.  There are two issues:
* SUSHI infers the wrong minimum cardinality for the element that is sliced in the profile
* SUSHI does not recognize re-slices in an instance as fulfilling cardinality of the parent slice

Given this example (https://fshschool.org/FSHOnline/#/share/3opCADp):
```
Alias: HeartRate = http://hl7.org/fhir/StructureDefinition/heartrate
Alias: LOINC = http://loinc.org

Profile: MyBundleProfile
Parent: Bundle
* type = #collection
* entry ^slicing.discriminator.type = #type
* entry ^slicing.discriminator.path = "resource"
* entry ^slicing.rules = #open
* entry contains observations 1..*
* entry[observations].resource only Observation
* entry[observations] ^slicing.discriminator.type = #profile
* entry[observations] ^slicing.discriminator.path = "resource"
* entry[observations] ^slicing.rules = #open
* entry[observations] contains heartrate 1..1 MS
* entry[observations][heartrate].resource only HeartRate

Instance: MyBundleInstance
InstanceOf: MyBundleProfile
* entry[observations][heartrate].resource = MyHeartRate

Instance: MyHeartRate
InstanceOf: HeartRate
* status = #final
* effectiveDateTime = "2021-01-28"
* valueQuantity = 44 '/min' "beats/min"

Instance: MyPatient
InstanceOf: Patient
* name.given = "Bob"
```

SUSHI produces these errors:
```
error Element Bundle.entry has minimum cardinality 2 but occurs 1 time(s).
  Line: 18 - 20

error Element Bundle.entry:observations has minimum cardinality 1 but occurs 0 time(s).
  Line: 18 - 20
```

Reported on Zulip: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Slicing.20on.20slices.3F

mint-thompson:
Fixes #742 and completes task [CIMPL-651](https://standardhealthrecord.atlassian.net/browse/CIMPL-651).

When constraining the cardinality of a reslice, it may be necessary to update the minimum cardinality of the parent slice, that is, the slice being resliced. Check the sliced element and all ancestor slices when constraining the cardinality, updating other cardinalities as necessary.

The new parentSliceName field on ElementDefinition is used to track slice lineage. A slice name may legally contain a forward slash, which is also used as the character to separate the slice name from the reslice name. Therefore, splitting a slice name on the forward slash character is not a reliable way of determining the slice name of ancestor slices.

When validating required child elements in an instance, the required elements may be on the element, a slice on that element, or a reslice on that element. All of these should count when determining how many elements are present for the purpose of validating element cardinality.


src/fhirtypes/ElementDefinition.ts
```diff
@@ -524,11 +525,27 @@ export class ElementDefinition {
...
+          slicedElement === el.slicedElement()
+      );
+      const newParentMin = min + sliceSiblings.reduce((sum, el) => sum + el.min, 0);
+      // if this is a reslice, the parent element will also be a slice of the sliced element.
+      // if this is not a reslice, the parent element is the sliced element.
+      const parentElement = this.parentSliceName
+        ? this.structDef.elements.find(
+            el => el.sliceName === this.parentSliceName && slicedElement === el.slicedElement()
+          )
+        : slicedElement;
```
cmoesel, line 541: Picking back up on the previous conversation about using `parentSliceName` to find the parent _vs_ using `/` in the `sliceName` to find the parent...  We've already established that `parentSliceName` will only be present for re-slices that were added by SUSHI in this profile.  So that means that if there are any re-slices in the parent profile, their elements will not have `parentSliceName`.  E.g., if base definition has slice `a` and re-slice `a/b`, and in FSH we re-slice `a/b` to make `a/b/c`, then we have:
* `a` with `parentSliceName === undefined`
* `a/b` with `parentSliceName === undefined` (since `a/b` is defined in the parent JSON)
* `a/b/c` with `parentSliceName === 'a/b'`

So... I think that means that when adding `a/b/c`, the cardinality constraints will properly be propagated from `a/b/c` to `a/b` (since `a/b/c` does have `parentSliceName`), but will not be properly propagated from `a/b` to `a` (since `a/b` does not have `parentSliceName`).

Similarly, if instead of reslicing `a/b` to create an `a/b/c`, you _added_ another reslice to `a`, making `a/b2`.  In this scenario, you would end up with:
* `a` with `parentSliceName === undefined`
* `a/b` with `parentSliceName === undefined` (since `a/b` is defined in the parent JSON)
* `a/b2` with `parentSliceName === 'a'`

Now we have a different problem in that in lines 528 - 533 above, the `a/b` re-slice will not be properly recognized as a sibling of `a/b2` since `a/b` does not have a `parentSliceName`.  This could also result in cardinality propagation errors since the sum of sibling slices would be incorrect.

The only way I see around this problem is to rely on using `/` to detect slice parents.  But if we need to do that anyway, then it seems like we would be overcomplicating things by attempting to use that approach _and_ `parentSliceName` at the same time.  It seems we should use `/` across the board.

Now, to mitigate things, we don't have to _only_ rely on `/`; we _can_ do some things to confirm that it is indeed a reslice.  If we see sliceName `x/y`, then there are a few things that will be true if it is a reslice:
* there will be an element w/ the same path and `sliceName === 'x'`
* that element will also have a `slicing` defined

If either of the above are not true, then `y` is _not_ a re-slice of `x`.

You may be able to keep much of this logic (and perhaps simplify some other logic) if you create a `getParentSlice` or `getParentSliceName` function that basically does what I described above. To support complex scenarios w/ multiple levels of re-slicing, I guess you would actually need to loop, walking the `/` backwards.  E.g., to find parent slice of `a/b/c/d`, you would first check:
* is `a/b/c` a valid slice w/ a slicing? If yes, it's the parent;
* if not, is `a/b` a valid slice w/ a slicing? If yes, it's the parent;
* if not is `a` a valid slice w/ a slicing? If yes, it's the parent;
* if not, this is not a re-slice, so it has not parent.

cmoesel (changes requested):
I think you found a good solution that works most of the time, but... there are some use cases it will not handle well (if I understand the code correctly).  I actually think it may be best to rely on `/` to detect parent slices.  While `/` can be a valid character in a slicename, I think that is rare, and I think there are things we can do to avoid being fooled by it.  I also think it is probably more rare than some of the other use cases that are actually broken by _not_ using `/` to detect parent slices.

See my individual comments for details.  Also, I definitely could be understanding the code wrong, and if so, let me know!

mint-thompson:
@cmoesel Thank you for these important points about slice name handling, particularly in regards to profiles that contain reslices that were not created by SUSHI. I think that case specifically has changed my mind on this matter: the `parentSliceName` field is not able to be used consistently enough, and it would be better to have more consistent behavior regardless of whether reslices came from SUSHI or not. I am willing to accept losing out on support for cases where reslices with unusual names are defined in SUSHI.

cmoesel:
@mint-thompson -- cool.  Sorry for the extra work!

cmoesel (approved):
This looks good and works well.  And it looks like it wasn't _too_ painful, so I'm glad for that!

ngfreiter (approved):
Looks good to me!

PR #747, opened 2021-02-03
long dull meeting

markkramerus:


src/utils/puns.ts
```diff
@@ -13,7 +13,12 @@ const CLEAN_RUN_PUNS = [
   'Fin-tastic job!',
   'You are dolphinitely doing great!',
   "It doesn't get any betta than this!",
-  "You're piranha roll now!"
+  "You're piranha roll now!",
+  "Valedictorian of the School of FSH",
+  "If you're FSHing for compliments, you got it!",
```
cmoesel, line 18: Unfortunately, puns must be 36 characters or less (else they don't fit in the box).  I should probably put in a unit test that enforces this.

src/utils/puns.ts
```diff
@@ -13,7 +13,12 @@ const CLEAN_RUN_PUNS = [
...
+  'Tank you for the great job',
+  "We don't have to print these puns, but we do it for the halibut",
```
cmoesel, line 20: This one exceeds the 36 character limit as well.

src/utils/puns.ts
```diff
@@ -24,7 +29,11 @@ const WARNING_PUNS = [
   'A bit pitchy, but tuna-ble.',
   'Do you sea the problem?',
   'You are skating on fin ice.',
-  'You should mullet over.'
+  'You should mullet over.',
+  'You might need some Vitamin Sea',
+  'Maybe you should scale back',
+  'Whale, whale, whale, what happened here?',
```
cmoesel, line 35: I really like this one, but it is 4 characters too long.

cmoesel (changes requested):
I'm sorry that you had to sit through a long dull meeting, but I love getting more puns!

There were a few issues:
* One case where you used `"` instead of `'`
* Several cases where you exceeded the (undocumented) 36-character limit
* Several cases missing ending punctuation

src/utils/puns.ts
markkramerus: fixed

src/utils/puns.ts
markkramerus: fixed

src/utils/puns.ts
markkramerus: DIdn't know that.

cmoesel: Yeah, I _thought_ I had put a comment about this at the top of the file, but apparently I thought wrong.

src/utils/puns.ts
```diff
@@ -13,7 +13,13 @@ const CLEAN_RUN_PUNS = [
   'Fin-tastic job!',
   'You are dolphinitely doing great!',
   "It doesn't get any betta than this!",
-  "You're piranha roll now!"
+  "You're piranha roll now!",
+  'Valedictorian of the School of FSH!',
+  'FSHing for compliments? Super job!',
+  'Tank you very much!',
+  "Why pun? Just for for the halibut.",
```
cmoesel, line 20:
```suggestion
  'Why pun? Just for for the halibut.',
```

cmoesel (approved):
Thank you for the awesome puns.  I think my top three favorites are:
* Did you do this on porpoise?
* Let minnow if you need some help
* Don't be koi about asking for help

PR #749, opened 2021-02-04
Avoid Regex Lookbehind 

jafeltra:
Safari does not yet support regular expression lookbehind assertions ([ugh](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#browser_compatibility)). Because of this, FSH Online does not load at all when we use the current version of SUSHI.

I tried to find a webpack configuration option, polyfill, or babel plugin that could get around this issue or fix it. I had no luck though. So, instead, I've updated SUSHI's handling of the one place we had a lookbehind. It's definitely much less elegant than what we had before, but it lets FSH Online load in Safari.

I'd like @mint-thompson to be one of the reviewers since this is code they originally wrote, and I know they've helped me through some other string escaping confusion in the past, so I'd like their input!

### DIFF
--- a/src/import/FSHImporter.ts
+++ b/src/import/FSHImporter.ts
@@ -1540,10 +1540,26 @@ export class FSHImporter extends FSHVisitor {
     const splitBackslash = ruleNoParens.split(/\\\\/g);
     // then, split the parameters apart with unescaped commas
     const splitComma = splitBackslash.map(substrBackslash => {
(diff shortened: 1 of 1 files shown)


cmoesel (approved):
This seems reasonable to me.  Ugly, but reasonable.  I can't think of a less ugly way to do it, so here we go. ;-)

mint-thompson (approved):

PR #753, opened 2021-02-04
Back out behavior to carry over extensions from contentreference targets

cmoesel:
Back out behavior to carry over extensions from contentreference targets.  Apparently this understanding of the behavior was born out of a misunderstanding. Authors should signal that they want constraints carried over by using a special extension -- but it's not clear to me yet how that extension works (and probably not clear to authors either), so... for now... just revert to the way it was.

See: https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/Clarification.20on.20contentReference/near/225076962


cmoesel:
For reference, here is the PR where I added this behavior: https://github.com/FHIR/sushi/pull/721

Note, however, there are other good things in that PR that we want to keep.  So this only backs out the specific carry-over behavior from that PR.

ngfreiter (dismissed):
Reverse!

jafeltra (dismissed):
This looks good to me. I noticed there were a few test fixtures that were added in that PR that were used in the test that was removed, but if the plan is to use those once we know how this works, I'm not worried about those. I'm also not that worried either way, so I'm approving as is!

cmoesel:
Good point @jafeltra.  Consider them _GONE!_

jafeltra (approved):

cmoesel:
I don't think we need re-approval from @ngfreiter, so I'm merging!

PR #754, opened 2021-02-08
Fix handling of empty string ruleset parameters

cmoesel:
To manually test this (and see broken behavior vs fixed behavior

* Locally edit `regression/repos-select.txt` so it contains only one line:
  ```
  openhie/case-reporting#master
  ```
* Run regression: `npm run regression`

Note that in the results, the master branch produces 12 errors and this branch produces 0 errors.  Also note that in the diff, this branch's questionnaire instance contains many items missing from the master branch's instance.

### DIFF
--- a/src/import/FSHImporter.ts
+++ b/src/import/FSHImporter.ts
@@ -1559,7 +1559,7 @@ export class FSHImporter extends FSHVisitor {
           // then, make all the replacements: closing parenthesis and comma
           return substrComma.replace(/\\\)/g, ')').replace(/\\,/g, ',');
         })
-        .filter(s => s); // Filter out any null values from incorrectly split escaped commas
+        .filter(s => s != null); // Filter out any null values from incorrectly split escaped commas
     });
     const paramList: string[] = [];
     // if splitComma has more than one list, that means we split on literal backslash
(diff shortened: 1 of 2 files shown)


src/import/FSHImporter.ts
```diff
@@ -1559,7 +1559,7 @@ export class FSHImporter extends FSHVisitor {
           // then, make all the replacements: closing parenthesis and comma
           return substrComma.replace(/\\\)/g, ')').replace(/\\,/g, ',');
         })
-        .filter(s => s); // Filter out any null values from incorrectly split escaped commas
+        .filter(s => s != null); // Filter out any null values from incorrectly split escaped commas
```
ngfreiter, line 1562: This comment is confusing me a bit. This filtering is handling empty parameter values right?

ngfreiter (approved):
This looks good to me, regression and the test worked as expected. I had one minor question, but it's minor enough for me to approve.

src/import/FSHImporter.ts
ngfreiter: Oh ok, and we were incorrectly filtering the `''`, which messed with empty params, gotcha that does make sense.

PR #757, opened 2021-02-12
Extensions can now be referenced by URL on Instances

### LINKED ISSUE #674: Instances do not properly handle extension URLs in the path
Reported on Zulip [here](https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/adding.20extension.20to.20Instance).

Based on [FSH 3.4.6 Extension Paths](https://hl7.org/fhir/uv/shorthand/reference.html#extension-paths), the following should work:
```
Instance: MyQ
InstanceOf: Questionnaire
* status = #active
* item[0].extension[http://hl7.org/fhir/StructureDefinition/questionnaire-itemControl].valueCodeableConcept = http://molit.eu/fhir/CodeSystem/questionnaire-codes-tbd#vas
* item[0].linkId = "foo"
* item[0].type = #group
```
SUSHI, however, reports:
> error The element or path you referenced does not exist: item.extension[http://hl7.org/fhir/StructureDefinition/questionnaire-itemControl].valueCodeableConcept
>   Line: 4
> 
> error y is undefined

If you use an alias instead of the URL directly, however, it works.

julianxcarter:
Fixes #674 and addresses [CIMPL-605](https://standardhealthrecord.atlassian.net/browse/CIMPL-605), fixing a bug where extensions would fail to be referenced by URL. On master, the following assignment rule on a Patient instance should generate an error...
```
* extension[http://hl7.org/fhir/us/core/StructureDefinition/us-core-ethnicity].extension[text].valueString = "Not Hispanic or Latino"
```
while simply setting an alias for the url would work just fine...
```
Alias:  USCoreEthnicity = http://hl7.org/fhir/us/core/StructureDefinition/us-core-ethnicity
* extension[USCoreEthnicity].extension[text].valueString = "Not Hispanic or Latino"
```
Both methods of referencing the extension should work in this PR. This is a pretty simple fix, simply creating a new slice on the extension element using the referenced extensions `id` property rather than its Url. This allows the extension being reference to be found by `findElementByPath` without any changes to that function.

### DIFF
--- a/src/fhirtypes/StructureDefinition.ts
+++ b/src/fhirtypes/StructureDefinition.ts
@@ -22,6 +22,7 @@ import {
 import { Fishable, Type } from '../utils/Fishable';
 import { applyMixins, parseFSHPath, assembleFSHPath } from '../utils';
 import { InstanceDefinition } from './InstanceDefinition';
+import { isUri } from 'valid-url';
 
 /**
  * A class representing a FHIR R4 StructureDefinition.  For the most part, each allowable property in a StructureDefinition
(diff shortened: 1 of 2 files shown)


ngfreiter (approved):
This looks good to me!

mint-thompson (approved):

Issue #751, opened 2021-02-04
Extension URL not getting set in instance


cmoesel:
In some cases, referencing an extension by its slice name in an instance is not properly populating the extension URL in the exported instance JSON.

To reproduce: https://fshschool.org/FSHOnline/#/share/3toDtQo
```
Alias: SCT = http://snomed.info/sct

Profile: MyDevice
Parent: Device
* extension contains StraumannExtensionAbutmentClass named abutmentclass 0..1
* type.extension contains StraumannExtensionAbutmentClass named abutmentclass 0..1

Extension: StraumannExtensionAbutmentClass
Id: straumann-extension-abutmentclass
Title: "Straumann Extension Abutment Class"
Description: "Straumann Extension Abutment Class"
* valueCodeableConcept only CodeableConcept
* valueCodeableConcept ^min = 1
* valueCodeableConcept from StraumannVSAbutmentClass (required)

ValueSet: StraumannVSAbutmentClass
// no definition; just here so reference resolves

CodeSystem: StraumannCSFDAProductCode
// no definition; just here so reference resolves

Instance: MyDeviceExample
InstanceOf: MyDevice
* extension[abutmentclass].valueCodeableConcept = SCT#468936006 "Dental implant suprastructure, permanent (physical object)"
* type = StraumannCSFDAProductCode#NHA "Abutment, Implant, Dental, Endosseous"
* type.extension[abutmentclass].valueCodeableConcept = SCT#468936006 "Dental implant suprastructure, permanent (physical object)"
```

PR #758, opened 2021-02-12
Fix Extension URL on Instance

ngfreiter:
Fixes https://github.com/FHIR/sushi/issues/751.

The bug described in the issue would happen because `type.extension.url` is an implied rule, and implied rules are applied before explicit rules. So this rule would be applied, but this line:
```
* type = StraumannCSFDAProductCode#NHA "Abutment, Implant, Dental, Endosseous"
```
would overwrite the value of `type` entirely, removing the `type.extension.url` value. To fix this, this PR makes it so that implied rules are applied to one copy of the instance, and explicit rules are applied to another. Then these two objects are merged together. If the implicit and explicit rules conflict, the explicit rules will win, but values that can coexist, like in this bug, are not overwritten.

To test this, you can run the example given in the issue on `master`, and notice that `type.extension[0].url` is not set, and an error is emitted about this. On this branch, that value does get set, and there is no error. You can also try the test on https://github.com/FHIR/sushi/commit/7b003f31ba6f857a9e175a96256b627d57b1a153 and note that it fails, but on the latest commit it will pass.

### DIFF
--- a/regression/run-regression.ts
+++ b/regression/run-regression.ts
@@ -8,7 +8,7 @@ import program from 'commander';
 import readlineSync from 'readline-sync';
 import extract from 'extract-zip';
 import opener from 'opener';
-import { union } from 'lodash';
+import { isEqual, union } from 'lodash';
 import { createTwoFilesPatch } from 'diff';
 
 // Track temporary files so they are deleted when the process exits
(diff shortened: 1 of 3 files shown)


mint-thompson (approved):

julianxcarter (approved):
Looks good to me!

PR #744, opened 2021-02-02
Update link in sushi --init

ngfreiter:
This fixes #738 by fixing the documentation link that we output when running `sushi --init`. Also had to make the boxes slightly bigger, because the link is longer now.

PR #761, opened 2021-02-15
Fix URL in SUSHI-generated config file

masnick:
The URL that points to the SUSHI configuration docs is wrong in the generated configuration file. I think this will fix that, along with cleaning up other references to that.

### DIFF
--- a/src/utils/init-project/sushi-config.yaml
+++ b/src/utils/init-project/sushi-config.yaml
@@ -1,6 +1,6 @@
 # ╭──────────────────────────────────────ImplementationGuide───────────────────────────────────────╮
 # │  The properties below are used to create the ImplementationGuide resource. For a list of       │
-# │  supported properties, see: https://fshschool.org/sushi/configuration/                         │
+# │  supported properties, see: https://fshschool.org/docs/sushi/configuration/                         │
 # ╰────────────────────────────────────────────────────────────────────────────────────────────────╯
 id: fhir.example
 canonical: http://example.org
--- a/src/utils/init-project/sushi-config.yaml
+++ b/src/utils/init-project/sushi-config.yaml
@@ -1,6 +1,6 @@
 # ╭──────────────────────────────────────ImplementationGuide───────────────────────────────────────╮
 # │  The properties below are used to create the ImplementationGuide resource. For a list of       │
-# │  supported properties, see: https://fshschool.org/docs/sushi/configuration/                         │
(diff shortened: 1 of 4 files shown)


cmoesel:
Hey @ngfreiter -- looks like @masnick beat you to it!

cmoesel:
Oh wait.  I thought you had done that already @ngfreiter.  Now I see you had in #744.  So... are there multiple paths to getting the wrong URL then?

cmoesel:
Ohh....  I think @ngfreiter's PR fixed it for `sushi --init`, but this one fixes it for when we generate a config based on an existing `package.json` (or IG JSON? I forget which)...

src/utils/init-project/sushi-config.yaml
```diff
@@ -1,6 +1,6 @@
 # ╭──────────────────────────────────────ImplementationGuide───────────────────────────────────────╮
 # │  The properties below are used to create the ImplementationGuide resource. For a list of       │
-# │  supported properties, see: https://fshschool.org/sushi/configuration/                         │
+# │  supported properties, see: https://fshschool.org/docs/sushi/configuration/                         │
```
jafeltra, line 3: I don't believe this actually affects the tests in any meaningful way, but the new link is a few characters longer than the old, so the `|` that closes the box on this line is a little too far out to correctly form the box. Would you mind removing a few of the trailing spaces between the link and the `|` so the boxes lines up here? Also, if you wouldn't mind, would you be able to make that change in the other two test fixtures below this one?

jafeltra (commented):
Thanks for the PR @masnick and for catching these extra incorrect links! I just had one small comment below. Let me know what you think!

src/utils/init-project/sushi-config.yaml
cmoesel: @masnick -- is this something you can quickly fix for us, or would you prefer we just pop in and do it?  I'm not sure if we have permission to push to your fork or not...

cmoesel:
OK.  I've fixed the border alignment.  Ready for your review whenever, @jafeltra!

jafeltra (approved):
```
╭──────────────Approval───────────────╮
│  Borders are aligned to perfection. │ 
╰─────────────────────────────────────╯
```

cmoesel:
```
╭──────────────Approval───────────────╮
│  Borders are perfect! │ 
╰─────────────────────────────────────╯
```
Doh!

PR #762, opened 2021-02-16
Logger warning on null values within arrays

### LINKED ISSUE #486: sushi should check and handle missing array elements
In my haste, I deleted array element 1 and forgot to renumber 

parameter[0].targetProfile = "CEQSubscriptionBundle"
parameter[2].name = #return

This caused sushi to insert a null, resulting in an IG Publisher error 
Not a JSON Object: null (00:13.0946)

Array checking may be worthwhile to ensure that missing elements don't cause IG Publisher crashes.

julianxcarter:
Fixes #486 and addresses [CIMPL-416](https://standardhealthrecord.atlassian.net/browse/CIMPL-416), checking and warning when a user has specified a resource with null values within one of its elements. Note that this does not happen on elements with a preceding `_`, as that signifies that an element within an array of primitives has an extension assigned to it. In that case, null values are expected behavior.

[This link](https://fshschool.org/FSHOnline/#/share/3dx0Emf) includes some example FSH that could be used to test these changes. When run on FSHOnline, no warning should be logged to the console. When run against this branch, you should get some warnings letting you know that each of these resources has some null values within their arrays.

### DIFF
--- a/src/utils/Processing.ts
+++ b/src/utils/Processing.ts
@@ -242,21 +242,23 @@ export function fillTank(rawFSHes: RawFSH[], config: Configuration): FSHTank {
 }
 
 export function checkNullValuesOnArray(resource: any, parentName = '', priorPath = ''): void {
-  const resourceName = parentName ? parentName : resource.id ?? resource.name;
+  const resourceName = parentName ?? resource.id ?? resource.name;
   for (const propertyKey in resource) {
     const property = resource[propertyKey];
     const currentPath = !priorPath ? propertyKey : priorPath.concat(`.${propertyKey}`);
     // If a property's key begins with "_", we'll want to ignore null values on it's top level
     // but still check any nested objects for null values
     if (propertyKey.startsWith('_')) {
-      if (isPlainObject(property))
+      if (isPlainObject(property)) {
         // If we encounter an object property, we'll want to check its properties as well
         checkNullValuesOnArray(property, resourceName, currentPath);
+      }
       if (Array.isArray(property)) {
         property.forEach((element: any, index: number) => {
-          if (isPlainObject(element))
+          if (isPlainObject(element)) {
             // If we encounter an object property, we'll want to check its properties as well
             checkNullValuesOnArray(element, resourceName, `${currentPath}[${index}]`);
+          }
         });
       }
     } else {
@@ -267,9 +269,10 @@ export function checkNullValuesOnArray(resource: any, parentName = '', priorPath
         const nullIndexes: number[] = [];
         property.forEach((element: any, index: number) => {
           if (element === null) nullIndexes.push(index);
-          if (isPlainObject(element))
+          if (isPlainObject(element)) {
             // If we encounter an object property, we'll want to check its properties as well
             checkNullValuesOnArray(element, resourceName, `${currentPath}[${index}]`);
+          }
         });
         if (nullIndexes.length > 0) {
           logger.warn(
--- a/src/utils/Processing.ts
+++ b/src/utils/Processing.ts
@@ -249,6 +249,9 @@ export function checkNullValuesOnArray(resource: any, parentName = '', priorPath
     // If a property's key begins with "_", we'll want to ignore null values on it's top level
     // but still check any nested objects for null values
     if (propertyKey.startsWith('_')) {
+      if (isPlainObject(property))
+        // If we encounter an object property, we'll want to check its properties as well
+        checkNullValuesOnArray(property, resourceName, currentPath);
       if (Array.isArray(property)) {
         property.forEach((element: any, index: number) => {
           if (isPlainObject(element))
--- a/src/utils/Processing.ts
+++ b/src/utils/Processing.ts
@@ -21,6 +21,7 @@ import {
 } from './InstanceDefinitionUtils';
 import { Configuration } from '../fshtypes';
 import { loadConfigurationFromIgResource } from '../import/loadConfigurationFromIgResource';
+import { InstanceDefinition } from '../fhirtypes';
 
 export function isSupportedFHIRVersion(version: string): boolean {
   // For now, allow current or any 4.x version of FHIR except 4.0.0. This is a quick check; not a guarantee.  If a user passes
@@ -240,6 +241,26 @@ export function fillTank(rawFSHes: RawFSH[], config: Configuration): FSHTank {
   return new FSHTank(docs, config);
 }
 
+function checkNullValuesOnArray(resource: any): void {
+  const resourceType = resource instanceof InstanceDefinition ? 'Instance' : resource.resourceType;
+  const resourceName =
+    resource instanceof InstanceDefinition ? resource._instanceMeta.name : resource.title;
+  for (const property in resource) {
+    if (!property.startsWith('_') && Array.isArray(resource[property])) {
+      const nullIndexes: number[] = [];
+      resource[property].forEach((element: any, index: number) => {
+        if (element === null) nullIndexes.push(index);
+      });
+      if (nullIndexes.length > 0)
+        logger.warn(
+          `Null values found at elements ${JSON.stringify(
+            nullIndexes
+          )} within '${property}' on ${resourceType} '${resourceName}'`
+        );
+    }
+  }
+}
+
 export function writeFHIRResources(
   outDir: string,
   outPackage: Package,
@@ -272,6 +293,7 @@ export function writeFHIRResources(
             predef.id === resource.id
         )
       ) {
+        checkNullValuesOnArray(resource);
         fs.outputJSONSync(path.join(exportDir, resource.getFileName()), resource.toJSON(snapshot), {
           spaces: 2
         });
--- a/src/utils/Processing.ts
+++ b/src/utils/Processing.ts
@@ -242,7 +242,7 @@ export function fillTank(rawFSHes: RawFSH[], config: Configuration): FSHTank {
 }
 
 export function checkNullValuesOnArray(resource: any, parentName = '', priorPath = ''): void {
-  const resourceName = parentName ?? resource.id ?? resource.name;
+  const resourceName = parentName ? parentName : resource.id ?? resource.name;
   for (const propertyKey in resource) {
     const property = resource[propertyKey];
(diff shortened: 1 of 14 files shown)


src/utils/Processing.ts
```diff
@@ -240,6 +241,26 @@ export function fillTank(rawFSHes: RawFSH[], config: Configuration): FSHTank {
   return new FSHTank(docs, config);
 }
 
+function checkNullValuesOnArray(resource: any): void {
+  const resourceType = resource instanceof InstanceDefinition ? 'Instance' : resource.resourceType;
+  const resourceName =
+    resource instanceof InstanceDefinition ? resource._instanceMeta.name : resource.title;
+  for (const property in resource) {
+    if (!property.startsWith('_') && Array.isArray(resource[property])) {
```
cmoesel, line 249: I know you're excluding `_` prefixes because they might represent extensions on primitive arrays, in which case `null` is allowed.  But other people (including "future me") might not know that, so can you leave a comment?

src/utils/Processing.ts
```diff
@@ -240,6 +241,26 @@ export function fillTank(rawFSHes: RawFSH[], config: Configuration): FSHTank {
...
+      const nullIndexes: number[] = [];
+      resource[property].forEach((element: any, index: number) => {
+        if (element === null) nullIndexes.push(index);
+      });
+      if (nullIndexes.length > 0)
+        logger.warn(
+          `Null values found at elements ${JSON.stringify(
+            nullIndexes
+          )} within '${property}' on ${resourceType} '${resourceName}'`
```
cmoesel, line 258: IIRC, `JSON.stringify` on an array will include the `[` / `]` delimiters.  So this message will end up something like:
> Null values found at elements [2,3] within 'foo' on StructureDefinition 'MyProfile'

I think people might find that confusing, so I might change that to read something more like:

> The array 'foo' in MyProfile is missing values at the following indices: 2, 3.

src/utils/Processing.ts
```diff
@@ -240,6 +241,26 @@ export function fillTank(rawFSHes: RawFSH[], config: Configuration): FSHTank {
   return new FSHTank(docs, config);
 }
 
+function checkNullValuesOnArray(resource: any): void {
+  const resourceType = resource instanceof InstanceDefinition ? 'Instance' : resource.resourceType;
+  const resourceName =
+    resource instanceof InstanceDefinition ? resource._instanceMeta.name : resource.title;
```
cmoesel, line 247: Instead of `resource.title` I would actually use `resource.name ?? resource.id`.  We rarely use `title` to identify a resource.

src/utils/Processing.ts
```diff
@@ -240,6 +241,26 @@ export function fillTank(rawFSHes: RawFSH[], config: Configuration): FSHTank {
...
+  for (const property in resource) {
```
cmoesel, line 248: It looks like you're only iterating top-level fields.  I think we want to detect missing values in arrays that are at nested paths too.  For example:
```json
{
  "foo": {
    "bar": [ "a", null, "c"]
  }
}
```

cmoesel (changes requested):
I left a few comments.  I think the two big asks are:

1. This seems to operate only on top-level properties, but we want it to operate recursively on all nested paths.
2. It would be great to have some unit tests for this.

src/utils/Processing.ts
```diff
@@ -240,6 +241,29 @@ export function fillTank(rawFSHes: RawFSH[], config: Configuration): FSHTank {
   return new FSHTank(docs, config);
 }
 
+export function checkNullValuesOnArray(resource: any, parentName = ''): void {
+  const resourceName = parentName ? parentName : resource.id ?? resource.name;
+  for (const property in resource) {
+    // We'll want to exclude properties beginning with an "_" as those are arrays containing extensions
+    // corresponding to a primitive array, per the FHIR spec
+    if (property.startsWith('_')) continue;
```
cmoesel, line 249: Unfortunately, I think it might be a little more complicated than this.  It is true that if a property starts with `_` and is an array, it is OK for it to have null values. But... each element of the array that is not null also might contain a sub-property that is an array and maybe should not have nulls.

For example, in this case:
```json5
{
  "_foo": [
    null, // this is ok
    {
      "extension": [
        null, // this is NOT ok
        { "url": "http://foo.org", "valueBoolean": true }
      ]
    }
  ]
} 
```

src/utils/Processing.ts
```diff
@@ -240,6 +241,29 @@ export function fillTank(rawFSHes: RawFSH[], config: Configuration): FSHTank {
...
+    if (_.isPlainObject(resource[property]))
```
cmoesel, line 250: I think the convention we've been using is to import just the functions we need rather than all of lodash.  So I'd change the import to:
```
import { isPlainObject } from 'lodash';
```
and just use `isPlainObject` here.

src/utils/Processing.ts
```diff
@@ -240,6 +241,29 @@ export function fillTank(rawFSHes: RawFSH[], config: Configuration): FSHTank {
...
+      // If we encounter an object property, we'll want to check its properties as well
+      checkNullValuesOnArray(resource[property], resourceName);
+    if (Array.isArray(resource[property])) {
+      const nullIndexes: number[] = [];
+      resource[property].forEach((element: any, index: number) => {
+        if (element === null) nullIndexes.push(index);
```
cmoesel, line 256: If the array item not null, and it's a plain object, we need to check that object for array properties w/ null values too, right?  E.g.,
```json5
{
  "foo": [
    {
      "coding": [
        null, // this is bad
        { "system": "http://foo.org", "code": "bar" }
      ]
    }
  ] 
}
```
The `foo` array is ok, but the `foo[0].coding` array is not ok.

src/utils/Processing.ts
```diff
@@ -240,6 +241,29 @@ export function fillTank(rawFSHes: RawFSH[], config: Configuration): FSHTank {
...
+      });
+      if (nullIndexes.length > 0) {
+        logger.warn(
+          `The array '${property}' in ${resourceName} is missing values at the following indices: ${nullIndexes}`
```
cmoesel, line 260: If the problem was in a deeply nested object, then `${property}` is only going to be the property name within the deeply nested object -- but I think it needs to be the path, right?  E.g.,
```json
{
  "resourceType": "Patient",
  "name": [ { "given": [ "Bob", null, "Robert" ] } ]
}
```
Would say:
> The array 'given' in Patient is missing values at the following indices: 1

But it probably should say:
> The array 'name[0].given' in Patient is missing values at the following indices: 1

cmoesel (changes requested):
Definitely a step in the right direction, but I think there are a couple more small things that need to be worked out further...  See individual comments for details.

julianxcarter:
All of your comments were great suggestions, and should be addressed now @cmoesel! The most recent commits broaden the scope of the warnings to include the other cases of nested null values that didn't occur to me before, makes the warning more useful by including the full path, and also adds a few more test cases to make sure this works as intended.

src/utils/Processing.ts
```diff
@@ -240,6 +241,46 @@ export function fillTank(rawFSHes: RawFSH[], config: Configuration): FSHTank {
...
+  const resourceName = parentName ? parentName : resource.id ?? resource.name;
+  for (const propertyKey in resource) {
+    const property = resource[propertyKey];
+    const currentPath = !priorPath ? propertyKey : priorPath.concat(`.${propertyKey}`);
+    // If a property's key begins with "_", we'll want to ignore null values on it's top level
+    // but still check any nested objects for null values
+    if (propertyKey.startsWith('_')) {
+      if (isPlainObject(property))
+        // If we encounter an object property, we'll want to check its properties as well
+        checkNullValuesOnArray(property, resourceName, currentPath);
```
jafeltra, line 254: This is a fairly small code-styling question, but I think we tend to use `{}` around `if` statements that are on more than one line. I'm not 100% if that's truly a pattern we follow or just something I tend to do normally. Do you know if this is a pattern or do you have a preference for keeping the `{}` off?

jafeltra (commented):
This looks really good. I didn't think of any cases that are not yet covered, but I did have one small code styling question below.

src/utils/Processing.ts
```diff
@@ -240,6 +241,46 @@ export function fillTank(rawFSHes: RawFSH[], config: Configuration): FSHTank {
   return new FSHTank(docs, config);
 }
 
+export function checkNullValuesOnArray(resource: any, parentName = '', priorPath = ''): void {
+  const resourceName = parentName ? parentName : resource.id ?? resource.name;
```
jafeltra, line 245: Oh this is one other really small code styling change, but I think you can chain the `??` together, so I think this line could be `parentName ?? resource.id ?? resource.name`. I'm not sure if there's any other benefit besides it looking simpler, though.

src/utils/Processing.ts
cmoesel: Yeah, I think it is a bit of an unwritten rule -- and I think I assumed eslint or prettier would enforce it -- but we pretty much always use `{` `}` except in one-liners.

cmoesel (commented):
I agree with @jafeltra's comments about styling.  Once those are addressed, this PR has my approval!

src/utils/Processing.ts
julianxcarter: This is definitely weird! Initially the if-statement was a one-liner, but prettier separated the clause out and I just ran with it. This has been fixed in the most recent commit.

src/utils/Processing.ts
julianxcarter: I tried switching this line over to chaining the `??`s together but that didn't work. I think this is because the ternary operator checks if `parentName` is truthy but the `??` operator only checks if it is defined, which it always will be. Because of that I think it'd be best to leave this line as is unless you think it'd be best for me rework my approach.

jafeltra: Ohh the truthy piece of this is definitely subtle and real and I missed it, so I'm good with keeping it how you have it.

jafeltra (approved):
Code styling updates look good to me! Thanks!

cmoesel (approved):
Boom!

PR #766, opened 2021-02-23
Support Canonical for Instances

### LINKED ISSUE #743: SUSHI should support `Canonical` for relevant Instances too
SUSHI should support the `Canonical(...)` lookup for instances w/ a canonical URL too.  This is not just _Conformance_ and _Terminology_ resources, but also things like `Questionnaire`.

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Failed.20canonical.20lookup

cmoesel:
Allow the Canonical(...) construct to target an instance.  This required updating the tank fisher to return instance url values when fishing for metadata.

If you want to test manually, you can use this modified example from the original bug reported on Zulip:
```
Instance: WhoCrQuestionnaireCovid19Surveillance
InstanceOf: Questionnaire
Description: "TODO"
Title: "Revised case report form for Confirmed Novel Coronavirus COVID-19"
Usage: #definition
* url = "http://openhie.github.io/covid-19/Questionnaire/WhoCrQuestionnaireCovid19Surveillance"
* status = #active

Profile:    WhoCrQuestionnaireResponseCovid19Surveillance
Parent:     QuestionnaireResponse
Description: "Who Case Reporting QuestionnaireResponse"
Title:      "Who Case Reporting QuestionnaireResponse"
* questionnaire = Canonical(WhoCrQuestionnaireCovid19Surveillance)
```
It will have an error on `master`:
> error Cannot use canonical URL of WhoCrQuestionnaireCovid19Surveillance because it does not exist. Be sure that WhoCrQuestionnaireCovid19Surveillance exists and it has a URL.

But it will succeed on this branch.

Fixes #743

### DIFF
--- a/src/import/FSHTank.ts
+++ b/src/import/FSHTank.ts
@@ -10,9 +10,10 @@ import {
   Mapping,
   Configuration
 } from '../fshtypes';
-import flatMap from 'lodash/flatMap';
+import { AssignmentRule } from '../fshtypes/rules';
 import { Type, Metadata, Fishable } from '../utils/Fishable';
 import { getUrlFromFshDefinition } from '../fhirtypes/common';
+import flatMap from 'lodash/flatMap';
 
 export class FSHTank implements Fishable {
   constructor(public readonly docs: FSHDocument[], public readonly config: Configuration) {}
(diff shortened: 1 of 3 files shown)


mint-thompson (approved):
Looks good to me! One thing I'm curious about: is it necessary for the referenced Instance to have `usage` of "Definition"?

ngfreiter (approved):
Radical 🛹

cmoesel:
> Looks good to me! One thing I'm curious about: is it necessary for the referenced Instance to have `usage` of "Definition"?

I thought about this too... I think it's possible that a definition might exists for example purposes only, and therefore should be usage `Example` (rather than `Definition`).  And other examples might need to reference that example definition by its canonical.  For example (no pun intended), maybe I have Example X that needs to refer to a `Questionnaire` -- but for the purpose of Example X, it doesn't need to refer to a _real_ `Questionnaire` -- therefore, it might refer to an example `Questionnaire` -- but still needs to refer to it by its canonical.  Anyway, I think it's better to be overly permissive than potentially too restrictive.

PR #770, opened 2021-02-25
Fix assigned value on sliced children bug

### LINKED ISSUE #748: Issue with assigning values on sliced instances
I'm not exactly sure what is going wrong here, but when you make the following instance:
```
Alias: Abrechnung = http://fhir.abda.de/eRezeptAbgabedaten/StructureDefinition/Abrechnungszeilen

Instance:  TestInstanceSlicing
InstanceOf: Abrechnung
Usage:  #example
* status = #issued
* type.coding.system = "http://fhir.abda.de/eRezeptAbgabedaten/CodeSystem/invoice-typ"
* type.coding.code.value = "abrechnungszeilen"
* lineItem[0].chargeItemCodeableConcept = http://fhir.de/CodeSystem/ifa/pzn#1
```
which is an instance of this this profile: https://simplifier.net/erezeptabgabedaten/abrechnungszeilen-duplicate-2 (which I've also attached as a text file), you get errors like:
```
error Cannot assign [object Object] to this element; a different Coding is already assigned: {"system":"http://TA1.abda.de"}.
  File: /input/fsh/abrechnungszeilen.fsh
  Line: 9
```

At the very least, the [object Object] is no good, but I'm also pretty sure this error should not exist at all. But this does require further investigation.

[Abrechnungszeile.txt](https://github.com/FHIR/sushi/files/5925759/Abrechnungszeile.txt)

ngfreiter:
This fixes #748. It is easiest to illustrate with an example. You can use the example cited in the issue, but here is a slightly simpler example I've been using:
```
Profile: MyProfile
Parent: Observation
* category.coding ^slicing.discriminator.type = #value
* category.coding ^slicing.discriminator.path = "coding"
* category.coding ^slicing.rules = #open
* category.coding contains slice1 0..1
* category.coding[slice1] = http://example.com#foo

Instance: MyInstance
InstanceOf: MyProfile
* category[0] = http://example.com#bar
* status = #active
* code = #final
```

In this example, `category.coding` is sliced, and the slice has a value set on it, but the slicing is `#open`, so any other values can also be added to the `category.coding` array. But, if you run this example on `master`, you will get an error like:
```
Cannot assign [object Object] to this element; a different Coding is already assigned: {"code":"foo","system":"http://example.com"}.
```
which is a bug, since the `http://example.com#foo` is assigned to a specific slice, not the entire `coding` array. If you run that example on this branch, you will see no errors, and the first instance rule will be correctly applied.

To further test this PR, change the slicing from `#open` to `#closed`. Now slicing is closed, so every element in the `coding` array must match a slice, and since the `http://example.com#bar` value does not, an error will be logged indicating that this value does not satisfy any slice, and cannot be assigned.

Now try changing `http://example.com#bar` to `http://example.com#foo`. This now does satisfy at least one slice in the closed slicing, so the assignment will be allowed.

### DIFF
--- a/src/errors/ValueConflictsWithClosedSlicingError.ts
+++ b/src/errors/ValueConflictsWithClosedSlicingError.ts
@@ -1,0 +1,7 @@
+export class ValueConflictsWithClosedSlicingError extends Error {
+  constructor(public requestedValue: boolean | number | string) {
+    super(
+      `Cannot assign ${requestedValue} to this element since it conflicts with all values of the closed slicing.`
+    );
+  }
+}
--- a/src/errors/index.ts
+++ b/src/errors/index.ts
@@ -44,3 +44,4 @@ export * from './InvalidResourceTypeError';
 export * from './InvalidExtensionParentError';
 export * from './InvalidTypeAccessError';
 export * from './NonAbstractParentOfSpecializationError';
+export * from './ValueConflictsWithClosedSlicingError';
--- a/src/fhirtypes/ElementDefinition.ts
+++ b/src/fhirtypes/ElementDefinition.ts
@@ -38,7 +38,8 @@ import {
   InvalidMappingError,
   InvalidFHIRIdError,
   DuplicateSliceError,
-  NonAbstractParentOfSpecializationError
+  NonAbstractParentOfSpecializationError,
+  ValueConflictsWithClosedSlicingError
 } from '../errors';
 import { setPropertyOnDefinitionInstance, splitOnPathPeriods } from './common';
 import { Fishable, Type, Metadata, logger } from '../utils';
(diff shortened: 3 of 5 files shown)


mint-thompson:
Something I'm curious about here is what ought to happen when the slice is required, that is:
```
* category.coding contains slice1 1..1
```

When I made that change on this branch, the resulting instance was the same as before, the test still passed, and no errors were added. But, that doesn't seem right, since the resulting instance doesn't have that required slice.

ngfreiter:
> Something I'm curious about here is what ought to happen when the slice is required, that is:
> 
> ```
> * category.coding contains slice1 1..1
> ```
> 
> When I made that change on this branch, the resulting instance was the same as before, the test still passed, and no errors were added. But, that doesn't seem right, since the resulting instance doesn't have that required slice.

I think that actually is right in this case on this PR: https://github.com/FHIR/sushi/pull/707, we made SUSHI less strict about slice checking. To know that the required slice is not present, we would need to check the value set here:
```
* category[0] = http://example.com#bar
 ```
against the slicing constraints. We decided this was just too dang tricky, so we don't say anything at all.

mint-thompson:
Ah, okay. Once numeric indices are involved, validation of required slices is no longer a concern. Got it.

mint-thompson (approved):

julianxcarter (approved):
Looks good to me!

PR #771, opened 2021-02-25
Clean up extensions on primitives

### LINKED ISSUE #752: SUSHI does not export extensions correctly when setting extension on primitive using caret rule
Easier to explain by example ([FSH Online](https://fshschool.org/FSHOnline/#/share/36L5EiC)):
```
Profile: MyProfile
Parent: Patient
// This serializes correctly:
* ^extension[MyBooleanExtension].valueBoolean = true
// But this does not serialize correctly:
* ^status.extension[MyBooleanExtension].valueBoolean = true

Extension: MyBooleanExtension
* value[x] only boolean
```

The first extension gets serialized correctly:
```json
{
  "url": "http://example.org/StructureDefinition/MyBooleanExtension",
  "valueBoolean": true
}
```

The second does not:
```json
{
  "_sliceName": "MyBooleanExtension",
  "url": {
    "assignedValue": "http://example.org/StructureDefinition/MyBooleanExtension",
    "_primitive": true
  },
  "valueBoolean": {
    "assignedValue": true,
    "_primitive": true
  }
}
```

mint-thompson:
Fixes #752 and completes task [CIMPL-668](https://standardhealthrecord.atlassian.net/browse/CIMPL-668).

When cleaning up properties on a StructureDefinition, properties are recursively traversed to find and remove those that were added by SUSHI to help in the export process. Properties that were added by SUSHI start with an underscore character. Properties that contain extra attributes of primitives also start with an underscore character, but will have a
sibling property with the same name without the underscore. Therefore, if such a sibling property is present, the property being examined is a true member of the StructureDefinition, and should be included in the cleanup traversal.

Basically, I'm _pretty_ sure that doing this won't ever get us into trouble, but I might be wrong.

### DIFF
--- a/src/export/StructureDefinitionExporter.ts
+++ b/src/export/StructureDefinitionExporter.ts
@@ -564,7 +564,15 @@ export class StructureDefinitionExporter implements Fishable {
     this.preprocessStructureDefinition(fshDefinition, structDef.type === 'Extension');
 
     this.setRules(structDef, fshDefinition);
-    cleanResource(structDef, (prop: string) => prop == 'elements' || prop.indexOf('_') == 0);
+    // Properties added by SUSHI should be skipped. These properties all start with "_",
+    // but so do the extra properties on primitive-type elements. So, check for a sibling
+    // property with the same name, but no underscore, to identify the primitive-type
+    // extra property objects that should _not_ get skipped.
+    cleanResource(
+      structDef,
+      (object: { [key: string]: any }, prop: string) =>
+        prop == 'elements' || (prop.indexOf('_') == 0 && object[prop.slice(1)] == null)
+    );
     structDef.inProgress = false;
 
     // check for another structure definition with the same id
--- a/src/export/StructureDefinitionExporter.ts
+++ b/src/export/StructureDefinitionExporter.ts
@@ -564,14 +564,10 @@ export class StructureDefinitionExporter implements Fishable {
     this.preprocessStructureDefinition(fshDefinition, structDef.type === 'Extension');
 
     this.setRules(structDef, fshDefinition);
-    // Properties added by SUSHI should be skipped. These properties all start with "_",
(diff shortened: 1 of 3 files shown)


ngfreiter (changes requested):
I think the assumption that the sibling property will exist can get us into trouble. If you check this part of the spec: https://www.hl7.org/fhir/json.html#primitive, it does say:
> Note: If the primitive has an id attribute or extension, but no value, only the property with the _ is rendered.

And for the following FHIR:
```
Profile: MyProfile
Parent: Patient
// This serializes correctly:
* ^extension[MyBooleanExtension].valueBoolean = true
// But this does not serialize correctly:
* ^status.extension[MyBooleanExtension].valueBoolean = true
* ^publisher.extension[MyBooleanExtension].valueBoolean = true

Extension: MyBooleanExtension
* value[x] only boolean
```
This PR works for `status`, but the problem still appears on `publisher`:
```
"_`publisher`": {
    "extension": [
      {
        "_sliceName": "MyBooleanExtension",
        "url": {
          "assignedValue": "http://example.org/StructureDefinition/MyBooleanExtension",
          "_primitive": true
        },
        "valueBoolean": {
          "assignedValue": true,
          "_primitive": true
        }
      }
    ]
  }
```
because there is no value for the primitive `publisher` element. I wonder if maybe we should just make the elements that SUSHI uses for tracking start more distinctively. In retrospect it was probably a bad decision to use a character that also has meaning in the FHIR spec (pretty sure this was me, so sorry for that one).

mint-thompson:
@ngfreiter thanks for catching that scenario. So, it seems like it'll be better to have the skip function directly check for the SUSHI-added properties like `_sliceName` and `_primitive`.

julianxcarter (approved):

ngfreiter (approved):
This looks good to me!

PR #773, opened 2021-03-02
Add Constructor Names

### LINKED ISSUE #750: Rule of type i cannot be applied to entity of type i
SUSHI emits the error `Rule of type i cannot be applied to entity of type i` in some cases.  This is quite cryptic.

Reproduce here: https://fshschool.org/FSHOnline/#/share/2O8g9pZ

This is a valid error -- but the error message is not helpful.  To fix the actual error, use a `^` on the rules that start with `expansion`.

jafeltra:
This PR fixes #750.

It turns out that this issue only showed up in FSH Online because we were using `constructor.name` in our error message, and the constructor names are minimized when we build FSH Online for production. I couldn't find a way to avoid this happening in the built code, so I decided to implement this change in SUSHI. This adds a string `constructorName` to each of the rules and entities that are used in the error message. This shouldn't change how SUSHI works, but should let the error message be correct when we get it in FSH Online. The linked issue has an example of how to replicate the error in SUSHI and FSH Online.

### DIFF
--- a/src/fhirtypes/common.ts
+++ b/src/fhirtypes/common.ts
@@ -477,7 +477,7 @@ export function applyInsertRules(
           expandedRules.push(cloneDeep(ruleSetRule));
         } else {
           logger.error(
-            `Rule of type ${ruleSetRule.constructor.name} cannot be applied to entity of type ${fshDefinition.constructor.name}`,
+            `Rule of type ${ruleSetRule.constructorName} cannot be applied to entity of type ${fshDefinition.constructorName}`,
             ruleSetRule.sourceInfo
           );
         }
--- a/src/fshtypes/Extension.ts
+++ b/src/fshtypes/Extension.ts
@@ -2,7 +2,6 @@ import { FshEntity } from './FshEntity';
 import { SdRule } from './rules';
 
 export class Extension extends FshEntity {
-  constructorName = 'Extension';
   id: string;
   parent: string;
   title?: string;
@@ -19,5 +18,9 @@ export class Extension extends FshEntity {
     this.parent = 'Extension'; // init to 'Extension'
     this.mixins = [];
     this.rules = [];
+  }
+
+  get constructorName() {
+    return 'Extension';
   }
 }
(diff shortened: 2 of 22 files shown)


ngfreiter (commented):
This looks good to me!

ngfreiter (dismissed):
Oops I commented instead of approving.

cmoesel (changes requested):
I think this makes sense, but I'd like for the `constructorName` properties to be `readonly` -- because they should never change, right?

jafeltra:
> I think this makes sense, but I'd like for the `constructorName` properties to be `readonly` -- because they should never change, right?

I think that makes sense, but the only slight caveat is that not every `constructorName` can be `readonly`. The ones that are set on classes that are then extended later on can't be `readonly` because then the children can't set their own. Does it make sense to add `readonly` to some, like `CardRule` and `AssignmentRule`, but not on others, like `Rule` and `ValueSetComponentRule`?

cmoesel:
> I think that makes sense, but the only slight caveat is that not every `constructorName` can be `readonly`.

Ah shoot.  I thought TypeScript might allow a child class to set an inherited `readonly` property on instantiation.  Apparently not. I _think_ I'd still rather have it be `readonly` where it can be, but I'm definitely less sure now.

How do you feel about using `getter` functions instead? This makes it effectively read-only, but also allows child classes to override it:
```ts
get constructorName() {
    return 'Rule';
}
```
and you still access it exactly the same way as if it were a normal property...

jafeltra:
> How do you feel about using `getter` functions instead?

I actually considered this approach and started out with it. I talked with @mint-thompson about potentially using an abstract field and when that didn't work out either, I guess I never went back. I'd be good with a `getter`!

jafeltra:
I was testing this out on a fork to make sure this change fixed the issue in the FSH Online production build, but something seems to be failing. I want to make sure that works before merging this.

jafeltra:
Okay false alarm! Something must have been out of sync because I re-packed this branch and used it in a FSH Online fork and the cryptic error message is no longer cryptic. I think this should be set for re-review @ngfreiter and @cmoesel. 

Also, the page it's built at is https://jafeltra.github.io/FSHOnline/#/ if you want to see the no-longer-cryptic error message.

ngfreiter (approved):
Tested it out, and didn't look cryptic to me 👍 .

cmoesel (approved):
Let's do this!

PR #775, opened 2021-03-08
OnlyRule with profile on sliced element uses profiles for type intersections on slices

### LINKED ISSUE #769: Incorrect slice elimination when applying profile to extension element.
```
Profile: MyQuestionnaire
Parent: http://hl7.org/fhir/StructureDefinition/cqf-questionnaire
* extension only http://hl7.org/fhir/StructureDefinition/cqf-library
```

When applying this profile to extension, the library slice on extension (which is defined on the parent as having profile http://hl7.org/fhir/StructureDefinition/cqf-library) is incorrectly determined to have all of its types eliminated.

mint-thompson:
Fixes #769 and completes task [CIMPL-674](https://standardhealthrecord.atlassian.net/browse/CIMPL-674).

When an OnlyRule with one or more profiles is applied to a sliced element, try to use the profile (rather than just the base type) when finding the type intersection between the new types and the types that currently exist on a slice. Previously, when the base type was used for this calculation, a slice which was constrained to a profile would find an empty type intersection, even when the type being applied to the sliced element was the same profile as the one on the slice.

The first test case uses FSH from the reported issue. The second test case is to show that when multiple profiles are allowed on the sliced element, the slice's type stays the same.


cmoesel (approved):
Well done.  Seems like the appropriate fix to me.  Thanks, @mint-thompson!

ngfreiter (approved):
This looks good to me!

PR #784, opened 2021-03-21
AddElementRule, Logical, and Resource classes

ngfreiter:
This completes [CIMPL-688](https://standardhealthrecord.atlassian.net/browse/CIMPL-688) by adding an `AddElementRule` class, a `Logical` class, and a `Resource` class.

I think @cmoesel should be one of the reviewers, since he designed the syntaxes that these classes represent, and I want to make sure I didn't misunderstand any of the requirements.

### DIFF
--- a/src/fshtypes/FshStructure.ts
+++ b/src/fshtypes/FshStructure.ts
@@ -1,12 +1,12 @@
 import { FshEntity } from './FshEntity';
-import { SdRule, AddElementRule } from './rules';
+import { Rule } from './rules';
 
-export abstract class Model extends FshEntity {
+export abstract class FshStructure extends FshEntity {
   id: string;
   parent?: string;
   title?: string;
   description?: string;
-  rules: (SdRule | AddElementRule)[];
+  rules: Rule[];
 
   constructor(public name: string) {
     super();
--- a/src/fshtypes/FshStructure.ts
+++ b/src/fshtypes/FshStructure.ts
@@ -15,6 +15,6 @@ export abstract class FshStructure extends FshEntity {
   }
 
   get constructorName() {
-    return 'Model';
+    return 'FshStructure';
   }
 }
--- a/src/fshtypes/Model.ts
+++ b/src/fshtypes/Model.ts
@@ -1,0 +1,20 @@
+import { FshEntity } from './FshEntity';
+import { SdRule, AddElementRule } from './rules';
+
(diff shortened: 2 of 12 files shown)


src/fshtypes/Model.ts
```diff
@@ -0,0 +1,20 @@
+import { FshEntity } from './FshEntity';
+import { SdRule, AddElementRule } from './rules';
+
+export abstract class Model extends FshEntity {
```
cmoesel, line 4: This is very close to the same definition as `Profile` (and only one more degree different than `Extension`).  Neither of those have a base class other than `Entity`.  Do you think it might make sense to have them all share a base class that has `Rules: Rule[]` and override that in the sub-classes?

I'm not sure if `Model` is quite the right word (whether shared among all four or just these two) -- but I'm struggling to find a better one...  I guess `Structure` might be more accurate in context, but maybe would be confusing due to the existence of a `StructureDefinition` class as well?  I guess my concern w/ `model` is that people might associate it exclusively with Logical Models.

cmoesel (commented):
I think this matches the syntax just fine.  I had one question I wanted @ngfreiter's further input on.

src/fshtypes/Model.ts
ngfreiter: Ok @cmoesel, https://github.com/FHIR/sushi/pull/784/commits/0235b08b4c970a81125832112f9e2419a08b9b21 adds a `FshStructure` class which `Profile` and `Extension` now also inherit from.

src/fshtypes/FshStructure.ts
```diff
@@ -0,0 +1,20 @@
...
+  rules: Rule[];
+
+  constructor(public name: string) {
+    super();
+    this.id = name; // init same as name
+    this.rules = [];
+  }
+
+  get constructorName() {
+    return 'Model';
```
cmoesel, line 18:
```suggestion
    return 'FshStructure';
```

cmoesel (changes requested):
Nice.  I like it.  The common base class cleans things up a bit -- so that looks good to me!  I just found one thing you missed when converting `Model` to `FshStructure`.  Once that is fixed, we're all good!

cmoesel (approved):
Yes please.

mint-thompson (approved):
hooray!!

Issue #759, opened 2021-02-15
Extra extensions in instances


cmoesel:
In some cases, two FSH rules that should be modifying the same extension are actually creating and modifying *different* extensions.  See [this test case](https://fshschool.org/FSHOnline/#/share/2Zhcys3):

```
Profile: TestMoney
Parent: Money
* value 1..1 MS 
* currency 1..1
* currency = #EUR

Extension: TestExtension
* extension contains money 1..1 and test 0..1
* extension[money].value[x] only TestMoney
* extension[money].value[x] 1..1
* extension[test].value[x] only string

Profile: TestProfile
Parent: Observation
* extension contains TestExtension named test 1..1

Instance: TestObs
InstanceOf: TestProfile
* status = #final
* code = #testcode
* extension[TestExtension].extension[money].valueMoney.value = 10
* extension[TestExtension].extension[money].valueMoney.currency = #EUR
```

The `TestObs` instance should have a single `TestExtension` extension with a single `money` sub-extension w/ value `10` and currency `#EUR`.  Instead, we end up with two `TestExtension` extensions, one with only the `currency` set and another with both `currency` and `value` set:
```json
{
  "resourceType": "Observation",
  "id": "TestObs",
  "meta": {
    "profile": [
      "http://example.org/StructureDefinition/TestProfile"
    ]
  },
  "extension": [
    {
      "url": "http://example.org/StructureDefinition/TestExtension",
      "extension": [
        {
          "url": "money",
          "valueMoney": {
            "currency": "EUR"
          }
        }
      ]
    },
    {
      "extension": [
        {
          "valueMoney": {
            "currency": "EUR",
            "value": 10
          },
          "url": "money"
        }
      ],
      "url": "http://example.org/StructureDefinition/TestExtension"
    }
  ],
  "status": "final",
  "code": {
    "coding": [
      {
        "code": "testcode"
      }
    ]
  }
}
```

Discovered while trying to reproduce a different bug in [this zulip thread](https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/value.5Bx.5D.20confusion.20on.20Instances/near/226386860).

cmoesel:
[...]

PR #787, opened 2021-03-26
Fix Extensions on ElementDefs

ngfreiter:
This fixes https://github.com/FHIR/sushi/issues/759.

The issue here was not actually related to child profiles, as is pointed out in the most recent comment. The issue was just that when we call the `cleanResource` function to clean a StructureDefinition of all of the extra properties that SUSHI puts on it, we ignored the `elements` array, because we didn't think those extra properties would end up there. But, in this example:
```
Profile: MyProfile
Parent: Patient
// This does not serialize correctly
* name ^extension[MyBooleanExtension].valueBoolean = true


Extension: MyBooleanExtension
* value[x] only boolean
```
Those elements do end up in the `elements` array, and are not correctly cleaned. This PR fixes that by now cleaning the `elements` array. We do still have to avoid the `structDef` element that exists on an ElementDefinition, because otherwise we will infinitely recurse.

You can test this using the example above, or by running the test on https://github.com/FHIR/sushi/commit/52609f626d8a8c8dbad9ec927cd79e10c94d9da5, where it should fail.

### DIFF
--- a/src/export/StructureDefinitionExporter.ts
+++ b/src/export/StructureDefinitionExporter.ts
@@ -567,7 +567,7 @@ export class StructureDefinitionExporter implements Fishable {
     // The elements list does not need to be cleaned up.
     // And, the _sliceName and _primitive properties added by SUSHI should be skipped.
     cleanResource(structDef, (prop: string) =>
-      ['elements', '_sliceName', '_primitive'].includes(prop)
+      ['structDef', '_sliceName', '_primitive'].includes(prop)
     );
     structDef.inProgress = false;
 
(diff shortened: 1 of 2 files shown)


julianxcarter (dismissed):
LGTM!

src/export/StructureDefinitionExporter.ts
```diff
@@ -567,7 +567,7 @@ export class StructureDefinitionExporter implements Fishable {
     // The elements list does not need to be cleaned up.
```
jafeltra, line 567: I might owe you something for this comment, but I don't think we need this line anymore.

jafeltra (commented):
I ran the regression on the selected repos and didn't see any changes! I did leave one _very_ small comment 😰

src/export/StructureDefinitionExporter.ts
ngfreiter: For 4 days I have toiled to fix this issue. But at long last, I think (_think_) I've come up with a solution.

jafeltra: After reviewing these changes for a full day, they look good to m!

jafeltra (approved):

julianxcarter (approved):

PR #788, opened 2021-03-26
Skip nonexistent files when traversing directories

### LINKED ISSUE #780: SUSHI chokes on emacs lock files
SUSHI seems to issue errors and ultimately fail processing when emacs lock files are in the SUSHI project path.  SUSHI should probably recognize these files and skip processing on them.

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/emacs.20love

mint-thompson:
Fixes #780 and completes task [CIMPL-708](https://standardhealthrecord.atlassian.net/browse/CIMPL-708).

- When is a door not a door?
- When it's a jar.
***
- When is a file not a file?
- When it's listed by `readdir`, but is a symbolic link with a nonexistent target.
***
A symbolic link with an invalid target will be listed when reading a directory, but it will not be possible to retrieve information about it. Therefore, it should be filtered out before continuing to traverse directories. This is primarily being done to resolve an issue with emacs lock files.

Additionally, I changed `ancestors` to `descendants` for the sake of accuracy in variable naming.

A note about the added test: I had to write the setup this way in order to get consistent behavior across systems, since non-admin Windows users can't create symbolic links. Hopefully, the test pipeline decides that what I've written is acceptable.

### DIFF
--- a/src/utils/Processing.ts
+++ b/src/utils/Processing.ts
@@ -455,8 +455,11 @@ export async function init(): Promise<void> {
 
 function getFilesRecursive(dir: string): string[] {
   if (fs.statSync(dir).isDirectory()) {
-    const ancestors = fs.readdirSync(dir, 'utf8').map(f => getFilesRecursive(path.join(dir, f)));
-    return [].concat(...ancestors);
+    const descendants = fs
+      .readdirSync(dir, 'utf8')
+      .filter(f => fs.existsSync(path.join(dir, f)))
+      .map(f => getFilesRecursive(path.join(dir, f)));
+    return [].concat(...descendants);
   } else {
     return [dir];
   }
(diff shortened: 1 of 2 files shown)


julianxcarter (approved):
This does the job for me! I tested this by creating an invalid symbolic link and placing it in the `fsh` folder of a test project, and sushi skips right over the file just as we'd want it to.

jafeltra (approved):
This looks great! Thanks for making the update to the tests - everything gets cleaned up as expected now! I also recreated the error with the released version of SUSHI and I get no such error on this branch!

julianxcarter (approved):
My approval still stands. Hooray for clean code!

litlfred:
As the emacs/fhir person that initiated this bug report, just wanted to say thanks!

Issue #106, opened 2020-01-19
Intensional Value Sets not expanding


markkramerus:
FHIR is not able to expand value sets defined by filter rules. Look at the HistologyMorphologyBehaviorVS in the current master branch, and compare to
http://build.fhir.org/ig/HL7/fhir-mCODE-ig/branches/master/ValueSet-onco-core-HistologyMorphologyBehaviorVS.html
I'm not sure if the problem is in the rule definitions, or the way we are producing the value sets. Here are the rules as the currently are written:
```
ValueSet:   HistologyMorphologyBehaviorVS
Title: "Histology Morphology Behavior Value Set"
Description: "Codes representing etc..."
* codes from system SCT where concept is-a #367651003 "Malignant neoplasm of primary, secondary, or uncertain origin (morphologic abnormality)"
* codes from system SCT where concept is-a #399919001 "Carcinoma in situ - category (morphologic abnormality)"
* codes from system SCT where
    concept is-a #399983006 "In situ adenomatous neoplasm - category (morphologic abnormality)" and
    concept is-not-a #450893003 "Papillary neoplasm, pancreatobiliary-type, with high grade intraepithelial neoplasia (morphologic abnormality)" and
    concept is-not-a #128640002 "Glandular intraepithelial neoplasia, grade III (morphologic abnormality)" and
    concept is-not-a #450890000 "Glandular intraepithelial neoplasia, low grade (morphologic abnormality)" and
    concept is-not-a #703548001 "Endometrioid intraepithelial neoplasia (morphologic abnormality)"
* SCT#399878004 "Malignant neoplasm with pilar differentiation (morphologic abnormality)"
```

cmoesel:
When I tested this on Friday, it worked.  Here is what the FHIR publisher produced for the HistologyMorphologyBehaviorVS:
![image](https://user-images.githubusercontent.com/2278253/72813477-609a2400-3c31-11ea-9df4-ba609a6b5402.png)

markkramerus:
Could not reproduce this bug. Closing.

PR #790, opened 2021-03-29
parseFSHPath regex fix

julianxcarter:
This is a companion PR to [PR #106 on the GoFSH side](https://github.com/FHIR/GoFSH/pull/106). The regex expression within `parseFSHPath` is currently `^[0-9]$` when it should really be `^[0-9]+$` to take multi-digit integers into account.


mint-thompson (changes requested):
Looks correct to me! Could you add a test case for this to the `resolveSoftIndexing` test suite?

jafeltra (approved):
The change and test cases look good to me too!

PR #792, opened 2021-04-01
Soft Indexing sequences can now being at non-zero indices

julianxcarter:
This PR addresses another issue found in [PR #106](https://github.com/FHIR/GoFSH/pull/106) on GoFSH. When unwinding soft indexing, Sushi currently makes the assumption that the first element being assigned by an array utilizing soft indexing will always be at index 0. In the event that sushi first encounters an array as it references a non-zero element, this could result in something like the following... 
```
* effective[x] ^constraint[1].key = "us-core-1"
* effective[x] ^constraint[=].severity = #error
* effective[x] ^constraint[=].human = "Datetime must be at least to day."
```
resolving to this:
```
* effective[x] ^constraint[1].key = "us-core-1"
* effective[x] ^constraint[0].severity = #error
* effective[x] ^constraint[0].human = "Datetime must be at least to day."
```
This PR fixes this bug, taking numeric indices into account at all times. The issue can be illustrated by running the unit test included in commit `6035ded`, which should fail in that commit but pass in commit `037cd63` which adds the fix.

### DIFF
--- a/src/utils/PathUtils.ts
+++ b/src/utils/PathUtils.ts
@@ -77,18 +77,17 @@ function convertSoftIndexes(element: PathPart, pathMap: Map<string, number>) {
     const existingNumericBracket = element.brackets?.find(bracket => indexRegex.test(bracket));
     if (existingNumericBracket) {
       pathMap.set(mapName, parseInt(existingNumericBracket));
-      return;
     } else {
       pathMap.set(mapName, 0);
-    }
-    if (element.brackets?.includes('+')) {
-      element.brackets[element.brackets.indexOf('+')] = '0';
-    } else if (element.brackets?.includes('=')) {
-      // If a sequence begins with a '=', we log an error but assume a value of 0
-      element.brackets[element.brackets.indexOf('=')] = '0';
-      throw new Error(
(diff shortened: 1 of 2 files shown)


jafeltra (approved):
These changes and the test make sense to me!

cmoesel:
This seems reasonable to me, but I'm going to run a full regression to ensure it doesn't affect any existing IGs (or if it does, that it affects them in a good way).

cmoesel:
This did well in regression.  In fact, it fixed a (previously undetected) bug in the mCODE IG generation because mCODE uses a numeric index followed by a `[=]` here: https://github.com/HL7/fhir-mCODE-ig/blob/master/input/fsh/EX_Bundle.fsh#L21-L22

So... way to go!

That said, @ngfreiter has asked that we not merge this until he has a chance to review and test it in combination w/ the GoFSH PR that raised this issue.

cmoesel (approved):
Approve, pending @ngfreiter's additional review.

ngfreiter (approved):
Sorry for the delay, I've confirmed that when running this with https://github.com/FHIR/GoFSH/pull/106, everything is working as we would want, so awesome!

PR #793, opened 2021-04-07
Changes to import-related code to support logical models and resources

joepaquette:


joepaquette:
CIMPL-689

npm-shrinkwrap.json
```diff
@@ -1,8 +1,11686 @@
 {
   "name": "fsh-sushi",
   "version": "1.3.1",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```
cmoesel, line 4: It looks like you are using NPM 7.  We attempted to move to NPM 7 previously (I forget exactly when), but ended up having to keep it on NPM 6 due to some compatibility issues (which I don't currently remember).  It's probably worth evaluating again, but maybe not as part of this PR.

Since you did not actually change any dependencies, can we just revert this file back to the version on `master` for now?

cmoesel (changes requested):
This is great, @joepaquette!  Really well-thought out code and tests.  No one would ever guess you're not a long-time member of this team!

I didn't really find anything _wrong_ with any of this -- but I did find a few opportunities for streamlining some of the code just a little bit more.  Let me know what you think!

ngfreiter (commented):
This looks great, I only found one small thing on top of Chris's comments. The only other thing we will need to track is how this related to https://github.com/FHIR/sushi/pull/797. That PR adds the capability to indent FSH to indicate context, so for example:
```
* name MS
  * family MS 
```
is the same as
```
* name MS
* name.family MS
```
I think we will want to support that on `Logical` and `Resource`. It's a little tricky to track since this is going into a feature branch and #797 is going into `master`. I think the most likely situation is that this PR goes into the feature branch, then #797 is merged into master, and since you'll already be on export at that point, I'll make a small PR to the feature branch to add support for context paths.

src/import/FSHImporter.ts
```diff
@@ -906,6 +978,69 @@ export class FSHImporter extends FSHVisitor {
...
+      location: this.extractStartStop(ctx)
+    });
+    return [];
+  }
+
+  visitAddElementRule(ctx: pc.AddElementRuleContext): AddElementRule {
+    const path = ctx.path() ? this.visitPath(ctx.path()) : '';
+    const addElementRule = new AddElementRule(path)
+      .withLocation(this.extractStartStop(ctx))
+      .withFile(this.currentFile);
```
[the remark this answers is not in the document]
joepaquette: Chris created the `logical-models` branch off `master` as an integration branch. All of my PRs will be to merge into the `logical-models` branch. Once I complete all of my work and everything has been signed off, we'll merge `master` into `logical-models` and then resolve all merge conflicts. Once those are resolved, we can merge `logical-models` into `master`.

test/import/FSHImporter.Logical.test.ts
```diff
@@ -0,0 +1,556 @@
...
+
+        // Since we'll do the same thing over and over (and over), create a function for it
+        const testToken = (token: string) => {
+          const input = `
+          Logical: ${token}
+          Parent: Observation
+          * value[x] only boolean
+          `;
+          const result = importSingleText(input);
+          expect(loggerSpy.getAllLogs('error')).toHaveLength(0);
```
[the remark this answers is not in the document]
joepaquette: Good catch! Copy/paste sometimes bites me 😃. 

The Jest setup/teardown methods **only** apply within a given test file regardless of the `describe` description text. There is a [specific ordering](https://jestjs.io/docs/setup-teardown) of how/when the various setup/teardown functions are called within a test file.

I don't know why the lack of a reset didn't cause any issues in the affected test suites, but I've added the missing `beforeEach()` to reset the `loggerSpy` mocks in the new test files to be safe.

ngfreiter (dismissed):
Looks good to me!

src/import/FSHImporter.ts
```diff
@@ -1509,21 +1491,34 @@ export class FSHImporter extends FSHVisitor {
...
+      if (FLAGS.includes(orType.type)) {
+        logger.warn(
+          `The targetType '${orType.type}' appears to be a flag value rather than a valid target data type.`,
+          {
+            file: this.currentFile,
+            location: this.extractStartStop(ctx)
+          }
+        );
+      }
+    });
```
cmoesel, line 1519: I think this is more appropriate in the specific code for `AddElementRule` rather than in this shared function.  The reason for this is because there is no ambiguity w/ the grammar for `OnlyRule`s.  If the FSH says `foo only MS` then there really is no confusion -- only a type would ever go after the word `only`.  As opposed to `AddElementRule` where `* foo 0..1 MS` is a little more ambiguous since the word after `0..1` can be a _flag_ or a _type_ (although, in the context of the rest of the phrase, I think we can still figure it out).

cmoesel (changes requested):
Awesome.  I only had one more little tweak of a change that I would suggest.  If you disagree w/ the suggestion, however, I'm open to discussing it!

src/import/FSHImporter.ts
joepaquette: Good catch - fixed.

cmoesel (approved):
I totally dig this PR.  Approved!

PR #796, opened 2021-04-08
More helpful error message on incorrect bindings

### LINKED ISSUE #538: SUSHI should provide more helpful error message when authors try to bind to a Code System
Attempting to bind a property to a Code System is a common enough mistake, that SUSHI should detect this and provide a more helpful error message.  For example, see: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Is.20not.20a.20valid.20URI

Similarly, there are cases where authors might try to use a Value Set where a Code System is needed (for example `MyVS#mycode`).  We should also consider more helpful error messages here.

julianxcarter:
Fixes #538 and addressed [CIMPL-470](https://standardhealthrecord.atlassian.net/browse/CIMPL-470). This PR makes it so that sushi outputs a more helpful error message when a user attempts to bind a property to a Code System rather than a Value Set. This more helpful error message is also output when a user attempts to assign the system on a coding to a Value Set rather than a Code System. 

The existing error messages that are output can be viewed by [running this example FSH on FshOnline](https://fshschool.org/FSHOnline/#/share/3cZW2nT). When run on this branch, you should get two error messages that let you know if you need a CS or a VS on that line.


jafeltra (approved):
Makes sense to me!

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1642,8 +1643,16 @@ export class ElementDefinition {
    * @throws {InvalidUriError} when the system being assigned is not a valid uri
    */
   private assignFshCode(code: FshCode, exactly = false, fisher?: Fishable): void {
-    if (code.system && !isUri(code.system.split('|')[0])) {
-      throw new InvalidUriError(code.system);
+    if (code.system) {
+      const csURI = code.system.split('|')[0];
+      if (fisher) {
+        const vsURI = fisher.fishForMetadata(code.system, Type.ValueSet)?.url.split('|')[0] ?? '';
+        if (isUri(vsURI)) {
```
ngfreiter, line 1650: Is it necessary to check if the `vsURI` is a URI here? I can't figure out why we need to do that. We are getting a `url`, so it is either going to exist or not, but if it exists won't it always be a `url`? Isn't the question here more if the input `code.system` resolves to a ValueSet? I don't think we need to then check that the ValueSet has a valid URI, but I could be missing something here.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1642,8 +1643,16 @@ export class ElementDefinition {
    * @throws {InvalidUriError} when the system being assigned is not a valid uri
    */
   private assignFshCode(code: FshCode, exactly = false, fisher?: Fishable): void {
-    if (code.system && !isUri(code.system.split('|')[0])) {
-      throw new InvalidUriError(code.system);
+    if (code.system) {
+      const csURI = code.system.split('|')[0];
+      if (fisher) {
+        const vsURI = fisher.fishForMetadata(code.system, Type.ValueSet)?.url.split('|')[0] ?? '';
```
ngfreiter, line 1649: Here we could probably just do:
```javascript
        const vsURI = fisher?.fishForMetadata(code.system, Type.ValueSet)?.url.split('|')[0] ?? '';
```
and avoid having the check for `fisher`.

src/export/StructureDefinitionExporter.ts
```diff
@@ -204,7 +206,15 @@ export class StructureDefinitionExporter implements Fishable {
             const target = structDef.getReferenceName(rule.path, element);
             element.constrainType(rule, this, target);
           } else if (rule instanceof BindingRule) {
-            const vsURI = this.fishForMetadata(rule.valueSet, Type.ValueSet)?.url ?? rule.valueSet;
+            const vsURI =
+              this.fishForMetadata(rule.valueSet, Type.ValueSet)?.url.split('|')[0] ??
+              rule.valueSet;
```
ngfreiter, line 211: I think this will end up changing the `vsURI` that we bind to so that the ValueSet cannot have a version appended. Here is an example of how this could happen:
```
Profile: MyPatient
Parent: Patient
* identifier.type from MyVS

ValueSet: MyVS
* ^url = "http://valueset.org|4.1"
```
on this branch, that `4.1` is lost when binding, but on `master`, it is preserved.

ngfreiter (changes requested):
I noticed a few things to change.

src/fhirtypes/ElementDefinition.ts
ngfreiter: I guess it won't always necessarily be a value URI, because a user could use a caret rule to set it to something invalid. But I'm still a bit unsure about if we need to actually be checking if these are URIs or not.

julianxcarter: No you're totally right! I could just check if the `vsURI` is truthy or not, rather than checking if the URI is valid.

src/export/StructureDefinitionExporter.ts
julianxcarter: Addressed in `e8b6e3f`.

src/fhirtypes/ElementDefinition.ts
julianxcarter: Addressed in `e8b6e3f`.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1642,8 +1643,16 @@ export class ElementDefinition {
...
+        if (isUri(vsURI)) {
+          throw new MismatchedBindingTypeError(code.system, this.path, 'Code System');
```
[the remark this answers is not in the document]
julianxcarter: I addressed this in the most recent commit; though I opted to just change the language used in `MismatchedBindingTypeError` when the target type is a Code System rather than creating a new error class. If you still think that it's weird to throw a `MismatchedBindingTypeError` in this case at all then I can create a new class!

src/fhirtypes/ElementDefinition.ts
julianxcarter: Addressed in `e8b6e3f`.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1642,8 +1643,14 @@ export class ElementDefinition {
    * @throws {InvalidUriError} when the system being assigned is not a valid uri
    */
   private assignFshCode(code: FshCode, exactly = false, fisher?: Fishable): void {
-    if (code.system && !isUri(code.system.split('|')[0])) {
-      throw new InvalidUriError(code.system);
+    if (code.system) {
+      const csURI = code.system.split('|')[0];
+      const vsURI = fisher?.fishForMetadata(code.system, Type.ValueSet)?.url ?? '';
+      if (vsURI) {
+        throw new MismatchedBindingTypeError(code.system, this.path, 'Code System');
```
ngfreiter, line 1650: I notice that here you use "Code System", but in the other spot you use "ValueSet". Would you be able to update this to be "CodeSystem"? I think I'd live a happier life if I could just let that " " go, but oh well.

src/errors/MismatchedBindingTypeError.ts
```diff
@@ -0,0 +1,9 @@
+export class MismatchedBindingTypeError extends Error {
+  constructor(public id: string, public path: string, public correctedType: string) {
+    super(
+      `Cannot ${
+        correctedType === 'Code System' ? 'assign' : 'bind'
```
ngfreiter, line 5: And you will need to update here too.

ngfreiter (commented):
Looks good, I did have one small comment.

cmoesel:
I appreciate the rhyming commit messages, @julianxcarter.  I think perhaps we should make that a requirement for all contributions.

ngfreiter (approved):
Looks good to me!

julianxcarter:
> I appreciate the rhyming commit message, @julianxcarter. I think perhaps we should make that a requirement for all contributions.

@cmoesel I wholeheartedly agree! @ngfreiter already committed a rhyming message yesterday so I think he'd be onboard as well.

cmoesel:
> I appreciate the rhyming commit message, @julianxcarter. I think perhaps we should make that a requirement for all contributions.

I'm sorry, it seems, that I broke my own rule.
For a comment's a contribution, so I am a fool.

This is what I should have said,
Had my body had its head:

> I appreciate that your commit messages rhyme.
> I think we should do this all of the time.

jafeltra (approved):
This looks good to me still!

PR #799, opened 2021-04-12
Loosen restrictions on constraining profiled types

### LINKED ISSUE #789: SUSHI is too strict when constraining profiled types
This is easiest to see via an example. If we have the following FSH:
```
Profile: MyDoc
Parent: ch-core-document
* entry[Composition].resource only MyComp

Profile: MyComp
Parent: http://fhir.ch/ig/ch-core/StructureDefinition/ch-core-composition
```
where the `Composition` slice is defined as:
```
   {
        "id" : "Bundle.entry:Composition.resource",
        "path" : "Bundle.entry.resource",
        "min" : 1,
        "type" : [
          {
            "code" : "Resource",
            "profile" : [
              "http://fhir.ch/ig/ch-core/StructureDefinition/ch-core-composition"
            ]
          }
        ]
      }
```
Then SUSHI does not allow the constraint on the `MyDoc` profile, giving the error:
```
The type "MyComp" does not match any of the allowed types: http://fhir.ch/ig/ch-core/StructureDefinition/ch-core-composition
```

This type constraint really should be allowed, because the `MyComp` profile is a sub-type of Resource, and it conforms to the desired profile. We are too strict in our type checking, since we require that `MyComp` is exactly the same type as the `type.code` (So in this case, changing that from `Resource` to `Composition` works), when we should probably just require that `MyComp` is a sub-type of the `type.code`.

ngfreiter:
Fixes #789 

If we have a situation like the one below:
```
Profile: MyBundle
Parent: Bundle
* entry.resource only http://hl7.org/fhir/StructureDefinition/resprate
* entry.resource ^type.code = "Resource"

Profile: MyResp
Parent: http://hl7.org/fhir/StructureDefinition/resprate

Profile: MyBundleChild
Parent: MyBundle
* entry.resource only MyResp
```
Then we get an error that looks like:
```
The type "MyResp" does not match any of the allowed types: http://hl7.org/fhir/StructureDefinition/resprate
```

This is because when we check if the `MyResp` profile can be applied as a type to `entry.resource`, we check that inherits from the current profile on `entry.resource`, and it does, since it is a child of `resprate`. But we also check that `MyResp`'s type exactly matches the `type.code` on the type constraint applied to `entry.resource`. That is not the case here, because `MyResp` is of type `Observation`, and the `type.code` is `Resource`. But, since `Observation` is a child of `Resource`, this should really still be allowed. This PR fixes this issue so that this type constraint is now allowed.

To test this, you can use the example above on `master` and observer the error, and then see that on this branch the error goes away. You can also run the tests added in 3f8bf05 and see that one fails (the other test which passes is just an implementation of an old skipped test which we had no reason to skip anymore). Then on this branch, all tests should pass.

### DIFF
--- a/src/fhirtypes/ElementDefinition.ts
+++ b/src/fhirtypes/ElementDefinition.ts
@@ -847,11 +847,13 @@ export class ElementDefinition {
             (t2.targetProfile == null || t2.targetProfile.includes(md.url))
         );
       } else {
-        // Look for exact match on the code (w/ no profile) or a match on the same base type with
+        // Look for exact match on the code (w/ no profile) or a match on an allowed base type with
         // a matching profile
         matchedType = targetTypes.find(t2 => {
           const matchesUnprofiledResource = t2.code === md.id && isEmpty(t2.profile);
-          const matchesProfile = t2.code === md.sdType && t2.profile?.includes(md.url);
+          const matchesProfile =
+            this.getTypeLineage(md.sdType, fisher).some(ancestor => ancestor.sdType === t2.code) &&
+            t2.profile?.includes(md.url);
           // True if we match an unprofiled type that is not abstract, is a parent, and that we are
           // specializing (the type does not match the sdType of the type to match)
           specializationOfNonAbstractType =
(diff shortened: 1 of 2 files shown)


jafeltra (approved):
Makes sense to me!

julianxcarter (approved):
This does the trick for me!

PR #802, opened 2021-04-19
Changes to export-related code to support logical models

joepaquette:


src/errors/InvalidProfileParentError.ts
```diff
@@ -0,0 +1,8 @@
+import { WithSource } from './WithSource';
+import { SourceInfo } from '../fshtypes';
+
+export class InvalidProfileParentError extends Error implements WithSource {
+  constructor(public childName: string, public parentName: string, public sourceInfo: SourceInfo) {
+    super(`Parent ${parentName} is not a valid Parent for Profile ${childName}.`);
```
ngfreiter, line 6: Here I am also not sure about this. Why _can't_ a profile just inherit from a logical model. I guess this isn't something we currently support in SUSHI, because we don't add logical models to our `FHIRDefinitions`, so you couldn't inherit from one. But I'm not able to figure out why we need to prohibit this.

src/errors/InvalidMustSupportError.ts
```diff
@@ -0,0 +1,10 @@
+import { Annotated } from './Annotated';
+
+export class InvalidMustSupportError extends Error implements Annotated {
+  specReferences = ['http://hl7.org/fhir/R4/profiling.html#mustsupport'];
```
ngfreiter, line 4: Is https://www.hl7.org/fhir/elementdefinition.html#interpretation another good reference to have for why this is disallowed?

src/export/StructureDefinitionExporter.ts
```diff
@@ -131,13 +375,31 @@ export class StructureDefinitionExporter implements Fishable {
...
+    structDef.abstract = false; // always reset to false, assuming most children of abstracts aren't abstract; can be overridden w/ rule
+    // context and contextInvariant only apply to extensions.
     // keep context, assuming context is still valid for child extensions
     // keep contextInvariant, assuming context is still valid for child extensions
-    // keep type since this should not change for profiles or extensions
+    if (!(fshDefinition instanceof Extension)) {
+      // Should not be defined for profiles, logical models, but clear just to be sure
+      delete structDef.context;
+      delete structDef.contextInvariant;
+    }
```
ngfreiter, line 390: Could this go in an `else` from the `if` on line 404?

src/export/StructureDefinitionExporter.ts
```diff
@@ -131,13 +375,31 @@ export class StructureDefinitionExporter implements Fishable {
...
+    // keep type since this should not change except for logical models or resources
+    if (fshDefinition instanceof Logical || fshDefinition instanceof Resource) {
+      // By definition, the 'type' is the same as the 'id'
+      structDef.type = fshDefinition.id;
+    }
     structDef.baseDefinition = baseURL;
-    structDef.derivation = 'constraint'; // always constraint since SUSHI only supports profiles/extensions right now
+    if (fshDefinition instanceof Logical || fshDefinition instanceof Resource) {
+      structDef.derivation = 'specialization';
```
ngfreiter, line 398: Do we want to just combine these two `if` statements since they have the same predicate.

src/export/StructureDefinitionExporter.ts
```diff
@@ -590,7 +977,7 @@ export class StructureDefinitionExporter implements Fishable {
   }
 
   /**
-   * Exports Profiles and Extensions to StructureDefinitions
+   * Exports Profiles, Extensions, and Logical models to StructureDefinitions
```
ngfreiter, line 980: And Resources right?

src/export/StructureDefinitionExporter.ts
```diff
@@ -488,82 +885,72 @@ export class StructureDefinitionExporter implements Fishable {
...
 
+    // At this point, 'structDef' contains the parent's metadata and elements; therefore,
+    // transform the current 'structDef' from the parent to the new StructureDefinition to
+    // be exported.
+    // Reset the parent's 'structDef.elements' as required.
+    this.resetParentElements(structDef, fshDefinition);
+    // Reset the original parent's metadata to that for the new StructureDefinition
     this.setMetadata(structDef, fshDefinition);
+    // Apply the AddElementRules to create the new elements before any other processing
+    this.applyAddElementRules(structDef, fshDefinition);
```
ngfreiter, line 929: I think having this up here before `applyInsertRules` on line 948 will cause us issues. Because what if an `addElementRule` is inside the `RuleSet` pointed to by an `insertRule`. But in trying to test this I am seeing that an `addElementRule` cannot be on a `RuleSet`. I know @joepaquette that grammar was set before you got here. I can't remember if that was intentional, or if we should add that. @mint-thompson any thoughts?

ngfreiter (commented):
Looks great so far. I have a few questions out of my initial review. Mainly related to what is allowed to be a parent of what.

joepaquette:
@ngfreiter raises good questions concerning restrictions on parents. I've updated my "Sushi Review Notes" based on his feedback. See the attached. We will need to discuss this further and also decide on any impacts to defining Parent on custom resources.
[Sushi Review Notes.md](https://github.com/FHIR/sushi/files/6417287/Sushi.Review.Notes.md)

src/errors/InvalidLogicalParentError.ts
```diff
@@ -0,0 +1,10 @@
+import { WithSource } from './WithSource';
+import { SourceInfo } from '../fshtypes';
+
+export class InvalidLogicalParentError extends Error implements WithSource {
+  constructor(public childName: string, public parentName: string, public sourceInfo: SourceInfo) {
+    super(
+      `Parent ${parentName} is not of type Logical or Element or Base, so it is an invalid Parent for Logical ${childName}.`
```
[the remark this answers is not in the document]
joepaquette: Good question... That is why Chris had me do the same experimentation you tried. I've updated my original comments (attached in separate comment). I think I've come to understand the issue isn't about "allowed parents" as it is whether parent elements can be constrained.

src/errors/InvalidProfileParentError.ts
joepaquette: Another good question. This should be included in the general discussion on allowed parents.

src/errors/InvalidMustSupportError.ts
joepaquette: Agree... will add

src/export/StructureDefinitionExporter.ts
joepaquette: Good catch. I'll move it

src/export/StructureDefinitionExporter.ts
joepaquette: Good catch... will combine

src/export/StructureDefinitionExporter.ts
joepaquette: Yep! will fix it

src/export/StructureDefinitionExporter.ts
joepaquette: Good question! My initial thought is could we end up with a chicken & egg problem? Could we define a rule set that includes an element that has not been added yet if we process insertRules before addElementRules?

cmoesel: I feel like we should support addElement rules in rulesets.  I could see a use case where someone has a common set of elements they want in every one of their logical models.

joepaquette: So, I'll move lines 928-929 down below the line containing `applyInsertRules(fshDefinition, this.tank);`

src/errors/InvalidProfileParentError.ts
joepaquette: If a profile inherits from a logical model, it should result in a logical model having a `derivation = 'constraint'` - essentially "profiling" the base logical model. Rather than trying to fit that in now, I recommend that it be deferred until we can get feedback from the community with actual use cases for "profiling logical models".

src/fhirtypes/ElementDefinition.ts
```diff
@@ -414,6 +416,101 @@ export class ElementDefinition {
...
+    this.constraint = [
+      {
+        key: 'ele-1',
+        severity: 'error',
+        human: 'All FHIR elements must have a @value or children',
+        expression: 'hasValue() or (children().count() > id.count())',
+        xpath: '@value|f:*|h:div',
+        source: 'http://hl7.org/fhir/StructureDefinition/Element'
+      }
+    ];
```
cmoesel, line 477: I think that maybe we should consider pulling this from an actual definition.  Since the source is `Element`, maybe we should fish for `Element`, grab this constraint off its first snapshot, and copy it here (adding the `source` since that is not in the original definition of the constraint).

I'm thinking we might want to do this because if a future version of FHIR tweaks the definition, we'd automatically reflect the correct constraint definition instead of inserting the old R4 variant of it.

src/fhirtypes/StructureDefinition.ts
```diff
@@ -665,6 +665,7 @@ export class StructureDefinition {
    * Looks for a slice within the set of elements that matches the fhirPath
    * @param {PathPart} pathPart - The path to match sliceName against
    * @param {ElementDefinition[]} elements - The set of elements to search through
+   * @param {Fishable} fisher - A fishable implementation for finding definitions and metadata
```
ngfreiter, line 668: Thanks for the cleanup here and elsewhere!

src/fhirtypes/ElementDefinition.ts
```diff
@@ -414,6 +416,101 @@ export class ElementDefinition {
...
+        referenceEdt = referenceEdt ?? new ElementDefinitionType('Reference');
+        const metadata = fisher.fishForMetadata(t.type, Type.Resource, Type.Profile);
+        if (metadata) {
+          targetProfiles.push(metadata.url);
+        } else {
+          throw new TypeNotFoundError(t.type);
+        }
+      } else {
+        const edt = new ElementDefinitionType(t.type);
+        initialTypes.push(edt);
```
cmoesel, line 504: What if the type provided by the rule is actually a _profile_?  For example, I *think* this is valid?
```
* patient 1..1 USCorePatientProfile "A US Core Patient"
```
I know this would *typically* be a reference, but in a logical model, anything goes.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -414,6 +416,101 @@ export class ElementDefinition {
...
+   * @param {Fishable} fisher - A fishable implementation for finding definitions and metadata
+   */
+  applyAddElementRule(rule: AddElementRule, fisher: Fishable): void {
+    // The constrainType() method applies a type constraint to an existing
+    // ElementDefinition.type. Since this is a new ElementDefinition, it
+    // does not yet have a 'type', so we need to assign an "initial" value
+    // that the constrainType() method can process.
+    this.type = this.initializeElementType(rule, fisher);
+    const target = this.structDef.getReferenceName(rule.path, this);
+    this.constrainType(rule, fisher, target);
```
cmoesel, line 433: I think I'm having a little trouble understanding the difference between what `initializeElementType` does and `constrainType` does...  In other words, what does `constrainType` do that `initializeElementType` did not already do?

cmoesel (changes requested):
PART 1.  I did not have time to do a full review today, but I felt like I gave enough feedback that it might be beneficial to go ahead and submit what I have...  I'll try to finish the review tomorrow.

src/export/StructureDefinitionExporter.ts
```diff
@@ -61,16 +78,244 @@ export class StructureDefinitionExporter implements Fishable {
...
+    }
+
+    let defnType: Type;
+    if (fshDefinition instanceof Extension) {
+      defnType = Type.Extension;
+    } else if (fshDefinition instanceof Logical) {
+      defnType = Type.Logical;
+    } else {
+      defnType = Type.Resource;
+    }
```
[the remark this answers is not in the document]
joepaquette: Good points! I really did not think this through. I'll fix this as suggested.

src/export/StructureDefinitionExporter.ts
```diff
@@ -61,16 +78,244 @@ export class StructureDefinitionExporter implements Fishable {
...
+
+    if (isEmpty(fshDefinition.parent)) {
+      // Handle cases where the parent is not specified by throwing an error.
+      // - Profile has a hard requirement to define a parent and does not have a default
+      //   parent defined in the Profile class. Therefore, we need to throw this error.
+      // - Extension, Logical, and Resource all have a default parent, so we should never
+      //   get to this point where it is not defined. That said, if something somewhere
+      //   in the code were to change/remove the default, this error will then be thrown.
+      throw new ParentNotProvidedError(fshDefinition.name, fshDefinition.sourceInfo);
+    }
```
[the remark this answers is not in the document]
joepaquette: I had not thought about regression issues when I changed this logic.
I intended to run a full regression on the integration branch after merging in master and resolving conflicts. I assumed that something would probably cause some regression issues.

src/ig/IGExporter.ts
```diff
@@ -178,6 +178,21 @@ export class IGExporter {
...
+    // Default 'autoload-resources' to false if it is not already defined
+    // and only if custom resources are defined. This is done to counter
+    // the IG Publisher behaving as if 'autoload-resources' is set to true
+    // and allowing the IG Publisher to attempt processing of custom resources.
+    if (
+      this.pkg.resources?.length > 0 &&
+      !this.ig.definition.parameter.some(param => {
+        return param.code === 'autoload-resources';
+      })
+    ) {
```
cmoesel, line 190: Can you put in a logging statement (`info`) indicating that `autoload-resources` is being set to false since this IG contains custom resources?

src/import/FSHImporter.ts
```diff
@@ -992,11 +996,30 @@ export class FSHImporter extends FSHVisitor {
...
 
   visitAddElementRule(ctx: pc.AddElementRuleContext): AddElementRule {
-    const addElementRule = new AddElementRule(this.visitPath(ctx.path()))
+    const path = this.visitPath(ctx.path());
+    const addElementRule = new AddElementRule(path)
       .withLocation(this.extractStartStop(ctx))
       .withFile(this.currentFile);
 
     const card = this.parseCard(ctx.CARD().getText(), addElementRule);
+    if (isNil(card.min) || Number.isNaN(card.min)) {
```
cmoesel, line 1005: I see that lodash `isNil` checks if a value is null or undefined.  We typically just use double-equals (e.g., `card.min == null`) when we check this elsewhere in the codebase.  Is there a difference?

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -1,34 +1,37 @@
-import { StructureDefinitionExporter, Package } from '../../src/export';
-import { FSHTank, FSHDocument } from '../../src/import';
+import { Package, StructureDefinitionExporter } from '../../src/export';
```
cmoesel, line 1: The diff displayed in GitHub is very messed up for this file.  Can I assume that the only substantial changes you made were to add the tests around logic/resource, and everything else was just moved around?

test/fhirtypes/ElementDefinition.applyAddElementRule.test.ts
```diff
@@ -0,0 +1,336 @@
...
+      newElement.applyAddElementRule(addElementRule, fisher);
+
+      expect(newElement.path).toBe('AlternateIdentification.org');
+      const expectedType = new ElementDefinitionType('Reference').withTargetProfiles(
+        'http://hl7.org/fhir/StructureDefinition/Organization',
+        'http://hl7.org/fhir/StructureDefinition/Practitioner'
+      );
+      expect(newElement.type).toStrictEqual([expectedType]);
+    });
+
```
cmoesel, line 150: As noted in a previous comment, I think we need to handle the case where the user specifies a profile as the type -- and we should have a test to ensure it is done correctly (i.e., the resource type is the type and the profile is in the profile array).

cmoesel (commented):
PART 2.  This is the remainder of my code review.  It is an addendum to PART 1.

One overall comment that applies to several files (in PART1 and PART2): I noticed that you seem to have re-arranged import statements in several files.  Do you have a plugin that does this? (If so, which one?).  I think that's fine (and perhaps even good), but... beware, it may cause some grief when we try to get this branch in sync with master!

My other overall comment: Wow!  This was a lot of work and it's looking great!  Thank you for taking the time to investigate all of the pertinent issues to ensure we get it right.  And thank you for all those tests!  It's always nice to see such a thorough set of unit tests!  You're awesome!

src/export/StructureDefinitionExporter.ts
```diff
@@ -61,16 +78,244 @@ export class StructureDefinitionExporter implements Fishable {
...
+
+    let parentJson = this.fishForFHIR(fshDefinition.parent);
+    if (
+      !parentJson &&
+      fshDefinition instanceof Logical &&
+      (fshDefinition.parent === 'Base' ||
+        fshDefinition.parent === 'http://hl7.org/fhir/StructureDefinition/Base')
+    ) {
+      parentJson = this.createR4BaseStructureDefinition();
+    }
```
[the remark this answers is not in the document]
joepaquette: Good point about R5. This specific logic is to create a non-existent Base StructureDefinition for R4. 

Base will exist if R5 and beyond. I looked at FHIRDefinition, and since R5 Base is `"kind": "complex-type"`, it will be added to the "types" map which will not really help here. I need to think this through, so I'm going to hold off making any changes until I can walk this through the code and come up with a proposal for you.

cmoesel:
One of @ngfreiter's comments got me thinking about valid and invalid element names.  Do we have any validation regarding what the user types in as a path?  Or is it anything goes?  Are there certain things we should disallow?

As I thought of that, I thought about names with a `.` in them -- which shouldn't be allowed since `.` is a path separator.  But that led me to try this:
```
Logical: Foo
* foo.bar 0..1 boolean "A bar"
```

This presupposes that an element with path `Foo.foo` exists, but... it does not.  Currently, SUSHI emits this error:
> Cannot read property 'id' of undefined

I think it would be more helpful if we could detect that situation as an invalid path because `foo` does not exist.

src/export/StructureDefinitionExporter.ts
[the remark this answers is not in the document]
cmoesel: OK.  Sounds good.  I commented elsewhere that I wondered if we should allow complex types as parents of logical models too.  This finding (the `Base` is a `complex-type`) would seem to suggest that perhaps we should.

cmoesel:
@ngfreiter said:
> One larger question that I didn't think of until now, do we need to support making an Instance of a Logical?

I don't _think_ so?  I think logical models are mainly intended to be used for communicating the conceptual structure of clinical models w/ SMEs moreso than actually exchanging instances of data.  But... I'm not 100% sure on that.  @joepaquette - what do you think?

src/fhirtypes/ElementDefinition.ts
joepaquette: ```* patient 1..1 USCorePatientProfile "A US Core Patient"```
error The type "USCorePatientProfile" does not match any of the allowed types: USCorePatientProfile

```* patient 1..1 http://hl7.org/fhir/us/core/StructureDefinition/us-core-patient "A US Core Patient"```
error The type "http://hl7.org/fhir/us/core/StructureDefinition/us-core-patient" does not match any of the allowed types: http://hl7.org/fhir/us/core/StructureDefinition/us-core-patient

```* patient 1..1 us-core-patient "A US Core Patient"```
SUCCESS ->
```
      {
        "id": "CustomLogicalModel.patient",
        "path": "CustomLogicalModel.patient",
        "short": "A US Core Patient",
        "definition": "A US Core Patient",
        "min": 1,
        "max": "1",
        "base": {
          "path": "CustomLogicalModel.patient",
          "min": 1,
          "max": "1"
        },
        "type": [
          {
            "code": "Patient",
            "profile": [
              "http://hl7.org/fhir/us/core/StructureDefinition/us-core-patient"
            ]
          }
        ],
        "constraint": [
          {
            "key": "ele-1",
            "severity": "error",
            "human": "All FHIR elements must have a @value or children",
            "expression": "hasValue() or (children().count() > id.count())",
            "xpath": "@value|f:*|h:div",
            "source": "http://hl7.org/fhir/StructureDefinition/Element"
          }
        ]
      },
```
ElementDefinition.findTypeMatch() only does matching logic based on the lineage.id (not name or url).

@cmoesel Not sure what you might want to do with this. Maybe a Jira issue on the backlog??

File: src/fhirtypes/ElementDefinition.ts [hunk 19]
```diff
@@ -414,6 +416,101 @@ export class ElementDefinition {
...
+   * @returns {ElementDefinitionType[]} The element types as defined by the AddElementRule
+   * @private
+   */
+  private initializeElementType(rule: AddElementRule, fisher: Fishable): ElementDefinitionType[] {
+    const initialTypes: ElementDefinitionType[] = [];
+    // If there are multiple reference types, only one EDT is created having
+    // one or more targetProfiles for the single reference type
+    let referenceEdt: ElementDefinitionType;
+    const targetProfiles: string[] = [];
+    rule.types.forEach(t => {
```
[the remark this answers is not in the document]
cmoesel: Yes, I agree -- error.

test/export/StructureDefinitionExporter.test.ts
joepaquette: Correct. I added `describe` for "#Parent", "#LogicalModel", "#Resource". and "#AddElementRule". I can't recall off the top of my head, but there might have been several minor fixes elsewhere.

test/fhirtypes/ElementDefinition.applyFlags.test.ts
```diff
@@ -19,7 +19,7 @@ describe('ElementDefinition', () => {
     fisher = new TestFisher().withFHIR(defs);
   });
   beforeEach(() => {
-    observation = fisher.fishForStructureDefinition('Observation');
+    observation = fisher.fishForStructureDefinition('us-core-observation-lab');
```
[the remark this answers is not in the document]
joepaquette: 'Observation' is a resource and 'us-core-observation-lab' is a profile... The resources causes the following failure so switched to a profile:.
`The MustSupport flag is not permitted on element Observation.note of Observation (allowed only in profiles and extensions)`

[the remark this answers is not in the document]
joepaquette: @cmoesel @ngfreiter I actually wrote myself a note about this (but it got buried). Many of the existing tests are using resources rather than profiles for test StructureDefinitions. In many cases, the scope of the test means it shouldn't matter, but in the above case, it did. It might make some sense later on to maybe make the tests a little more "rigerous" by using the appropriate type of StructureDefinitions when testing things like rules. Now that custom resources and logical models are in the mix, it might be beneficial to ensure we are testing with objects that the rules are intended to operate on/with to prevent something from slipping the the cracks down the road.

src/export/StructureDefinitionExporter.ts
```diff
@@ -61,16 +78,244 @@ export class StructureDefinitionExporter implements Fishable {
...
+    } else if (
+      fshDefinition instanceof Logical &&
+      !(
+        (['logical', 'resource'].includes(parentJson.kind) &&
+          parentJson.derivation === 'specialization') ||
+        parentJson.type === 'Base' ||
+        parentJson.type === 'Element'
+      )
+    ) {
+      // A logical model can only have another logical model as a parent
```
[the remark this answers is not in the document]
joepaquette: Fixed comment adding resource - left out type until we are sure that is something we should do

src/export/StructureDefinitionExporter.ts
```diff
@@ -61,16 +78,244 @@ export class StructureDefinitionExporter implements Fishable {
...
+        fshDefinition.sourceInfo
+      );
+    } else if (
+      fshDefinition instanceof Logical &&
+      !(
+        (['logical', 'resource'].includes(parentJson.kind) &&
+          parentJson.derivation === 'specialization') ||
+        parentJson.type === 'Base' ||
+        parentJson.type === 'Element'
+      )
```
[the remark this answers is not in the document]
cmoesel: > From a practical perspective, the purpose of logical models is to eventually use it to define a FHIR resource (either as new resource or as a profile for an existing resource).

I think logical models are intended as a communication mechanism w/ clinicians and SMEs -- which, as you said, will eventually get realized as FHIR resources and profiles.  If a modeler wants to  represent things like  `HomeAddress` and `WorkAddress`, they might want to base them on `Address`.  I guess the modeler could just define them as profiles from the get-go, but then if they want to add properties, they need to use extensions -- which are not very friendly as a communication mechanism w/ clinicians and SMEs.

> Using a base of a complex-type implies defining a "data type".

I think it more likely implies defining a profile on a data type, which is not totally uncommon in IGs.

src/import/FSHImporter.ts
cmoesel: Yeah, I think I'd prefer the consistency of using `==`.  Thanks!

test/fhirtypes/ElementDefinition.applyFlags.test.ts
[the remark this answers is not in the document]
cmoesel: Ah.  Interesting.  I guess I had never noticed that we have tests that apply constraints to resources (rather than profiles).  I agree -- this is something we should try to avoid.  Where possible, our tests should represent at least _good_ practice, if not _best_ practice!

I logged this as #815.

src/fhirtypes/ElementDefinition.ts
cmoesel: I think the issue is that our code expects `type.code` to be an `id` of a FHIR type.  In most cases (at least up until logical models) this was guaranteed to be true.  Consider the definition of `type.code`:
> URL of Data type or Resource that is a(or the) type used for this element. References are URLs that are relative to http://hl7.org/fhir/StructureDefinition e.g. "string" is a reference to http://hl7.org/fhir/StructureDefinition/string. Absolute URLs are only allowed in logical models.

I expect this is why using the name or url did not work -- because it was getting put into `type.code` and then the `constrainType` function could not look it up (since it thinks it should be an `id`).  I think maybe this would be fixed if you looked up the type (by whatever was passed in) to determine if it is a specialization or a constraint.  If it's a constraint, then the `type.code` should be the `SD.type` and the `type.profile` array should contain the `SD.url`.

src/fhirtypes/ElementDefinition.ts
cmoesel: OK.  I've thought about this some more and have gone back and forth -- and I think your initial inclination was right, but the execution isn't quite there yet.  The `constrainType` function is very specific to the use case of constraining a broad set of types to a narrower set of types.  So your `initializeElementType` should just establish that first *broad* set of types.  I think that right now, you're maybe setting it up too narrowly.

What we really want is for the initial set of types just to represent the set of base types that the more specific types will either match exactly or match as a profile.  For example, if the types in AddElement are:
* MyCustomLogicalModel
* MyPatientProfileA (a profile on Patient)
* MyPatientProfileB (a profile on Patient)
* Reference(Practitioner)
* Reference(Group)
Then you want an initial type array of:
```
[
  { code: 'MyCustomLogicalModel' },
  { code: 'Patient' },
  { code: 'Reference' }
]
```
because that is the most basic set of types that the incoming types can all cleanly map to.

I think you can implement this by looping through the types:
* if the type `isReference`, add `{ code: 'Reference' }` to the initial types
* else `fshForMetadata` on the type and add `{ code: meta.sdType }` to the initial types
* either avoiding duplicates along the way or removing duplicates at the end

Then after that, I think `constrainType` should work as expected with the types from the `addElementRule`.

src/fhirtypes/ElementDefinition.ts
cmoesel: I updated my last comment in light of the approach I am suggesting in the next comment.

src/export/StructureDefinitionExporter.ts
```diff
@@ -155,18 +412,187 @@ export class StructureDefinitionExporter implements Fishable {
...
+          logger.error(
+            `FHIR prohibits constraining parent elements. Skipping '${rule.constructorName}' at path '${rule.path}' for '${fshDefinition.name}'.`,
+            rule.sourceInfo
+          );
+          continue;
+        }
+      }
+
+      // CaretValueRules apply to both StructureDefinitions and ElementDefinitions; therefore,
+      // rule handling for CaretValueRules must be outside the 'if (element) {...}' code block.
```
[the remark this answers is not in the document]
joepaquette: Required due to separating out AddElementRules from all other rules. Reverted as part of adding AddElementRule processing within the "setRules" method.

joepaquette:
> One overall comment that applies to several files (in PART1 and PART2): I noticed that you seem to have re-arranged import statements in several files. Do you have a plugin that does this? (If so, which one?). I think that's fine (and perhaps even good), but... beware, it may cause some grief when we try to get this branch in sync with master!

Great catch! Out of habit, I might have hit the WebStorm "Organize Imports" feature.   If time allows, I'll try to do some cleanup of the worst offenders...

joepaquette:
> > ```
> > Logical: Foo
> > * foo.bar 0..1 boolean "A bar"
> > ```
> > This presupposes that an element with path `Foo.foo` exists, but... it does not. Currently, SUSHI emits this error:
> > > Cannot read property 'id' of undefined
> > 
> > 
> > I think it would be more helpful if we could detect that situation as an invalid path because `foo` does not exist.
> 
> Yeah I think we can probably follow the spec on this. On this page: https://www.hl7.org/fhir/elementdefinition-definitions.html#def, it looks like `eld-19` gives a regex that rules out certain special characters that we can also disallow in our element names. I don't think we can use that regex directly, since a FSH path can also not include a `.`, we don't want people defining `foo.bar` at once like Chris is showing. But it is a good reference for which characters can be considered an error. Then `eld-20` we can use to justify issuing a warning on any non-alphanumeric characters, but I think we should limit that to just a warning, and not an error.

So I traced where the `Cannot read property 'id' of undefined` was being thrown - it was in the `StructureDefinition.findObsoleteChoices(...)` --> `const parentId = baseElement.parent().id;`
The `parent()` returned undefined.  After pulling on this thread, I decided to add a parent check in `ElementDefinition.applyAddElementRule(...)` and throw an error if a parent element was not found since every added element has to have a parent, even if it is the root element.

src/export/StructureDefinitionExporter.ts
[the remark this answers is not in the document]
joepaquette: TODO: PAQ - Resolve fishForMetadata issue:
Searches for the Metadata associated with the passed in name/id/url.  It will first search through the local package (which contains FHIR artifacts exported so far), then through the tank, then through the external FHIR definitions.
This means that when fishing for metadata to get the "result?.url for throwing the `ParentDeclaredAsNameError` and `ParentDeclaredAsIdError`, it will get the fshDefinition's URL rather than the parent's.

src/export/StructureDefinitionExporter.ts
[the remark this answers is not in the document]
joepaquette: I missed this conversation as I went back through to make sure I covered all of the comments. I will go back in and add support for complex-types for logical models.

src/export/StructureDefinitionExporter.ts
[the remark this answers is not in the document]
joepaquette: @cmoesel @ngfreiter The current commits of this PR now have support for complex types. That pretty much solved this issue since we now will fish for complex-types for parents of logical models. I did some testing with both R4 and R5 in my .fhir cache. When I set the fhirVersion in my sushi-config.yaml to 4.0.1, my logical model used the "hard-coded" 4.0.1 version of `Base` and when I set it to 4.6.0, it used the R5 version of `Base` as it should have. So, there was nothing to change to properly handle the use of the R4 vs the R5 `Base` parent.

src/errors/InvalidLogicalParentError.ts
```diff
@@ -4,7 +4,7 @@ import { SourceInfo } from '../fshtypes';
 export class InvalidLogicalParentError extends Error implements WithSource {
   constructor(public childName: string, public parentName: string, public sourceInfo: SourceInfo) {
     super(
-      `Parent ${parentName} is not of type Logical or Resource or Element or Base, so it is an invalid Parent for Logical ${childName}.`
+      `Invalid parent ${parentName} specified for logical model ${childName}. The parent of a logical model must be Element, Base, another logical model, or a resource.`
```
cmoesel, line 7: ... or a type.

src/errors/InvalidProfileParentError.ts
```diff
@@ -3,6 +3,8 @@ import { SourceInfo } from '../fshtypes';
 
 export class InvalidProfileParentError extends Error implements WithSource {
   constructor(public childName: string, public parentName: string, public sourceInfo: SourceInfo) {
-    super(`Parent ${parentName} is not a valid Parent for Profile ${childName}.`);
+    super(
+      `Invalid parent ${parentName} specified for profile ${childName}. The parent of a profile must be a resource or another profile.`
```
cmoesel, line 7: Technically I think we allow profiles on types too (like `Address`).

src/export/StructureDefinitionExporter.ts
```diff
@@ -105,47 +91,54 @@ export class StructureDefinitionExporter implements Fishable {
...
+    // Type.Profile when FSH definition is an instanceof Profile) because the result of
+    // the fishing would only find itself.
+
+    let possibleParentMeta: Metadata;
+    if (fshDefinition instanceof Logical) {
+      // Logical models can have resources, complex-types, and other logical models as a parent
+      possibleParentMeta = this.fishForMetadata(fshDefinition.parent, Type.Resource, Type.Type);
+    } else if (fshDefinition instanceof Profile) {
+      // Profiles can have resources and other profiles as a parent
+      possibleParentMeta = this.fishForMetadata(fshDefinition.parent, Type.Resource);
```
cmoesel, line 108: This cleaned things up nicely.  Thank you!

I do have one small concern in that we are now doing this fishing operation for every logical model and every profile, regardless of whether or not their name/id matches the parent.  In the very large majority of cases, there isn't a match, so this is a lot of unnecessary fishing.

We should probably wrap this whole section (lines 102 - 128) in an `if` block that first checks if the parent matches the name or id.

cmoesel (changes requested):
Wow.  This is looking really nice.  Thank you for responding to my previous feedback _so elegantly_.  I really like what you've done with the place! ;-)

This time around I only found a few nitpicky things!

src/export/StructureDefinitionExporter.ts
joepaquette: Yep... You did mention that and I forgot about it. The above reminder about profiles having a type parent allowed resulted in further cleanup...

cmoesel:
Thanks for all the changes and quick responses, Joe.  All the code looks great.  Next up, I'm going to take it for a spin -- coding up some logical models and resources, etc... Once I've confirmed that all works as expected, you'll have my approval.

cmoesel (changes requested):
Today was mostly full of meetings, so I did not get to try too many things, but I did try a custom resource modeled after the real `Patient` resource (but only a few fields because... so tedious!).

I uncovered a couple of issues that you might want to look into:

1. The definition portion of the AddElement rule does not seem to allow multi-line strings using `"""`.  For example, this fails:
```
* active 0..1 ?! SU boolean "Whether this patient's record is in active use"
    """
    Whether this patient record is in active use. Many systems use this property to mark
    as non-current patients, such as those that have not been seen for a period of time based
    on an organization's business rules. It is often used to filter patient lists to
    exclude inactive patients. Deceased patients may also be marked as inactive for the
    same reasons, but may be active for some time after death.
    """
```
with this error:
> error extraneous input '"""\n    Whether this patient record is in active use. Many systems use this property to mark\n    as non-current patients, such as those that have not been seen for a period of time based\n    on an organization's business rules. It is often used to filter patient lists to\n    exclude inactive patients. Deceased patients may also be marked as inactive for the\n    same reasons, but may be active for some time after death.\n    """' expecting {<EOF>, KW_ALIAS, KW_PROFILE, KW_EXTENSION, KW_INSTANCE, KW_INVARIANT, KW_VALUESET, KW_CODESYSTEM, KW_RULESET, KW_MAPPING, KW_LOGICAL, KW_RESOURCE}

I think this is something that actually needs to be fixed in the grammar (and then might require changes in the importer too, but I can't remember for sure).

2. I noticed that the elements in the differential seem to have some properties injected automatically, like `base`, `constraints`, `mapping`, `isSummary`, `isModifier`, ... but when I look at the differential in real FHIR resources (like the real `Patient` resource), I often don't see these things.  I'm guessing that you are inheriting some of these from the parent (`DomainResource`) or the source `ElementDefinition` -- in which case they probably _shouldn't_ be in the differential (since they're not different).  Maybe you need to call `captureOriginal()` at some point earlier in the process (but still before setting the author-specified values)?

Here is a gist of the code I tested this with.  I commented out the `"""` strings causing issues: https://gist.github.com/cmoesel/859ba812389ff7dc64effae5288e264a

src/export/StructureDefinitionExporter.ts
```diff
@@ -364,10 +357,32 @@ export class StructureDefinitionExporter implements Fishable {
...
+    // Hack Alert:
+    // A material change is required to an element AFTER the execution of the
+    // structDef.captureOriginalElements() function to ensure the differential
+    // is properly rendered.
+    // To ensure the differential root element contains only the appropriate
+    // attributes, we are setting an initial value for the prohibited "maxLength"
+    // element attribute that will then be deleted below. The "maxLength"
+    // attribute was selected because it is prohibited in type definitions (i.e.,
+    // logical models and resources).
+    // Ref: http://hl7.org/fhir/elementdefinition.html#interpretation
```
cmoesel, line 369: Clever approach.  I think this works...

I know you considered some other approaches.  Two that I can think of are:
1. `StructureDefinition.toJSON`: As it loops through the elements checking `hasDiff()` (to generate the differential), could it also check if it is the element at index 0 in a specialization?
2. `Element.hasDiff`: Could it check if it is the root element in a specialization and return true if so?

I think I like option 1 above best.  Are there reasons this would not be sufficient?

cmoesel (commented):
I think you're still working on this, but thought I'd send my comment about your most recent commit.  I also want to clarify that I was seeing those extra elements (`base`, `constraints`, etc.) in all elements (not just the root element).

src/export/StructureDefinitionExporter.ts
joepaquette: Thank you! I hate hacks! I guess I had some tunnel vision. I agree approach #1 is the better option, so I've implemented that approach and removed the hack.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -467,21 +477,36 @@ export class ElementDefinition {
...
+    // REF: http://hl7.org/fhir/elementdefinition.html#interpretation
+    // However, convention, based on looking at FHIR resources, is to exclude the 'base'
+    // attribute from the differential and to only include constraint attributes when
+    // specifically defined for an element.
+    // The 'elements' collection in the FSH StructureDefinition is split out into snapshot
+    // and differential in the StructureDefinition.toJSON() method. Therefore, to achieve the
+    // observed convention, we place the 'base' and the "base" element's constraints into
+    // temp holding properties. During execution of the StructureDefinition.toJSON() method,
+    // we use these temp holding properties to properly set the 'base' and 'constraint'
+    // attributes in the snapshot elements.
```
cmoesel, line 493: I think we might be able to accomplish this without adding new temporary properties.

Since this function (`applyAddElementRule`) is always called on a brand new element (w/ no changes yet applied), I think we can safely set `base` and `constraint` at the very top of this function and then call `this.captureOriginal()` _before_ setting any of the other things.  This will have the affect of treating them as part of the original definition -- which will keep them in the snapshot but suppress them from the differential.

We already treat `constraint` as an "additive" property ([implementation](https://github.com/joepaquette/sushi/blob/5bb3cd8d0b5c9bdef90ba641abfc298d54dc19fe/src/fhirtypes/ElementDefinition.ts#L389-L396)), which means that when we generate the differential, we only include those constraints that are new (i.e., not in the "original").  So following this approach we should also still get the desired behavior of the differential only including the constraints that were directly set on the element.

Assuming this approach works (and I haven't missed something important), I think it would be a good bit cleaner (and more efficient).

cmoesel (commented):
Thanks, @joepaquette!  I've left another comment about a different approach to the `base` and `constraint` issue.  It seems to me like it should work, but let me know if you think I'm missing something important!

src/fhirtypes/ElementDefinition.ts
joepaquette: It did work and the code is definitely much cleaner. Thanks for the help!

src/fhirtypes/ElementDefinition.ts
```diff
@@ -519,7 +498,11 @@ export class ElementDefinition {
    */
   private initializeElementType(rule: AddElementRule, fisher: Fishable): ElementDefinitionType[] {
     if (rule.types.length > 1 && !rule.path.endsWith('[x]')) {
-      throw new InvalidChoiceTypeRulePathError(rule);
+      const referenceTypes = rule.types.filter(t => t.isReference);
+      // Reference data type with multiple targets is not considered a choice data type.
+      if (referenceTypes.length < rule.types.length) {
```
cmoesel, line 503: I think this could be simplified as:
```
if (rule.types.some(t => ! t.isReference)) {
```
or
```
if (! rule.types.every(t => t.isReference)) {
```

cmoesel: But that could be just my opinion.  If you feel that the approach you took is easier to understand then feel free to ignore.

joepaquette: good suggestion - fixed

cmoesel (commented):
Awesome.  I only have a few relatively minor comments for your consideration.

src/fhirtypes/common.ts
```diff
@@ -645,6 +645,50 @@ export function getUrlFromFshDefinition(
...
+
+  for (const rule of fshDefinition.rules) {
+    if (rule instanceof CaretValueRule && rule.path === '' && rule.caretPath === 'type') {
+      // this value should only be a string, but that might change at some point
+      return rule.value.toString();
+    }
+  }
+  // Default type for logical model to the StructureDefinition url;
+  // otherwise default to the id meta property.
+  // Ref: https://chat.fhir.org/#narrow/pm-with/191469,210024,211704,239822-group/near/240237602
```
ngfreiter, line 674: I know there has been a lot of discussion about this that I was out of office for much of, but it seems like setting `type` to an absolute URL causes problems for the IG Publisher:
image

Since the Publisher seems to expect that an element path will start with the `type`, but the element path starts with the `id` of the element, `MyLogicalPatient` in this case. Is this an issue we are aware of?

ngfreiter: Then of course, if you do try to put the full url in the path, the IG Publisher is mad at the illegal characters in that URL.

cmoesel: Ugh.  Grahame has said on multiple occasions that it should be a URL.

image

([source](https://chat.fhir.org/#narrow/stream/179177-conformance/topic/StructureDefinition.2Etype.20for.20logical.20model.3F))


image

([source](https://chat.fhir.org/#narrow/stream/179166-implementers/topic/Logical.20models.20-.20Type.20Element))


image

([source](https://chat.fhir.org/#narrow/stream/179177-conformance/topic/StructureDefinition.2Etype.20for.20Logical.20Models.2FCustom.20Resources))


And the spec also says it is supported and how it relates to the paths:
image

([source](http://hl7.org/fhir/R4/structuredefinition-definitions.html#StructureDefinition.type))

I'll ask on Zulip.

cmoesel: @ngfreiter -- what version of the IG Publisher are you using?  I just tried this with a logical model and it worked fine.

ngfreiter: I was using an older version, as you've seen (and I've also confirmed) looks like this works in more recent versions. Sorry I didn't realize mine was out of date!

cmoesel:
OK.  I've done some manual testing of this and ran into a couple of issues specifically w/ using one logical model as the type of an element in another.

If you run SUSHI on this:
```
Logical: Bar
* height 1..1 Quantity "The height of the bar"
* width 1..1 Quantity "The width of the bar"
* depth 1..1 Quantity "The depth of the bar"

Logical: Foo
* bars 0..* Bar "The bars of the foo"
```
then you get this error:
> error The type "Bar" does not match any of the allowed types: http://logicals.org/StructureDefinition/Bar

The `Foo.bars` element it outputs in the SD has the correct type (w/ the URL as the `type.code`), but is missing the short and definition:
```json
{
  "id": "Foo.bars",
  "path": "Foo.bars",
  "type": [{
    "code": "http://logicals.org/StructureDefinition/Bar"
  }]
}
```

But... **plot twist**... if you swap the order of the definitions in the FSH:
```
Logical: Foo
* bars 0..* Bar "The bars of the foo"

Logical: Bar
* height 1..1 Quantity "The height of the bar"
* width 1..1 Quantity "The width of the bar"
* depth 1..1 Quantity "The depth of the bar"
```
then you get _no errors_ in SUSHI, and the `Foo.bars` element renders the type _incorrectly_ (URL should be the code value, not profile value), but does have the short and definition:
```json
{
  "id": "Foo.bars",
  "path": "Foo.bars",
  "short": "The bars of the foo",
  "definition": "The bars of the foo",
  "min": 0,
  "max": "*",
  "type": [{
    "code": "Base",
    "profile": [
      "http://logicals.org/StructureDefinition/Bar"
    ]
  }]
}
```

src/export/StructureDefinitionExporter.ts
```diff
@@ -155,18 +310,119 @@ export class StructureDefinitionExporter implements Fishable {
...
+        continue;
+      }
+
       const element = structDef.findElementByPath(rule.path, this);
+
+      if (element && (fshDefinition instanceof Logical || fshDefinition instanceof Resource)) {
+        // The FHIR spec prohibits constraining any parent element in a 'specialization'
+        // (i.e., logical model and resource), therefore log an error if that is attempted
+        // and continue to the next rule.
+        if (element.path !== element.base.path) {
```
ngfreiter, line 415: I'm noticing that this check prevents something like this:
```
Logical: MyLogicalBase
Parent: Base
* foo 1..1 HumanName "short"
* foo.given 1..1
```
which may be too aggressive? I'm not entirely sure. Someone could work around this by profiling HumanName and referring to that type instead, so it isn't impossible to represent this idea, but maybe more difficult. What do you think @joepaquette, should this be allowed?

joepaquette: Great question! I think this is probably a reasonable expectation so I think I should fix this. @cmoesel ??

joepaquette: @cmoesel @ngfreiter  Another question... What constitutes a parent? In the example above, we could also try `* foo.id 1..1`. In this case, I don't think this should be allowed since `Element` is the parent of `HumanName`. Does this make sense?

cmoesel: I should look back at the part of the spec where this is discussed, but my interpretation was that you cannot constrain any path that would be a valid path in the parent SD.  So, since `foo` was only introduced in `MyLogicalBase`, then I think you could constrain _any_ path that starts with `foo` inside `MyLogicalBase` -- because there was no `foo` anything in the parent.

cmoesel: I'm trying to find where we got the idea that you cannot constrain elements from the parent.  And having trouble finding it (although I _know_ we discussed this and had justification for it).  I wouldn't make any code changes about this until we chase it down a little further.

cmoesel: [Found it] and asked a follow-up question.

src/fhirdefs/FHIRDefinitions.ts
```diff
@@ -24,6 +25,11 @@ export class FHIRDefinitions implements Fishable {
     this.codeSystems = new Map();
     this.implementationGuides = new Map();
     this.predefinedResources = new Map();
+
+    // FHIR R4 does not have a StructureDefinition that defines "Base" but FHIR R5 does.
+    // We have defined a "placeholder" StructureDefinition for "Base" for R4.
+    // Inject the R4 "Base" placeholder StructureDefinition
+    this.add(JSON.stringify(STRUCTURE_DEFINITION_R4_BASE));
```
cmoesel, line 32: I couldn't figure out why the order was still causing problems for me when I did manual testing using R5 (but somehow passed the R4-based unit tests).  It turns out that this line is the culprit.  The argument to `add` should be a JSON _object_, not a string.  When you do it like this (as a string), then `Base` doesn't get registered in the definitions at all (for R4) -- so lookups for `Base` still return `undefined` in R4.

Once you fix this, it cause's the tanks `getSdType` function to operate a little differently -- which, in turn, causes an error.  We _do_ want to be able to fish for `Base` in R4, so I think you need to fix this here (by removing the `stringify`) and then also make a fix in `getSdType` to get the now-breaking test passing again.

cmoesel (changes requested):
The ordering issue isn't quite fixed yet.  See my individual comment.

cmoesel:
The original FSH seems to work correctly now, but I've noticed that if the instance sets a _different_ value at element 0, then it overwrites the supplied fixed value and SUSHI doesn't complain.  E.g., if I don't set `code.coding` at all in the `Instance`, then I get this -- which matches the pattern:
```
  "code": {
    "coding": [
      {
        "code": "8302-2",
        "system": "http://loinc.org"
      }
    ]
  },
```
That's good!  BUT... if I set the coding at index 0 to something else:
```
  "code": {
    "coding": [
      {
        "code": "8306-3",
        "system": "http://loinc.org",
        "display": "Body height --lying"
      }
    ]
  },
```
then it just happily overwrites the value from the pattern and doesn't complain at all.

I think the right behavior would probably be to either: (a) add the required value from the pattern to the end of the array, or (b) keep the required value at element 0, but then issue an error if anyone tries to assign a different value to element 0.  I think there are upsides and downsides to both approaches.

I don't know if we should treat this as a separate bug (and approve this PR as-is) or try to fix it in this PR.  Technically speaking, I think this bug was introduced by the fix in this PR.  Before, SUSHI was too strict; now it is too lenient.

mint-thompson (approved):

src/utils/MasterFisher.ts
```diff
@@ -79,6 +79,12 @@ export class MasterFisher implements Fishable {
     const history = [meta];
     let [sdType, parent] = [meta.sdType, meta.parent];
     while (sdType == null && parent != null) {
+      // A logical model with the Base parent will not have a sdType value. In this case, we know
+      // that the sdType should be the logical model's url and not be based on its parent.
+      if (parent === 'Base') {
+        [sdType, parent] = [meta.url, parent];
+        continue;
+      }
```
cmoesel, line 87: I think this is too specific.  We want to do this for *every* logical model (whether it derives from `Base` or not).  If it is a logical model, then its type should be its own URL, right?  As it is now, I'm still getting the wrong type when I do this:
```
Logical: FooFromElement
Parent: Element
* bars 0..* BarFromElement "The bars of the foo"

Logical: BarFromElement
Parent: Element
* height 1..1 Quantity "The height of the bar"
```
Result:
```
"type": [
  {
    "code": "Element",
    "profile": [
      "http://logicals.org/StructureDefinition/BarFromElement"
    ]
  }
]
```

Or this:
```
Logical: Other
* thing 1..1 boolean "Is it a thing?"

Logical: FooFromOther
Parent: Other
* bars 0..* BarFromOther "The bars of the foo"

Logical: BarFromOther
Parent: Other
* height 1..1 Quantity "The height of the bar"
```
Result:
```
"type": [
  {
    "code": "http://logicals.org/StructureDefinition/Other",
    "profile": [
      "http://logicals.org/StructureDefinition/BarFromOther"
    ]
  }
]
```

test/export/StructureDefinition.LogicalExporter.test.ts
```diff
@@ -328,6 +328,82 @@ describe('LogicalExporter', () => {
...
+    logicalFoo.parent = 'Element';
+    const addElementRuleBars = new AddElementRule('bars');
+    addElementRuleBars.min = 0;
+    addElementRuleBars.max = '1';
+    addElementRuleBars.types = [{ type: 'Bar' }];
+    addElementRuleBars.short = 'short of property bars';
+    logicalFoo.rules.push(addElementRuleBars);
+    doc.logicals.set(logicalFoo.name, logicalFoo);
+
+    const logicalBar = new Logical('Bar');
```
cmoesel, line 342: To really test your logic, I think it's actually `Bar` that needs to have the parent set to `Element` -- because `Bar` is the type we are attempting to constrain (and get the `sdType` in the process).  If you set `logicalBar.parent` to `'Element'` here the test will fail.

test/export/StructureDefinition.LogicalExporter.test.ts
```diff
@@ -328,6 +328,82 @@ describe('LogicalExporter', () => {
...
+    logicalFoo.parent = 'AlternateIdentification';
+    const addElementRuleBars = new AddElementRule('bars');
+    addElementRuleBars.min = 0;
+    addElementRuleBars.max = '1';
+    addElementRuleBars.types = [{ type: 'Bar' }];
+    addElementRuleBars.short = 'short of property bars';
+    logicalFoo.rules.push(addElementRuleBars);
+    doc.logicals.set(logicalFoo.name, logicalFoo);
+
+    const logicalBar = new Logical('Bar');
```
cmoesel, line 380: To really test your logic, I think it's actually `Bar` that needs to have the parent set to `AlternateIdentification` -- because `Bar` is the type we are attempting to constrain (and get the `sdType` in the process).  If you set `logicalBar.parent` to `'AlternateIdentification'` here the test will fail.

cmoesel (changes requested):
Good idea for the other tests when parent is Element or another logical model.  I hadn't thought of that scenario, and it turns out that there are still issues there.  Unfortunately, your tests don't show the issues because they set the parent on `Foo` instead of `Bar`.  But... I think it's an easy fix?

cmoesel (approved):
This.  Looks.  Great!  No further comment!  Approved!

cmoesel:
Once @ngfreiter gives his approval, we can squash-merge!

ngfreiter (approved):
Great stuff, thank you @joepaquette!

PR #803, opened 2021-04-21
Adding profile element extension support

### LINKED ISSUE #767: Support elementdefinition-profile-element extension
The [elementdefinition-profile-element extension](http://hl7.org/fhir/R4/extension-elementdefinition-profile-element.html) refers to the:
> specific element to use in the referenced profile. This is used when a backbone element is being profiled, rather than an established type.

More specifically, it can be used to indicate that content reference element should take on the constraints of the thing that it targets.  See: https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/Clarification.20on.20contentReference/near/225076962

julianxcarter:
Fixes #767 and addresses [CIMPL-678](https://standardhealthrecord.atlassian.net/browse/CIMPL-678), adding support for the [profile element extension](https://www.hl7.org/fhir/extension-elementdefinition-profile-element.html). When applied to an ElementDefinition's `type.profile` element, this extension allows for that ElementDefinition to take on the constraints of the ElementDefinition pointed to by this extension. Additionally, this extension allows for each recursive contentReference element to take on the constraints of the original element (e.g: when the extension is applied to `Questionnaire.item`, `Questionnaire.item.item` will pick up it's constraints as well). 

The effects of this can be seen by running [this example on FSHOnline](https://fshschool.org/FSHOnline/#/share/3v7riaz). This FSH demonstrates the profile-element extension in two ways:

1. By creating a new `Patient` profile and using the profile-element extension to apply the constraints on `USCorePatient.name` to the new profile's name element. US Core's Patient profile gives the `name` element a minimum cardinality of 1, so the following instance should generate an error since it does not set a `name` element at all.
2. By creating a new `Questionnaire` profile and using the profile-element extension to apply constraints on `Questionnaire.item` recursively. This profile gives the `item-ctl` extension a minimum cardinality of 1 on `Questionnaire.item`  and all of its children, so the following instance should generate an error since `Questionnaire.item.item` does not include this extension.

When run against master, this FSH will generate no errors since the constraints were not copied over, however when run against this branch there should be errors related to the missing elements.


src/export/StructureDefinitionExporter.ts
```diff
@@ -493,6 +502,64 @@ export class StructureDefinitionExporter implements Fishable {
     return this.fisher.fishForMetadata(item, ...types);
   }
 
+  applyElementProfiles(caretRules: CaretValueRule[], structDef: StructureDefinition): void {
+    const profileElementExtension =
+      'http://hl7.org/fhir/StructureDefinition/elementdefinition-profile-element';
```
ngfreiter, line 507: According to those rule followers (cough @cmoesel cough) we should make this a constant at the top of the file I think? Alternatively I think in this case you could just use that string directly where it is used, since it only appears to be used one place.

src/export/StructureDefinitionExporter.ts
```diff
@@ -569,6 +636,11 @@ export class StructureDefinitionExporter implements Fishable {
     applyInsertRules(fshDefinition, this.tank);
     this.preprocessStructureDefinition(fshDefinition, structDef.type === 'Extension');
 
+    const caretRules: CaretValueRule[] = fshDefinition.rules.filter(
+      rule => rule instanceof CaretValueRule
+    ) as CaretValueRule[];
+    this.applyElementProfiles(caretRules, structDef);
```
ngfreiter, line 642: Alternatively we could just do:
```javascript
    this.applyElementProfiles(
      fshDefinition.rules.filter(rule => rule instanceof CaretValueRule) as CaretValueRule[],
      structDef
    );
```

src/export/StructureDefinitionExporter.ts
```diff
@@ -493,6 +502,64 @@ export class StructureDefinitionExporter implements Fishable {
     return this.fisher.fishForMetadata(item, ...types);
   }
 
+  applyElementProfiles(caretRules: CaretValueRule[], structDef: StructureDefinition): void {
+    const profileElementExtension =
+      'http://hl7.org/fhir/StructureDefinition/elementdefinition-profile-element';
+
+    caretRules.forEach(rule => {
+      // If the rule isn't applied to an element, it's not relevant
+      if (rule.path) {
```
ngfreiter, line 511: It could save you a level of nesting to just do something like:
```javascript
if (!rule.path) {
  return;
}
```
that way the rest of the code does not have to be nested at that level

src/export/StructureDefinitionExporter.ts
```diff
@@ -493,6 +502,64 @@ export class StructureDefinitionExporter implements Fishable {
...
+        if (
+          parsedCaret.length === 2 &&
+          parsedCaret[0].base === 'type' &&
+          parsedCaret[1].base === 'profile'
+        ) {
+          const associatedCaretPaths = [
+            `${assembleFSHPath(parsedCaret)}.extension[0].url`,
+            `${assembleFSHPath(parsedCaret)}.extension.url`,
+            `${assembleFSHPath(parsedCaret)}.extension[0].valueString`,
+            `${assembleFSHPath(parsedCaret)}.extension.valueString`
```
ngfreiter, line 524: Instead of reassembling the parsed path, could we just use the original path here?

ngfreiter (commented):
I have a question about the approach at a high level, let me know what you think.

ngfreiter:
Yikes it appears GitHub wiped away my final comment, which contained my question about the overall approach, thanks GitHub. So the question was about the following situation:
```
Profile: MyNamePatient
Parent: Patient
* name.given 1..1

Profile: HumanNameProfile
Parent: HumanName
* given 1..1

Profile: ExtensionPatient
Parent: Patient
* name ^type.profile = "MyNamePatient"
* name ^type.profile.extension.url = http://hl7.org/fhir/StructureDefinition/elementdefinition-profile-element
* name ^type.profile.extension.valueString = "Patient.name"

Profile: ProfilePatient
Parent: Patient
* name only HumanNameProfile
```
On the `ProfilePatient`, we do not unfold the elements of `HumanNameProfile` unless we actually add a rule that digs into those elements, `* name.given MS` for example. On the `ExtensionPatient`, it seems like in the `StructureDefinitionExporter` we are prematurely unfolding all of the elements on the `Patient.name` element on the `MyNamePatient`. Shouldn't we only be unfolding those elements once we actually need to access those elements?

src/export/StructureDefinitionExporter.ts
julianxcarter: We certainly could!

julianxcarter:
@ngfreiter Yes that is what we've been doing in Sushi thus far now that I'm thinking of it. I'll update this approach so that the constraints on the parent element referenced by the profile element extension are still copied over in the `StructureDefinitionExporter` when the extension is used, but that the actual children elements are not unfolded until they're accessed.

mint-thompson (commented):
I have no findings beyond what @ngfreiter has already mentioned. I'll take another look once you've addressed those issues.

cmoesel:
I only had time for a quick scan, but I do have some questions...

So, in StructureDefinitionExporter, I think you are detecting when the author uses a caret rule to apply constraints from another profile, and then you are actually directly applying those constraints to the element, right?  So in a way, there is a little redundancy there (since the constraints are represented by the presence of the extension, but also directly manifested in the element) -- right?  I'm not sure if that is what the spec expects for behavior or not.  The spec might think that just having the extension there is enough (and not require that you reflect the constraints).  I think your approach makes sense and simplifies things, so I'm not complaining -- but I am wondering...

Given this approach, to clarify @ngfreiter's last comment...  I think that if the current profile _has_ child elements already there, then the constraints from the target profile _should_ still be copied to them.  But if they're not there already, then we shouldn't create them just for that purpose -- but rather wait until we need to unfold them before we do it.

Which brings me to my other comment.  It looks to me like the special profile element extension will only be applied during an unfold operation when it is unfolding a content reference.  But really, I think that it needs to happen in other cases too in case we are unfolding children under a normal element that uses one of these special extensions.

ngfreiter:
> I think that if the current profile _has_ child elements already there, then the constraints from the target profile _should_ still be copied to them.

I might be misunderstanding the ask here, but to me, this sounds like it is expanding the scope of this task in a way I'm not sure we want. Basically, I don't think we do this for an `only` rule, and so I'm not sure we should do it for these caret rules. See this example
```
Profile: HumanNameProfile
Parent: HumanName
* given 0..1

Profile: ProfilePatient1
Parent: Patient
* name.given 0..2
* name only HumanNameProfile

Profile: ProfilePatient2
Parent: Patient
* name only HumanNameProfile
* name.given 0..2
```
On `ProfilePatient1`, we apply a rule to `name.given`, so it exists, but then when we type constraint `name`, we do not go back and apply the constraints introduced by `HumanNameProfile` to `name.given`, even though `name.given` already exists. If you test this out, `ProfilePatient1` will have `name.given` set to `0..2`, not `0..1`.

I could definitely be missing something here, but I don't see why this extension should work any different than an only rule that is constraining something to a profile. To me this extension is just a way to apply an only rule to something that doesn't have an actual type, aka it is a `BackboneElement`.

cmoesel:
@ngfreiter -- What you describe is what I actually expected at first.  But then I saw that in the PR, it seems that the constraints from the targeted profile _are_ being explicitly applied to the element which holds the extension.  So... it feels inconsistent to directly apply _some_ of the constraints (those on the root element), but not _all_ of the constraints (those on the child paths).  It seems to me like maybe we should either apply none or all.

The upside of applying all constraints is that SUSHI does correct validation on subsequent actions on child elements (because those constraints are physically there), whereas not applying them means SUSHI acts as if they are not there at all.  The downsides are (a) more complexity in implementation (it's basically a merge operation), and (b) I don't know if the IG Publisher expects that or not.

I agree it is inconsistent w/ what we currently do when applying a profile, and it would be actually impossible to do when applying more than one profile (since multiple profiles means "pick one", _not_ "do them all").  We _do_ apply the constraints when unfolding elements -- but I think we reset the "original" so they don't show up in the diff (and we actually still have the probably when there are multiple profiles).

So...  yeah, perhaps this needs more thought and discussion -- and maybe a tad bit of investigation w/ the IG Publisher too?

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1819,13 +1823,30 @@ export class ElementDefinition {
...
-
         // Get the original resource JSON so we unfold unconstrained reference
         const type = this.structDef.type;
         const json = fisher.fishForFHIR(type, Type.Resource);
-        if (json) {
+        // contentReference elements will not contain a type field, so we must fish for the StructDef and
+        // check the differential
+        const profileJson = fisher.fishForFHIR(this.structDef.id, Type.Profile);
+        if (this.hasProfileElementExtension(profileJson)) {
+          const def = StructureDefinition.fromJSON(profileJson);
```
ngfreiter, line 1833: It isn't clear to me that this is necessary. From what I am seeing `this.structDef` does already have the element with the required constraints on it. I'm testing with this example:
```
Profile: MyQ
Parent: Questionnaire
* item.code 0..2
* item ^type[0].profile[0] = http://example.org/StructureDefinition/MyQ
* item ^type[0].profile[0].extension[0].url = http://hl7.org/fhir/StructureDefinition/elementdefinition-profile-element
* item ^type[0].profile[0].extension[0].valueString = "Questionnaire.item"
* item.item.code 0..3
```
And I'm finding that just doing:
```
          const def = this.structDef;
```
works the same for me, and the new tests pass as well. Let me know if you have an example of why we really do need to fish for the profile.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1819,13 +1823,30 @@ export class ElementDefinition {
...
+            eClone.structDef = this.structDef;
+            eClone.captureOriginal();
+            return eClone;
+          });
+          if (newElements.length > 0) {
+            // If we successfully unfolded, this element is no longer a content reference
+            this.type = referencedElement.type;
+            delete this.contentReference;
+          }
+        } else if (json) {
```
ngfreiter, line 1849: As far as I can tell, this code all repeats the code below, with the only major difference being which `def` we are using. Would we be able to update this so that we don't have the repeated code?

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1864,7 +1885,18 @@ export class ElementDefinition {
...
+          const def = StructureDefinition.fromJSON(json);
+          const targetElement = def.findElement(backboneProfile as string);
+          newElements = targetElement?.children().map(e => {
+            const eClone = e.clone();
+            eClone.id = eClone.id.replace(targetElement.id, this.id);
+            eClone.structDef = this.structDef;
+            eClone.captureOriginal();
+            return eClone;
+          });
+        } else if (json) {
```
ngfreiter, line 1899: So as I understand it, this case is handling unfolding when the element isn't a content reference, so an example like:
```
Profile: MyPatient
Parent: Patient
* name.given 1..1

Profile: MyPatient2
Parent: Patient
* name ^type[0].profile[0] = http://example.org/StructureDefinition/MyPatient
* name ^type[0].profile[0].extension[0].url = http://hl7.org/fhir/StructureDefinition/elementdefinition-profile-element
* name ^type[0].profile[0].extension[0].valueString = "Patient.name"
* name.given 1..2
```
I thought it was still an open question on if we were going to allow this extension on non-`BackboneElement` type elements. Did we get clarity on if this extension can be used in that context? Very possible I am just forgetting the resolution of that.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1889,6 +1921,52 @@ export class ElementDefinition {
     return [];
   }
 
+  /**
+   * Checks a StructureDefinition's differential to determine if the profile element extension has been used on
+   * an element.
+   * @param {Object} profileJson - The json reprsentation of this ElementDefinition's structDef
```
ngfreiter, line 1927: Typo on "representation".

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1889,6 +1921,52 @@ export class ElementDefinition {
...
+   * @returns {boolean | string} If the profile element extension is not found, false if returned. If it is found,
+   * the id of the BackBone Element being profiled is returned.
+   */
+  private hasProfileElementExtension(profileJson?: any): boolean | string {
```
ngfreiter, line 1931: I think it makes sense to clarify the name of this function and the return type. If we want it to return the `id` of the `BackboneElement` being profiled, let's just make it do that, return only a string, and rename it something like `getProfileElementExtensionTarget`. Or along those lines.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -2058,6 +2136,47 @@ export class ElementDefinition {
     return slice;
   }
 
+  copyFromParent(profiledElement: ElementDefinition, structDef: StructureDefinition): void {
```
ngfreiter, line 2139: Am I missing something major, or is `copyFromParent` only used in `copyFromParent`?

src/fhirtypes/ElementDefinition.ts
ngfreiter: I think this is important to figure out, because if this extension can truly only be used on a `BackboneElement`, then the _only_ place we would have to unfold would be on a `contentReference`, so I think this code down here could actually go, and it would simplify things more. Now there is still an open question of if we would then want to add functionality so that we better handle applying constraints to elements that are already unfolded when type constraints are made, but I think that is a similarly open question for `only` rules, and so more of a separate task.

ngfreiter (dismissed):
I have a few questions and suggestions, and a few things we may have already resolved that I am just forgetting.

src/fhirtypes/ElementDefinition.ts
julianxcarter: `this.structdef` does in fact have the element with the required constraints on it, however that element does not contain the profile-element extension on it's `type.profile` property as the `_profile` array is not present (shown in the attached screenshot). To check and see if the extension is there, I found that I had to fish for that StructureDefinition separately and then check the differential array.  

There's no need to convert that JSON back into a `StructureDefinition` on line 1833 as I'm currently doing, but I do think there's a need to fish for the `profileJson` in order to check for the extension and know to unfold from the element on the current profile, rather than the parent profile.
Screen Shot 2021-05-11 at 3 56 39 PM

src/fhirtypes/ElementDefinition.ts
julianxcarter: > I thought it was still an open question on if we were going to allow this extension on non-`BackboneElement` type elements. Did we get clarity on if this extension can be used in that context? Very possible I am just forgetting the resolution of that.

[I posed the question on Zulip](https://chat.fhir.org/#narrow/stream/179177-conformance/topic/profile-element.20Extension/near/238329205) and we do want to scope the use of this extension to only include `BackboneElement`s. So, I've removed any unfolding that would've happened on non-contentReference elements. Because we haven't made a practice of altering the constraints on pre-existing children before, we've decided to leave the non-contentReference use of this extension for another day.

src/fhirtypes/ElementDefinition.ts
julianxcarter: I chose to just simplify the return type to only be a boolean, returning `true` if the extension is found and `false` otherwise. Since we're only including support for the profile-element extension on `contentReference` elements, there shouldn't be a need to return the `id` of the `BackboneElement` anyways (it will always be the same as the `contentReferenceId`). When we expand support to include other `BackboneElement`s, it'd probably make sense to just make `getProfileElementExtensionTarget` a separate function.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1889,6 +1892,69 @@ export class ElementDefinition {
...
+  }
+
+  /**
+   * Checks a StructureDefinition's differential to determine if the profile element extension has been used on
+   * an element.
+   * @param {Object} profileJson - The json representation of this ElementDefinition's structDef
+   * @returns {boolean} True if the profile element extension is found on this elements profile property, false
+   * if the extension is not found
+   */
+  private hasProfileElementExtension(profileJson?: any): boolean | string {
```
ngfreiter, line 1919: The return type here should just be `boolean` now. And also is `profileJson` still an optional argument? I think at this point, won't it always be passed?

I am also looking through this function, and I know before it was also handling unfolding on non-`BackboneElement` data types. Are there pieces of this function that could be simplified now that we know this is only getting called on `contentReference` elements? Or do you feel we can't/shouldn't make those simplifications.

ngfreiter (commented):
Thanks for all the work, it's looking great.

mint-thompson:
With the latest updates, things are looking good! No comments beyond what @ngfreiter has asked in the most recent post.

src/fhirtypes/ElementDefinition.ts
julianxcarter: There definitely are simplifications that could be made to the function, but I thought that we would still want the function to identify the profile-element extension on non-`contentReference` elements if/when we go back and broaden support for the extension. That's why I kept `profileJson` as an optional parameter, since it's necessary for `contentReference` elements but not needed for regular elements.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1889,6 +1892,60 @@ export class ElementDefinition {
...
+    const profileCanonical = elementType.profile?.[0];
+    let extensionUrl, targetElement: string;
+
+    // If there is no _profile array, the profile-element extension is not present
+    if (elementType._profile) {
+      extensionUrl = elementType._profile[0].extension[0].url;
+      targetElement = elementType._profile[0].extension[0].valueString;
+    } else {
+      return false;
+    }
```
mint-thompson, line 1940: Is it okay to only be checking `profile[0]` and `_profile[0]` here, or is it possible that the profile element extension is at another index in these arrays?

julianxcarter: That could certainly create some problems, I'll change this so that the extension will be found even if it's not the first element in the `_profile` array.

julianxcarter: Addressed in `b4d43c8`

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1889,6 +1892,73 @@ export class ElementDefinition {
...
+
+    // If the profile and _profile arrays are not present, the profile-element extension is not present
+    if (elementType.profile && elementType._profile) {
+      const profileIndex = elementType._profile.findIndex(
+        (profile: any) =>
+          profile &&
+          profile.extension.find(
+            (extension: any) =>
+              extension.hasOwnProperty('url') && extension.hasOwnProperty('valueString')
+          ) !== undefined
```
mint-thompson, line 1941: This can be shortened a bit, since we don't need to actually use the found value:
```
(profile: any) => profile.extension?.some( ... )
```
Alternatively, since we do eventually want the index of that extension, it may be possible to rewrite this and the following search in order to grab both indices at the same time.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1889,6 +1892,73 @@ export class ElementDefinition {
...
+      );
+
+      const extensionIndex = elementType._profile[profileIndex].extension?.findIndex(
+        (extensionObj: any) =>
+          extensionObj.hasOwnProperty('url') && extensionObj.hasOwnProperty('valueString')
+      );
+      extensionUrl = elementType._profile[profileIndex].extension[extensionIndex].url;
+      targetElement = elementType._profile[profileIndex].extension[extensionIndex].valueString;
+      profileCanonical = elementType.profile[profileIndex];
```
mint-thompson, line 1950: I suspect bad things might happen here if the search for `profileIndex` is unsuccessful. I haven't yet been able to craft a test case that proves this, though.

mint-thompson (changes requested):
So, I may not have been sufficiently diligent in my previous reviews, as I only now discovered something about `hasProfileElementExtension` that needs to be investigated.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1889,6 +1892,73 @@ export class ElementDefinition {
...
+  }
+
+  /**
+   * Checks a StructureDefinition's differential to determine if the profile element extension has been used on
+   * an element.
+   * @param {Object} profileJson - The json representation of this ElementDefinition's structDef
+   * @returns {boolean} True if the profile element extension is found on this elements profile property, false
+   * if the extension is not found
+   */
+  private hasProfileElementExtension(profileJson: any): boolean {
```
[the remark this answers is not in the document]
julianxcarter: You're totally right. I've updated the unit test to properly test these additions in commit `1a05ba2`, the added code should be covered now!

src/fhirtypes/ElementDefinition.ts
julianxcarter: This is definitely correct. I've added in a safety check for `profileIndex` in commit `5504e91`. `extensionIndex` should not need a check of its own, since `profileIndex` existing implies that the profile-element extension exists somewhere in that array.

src/fhirtypes/ElementDefinition.ts
julianxcarter: Thank you! I've switched over to using the `some` function in commit `5504e91`, though I did not combine the two searches into one since I don't think either `some` or `findIndex` actually allow for values to be set in their callback functions.

ngfreiter (dismissed):
Looks good to me!

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1889,6 +1892,77 @@ export class ElementDefinition {
...
+    // If the profile and _profile arrays are not present, the profile-element extension is not present
+    if (elementType.profile && elementType._profile) {
+      const profileIndex = elementType._profile.findIndex(
+        (profile: any) =>
+          profile &&
+          profile.extension.some(
+            (extension: any) =>
+              extension.hasOwnProperty('url') && extension.hasOwnProperty('valueString')
+          )
+      );
```
mint-thompson, line 1942: I think this will need to be adjusted, because if there's more than one extension with a string value on the element, the one that gets found might not be the profile element extension. I got the test case in `StructureDefinitionExporter.test.ts` on line 248 by having these rules on the profile:
```ts
    const otherExtensionRule = new CaretValueRule('item');
    otherExtensionRule.caretPath = 'type.profile.extension[0].url';
    otherExtensionRule.value = 'http://hl7.org/fhir/us/minimal/StructureDefinition/FooExtension';

    const otherExtensionValue = new CaretValueRule('item');
    otherExtensionValue.caretPath = 'type.profile.extension[0].valueString';
    otherExtensionValue.value = 'Foo';

    const extensionRule = new CaretValueRule('item');
    extensionRule.caretPath = 'type.profile.extension[1].url';
    extensionRule.value =
      'http://hl7.org/fhir/StructureDefinition/elementdefinition-profile-element';
    const targetElementRule = new CaretValueRule('item');
    targetElementRule.caretPath = 'type.profile.extension[1].valueString';
    targetElementRule.value = 'Questionnaire.item';
```

The function finds the first extension, which isn't the profile element extension, so the function returns false.

mint-thompson (changes requested):
Okay, hopefully this is the last thing that needs fixing: a case where the profile element extension _is_ present, but isn't found.

src/fhirtypes/ElementDefinition.ts
julianxcarter: This is totally right, thank you for finding this! In commit `96fa4a6` I've added in a third condition to the search on line 1940 that ensures that the extension's `url` property is the profile-element extension's canonical. That should address this case, and I've also added in a unit test that tests this scenario.

mint-thompson (approved):

Issue #756, opened 2021-02-11
Provide a mechanism to output pre-processed FSH


cmoesel:
There are a number of things SUSHI does in pre-processing, but most the most notable ones are:
* expansion of RuleSets
* numbering of soft-indexing

Some users have reported that it would be easier to debug issues if they could see the expanded output.  For example, if they get an error about Bundle.entry[25], it's hard to know which one that is when you've used soft indexing.  A "spelled-out" edition of the FSH would help with that.

This could happen automatically every time or it could be supported via some flag or something.

cmoesel:
This was fixed by #804 a long time ago.

PR #804, opened 2021-04-21
Add option to output preprocessed FSH

mint-thompson:
Completes #756 and [CIMPL-669](https://standardhealthrecord.atlassian.net/browse/CIMPL-669).

By setting a flag at the command line, preprocessed FSH is written to a directory named `_preprocessed` in the output directory. Preprocessed FSH represents the in-memory entities that exist after the processing steps have been completed, such as Alias resolution, RuleSet insertion, and soft index calculation. As a result, it won't exactly match the input FSH. But, in this form, it is useful for troubleshooting problems that occur when other programs, such as the IG publisher, run into problems with the FHIR produced by SUSHI.

A `toFSH` method is added to FSH rules and entities. These may eventually be of use to goFSH, but that's a separate question. This does mean that a lot of classes have new methods and tests. But, they are all following a similar pattern, and have minimal interdependencies, so it hopefully shouldn't be too complicated to understand the changes to any individual class.

There is some reorganization of commonly used components, as well as changes to some import statements to target specific files rather than groups of files. These changes help to avoid circular imports, which causes problems when running tests.

I've written tests for the new functions, but since this is also a new command line option, I recommend running with that option a few times to see it in action.


src/fshtypes/rules/ContainsRule.ts
```diff
@@ -11,6 +12,24 @@ export class ContainsRule extends Rule {
...
+      } else {
+        line += `${item.name}`;
+      }
+      return line;
+    });
+
+    return `* ${this.path} contains${
+      itemsWithAssociatedRules.length > 1 ? `${EOL}    ` : ' '
+    }${itemsWithAssociatedRules.join(` and${EOL}    `)}`; // New line and indent each
+  }
```
ngfreiter, line 32: I think this ends up writing syntactically invalid `ContainsRules` because a `ContainsRule` is always supposed to be followed by a cardinality.

ngfreiter (commented):
I noticed a few things, but I want to clarify something, is the intent that the FSH in `_preprocessed` could be run through SUSHI and produce the exact same output as the original FSH? I wasn't sure if that was a requirement of this feature, or if the stuff in `_preprocessed` is really only an intermediate step that cannot be reliably trusted?

src/fshtypes/rules/AssignmentRule.ts
```diff
@@ -27,4 +31,29 @@ export class AssignmentRule extends Rule {
...
+  toFSH(): string {
+    let printableValue = '';
+    if (
+      typeof this.value === 'boolean' ||
+      typeof this.value === 'number' ||
+      typeof this.value === 'bigint'
+    ) {
+      printableValue = String(this.value);
+    } else if (typeof this.value === 'string') {
+      printableValue = this.isInstance ? this.value : `"${fshifyString(this.value)}"`;
```
ngfreiter, line 44: It doesn't look like `fshifyString` handles the case that `this.value` has quotations in it. So the preprocessed string is written with un-escaped quotation marks.

src/fshtypes/rules/ContainsRule.ts
mint-thompson: That's a good point. The FSH produced here is not generally intended to be run, but it should still be syntactically valid. Is it safe to always include a cardinality of `0..` as part of the `ContainsRule` with the expectation that a later `CardRule` will get the correct cardinality on to the slice? I feel like that's safe.

src/fshtypes/rules/AssignmentRule.ts
mint-thompson: I think `fshifyString` handles this with [the second replacement](https://github.com/FHIR/sushi/pull/804/files#diff-6fca0c9ea8a1b6c0b3966479a503a1c9e7427f1074e9e2569f7862807fddcfb4R17). Also, if it weren't handling these strings correctly, wouldn't [this test for string values on an AssignmentRule](https://github.com/FHIR/sushi/pull/804/files#diff-1624e0922423d698734d0b7e59bdde701578748bce84498b3c1c44b3927d9a3aR37-R43) fail?

src/fhirtypes/CodeSystem.ts
```diff
@@ -3,8 +3,8 @@ import { Meta } from './specialTypes';
 import { Extension } from '../fshtypes';
 import { Narrative, Resource, Identifier, CodeableConcept, Coding } from './dataTypes';
 import { ContactDetail, UsageContext } from './metaDataTypes';
-import { HasName, HasId } from './common';
-import { applyMixins } from '../utils';
+import { HasName, HasId } from './mixins';
+import { applyMixins } from '../utils/Mixin';
```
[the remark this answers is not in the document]
mint-thompson: We figured this would be likely to happen. One way or another, we'll have to resolve that merge conflict when it appears, but I don't think it'll be too difficult to do that.

src/fshtypes/rules/AssignmentRule.ts
mint-thompson: Oh, I think I found what you're talking about: `CaretValueRule` is missing this usage of `fshifyString`. I'll add that in and update the test to prove it.

julianxcarter:
> I noticed a few things, but I want to clarify something, is the intent that the FSH in `_preprocessed` could be run through SUSHI and produce the exact same output as the original FSH? I wasn't sure if that was a requirement of this feature, or if the stuff in `_preprocessed` is really only an intermediate step that cannot be reliably trusted?

I was under the impression that the purpose of the feature wasn't so that the FSH could be run, but rather so that users could have something exact to reference if the IG Publisher were to log an error at a specific line number or something of the sort.

test/fshtypes/AllowedRules.test.ts
```diff
@@ -158,7 +158,7 @@ describe('isAllowedRule', () => {
       expect(isAllowedRule(m, new MappingRule('foo'))).toBeTrue();
     });
 
-    it('should not allow invalid rules on a Maping', () => {
+    it('should not allow invalid rules on a Mapping', () => {
```
julianxcarter, line 161: Nice catch!

julianxcarter (commented):
This works well for me! Apart from the what Nick pointed out in regards to CaretValueRules, I can't think of anything else that I'd like to see changed.

ngfreiter:
> I was under the impression that the purpose of the feature wasn't so that the FSH could be run, but rather so that users could have something exact to reference if the IG Publisher were to log an error at a specific line number or something of the sort.

Yeah I understand that this is really meant to be looked at. My question is more about how strict we plan on being here. In a perfect world I think this FSH should be able to be run through SUSHI and produce the same output as the un-preprocessed content. It is a good way to test if the preprocessed content is correctly capturing the original content. But I just wasn't sure how strict we should be. I guess we can wait and see how the updates to `Contains` and `Caret` rules change things.

src/fshtypes/rules/ContainsRule.ts
ngfreiter: I also feel like that's safe, good idea!

src/utils/Processing.ts
```diff
@@ -355,6 +356,36 @@ export function writeFHIRResources(
...
+      ...doc.mappings.values()
+    ];
+    // Sort entities by original line number, then write them out.
+    sortBy(entities, 'sourceInfo.location.startLine').forEach(entity => {
+      fileContent += `// Originally defined on lines ${entity.sourceInfo.location.startLine} - ${entity.sourceInfo.location.endLine}${EOL}`;
+      fileContent += `${entity.toFSH()}${EOL}${EOL}`;
+    });
+    const outPath = path.relative(inDir, doc.file);
+    fs.ensureFileSync(path.join(preprocessedPath, outPath));
+    fs.writeFileSync(path.join(preprocessedPath, outPath), fileContent);
```
ngfreiter, line 384: I think before writing the file, we want to check that it actually has content. Something like a file of only aliases will end up empty after preprocessing. Then we can either not write this file at all, or write a comment like: 
```
// This file has no content after preprocessing
```
Personally I think I prefer the comment for transparency.

ngfreiter: Actually, I think we may need to make some changes to how aliases are handled in general. The one issue is that the way we use aliases, they are not actually all resolved in the importer. There are cases like this:
```
* extension[USCoreRace].extension[ombCategory].valueCoding = OmbRaceCat#2106-3 "White"
```
where `USCoreRace` is an alias, but we don't resolve that in the importer, since the importer just treats the entire path as a string, but rather when we go to `fish` for the `USCoreRace` definition. I'm not totally sure how you want to handle this. The options I see are:
1. Do still output aliases in the preprocessed output
2. Somehow figure out a way to resolve these aliases in the rules we are `toFSH`ing
I think 2 seems harder than 1, but I'm not totally sure. Alternatively there is an option 3 which would be that we don't worry too much about this, but on the other hand someone looking at this output could get confused wondering what `USCoreRace` points to, since that alias doesn't exist, and nothing in the dependencies or FSH is defined with that name. What do you think?

ngfreiter (commented):
This is looking great to me. I tried running the preprocessed version of mCODE back through SUSHI, and it almost perfectly matches the original output. There is just one minor thing with how we handle aliases.

src/app.ts
```diff
@@ -44,6 +45,7 @@ async function app() {
     .usage('[path-to-fsh-defs] [options]')
     .option('-o, --out <out>', 'the path to the output folder')
     .option('-d, --debug', 'output extra debugging information')
+    .option('-p, --preprocessed', 'output FSH produced by pre-processing steps')
```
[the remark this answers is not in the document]
ngfreiter: Reminder that we need this in the README.md. Sorry if I'm reminding you of something you were planning to do, but didn't see it in the latest commit.

src/utils/Processing.ts
mint-thompson: I'm not seeing any cases like this in the most recent version on the `fsh-mcode` repo, and I don't entirely understand what you're trying to explain here. Are there any other examples you can give me of what you're talking about?

ngfreiter: I think if you check out the https://github.com/HL7/fhir-mCODE-ig repo, you should be able to see an example of this. On [this](https://github.com/HL7/fhir-mCODE-ig/blob/c2e1fdcaee8e1734720b7213b1d936318090a1be/input/fsh/EX_ExtendedExample.fsh#L743) line, `USCoreRace` is used as an alias, defined in the `AL_USCore.fsh` file. If you check out SUSHI, in `FSHTank.ts`'s `fish` function, we do this:
```
    item = this.resolveAlias(item) ?? item;
```
and that is the line that resolves the alias, it doesn't happen on the import side.

mint-thompson: Ah, that clears things up! So, since we're not entirely done with Aliases after these preprocessing steps, I think it makes sense to include them in the output. The fishing is happening in `setPropertyOnInstance`, which doesn't get the rule itself, just a `PathPart[]` that we've built. The FSH output by this feature is meant to represent the objects we have in memory, and the Rule's path does not change as a result of this Alias resolution.

Fortunately, a file with nothing but RuleSets should still be empty after preprocessing, so that will be useful for demonstration in a test case.

ngfreiter: Ok great, makes sense to me.

README.md
```diff
@@ -11,7 +11,7 @@ For more information about the evolving FSH syntax see the [FHIR Shorthand Refer
 
 # Installation for SUSHI Users
 
-SUSHI requires [Node.js](https://nodejs.org/) to be installed on the user's system.  Users should install Node.js 12 (LTS), although the previous LTS version (Node.js 10) is also expected to work.
+SUSHI requires [Node.js](https://nodejs.org/) to be installed on the user's system. Users should install Node.js 12 (LTS), although the previous LTS version (Node.js 10) is also expected to work.
```
ngfreiter, line 14: Glad to have another ally in the war against double space.

ngfreiter: Oh I see the ally may have been an automatic formatter, but glad to have the ally all the same. And thanks for formatting this, looks more readable now.

ngfreiter (approved):
Looks good to me!

julianxcarter (approved):

PR #807, opened 2021-04-25
Remove ValueSet expansion

mint-thompson:
Completes task [INT-999](https://standardhealthrecord.atlassian.net/browse/INT-999).

The ValueSet expansion feature that was added in PR #763, which merged in branch `cimpl-655-valueset-expansion`, is removed. The feature did not provide the desired behavior when the resulting ValueSet resources were processed by the IG publisher, and  allowing the feature to remain could be misleading and confusing to users.


cmoesel:
I thought that when we discussed this, we decided to start by putting in a deprecation warning, and then we would actually _remove_ the functionality w/ the next major release.  But I know we went back and forth a few times, so I might be remembering the wrong conclusion?

cmoesel:
You could just pop this PR into DRAFT mode and issue a new one w/ the deprecation.  I hope we'll do the next major release soon, so I'm not too concerned about this PR hanging out for a bit.

jafeltra (approved):
Sad, but all relevant pieces are removed!

ngfreiter (approved):

PR #808, opened 2021-04-26
Deprecate ValueSet expansion

mint-thompson:
Addresses task [INT-999](https://standardhealthrecord.atlassian.net/browse/INT-999).

A warning is emitted when exporting a ValueSet if the ValueSet has the expansion parameters set. This function is being deprecated because it doesn't actually solve the problem it was intended to solve when sending ValueSet resources to the IG publisher. Keeping the feature in could be confusing, so it is being deprecated.


ngfreiter (approved):
Looks good to me!

cmoesel (approved):
A++.  Would purchase from again.

PR #809, opened 2021-04-27
Import CodeSystems with hierarchical codes

mint-thompson:
Completes task [CIMPL-725](https://standardhealthrecord.atlassian.net/browse/CIMPL-725).

Concepts in a CodeSystem can exist in a hierarchy. This PR adds new FSH syntax to support building such a hierarchy. Additionally, there is new caret value syntax in order to set values on concepts within the hierarchy. This is only the import side, so the export side won't arrange the hierarchy. and it won't do anything with the caret value rules on concepts.

To arrange the hierarchy, list out the concepts that precede the new one in the hierarchy:
```
* #top "Top level concept"
* #top #middle "Concept in middle of hierarchy"
* #top #middle #more "You can keep going with more nesting."
```
The hierarchy must be built in order, and an error is thrown if the supporting hierarchy needed for a concept doesn't exist:
```
* #this #that "Won't work, since the preceding concept doesn't exist"
* #this "If the order of these concept rules were switched, it would work."
```
To set values on concepts that can't be set using normal FSH syntax, the new `CodeCaretValueRule` can be used:
```
* #something "Some concept"
* #something ^property[0].code = "someProperty"
* #something #else "Some other concept"
* #something #else ^designation[0].value = "Nested concepts can be used too."
```
Note that `CodeSystem.concept` does not inherit from `ElementDefinition`, so the set of properties available is a little different.

I implemented this as a new type of rule because of the differences between traversing by structure definition path and traversing by a sequence of concepts. Additionally, this type of rule should only be allowed on CodeSystems and RuleSets, so it did not make sense to me to have this new class inherit from `CaretValueRule`. Storing the path as a dot-separated list of concepts would not be reliable, as a concept's code may contain a `.` character, hence the list of strings.


antlr/src/main/antlr/FSH.g4
```diff
@@ -100,7 +101,7 @@ strength:           KW_EXAMPLE | KW_PREFERRED | KW_EXTENSIBLE | KW_REQUIRED;
 value:              STRING | MULTILINE_STRING | NUMBER | DATETIME | TIME | reference | canonical | code | quantity | ratio | bool | name;
 item:               name (KW_NAMED name)? CARD flag*;
 code:               CODE STRING?;
-concept:            STAR code (STRING | MULTILINE_STRING)?;
+concept:            STAR CODE+ STRING? (STRING | MULTILINE_STRING)?;
```
ngfreiter, line 104: Would it work to make `CODE` optional here? So `CODE*`. Because that way couldn't we eliminate `CaretValueRule` from `CodeSystem` altogether and process everything via `codeCaretValueRule`?

ngfreiter: I guess we couldn't remove `CaretValueRule` actually until the export side is done too, but does this idea make sense?

src/import/FSHImporter.ts
```diff
@@ -1215,10 +1222,23 @@ export class FSHImporter extends FSHVisitor {
...
+      concept.display = availableStrings[0];
+    }
+    if (availableStrings.length > 1) {
+      concept.definition = availableStrings[1];
+    } else if (ctx.MULTILINE_STRING()) {
+      concept.definition = this.extractMultilineString(ctx.MULTILINE_STRING());
+    }
     if (codePart.system) {
       logger.error(
         'Do not include the system when listing concepts for a code system.',
```
ngfreiter, line 1244: Since we do this check for the `codePart`, should we do it for each code in the hierarchy for consistency?

src/fshtypes/FshCodeSystem.ts
```diff
@@ -26,6 +28,24 @@ export class FshCodeSystem extends FshEntity {
...
+      if (
+        !this.rules.some(rule => {
+          return (
+            rule instanceof ConceptRule &&
+            rule.code === predecessor &&
+            isEqual(rule.hierarchy, newConcept.hierarchy.slice(0, i))
+          );
+        })
+      ) {
+        throw new CodeSystemIncorrectHierarchyError(this.id, newConcept.code);
```
ngfreiter, line 46: I'm not sure we can safely do this check here, what about RuleSets:
```
CodeSystem: MyCS
* insert MyRS
* #code1 #code2 "foo"

RuleSet: MyRS
* #code1
```
Can we do this on the export side instead?

test/import/FSHImporter.CodeSystem.test.ts
```diff
@@ -392,6 +469,58 @@ describe('FSHImporter', () => {
       });
     });
 
+    describe('#codeCaretValueRule', () => {
+      it('should parse a code system that uses a CodeCaretValueRule on a top-level concept', () => {
+        const input = `
+        CodeSystem: ZOO
+        * #anteater "Anteater"
+        * #anteater ^property[0].valueString = "Their threat pose is really cute."
```
ngfreiter, line 477: Gotta go look up anteater threat pose now!

ngfreiter (commented):
Looking good, I had a few questions and suggestions.

One thing to consider is that this will have to work with the indent blocks in #797. That code works on the assumption that the context of a rule is going to be a single string, and the path of the rule is going to be a single string. It isn't immediately obvious to me how those two PRs are going to work together. I think we will be able to figure something out, but wanted to make you aware of that, and you may want to take a look at #797 and double check that you have an idea of how the representation of hierarchy in this PR will work with that.

src/import/FSHImporter.ts
mint-thompson: Yes, I think that makes sense. We want authors to write `#first #second #third`, without systems attached to any of them.

src/fshtypes/FshCodeSystem.ts
mint-thompson: So, this is something that I also realized recently: InsertRules can make all kinds of things happen on CodeSystems that shouldn't be able to happen. I opened [a ticket](https://standardhealthrecord.atlassian.net/browse/CIMPL-726) for this, and I think your comment here would be in scope for that task. I think leaving the implementation like this for this PR is fine, as it is consistent with the existing implementation, and they will both be addressed by that new task.

antlr/src/main/antlr/FSH.g4
mint-thompson: I assume that you're talking about `codeCaretValueRule` here? So then those would be handled like this:
- if there are no `CODE` tokens, handle them the way we currently handle `caretValueRule` on a `CodeSystem`, with the bonus of knowing that the user didn't accidentally include a `path`.
- if there are `CODE` tokens, handle them using the new processing that applies values to concepts within the `CodeSystem`.

I think I like it, actually? It feels kind of weird, but at the same time, it sort of makes it clearer how we want folks to think about a `CodeSystem`: there are the elements on the `CodeSystem` itself that should be accessed with a `caretPath`, and there are the concepts that should be accessed with a `CODE+` path. The main side effect would be that a regular caret value rule would now be syntactically incorrect. I would want to make sure the error message in that case is still clear.

mint-thompson:
@ngfreiter thanks for the review! As far as the interactions with indent blocks, my expectation is that with both this PR and the indent block PR merged in, there will not yet be a way to specify concept hierarchies using indentation. This will have to be added separately, both for defining concepts and for caret value rules on concepts. There will probably need to be a `codePathRule` for CodeSystems that works similarly to `pathRule` for other entities.

jafeltra:
This is looking good! I haven't noticed anything else beyond what Nick pointed out, so I'll just plan to review again after any of those updates!

ngfreiter (approved):
Looks good!

jafeltra (approved):
Looks good to me!

ngfreiter:
> I know I approved, but please hold I want to just double check and make sure we've run a regression on this before merging.

No diff!

PR #811, opened 2021-04-29
Add more repos to repos-all.txt

ngfreiter:
This adds more repos to repos-all.txt, since a good deal more were available after running find-repos. Although running regressions takes a while, I'm going to request that reviewers do it in the background before approving, just to make sure the new repos work. I've tested it for me and they seem to.


jafeltra:
Running the regressions now!

jafeltra (approved):
Perfection!

PR #814, opened 2021-05-10
Export CodeSystems with hierarchical codes

mint-thompson:
Completes task [CIMPL-697](https://standardhealthrecord.atlassian.net/browse/CIMPL-697).

Concepts within a code system are arranged hierarchically based on the ConceptRule's specified hierarchy. The path to a hierarchical code must exist in order to be added, and this is checked when adding the ConceptRule to the FshCodeSystem.

Additionally, a CodeCaretValueRule may be used to specify values on a concept anywhere in the hierarchy. The codePath on this rule is traversed after all the concepts are added so that the correct element in the CodeSystem can be located. Just like for CaretValueRules, the value to set is validated before being assigned.


jafeltra (approved):
This looks great to me! The code makes sense, the tests cover the cases I could think of, and I ran the regression on all the repos and there were not changes!

src/export/CodeSystemExporter.ts
```diff
@@ -54,10 +71,58 @@ export class CodeSystemExporter {
...
+        );
+        setPropertyOnInstance(codeSystem, pathParts, assignedValue, this.fisher);
+      } catch (e) {
+        logger.error(e.message, rule.sourceInfo);
+      }
+    }
+  }
+
+  private findConceptPath(codeSystem: CodeSystem, codePath: string[]): string {
+    const conceptPath: number[] = [];
```
cmoesel, line 96: This is a very clever approach.  I like what you've done here.  My only suggestion is that the name `conceptPath` doesn't seem a perfect fit since it's really an array of indices.  I'd suggest renaming to something more descriptive (maybe `conceptIndexPath`?) or changing line 103 so it pushes the string `concept[${conceptIndex}]` instead of mapping a numeric index array later.

src/export/CodeSystemExporter.ts
```diff
@@ -26,19 +27,35 @@ export class CodeSystemExporter {
...
+        if (concept.definition) {
+          newConcept.definition = concept.definition;
+        }
+        concept.hierarchy.forEach(ancestorCode => {
+          const ancestorConcept = conceptContainer.find(
+            ancestorConcept => ancestorConcept.code === ancestorCode
+          );
+          if (!ancestorConcept.concept) {
+            ancestorConcept.concept = [];
+          }
```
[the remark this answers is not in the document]
cmoesel: I tested manually and it looks like import process prevents us from getting this far with an invalid hierarchy.  And, in fact, it's the FSH CodeSystem that prevents it.  Still... I wonder, should we be defensive here just in case?  In theory, if someone constructed a `FshCodeSystem` by pushing to the `rules` array directly (instead of through `addConcept`), we could get to that case.

src/export/CodeSystemExporter.ts
```diff
@@ -54,10 +71,58 @@ export class CodeSystemExporter {
     }
   }
 
+  private setCodeCaretRules(
+    codeSystem: CodeSystem,
+    csStructureDefinition: StructureDefinition,
+    rules: CodeCaretValueRule[]
+  ) {
+    resolveSoftIndexing(rules);
```
cmoesel, line 79: What soft indices might exist that need to be resolved?  We currently don't support soft indices on caret rules, and I don't think concept paths can hold soft indices either.

Or is this forward-looking so it is already in place for if/when we do support soft indices on caret paths?

cmoesel (changes requested):
This looks good and I really like the approach!

I had a few minor questions/comments and one small suggestion...

src/export/CodeSystemExporter.ts
mint-thompson: I probably put this line in when initially writing the function as a mirror of what happens in `setCaretRules`, but it isn't needed right now, so there's no reason for it to stay.

cmoesel (approved):
So cool.  People are going to be so glad to see this.  Having to resort to `Instance` is a terrible way to develop a CodeSystem.  You fixed that!  Hurray!

PR #818, opened 2021-05-20
Support indentation in CodeSystem rules

mint-thompson:
Completes task [CIMPL-709](https://standardhealthrecord.atlassian.net/browse/CIMPL-709).

The features for CodeSystems added by #809 were developed at around the same time as the support for indentation added in #797, so this PR adds the power of indentation to concept rules and code caret value rules in code systems. These rules make use of a code path context, which is distinct from a regular path context. Additionally, there is a small but important change in the FSH grammar that allows for the code path equivalent of a path rule, where a code path context is set without creating any new rules. Because the grammar for this is indistinguishable for defining a new concept, this somewhat overlaps with detecting duplicate codes being added to the code system. Thus, a duplicate code will _not_ be detected as an error if the definitions after the first do not include a display or definition. There are test cases to demonstrate this behavior. I think that this is the best way to allow for the setting of code path contexts without requiring additional FSH that would distinguish between defining a code and referring to an existing code.

A somewhat odd but very convenient side effect of supporting code path contexts is that a code caret value rule can now legally be written without any codes at all, if it makes use of a code path context. As such, the grammar for code system rules no longer accepts regular caret value rules. The actual `FshCodeSystem` class still uses `CaretValueRule` the same way it did before. Logic in the visitor for code caret value rule determines whether to construct a `CodeCaretValueRule` or `CaretValueRule`.

All the rules from code hierarchies still apply, of course. The indentation simply allows for some amount of saved typing by using an established code path context.


src/fshtypes/FshCodeSystem.ts
```diff
@@ -27,8 +27,22 @@ export class FshCodeSystem extends FshEntity {
...
+      // the user may simply be referencing the existing concept to establish context.
+      if (
+        newConcept.display == null &&
+        newConcept.definition == null &&
+        isEqual(existingConcept.hierarchy, newConcept.hierarchy)
+      ) {
+        return;
+      } else {
+        throw new CodeSystemDuplicateCodeError(this.id, newConcept.code);
+      }
```
ngfreiter, line 45: Something like this:
```
CodeSystem: MyCS
* #foo
  * #bar
* #bar
```
counts as a duplicate, even though the first `bar` is a part of a hierarchy. Is that correct? It seems like you've considered this case and decided it is correct, but just want to double check.

src/import/FSHImporter.ts
```diff
@@ -1498,19 +1501,26 @@ export class FSHImporter extends FSHVisitor {
...
+      ? ctx.CODE().map(code => {
+          return this.parseCodeLexeme(code.getText(), ctx).code;
+        })
+      : [];
+    const fullCodePath = this.getCodePathWithContext(localCodePath, ctx);
+    // If there is a code path, use it to make a CodeCaretValueRule.
+    // Otherwise, make a CaretValueRule.
+    const caretRule = (
+      fullCodePath.length > 0 ? new CodeCaretValueRule(fullCodePath) : new CaretValueRule('')
+    )
```
ngfreiter, line 1515: I can't comment on the actual line, but in the `CodeSystemExporter`, in `setCodeCareRules`, we don't do `resolveSoftIndexing`. I think I would expect something like: 
```
CodeSystem: MyCS
* #foo
  * #bar
    * #bam
    * ^designation[+].value = "index 0"
    * ^designation[+].value = "index 1"
```
to work, considering something like:
```
CodeSystem: MyCS
* ^identifier[+].value = "foo"
* ^identifier[+].value = "bar"
```
does work.

ngfreiter: I'm also not sure why we need `CaretValueRules` at all. Can't a `CodeCaretValueRule` represent everything that a `CaretValueRule` can, just by having an empty `codePath`? On the export side, it looks like the processing is almost exactly the same.

src/import/FSHImporter.ts
```diff
@@ -2059,6 +2069,67 @@ export class FSHImporter extends FSHVisitor {
...
+    const location = this.extractStartStop(parentCtx);
+    const currentIndent = location.startColumn - this.baseIndent;
+
+    if (currentIndent > 0 && this.codePathContext.length === 0) {
+      logger.error(
+        'The first rule of a definition cannot be indented. The rule will be processed as if it is not indented.',
+        { location, file: this.currentFile }
+      );
+      return codePath;
+    }
```
ngfreiter, line 2088: Maybe this block and the subsequent repeated blocks could be factored out into some sort of `validateContext` or `validateIndent` function? It just seems very similar to the three `if` statements` in the `prependPathContext` function. I think if you do that, it would be fine to drop the specific language of "code path" for simplicity.

ngfreiter (commented):
Thanks for taking this, sorry the previous indenting work may not have made it easy as possible, I had a few comments.

julianxcarter (commented):
This looks good to me. I agree with Nick in saying that soft indexing should probably be supported in `CodeCaretRule`s as well. Aside from that addition, I don't see anything else that I'd like changed. Nice work!

src/fshtypes/FshCodeSystem.ts
mint-thompson: Yes, that is correct. Based on my understanding of [the csd-1 constraint](https://hl7.org/fhir/codesystem.html#invs), the above would not be allowed. When a concept in a CodeSystem is referenced, the reference doesn't contain any hierarchy information, so there would be no way to differentiate between the two uses of `#bar`.

src/import/FSHImporter.ts
mint-thompson: Huh, somehow I was under the impression that soft indexing was not allowed in caret paths, which is not true at all. So yes, since soft indexing is allowed on caret paths in `CaretValueRule`, it should also be allowed on caret paths in `CodeCaretValueRule`.

mint-thompson: And, I think that `CaretValueRule` still needs to be allowed on a CodeSystem so that inserted RuleSets can still provide them, but on the import side, I think it should be fine to always provide a `CodeCaretValueRule`. I'll try it out and see if anything bad happens.

ngfreiter (approved):
This looks good to me!

julianxcarter (dismissed):
This looks great!

cmoesel:
One CI is stuck on checking out the code -- and this has already been approved by Nick and Julian -- so I am merging!

PR #823, opened 2021-06-03
Fix vs filter list bug

ngfreiter:
Found a bug in the grammar for ValueSet filters that we want to fix. We were not allowing chained filters.


cmoesel (approved):
Whoopsies.  Thanks for catching and fixing this!

jafeltra (approved):
Good catch!

PR #830, opened 2021-06-07
Change HasFlags mixin to FlagCarryingRule superclass

cmoesel:
The HasFlags mixin causes strange errors in FSH Online that are difficult (or possibly impossible) to debug. This abandons the mixin in favor of a common superclass, which is a reasonable solution to this problem.  At least for now.  Since we want to fix FSH Online *tonight*.


jafeltra (approved):
Packaged the code on this branch and used it in a static build of FSH Online and the rules with Flags and on LogicalModels work!

PR #831, opened 2021-06-07
Change HasFlags mixin to FlagCarryingRule superclass

cmoesel:
**NOTE: This is the same as #830, but into _master_ (since we don't want the version bump in the fshonline-hotfix branch)**

The HasFlags mixin causes strange errors in FSH Online that are difficult (or possibly impossible) to debug. This abandons the mixin in favor of a common superclass, which is a reasonable solution to this problem.  At least for now.

### DIFF
--- a/src/fshtypes/common.ts
+++ b/src/fshtypes/common.ts
@@ -19,24 +19,3 @@ export function fshifyString(input: string): string {
     .replace(/\r/g, '\\r')
     .replace(/\t/g, '\\t');
 }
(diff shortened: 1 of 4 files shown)


jafeltra (approved):
I approve this one, too! It fixes the problem in FSH Online, even though we lost the nice mixin.

julianxcarter (approved):
Looks good to me! Great job tracking this issue down guys!

PR #837, opened 2021-06-14
Binary Adjunct strings allowed on base64Binary types

### LINKED ISSUE #819: SUSHI should allow special binary adjunct "ig-loader-[filename]" string as base64 content
The [Binary Adjunct](https://confluence.hl7.org/pages/viewpage.action?pageId=66938614#ImplementationGuideParameters-BinaryAdjunctFiles) feature allows authors to use a special string format to tell the IG publisher to load a separate file and encode it as binary.

When using this with most resources, authors just set the `id` field of an `Attachment` value -- but when using it with the `Binary` resource, authors must set the `data` (`Base64Binary`) value to `"ig-loader-[filename]"`.  SUSHI does not like this, however, because it (correctly) detects that `"ig-loader-[filename]"` is not a valid Base64 string.

We need to update SUSHI to support this use case w/ the `Binary` resource.

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Binary.20Adjunct

julianxcarter:
Fixes #819, allowing strings using the binary adjunct reference to be set on elements of type base64Binary.

### DIFF
--- a/src/fhirtypes/ElementDefinition.ts
+++ b/src/fhirtypes/ElementDefinition.ts
@@ -1714,8 +1714,8 @@ export class ElementDefinition {
       (type === 'uri' && /^\S*$/.test(value)) ||
       (type === 'url' && /^\S*$/.test(value)) ||
       (type === 'canonical' && /^\S*$/.test(value)) ||
-      (type === 'base64Binary' && (/^(\s*([0-9a-zA-Z\+\/=]){4}\s*)+$/.test(value)) ||
-      value.startsWith('ig-loader-')) ||
+      (type === 'base64Binary' &&
+        (/^(\s*([0-9a-zA-Z\+\/=]){4}\s*)+$/.test(value) || value.startsWith('ig-loader-'))) ||
       (type === 'instant' &&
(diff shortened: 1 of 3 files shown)


src/fhirtypes/ElementDefinition.ts
```diff
@@ -1715,6 +1715,7 @@ export class ElementDefinition {
       (type === 'url' && /^\S*$/.test(value)) ||
       (type === 'canonical' && /^\S*$/.test(value)) ||
       (type === 'base64Binary' && /^(\s*([0-9a-zA-Z\+\/=]){4}\s*)+$/.test(value)) ||
+      value.startsWith('ig-loader-') ||
```
cmoesel, line 1718: Due to operator precedence in JS, I think you need to add parentheses around the "or" expression.  E.g.,
`false && false || true` is actually true in JS, so we want something more like `false && (false || true)` to accomplish the boolean logic we're looking for.
```suggestion
      (type === 'base64Binary' && (/^(\s*([0-9a-zA-Z\+\/=]){4}\s*)+$/.test(value)) ||
      value.startsWith('ig-loader-')) ||
```

cmoesel (changes requested):
This looks pretty good, but I have suggested one minor tweak to the logic.

src/fhirtypes/ElementDefinition.ts
julianxcarter: Ooh yes, thank you for catching this!

cmoesel (approved):
Yup.  Looks good!  Thanks!

PR #839, opened 2021-06-15
Merge CodeCaretValueRule with CaretValueRule

### LINKED ISSUE #834: Problems w/ indented ^ rules in RuleSets applied to CodeSystems
This is similar to #833, but also uses indentation to establish the `^` rules within a specific code context.  When you attempt to do that, you get errors like:

> error Errors parsing insert rule with parameterized RuleSet ConceptWithGermanSynonym
> - The first rule of a definition cannot be indented. The rule will be processed as if it is not indented.
and

> error The element or path you referenced does not exist: designation.language

Example:
```
Alias:   $SCT = http://snomed.info/sct

CodeSystem: SimpleSystem
Title:     "Simple System"
Description: "A simple system"
* insert ConceptWithGermanSynonym(
    #Fish,
    "Fish",
    "It is a fish",
    "Fisch")
* insert ConceptWithGermanSynonym(
    #Seal,
    "Seal",
    "It is a seal",
    "Seehund")

RuleSet: ConceptWithGermanSynonym(code, display, definition, synonym)
* {code} {display} {definition}
  * ^designation.language = #de
  * ^designation.use = $SCT#900000000000013009 "Synonym"
  * ^designation.value = {synonym}
```

https://fshschool.org/FSHOnline/#/share/3wbdgpB

ngfreiter:
Fixes #834 by eliminating conflicts between `CaretValueRule` and `CodeCaretValueRule`. The referenced bug happened because on a `RuleSet`, there was no way for us to tell if a rule was a `CaretValueRule`, or a `CodeCaretValueRule` when there was no path before the `^`. The grammar is organized such that a `CaretValueRule` was always preferred in this case, so in the example:
```
* {code} {display} {definition}
  * ^designation.language = #de
```
the first line (once params were added) was processed as a `ConceptRule`, and set context in the `codePathContext` array, but then the second line was processed as a `CaretValueRule`, and so would attempt to get context from the `pathContext` array, find no context (since it was looking in the wrong spot) and then cause errors.

I played around with trying to just select which context to use based on where the `RuleSet` was being `inserted`, but this actually doesn't work if the `RuleSet` has no parameters, because in that case we don't know where the `RuleSet` is `inserted` until the export side. So in the end the easiest solution seemed to be just refactoring so that `CodeCaretValueRule` is removed, and all of its functionality is handled by `CaretValueRule`. This refactor ended up being bigger than I thought, but I _think_ it was an improvement, more code went away than got added so that's a good sign. But please let me know if you think this refactor made things worse, not better.

Note that I left `codeCaretValueRule` in the grammar, because I think on the import side it is still easier to handle these separately from `caretValueRule`s (even though both grammar rules end up being stored in the same `CaretValueRule` class).


jafeltra (commented):
This looks good to me! I like the refactor and I think it simplifies things since it combines some places we were doing very similar processing.

The only question I had is about a potential test for the issue that was reported. Is there a way to test that these changes can properly handle inserting CaretValueRules from a RuleSet on a CodeSystem? Or is that too indirect to warrant its own test?

ngfreiter:
> The only question I had is about a potential test for the issue that was reported. Is there a way to test that these changes can properly handle inserting CaretValueRules from a RuleSet on a CodeSystem? Or is that too indirect to warrant its own test?

Ok I've added a test in https://github.com/FHIR/sushi/pull/839/commits/f9a10de477c7b1257974b835a9aee7f407034cb2. If you copy that test onto `master`, you will have to tweak it a bit to make TypeScript happy, but once you do you will see that it fails because no path gets set on the indented rule.

jafeltra:
This is all looking pretty good to me. The one thing I noticed in the final moments is that some of the regressions break on this branch.  I see a number of differences that look similar to this one:

image

I haven't had a chance to look into why that might be happening on this branch, but I wanted to post this before heading out!

ngfreiter:
> This is all looking pretty good to me. The one thing I noticed in the final moments is that some of the regressions break on this branch. I see a number of differences that look similar to this one:

Good catch. It appears this was happening because the new logic did some splitting on `.`, and this would fail when `.` was used to indicate a path (as in `* . ^short = "foo"`). https://github.com/FHIR/sushi/pull/839/commits/df1bd4ddad0edacad915f26deac7e3e573fd1d7c should fix that bug.

jafeltra (approved):
Alrighty. I'm feeling bold. There are no more regressions, the changes here make sense to me, and the reported issue is resolved. I'm approving!

julianxcarter (approved):
Looks good to me!

test/import/FSHImporter.context.test.ts
```diff
@@ -678,14 +685,38 @@ describe('FSHImporter', () => {
...
+      assertConceptRule(codeSystem.rules[0], 'anteater', 'Anteater', undefined, []);
+      expect(codeSystem.rules[0].sourceInfo.file).toBe('Zoo.fsh');
+      assertCaretValueRule(
+        codeSystem.rules[1],
+        'anteater',
+        'property[0].valueString',
+        'Their threat pose is really cute.',
+        false,
+        ['anteater']
+      );
```
mint-thompson, line 717: There is a bit of a problem with the rule here: its `isCodeCaretRule` property is false, since the rule gets parsed by `caretValueRule`. I think that when parsing a RuleSet, it may be necessary to check what the previous rule was in order to correctly determine what kind of context applies.

mint-thompson (changes requested):

test/import/FSHImporter.context.test.ts
ngfreiter: Yeah this brings up a good point. So at this point in the code, the `isCodeCaretRule` flag isn't actually necessary. That property is only used in the `toFSH` method of `CaretValueRule`, so specifically it is relevant to creating preprocessed FSH. So that flag is only set in the `CodeSystemExporter`, so that by the time the `toFSH` method is called, the flag is set. Because you're right, on the import side, it is hard to tell which context applied without checking adjacent rules.

I _think_ the approach should still work fine, but I feel like the name of the flag is pretty confusing, since for much of the lifespan of the `CaretValueRule`, the flag is always `false`. Maybe I could change it to something like `printAsCodeCaretRule`? Any other ideas?

mint-thompson: Ah, okay, I think you're right that it still ends up fine in terms of the FHIR output. Since a `CodeSystem` can't have a `CaretValueRule` with a regular path (since it doesn't have any elements to reference with a regular path), it necessarily wants to use the path parts as codes. Or put another way, every `CaretValueRule` that ends up on a `CodeSystem` is a code caret rule. And since the writing of preprocessed FSH only happens after export, and that export will set the flag, it'll end up fine in that part of the output as well.

This may end up making some exciting things happen once insert rules can be indented, though, so I'll have to look carefully at that PR.

mint-thompson (approved):

PR #841, opened 2021-06-16
Use Version with FSH Code System

### LINKED ISSUE #816: {CodeSystem name}|{version}#{code} "{display}" not working correctly
See Zulip: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Using.20Coding.20with.20.2Eversion/near/239067905

It appears that specifying the version in FSH concept syntax does not work when referencing a code from FSH-defined CodeSystems.  For example, this should work but does not:
```
CodeSystem: Test
* ^version = "1.0"
* #testcode "test-code" "A code for testing"


Instance: TestPatient
InstanceOf: Patient
* maritalStatus = Test|1.0#testcode
```

It results in:
> error Resolved value "Test|1.0" is not a valid URI.

jafeltra:
Fixes #816 

This PR supports parsing out the CodeSystem system from the version when fishing for CodeSystems while resolving references. Previously, the full `system|version` text was used, so fishing for the code system was unsuccessful. This instead uses just the `system` portion if the version is included, and then adds the version back after the URL is resolved.

### DIFF
--- a/src/fhirtypes/common.ts
+++ b/src/fhirtypes/common.ts
@@ -299,7 +299,8 @@ export function replaceReferences<T extends AssignmentRule | CaretValueRule>(
       assignedReference.sdType = instanceMeta.sdType;
     }
   } else if (value instanceof FshCode) {
-    const [system, version] = value.system?.split('|') ?? [];
+    const [system, ...versionParts] = value.system?.split('|') ?? [];
+    const version = versionParts.join('|');
     const codeSystem = tank.fish(system, Type.CodeSystem);
     const codeSystemMeta = fisher.fishForMetadata(codeSystem?.name, Type.CodeSystem);
     if (codeSystem && codeSystemMeta) {
--- a/src/fhirtypes/common.ts
+++ b/src/fhirtypes/common.ts
@@ -299,12 +299,13 @@ export function replaceReferences<T extends AssignmentRule | CaretValueRule>(
       assignedReference.sdType = instanceMeta.sdType;
     }
   } else if (value instanceof FshCode) {
-    const codeSystem = tank.fish(value.system, Type.CodeSystem);
+    const [system, version] = value.system?.split('|') ?? [];
+    const codeSystem = tank.fish(system, Type.CodeSystem);
     const codeSystemMeta = fisher.fishForMetadata(codeSystem?.name, Type.CodeSystem);
     if (codeSystem && codeSystemMeta) {
       clone = cloneDeep(rule);
       const assignedCode = getRuleValue(clone) as FshCode;
-      assignedCode.system = codeSystemMeta.url;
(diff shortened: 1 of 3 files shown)


src/fhirtypes/common.ts
```diff
@@ -299,12 +299,13 @@ export function replaceReferences<T extends AssignmentRule | CaretValueRule>(
       assignedReference.sdType = instanceMeta.sdType;
     }
   } else if (value instanceof FshCode) {
-    const codeSystem = tank.fish(value.system, Type.CodeSystem);
+    const [system, version] = value.system?.split('|') ?? [];
```
mint-thompson, line 302: I think this could lose part of the version if the version contains a `|` character. I think something like this would work:
```js
const [system, ...versionParts] = value.system?.split('|') ?? [];
const version = versionParts.join('|');
```

mint-thompson (changes requested):
This looks good! I have one minor request for change based on an edge case with version strings.

ngfreiter (commented):
Looks good to me, just holding approval until Mint's comment is addressed.

src/fhirtypes/common.ts
jafeltra: This makes sense to do, but I do have one question. If I include that here, it looks like the `toFHIRCoding` method of `FshCode` (in [this line](https://github.com/FHIR/sushi/blob/master/src/fshtypes/FshCode.ts#L21)) will again lose part of the version if there is a `|` character. That would result in part of the version not included in the output.

I've updated the line, but it does pretty explicitly use `.split('|', 2)`, which will cut off anything after the first `|`. Is there any reason why we had that and I shouldn't have taken it out?

cmoesel: If I wrote that code, I probably was thinking that JavaScript's `string.split` worked like Java's `[string.split](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html#split(java.lang.String,int))`.  In Java, if you use a limit (e.g., `2`) then the second item in the returned array contains _the rest of the string_ even if it includes the delimiter character.  It appears JavaScript does not work that way.

jafeltra: It was indeed you who wrote that code, and that is a bummer, because that's exactly what I wanted `string.split` to do, but alas, it is exactly not what Javascript's [`string.split`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split) does:

> Limit: A non-negative integer specifying a limit on the number of substrings to be included in the array. If provided, splits the string at each occurrence of the specified `separator`, but stops when `limit` entries have been placed in the array. Any leftover text is not included in the array at all.

But this at least makes me more comfortable updating that line! So I think this is set to be re-reviewed.

mint-thompson (approved):

ngfreiter (approved):
Unholding approval!

PR #843, opened 2021-06-16
Support code values in =, in, and not-in filters

### LINKED ISSUE #812: Filter operator "in" can have a code value
As described here:
https://www.hl7.org/fhir/snomedct.html#4.3.1.0.8.2
the filter operator can take a concept as a value, not just a string.

https://github.com/FHIR/sushi/blob/d785865cf9d84e89316b0abec9ff2c9671fd5871/src/import/FSHImporter.ts#L1853

jafeltra:
This PR fixes #812 by adding support for code values in the `=`, `in`, and `not-in` filter operators. Based on the [documentation](http://hl7.org/fhir/R4/valueset-definitions.html#ValueSet.compose.include.filter.value) and discussion, we believe these filter operators should _only_ accept code values. We will continue support string values for backwards compatibility, but we will officially deprecate support for strings and log a warning. There are multiple projects, including mCODE, that use string values for these filters which would break if we removed string values immediately. The `is-a`, `is-not-a`, `descendent-of` and `generalizes` remain supporting only code values since we believe that is the correct interpretation of the documentation.

### DIFF
--- a/src/import/FSHImporter.ts
+++ b/src/import/FSHImporter.ts
@@ -1933,6 +1933,10 @@ export class FSHImporter extends FSHVisitor {
       case VsOperator.EQUALS:
       case VsOperator.IN:
       case VsOperator.NOT_IN:
+        // NOTE: We believe that =, in, and not-in operators should ONLY support code values
+        // based on the filter value documentation:
+        // http://hl7.org/fhir/R4/valueset-definitions.html#ValueSet.compose.include.filter.value
+        // Both string and code are supported for now in order to maintain backwards compatibility.
         if (typeof value !== 'string' && !(value instanceof FshCode)) {
           throw new ValueSetFilterValueTypeError(operator, ['string', 'code']);
         }
--- a/src/import/FSHImporter.ts
+++ b/src/import/FSHImporter.ts
@@ -1940,11 +1940,12 @@ export class FSHImporter extends FSHVisitor {
         if (typeof value === 'string') {
           logger.warn(
             `The match value of filter operator "${operator}" must be a code. ` +
-              'Support for string values has been deprecated and will be removed in a future release.'
+              'For string valued properties, use the regex filter operator with a regular expression value. ' +
+              'Support for using strings as filter values has been deprecated and will be removed in a future release.'
           );
         }
         if (typeof value !== 'string' && !(value instanceof FshCode)) {
-          throw new ValueSetFilterValueTypeError(operator, ['string', 'code']);
+          throw new ValueSetFilterValueTypeError(operator, 'code');
         }
         break;
       case VsOperator.IS_A:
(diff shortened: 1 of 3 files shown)


src/import/FSHImporter.ts
```diff
@@ -1933,26 +1933,36 @@ export class FSHImporter extends FSHVisitor {
...
+        // NOTE: We believe that =, in, and not-in operators should ONLY support code values
+        // based on the filter value documentation:
+        // http://hl7.org/fhir/R4/valueset-definitions.html#ValueSet.compose.include.filter.value
+        // Both string and code are supported for now in order to maintain backwards compatibility.
+        if (typeof value === 'string') {
+          logger.warn(
+            `The match value of filter operator "${operator}" must be a code. ` +
+              'Support for string values has been deprecated and will be removed in a future release.'
+          );
+        }
```
ngfreiter, line 1945: So I've gone back and forth on this a lot, but it feels a little bit harsh to me for us to go from _requiring_ `string`, to now saying that `string` is deprecated, and you have to use `code`. Especially since there isn't really a super strong reason why we can't allow `string`, the `code` just gets turned into a `string` anyway. I know there is an argument for readability of having the types be explicit in FSH, but in this case I'm not sure the upside is worth deprecating something that we really could just leave in. But I could go either way honestly, and I've already changed my mind multiple times. What do others thing?

ngfreiter: Also not just deprecating something, but deprecating something that we know people must have been using, because it was the only way to use an `in`, `=` or `not-in` operator.

ngfreiter (commented):
I am flipping and flopping all over, see my comment below.

src/import/FSHImporter.ts
jafeltra: This is a fair point. I think based on my reading of the definition (which admittedly I'm not very confident in), the type we _should_ be supporting is just `code`. In that case, we should somehow start advising people to shift from `string` to `code`. Maybe deprecating is too harsh and we should just say it is not advised to use a string and update to a code?

julianxcarter: I am... also unsure of this. As @ngfreiter said, the `code` will only be turned into a `string` anyways so I lean on the side of not deprecating this. I think a simple warning letting users know that the `string` will be interpreted as a `code`, but that using a `string` value for these operators is inappropriate would suffice.

jafeltra: Based on our conversation this morning, I've kept the warning message for deprecating `string` values and updated the warning message. I also reverted the support for informing users about a choice of values since we only want them using a `code` for the majority of the filter operators. Let me know what you think.

ngfreiter (approved):
This looks good to me, I have always supported the idea of deprecating strings, and I always will.

julianxcarter (approved):

PR #847, opened 2021-06-17
Fix find-repos script

cmoesel:
This fixes the find-repos.ts script so that it does not list invalid repos (that then caused the regression to crash).  Primary changes are:
* removes repos from the existing list if they were not found in the latest crawl
* does not re-add repos that were previously commented out
* retains all blank and commented lines as-is
* guesses default branch by doing an HTTP `HEAD` request on the expected branch zip


mint-thompson (approved):

ngfreiter (approved):
Looks good to me!

PR #849, opened 2021-06-18
Constraints on Resource and Logical Model's own elements now allowed

julianxcarter:
Addresses [CIMPL-752](https://standardhealthrecord.atlassian.net/browse/CIMPL-752), allowing Logical Models and Resources to add constraints to elements introduced by themselves but still disallowing constraints on elements introduced by its parent. Can be tested by running [this FSH](https://fshschool.org/FSHOnline/#/share/3gBZEOP).


src/export/StructureDefinitionExporter.ts
```diff
@@ -414,14 +414,38 @@ export class StructureDefinitionExporter implements Fishable {
...
+              fshDefinition.parent
+            );
+            const parentElement = parentResource.findElement(parentElementPath);
+            if (parentElement) {
+              logger.error(
+                `FHIR prohibits logical models and resources from constraining parent elements. Skipping '${rule.constructorName}' at path '${rule.path}' for '${fshDefinition.name}'.`,
+                rule.sourceInfo
+              );
+              continue;
+            }
```
mint-thompson, line 441: Based on the coverage report, there don't appear to be any test cases that go into this block. Could you add tests for `BackboneElement` elements? There should probably be a test for one that is newly-defined and one that is inherited.

src/export/StructureDefinitionExporter.ts
```diff
@@ -414,14 +414,38 @@ export class StructureDefinitionExporter implements Fishable {
...
         // and continue to the next rule.
-        if (element.path !== element.base.path) {
-          // The AddElementRule always sets the newElementBase.path to the value of element.path.
-          // All parent elements will have the element.base.path pointing to the parent.
-          logger.error(
-            `FHIR prohibits logical models and resources from constraining parent elements. Skipping '${rule.constructorName}' at path '${rule.path}' for '${fshDefinition.name}'.`,
-            rule.sourceInfo
-          );
-          continue;
+        const baseElementPath = element.path.replace(element.structDef.id, '').split('.')[1];
```
mint-thompson, line 417: I'm not sure that just taking the element at index 1 is correct. Is it possible for the element we're looking at to be more deeply nested, such that its parent path would include more parts?

mint-thompson (changes requested):

src/export/StructureDefinitionExporter.ts
julianxcarter: That's certainly possible. My thinking here was that for elements other than `BackboneElement`s, it would actually be preferable to check for the top-level element rather than the entirety of the path. That was based on an assumption that you couldn't use an `addElementRule` to add nested elements to a pre-defined fhir type, e.g `HumanName.foo`. 

When I just checked on that though, I wasn't able to find anything in the spec that rules that out. It looks like the IGPublisher doesn't have a problem with it either, so I'll rework this to be a bit more granular. What I was trying to avoid was a situation where a parent `Resource`/`Logical` adds an element of type `HumanName` for example, then a child `Resource`/`Logical` constrains a sub-element of `HumanName` and that's allowed because the path _looks_ like a new element even though it's technically inherited. I'll address this now!

cmoesel: Sorry... just seeing this now, but adding new subelements to concrete types (like `HumanName`) is **illegal**, even if IG Publisher does not complain about it.  I asked this in Zulip a few weeks back: https://chat.fhir.org/#narrow/stream/179177-conformance/topic/Logical.20Models.3A.20Adding.20new.20sub-elements.20to.20existing.20types.3F

So, it might be ok to add a new subelement to an element w/ type `Element` or `BackboneElement`, but I think that's it.

src/export/StructureDefinitionExporter.ts
```diff
@@ -414,9 +414,9 @@ export class StructureDefinitionExporter implements Fishable {
         // The FHIR spec prohibits constraining any parent element in a 'specialization'
         // (i.e., logical model and resource), therefore log an error if that is attempted
         // and continue to the next rule.
-        if (element.path !== element.base.path) {
-          // The AddElementRule always sets the newElementBase.path to the value of element.path.
-          // All parent elements will have the element.base.path pointing to the parent.
+        const baseElementPath = element.path.replace(`${element.structDef.id}.`, '');
```
cmoesel, line 417: I think that technically the first part of the path is the `element.structDef.type`.  That will be the same as the `element.structDef.id` in most (or all?) cases for logical models and resources -- but I still think it's more correct to use `type` here rather than `id`.

src/export/StructureDefinitionExporter.ts
```diff
@@ -414,9 +414,9 @@ export class StructureDefinitionExporter implements Fishable {
...
+        const addElementRules = fshDefinition.rules.filter(rule => rule instanceof AddElementRule);
```
cmoesel, line 418: This seems like kind of an expensive operation to do for every iteration through the loop.  Can we put it outside the loop?

src/export/StructureDefinitionExporter.ts
julianxcarter: Commit `b765ed5` changes the approach to only use a single if-statement, so the existing tests should be alright now!

src/export/StructureDefinitionExporter.ts
```diff
@@ -414,9 +414,9 @@ export class StructureDefinitionExporter implements Fishable {
...
+        if (rule.path && !addElementRules.some(rule => baseElementPath.includes(rule.path))) {
```
cmoesel, line 419: I think that technically we want to use `startsWith` rather than `includes`, right?  Because we don't want to potentially catch matches in the middle or end of the path (without the beginning of the path).  In fact, to really do this right, I think we would check for an exact match or a startsWith having the path and a trailing `.`.  You might be able to even remove the `baseElementPath` const on line 417 and just do this instead:
```
if (rule.path && !addElementRules.some(rule => {
  return
    element.path === `${element.structDef.type}.${rule.path}`
    || element.path.startsWith(`${element.structDef.type}.${rule.path}.`)
}));
```

src/export/StructureDefinitionExporter.ts
julianxcarter: I figured this was the case! I opted to just determine if a rule is legal by checking if an `addElementRule` is present with a matching path. That seemed like a pretty surefire way to know if the element was inherited or not.

cmoesel (changes requested):
I'm glad to see that the solution seems like it is pretty simple (or at least small).  I do have a couple of suggested tweaks.  And I agree with @mint-thompson that it would be good to have a test or two.

julianxcarter:
Commit `47cda16` adds in your suggestions, and also adds two unit tests to test that constraints on newly added elements are allowed on Logical Models and Resources. @cmoesel

src/export/StructureDefinitionExporter.ts
```diff
@@ -414,9 +415,12 @@ export class StructureDefinitionExporter implements Fishable {
         // The FHIR spec prohibits constraining any parent element in a 'specialization'
         // (i.e., logical model and resource), therefore log an error if that is attempted
         // and continue to the next rule.
-        const baseElementPath = element.path.replace(`${element.structDef.id}.`, '');
-        const addElementRules = fshDefinition.rules.filter(rule => rule instanceof AddElementRule);
-        if (rule.path && !addElementRules.some(rule => baseElementPath.includes(rule.path))) {
+        if (
+          rule.path &&
+          !addElementRules.some(rule =>
+            element.path.startsWith(`${element.structDef.pathType}.${rule.path}`)
```
cmoesel, line 421: This isn't quite right.  Consider:
* elementPath: `'Foo.bars'`
* rule.path: `'bar'`
Then the above line is equivalent to:
```ts
'Foo.bars'.startsWith('Foo.bar')
```
which is true, but that's not the intent.  I think you want:
```ts
element.path === `${element.structDef.pathType}.${rule.path}` ||
element.path.startsWith(`${element.structDef.pathType}.${rule.path}.`)
```
which, in our example would be equivalent to:
```ts
'Foo.bars' === 'Foo.bar' || 'Foo.bars'.startsWith('Foo.bar.')
```
That is false, which is what I think we intend since there is no subpath relationship here.

test/export/StructureDefinition.ResourceExporter.test.ts
```diff
@@ -164,6 +164,33 @@ describe('ResourceExporter', () => {
     );
   });
 
+  it('should allow constraints on newly added elements and sub-elements', () => {
+    const resource = new Resource('ExampleModel');
+    resource.id = 'ExampleModel';
```
cmoesel, line 169: For consistency, this should probably be called `ExampleResource`.

cmoesel (changes requested):
Close, but still doesn't quite cover all the edge cases.  See my individual comment for details.

src/export/StructureDefinitionExporter.ts
julianxcarter: This is totally right, thank you for finding this!

cmoesel (changes requested):
I tested this manually and discovered an edge case that we missed. This is wrongfully capturing rules on the `.` path.  E.g., this should be allowed in all cases, but is not:
```
* . ^alias = "MyAlias"
```
You end up with:
> FHIR prohibits logical models and resources from constraining parent elements. Skipping 'CaretValueRule' at path '.' for 'MyLogical'.

The root element is special.  Even though the parent has a root element too, a child's root element is its own (and should not be considered as "inherited" from the parent). So we need to account for that.

Here is the full test FSH I used for this.  Note that there should be some errors in "MyBadChild", but "MyParent" and "MyGoodChild" should be clean.

```
Logical: MyParent
* . ^alias = "MyFolk" // <-- this should be allowed (but isn't)
* fieldA 0..1 HumanName "Field A"
* fieldB 0..1 BackboneElement "Field B"
* fieldB.fieldB1 1..1 Address "Field B1"

Logical: MyGoodChild
Parent: MyParent
* . ^alias = "MyGoodKid" // <-- this should be allowed (but isn't)
* fieldC 0..1 HumanName "Field C"
* fieldC ^alias = "fieldCharlie" // <-- this is allowed (yay)
* fieldC.family 0..0
* fieldD 0..1 BackboneElement "Field D"
* fieldD.fieldD1 1..1 Address "Field D1"
* fieldD.fieldD1 ^alias = "fieldDelta1" // <-- this is allowed (yay)
* fieldD.fieldD1.line SU

Logical: MyBadChild
Parent: MyParent
* . ^alias = "MyBadKid" // <-- this should be allowed (but isn't)
* fieldC 0..1 HumanName "Field C"
* fieldA.family 0..0 // <-- this should NOT be allowed
* fieldA ^alias = "fieldAlpha" // <-- this should NOT be allowed
* fieldB.fieldB1.line SU // <-- this should NOT be allowed
* fieldB.fieldB1 ^alias = "fieldBeta1" // <-- this should NOT be allowed
```

mint-thompson (changes requested):
This looks like the correct change. Could you add rules on the root element to your test cases?

cmoesel (dismissed):
Looks good to me and works as expected.  I am approving, but I agree w/ @mint-thompson that a test case would be nice!

julianxcarter:
Commit `16c06a2` adds the suggested test cases that @mint-thompson suggested. I've also updated this branch to be in line with master so both of your reviews have been dismissed, however everything should still be essentially the same! @cmoesel

cmoesel (approved):
Looks good!  Thanks, @julianxcarter!

mint-thompson (approved):

Issue #828, opened 2021-06-05
Should be able to use indented insert rules 


markkramerus:
There may be a showstopper down the road somewhere, but it seems like I should be able to replace this:
```
* rest.resource[0]
  * interaction[0]
    * code = #read
    * extension[0].url = "http://hl7.org/fhir/StructureDefinition/capabilitystatement-expectation"
    * extension[0].valueCode = #SHALL
   ```
with this:
```
RuleSet: SHALL
* extension[0].url = "http://hl7.org/fhir/StructureDefinition/capabilitystatement-expectation"
* extension[0].valueCode = #SHALL

//snip

  * interaction[0]
    * code = #read
    * insert SHALL
```
Is there a reason this would not/should not be feasible?

cmoesel:
Right now the reason is because there is no "long form" syntax for it.

When we originally proposed the indent syntax, we actually _did_ also suggest adding support for path-based inserts (e.g., `rest.resource[0].interaction[0] insert SHALL`) so that we _could_ do something like you suggest above.  But there was some confusion and we thought maybe we were introducing too many things at the same time -- so we scaled back and dropped the path-based inserts.

Personally I would like to see support for path-based inserts -- so I'm definitely game for picking them back up again.

PR #853, opened 2021-06-23
Indented InsertRules

ngfreiter:
This fixes https://github.com/FHIR/sushi/issues/828 by adding support for paths in an `InsertRule`. This then automatically gives us support for indentation. The path is then carried through until the `RuleSet` is applied, and at that point prepended to the path of each of the rules in the `RuleSet`. The syntax is as you would expect:
```
RuleSet: R
* family MS
* given MS

Profile: P1
Parent: Patient
* name insert R
```
and the path indentation should work as in other rules, so this definition of `P2` is equivalent to the above definition of `P1`:
```
Profile: P2
Parent: Patient
* name
  * insert R
```
The context does not leak out of a `RuleSet`, so in the following:
```
Profile: P3
Parent: Patient
* name
  * insert R
  * id = "foo"
```
the full path for the third rule is `name.id`, no matter what is contained in the `RuleSet` pointed to by the `insert` rule.


src/fhirtypes/common.ts
```diff
@@ -469,7 +469,9 @@ export function applyInsertRules(
         ruleSetRule.sourceInfo.appliedFile = rule.sourceInfo.file;
         ruleSetRule.sourceInfo.appliedLocation = rule.sourceInfo.location;
         if (isAllowedRule(fshDefinition, ruleSetRule)) {
-          expandedRules.push(cloneDeep(ruleSetRule));
+          const ruleSetRuleClone = cloneDeep(ruleSetRule);
+          ruleSetRuleClone.path = `${rule.path ? rule.path + '.' : ''}${ruleSetRuleClone.path}`;
```
cmoesel, line 473: Not a biggie, but I feel like it's much easier to see what's going on if this is just expressed more simply as:
```ts
if (rule.path) {
  ruleSetRuleClone.path = `${rule.path}.${ruleSetRuleClone.path}`;
}
```

cmoesel (changes requested):
Nice feature!  I'm not sure it's working as expected (or at least as _I_ expect) when used in conjunction w/ soft indices.  For example, I wrote this FSH:
```
RuleSet: Question(linkId, text, type)
* linkId = "{linkId}"
* text = "{text}"
* type = #{type}

Instance: ExampleQuestionnaire
InstanceOf: Questionnaire
* name = "Questionnaire"
* status = #draft
* item[+] insert Question(q1, Are you from Wisconsin?, boolean)
* item[+] insert Question(q2, Do you like cheese?,     boolean)
```

I expected it to translate to something along the lines of:
```
Instance: ExampleQuestionnaire
InstanceOf: Questionnaire
Usage: #example
* name = "Questionnaire"
* status = #draft
* item[0].linkId = "q1"
* item[0].text = "Are you from Wisconsin?"
* item[0].type = #boolean
* item[1].linkId = "q2"
* item[1].text = "Do you like cheese?"
* item[1].type = #boolean
```

But instead, I got:
```
Instance: ExampleQuestionnaire
InstanceOf: Questionnaire
Usage: #example
* name = "Questionnaire"
* status = #draft
* item[0].linkId = "q1"
* item[1].text = "Are you from Wisconsin?"
* item[2].type = #boolean
* item[3].linkId = "q2"
* item[4].text = "Do you like cheese?"
* item[5].type = #boolean
```

So it looks like each rule in the RuleSet is getting the path as-is (w/ the `[+]`), resulting in extra incrementing.  I think that maybe the application of the paths needs to be smarter so the first time it prepends the path, it retains any `[+]` in the path; but all subsequent times it should replace `[+]` with `[=]`.

Do you agree with my expectation or is it actually working the way you would have expected it to?

src/fhirtypes/common.ts
cmoesel: I wrote this before I noticed the issue w/ `[+]` but I'm guessing that the fix for that will be somewhere in this area too?

ngfreiter: Yes I agree, that soft indexing isn't working as I would expect. So I will fix that, and I think it will be around this area, yes. So I will make sure however this is changed, it makes more sense than the conditional within a template string.

ngfreiter: This should be fixed in https://github.com/FHIR/sushi/pull/853/commits/7869bd913a5dd54af2ff7753edb617505eb96dba.

jafeltra (commented):
This looks really good! I only had one really small comment on a semi-unrelated test.

test/import/FSHImporter.CodeSystem.test.ts
```diff
@@ -407,10 +407,8 @@ describe('FSHImporter', () => {
         CodeSystem: ZOO
         * somepath ^publisher = "Marky Mark"
         `;
-        const result = importSingleText(input, 'Simple.fsh');
-        const codeSystem = result.codeSystems.get('ZOO');
-        expect(codeSystem.rules).toHaveLength(0);
-        expect(loggerSpy.getLastMessage('error')).toMatch(/File: Simple\.fsh.*Line: 3\D*/s);
+        importSingleText(input, 'Simple.fsh');
+        expect(loggerSpy.getLastMessage('error')).toMatch(/File: Simple\.fsh.*Line: 3*/s);
```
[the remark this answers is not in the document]
ngfreiter: Righto, done in 96ee92b.

jafeltra (approved):
This looks good to me and worked in cases that I tried!

cmoesel (approved):
Looks good!  Nice addition to FSH!

Issue #842, opened 2021-06-16
Warn more on custom resource definition and use


cmoesel:
When an author defines a custom resource (i.e., not in the `http://hl7.org/fhir/StructureDefinition/` namespace), SUSHI should provide an impossible-to-miss warning about definition of custom resources (that they are not conformant, that they are incompatible w/ standard XML/JSON schemas, public test servers, public registries, standard validation or publication tooling, etc.).  It should also (ideally) warn on cases where a standard FHIR resource (or profile or extension) constrains a reference to refer to a custom resource, or when an instance of a standard FHIR resource (or profile or extension) refers to a custom resources.

Since these warnings could be annoying, users suggest we provide a way to suppress them in `sushi-config.yaml`. We may or may not want to tackle that now, as we should probably consider a general feature for suppressing warnings in FSH.

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/References.20to.20custom.20resources

PR #859, opened 2021-06-29
Warn on Custom Resource definition and use

ngfreiter:
This fixes https://github.com/FHIR/sushi/issues/842 by adding warnings on definition and use of custom `Resources`.

Whenever anyone defines a custom resource using the `Resource` keyword which is not in the `http://hl7.org/fhir/StructureDefinition` domain, they will get a warning telling them a lot of scary things about how doing that is non-conformant to FHIR and a lot of stuff can be expected to break. The warning will only be logged once, and list all defined custom resources.

Whenever anyone uses a custom resource which is not in the `http://hl7.org/fhir/StructureDefinition` domain via a `Reference`, either on a `Profile`, `Extension`, or `Instance`, they will get a warning _if_ their `Profile`, `Extension`, `Instance` is derived from core FHIR. So if you make a `Profile` or your own custom resource, and that resource references another custom resource, then no warning will be logged. For this case, there is one warning for every usage of a resource.

These warnings can be disabled with the `suppressWarnings` configuration property, which takes an array of strings which represent types of warnings that can be suppressed. For example:
```
suppressWarnings:
  - "nonConformantResources"
```
is used to suppress the messages added in this PR.

I'm not really sure about the logging of messages upon use of a custom resource. I guess if we didn't do this, someone could pull a custom resource in as a dependency (probably not via a package but just by pasting the resource into `input`) and then use that resource in their FSH without being warned. But I feel like 99% of the time, the warning upon definition of the custom resource will be sufficient to alert users to the issue. So don't feel bad if in your review you think that should be taken out, because I'm probably on your side.

Here are some cases you will want to verify:
* Warning when defining `Resource` with non-HL7 URL
* No warning when defining `Resource` with `http://hl7.org/fhir/StructureDefinition` URL
* Warning when referencing non-HL7 custom resource on `Profile` derived from FHIR
* No warning when referencing HL7 custom resource on `Profile` derived from FHIR
* No warning when referencing non-HL7 custom resource on `Profile` derived from custom resource
* Warning when referencing non-HL7 custom resource on `Instance` derived from FHIR
* No warning when referencing HL7 custom resource on `Instance` derived from FHIR
* No warning when referencing non-HL7 custom resource on `Instance` derived from custom resource
* All of the above warnings go away if the `suppressWarnings` property described above is added to the config file


src/export/StructureDefinitionExporter.ts
```diff
@@ -71,6 +72,43 @@ export class StructureDefinitionExporter implements Fishable {
     private readonly fisher: MasterFisher
   ) {}
 
+  /**
+   * Checks generated resources for any custom resources which are not in the
+   * http://hl7.org/fhir/StructureDefinition namespace, and are therefore not
+   * comformant.
+   */
+  private warnOnNonComformantResourceDefinitions(): void {
```
mint-thompson, line 80: Minor typo here, and in a few other spots in this function:
```suggestion
  private warnOnNonConformantResourceDefinitions(): void {
```

test/export/StructureDefinition.ResourceExporter.test.ts
```diff
@@ -230,4 +231,34 @@ describe('ResourceExporter', () => {
...
+    const resource = new Resource(
+      'SupercalifragilisticexpialidociousIsSurprisinglyNotEvenLongEnoughOnItsOwn'
+    );
+    doc.resources.set(resource.name, resource);
+    const exported = exporter.export().resources;
+    expect(exported.length).toBe(1);
+    expect(loggerSpy.getLastMessage('warn')).toMatch(
+      /non-conformant Resource.*- SupercalifragilisticexpialidociousIsSurprisinglyNotEvenLon\.\.\./s
+    );
+  });
```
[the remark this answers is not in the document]
ngfreiter: Done in https://github.com/FHIR/sushi/pull/859/commits/465c001f3d6b57051f44e46d0d5c0d3addfc88b2

mint-thompson (changes requested):
I wasn't able to suppress the warning messages with the configuration option you provided. Is this something that it would be possible to make a test for?

cmoesel:
@mint-thompson - the ability to suppress the message was put on hold until we can think more about a message suppression mechanism in general.  So it is expected not to work.

src/export/StructureDefinitionExporter.ts
ngfreiter: Oopsy, looks like my spelling wasn't too conformant am I right.

ngfreiter: Better spelling in https://github.com/FHIR/sushi/pull/859/commits/4f0e3013eae8f4d0e0651130f2501e981ec375ac.

jafeltra (approved):
Great box!

PR #860, opened 2021-06-29
Add Support for FHIR R4B

### LINKED ISSUE #855: Look for R4B in packages2.fhir.org
R4B ballot edition (4.1.0) is currently only in packages2.fhir.org.  We should look there for it like we do for r5.  Also, since it is intended to be temporary, consider looking in packages.fhir.org and then _falling back_ to packages2 if we don't find it.

cmoesel:
Adds support for FHIR 4.1.0 (R4B prerelease) and the eventually coming FHIR 4.3.0 (R4B official). Also changes logic to only use packages2 after first trying packages.

To test, specify this in your `sushi-config.yaml`:
```
fhirVersion: 4.1.0
```
and include the following FSH definition that uses things changed in R4B:
```
Profile: SubstanceDefinitionProfile
Parent: SubstanceDefinition
* status 1..1

Profile: SubstanceProfile
Parent: Substance
* code only Reference(SubstanceDefinitionProfile)
```

On `master` branch, it fails at the point of trying to download R4B.  On this branch, the download succeeds and the profiles compile w/ no errors.

Fixes #855


mint-thompson (approved):
Confirmed that I get the expected results on this branch when specifying FHIR version 4.1.0. Version 4.3.0 still gives a 404, but that's currently expected, since it doesn't exist yet. Hooray for version support!

cmoesel:
> Version 4.3.0 still gives a 404, but that's currently expected, since it doesn't exist yet.

Correct!  Although I have a mock in there to pretend it exists for testing purposes.

jafeltra (approved):
Works as expected!

PR #861, opened 2021-06-29
Warn when unfolding non-choice element with multiple profiles

### LINKED ISSUE #794: Improve instance support for elements with multiple profiles
If an element specifies multiple profiles, SUSHI seems to not allow you to assign any values to subpaths of that element.  Instead, SUSHI should just use the base type and emit a warning.

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/sushi.3A.20problem.20with.20external.20profiles.20containing.20profiles

mint-thompson:
Fixed #794 and completes task [CIMPL-716](https://standardhealthrecord.atlassian.net/browse/CIMPL-716).

When assigning a value on an Instance's element, if that element is nested under a non-choice element with multiple profiles, it cannot be determined which of the profiles should apply when unfolding the profiled element. Therefore, ignore all profiles and use the base type. Warn the user that this is being done.

Previously in this scenario, the unfolding would not occur, which resulted in an error being logged indicating that the path to the nested element did not exist. Such an error message was misleading and unhelpful.

You may be wondering why choice and non-choice elements need to be handled differently. Existing logic in `StructureDefinition.findElementByPath` expects choice elements with multiple profiles to get an empty array back from `ElementDefinition.unfold` so it can proceed with [separate logic for choice elements](https://github.com/FHIR/sushi/blob/master/src/fhirtypes/StructureDefinition.ts#L219-L226). In order to limit the code changes for this PR to the `ElementDefinition.unfold` method, the behavior for choice elements is preserved. A [test case for StructureDefinitionExporter](https://github.com/FHIR/sushi/blob/master/test/export/StructureDefinitionExporter.test.ts#L3450) demonstrates a scenario of multiple profiles on a choice element.


src/fhirtypes/ElementDefinition.ts
```diff
@@ -1962,10 +1962,23 @@ export class ElementDefinition {
    *   be resolved.
    */
   unfold(fisher: Fishable): ElementDefinition[] {
-    if (
-      (this.type?.length === 1 && (this.type[0].profile ?? []).length <= 1) ||
-      this.contentReference
-    ) {
+    // if we have a choice element where id ends with [x], proceed if there is at most one profile, since it expects to get [] back in that case
+    // if we don't have a choice element, proceed with any number of profiles
+    const isChoiceElement = this.id.endsWith('[x]');
```
cmoesel, line 1967: I don't think I fully understand why we wouldn't want to take this same approach for choice elements _if_ they have already been constrained down to one type?  E.g., if this was done:
```
* value[x] only MyQuantityProfile1 or MyQuantityProfile2
```
then even though it is a choice (e.g., ends in `[x]`), it only has one type (`Quantity`), so at this point it acts like a single-type element w/ two profiles. And I think it would be appropriate to treat it as such -- issuing a warning and then using the base type (`Quantity`).

cmoesel (changes requested):
I think maybe we're missing one case where this can still be applied.  See my individual comment.

src/fhirtypes/ElementDefinition.ts
mint-thompson: Yes, I think that makes sense. So, a choice element with multiple types shouldn't be unfolded here, but a choice element with one type should proceed with unfolding (and issue a warning if there are multiple profiles on that type). So, when the choice element has multiple types and ends up in `unfoldChoiceElementTypes`, should the same warning be issued if any of those types have multiple profiles on them? The current behavior is to just always use the common ancestor type among all of the types and profiles (which might end up just being `Element`), and that will still happen, just maybe with a warning on top if one of the types has multiple profiles on it.

cmoesel: Re: unfolding a choice w/ multiple types... it seems that if we already have to unfold at a higher level than a single type, that maybe the multiple profiles really doesn't matter as much anyway.  I mean, if the choice is A or B and B, I don't think our logic changes whether or not either one has multiple profiles, right?  Because the common ancestor is higher than the `type.code`s represented?  So in that case, maybe a warning is appropriate in general, but not necessarily because of the profiles.  But I feel like that's starting to go out of scope of this PR, so I wouldn't really worry about it...

src/fhirtypes/ElementDefinition.ts
```diff
@@ -2003,8 +2016,8 @@ export class ElementDefinition {
         newElements = this.cloneChildren(slicedElement);
       }
       if (newElements.length === 0) {
-        // If it has a profile, use that, otherwise use the code
-        const type = this.type[0].profile?.[0] ?? this.type[0].code;
+        // If we have exactly profile to use, use that, otherwise use the code
```
jafeltra, line 2019:
```suggestion
        // If we have exactly one profile to use, use that, otherwise use the code
```

This might be the smallest comment I could make.

jafeltra (commented):
Besides adding the same approach for choice elements that are only one type, I just had one incredibly small comment.

cmoesel (approved):
OK!  Looks good and I've verified that single element choices already work (and still work in this PR).  Thanks!

jafeltra (approved):
Top notch!

PR #866, opened 2021-07-02
Fix bug with applying path to insert rules with empty paths

ngfreiter:
When a rule in a `RuleSet` has an empty path (such as a `^` rule), and that `RuleSet` is applied from an `insert` rule that has a path context, we were incorrectly joining the empty path with the context, resulting in paths such as `name.`, which are incorrect. This fixes that and adds a test.

See this [thread](https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Caret.20paths.20in.20.22deeply.20inserted.22.20rulesets) for context.

And here is some example FSH that does not work on `master`, but should work on this branch:
```
RuleSet: rs
* ^short = "test"

Profile: TestPatient
Parent: Patient
* name insert rs
```


cmoesel (commented):
Thanks for the fix and the unit test.  What do you think should happen when we have this?
```
RuleSet: Bar
* . ^short = "foo"

Profile: Foo
Parent: Observation
* category insert Bar
```

Currently we get:
> error No element found at path category.. for CaretValueRule in Foo, skipping rule

I think the right behavior is ambiguous because:
* On one hand, people might think of `.` path as "this element", which would imply the short should be applied to `category` in the example above.
* On the other hand, we say the `.` represents the "root" element, and so in a way, the previous context doesn't matter if `.` is always "root".

So, given the ambiguity, maybe it is right to just disallow it.  But... perhaps we should have a friendlier error message?  Something like "The special dot path is only allowed in top-level rules (i.e., rules without any path context)"?  What do you think?

ngfreiter:
> I think the right behavior is ambiguous because:
> 
> * On one hand, people might think of `.` path as "this element", which would imply the short should be applied to `category` in the example above.
> * On the other hand, we say the `.` represents the "root" element, and so in a way, the previous context doesn't matter if `.` is always "root".

I had thought of it as `.` representing the path to the "root" element, so in that way I still think it does matter if you put previous context in front of it. Just like with any other path to a specific element, if I then prepend some other path, I'm going to end up with a broken path. So I'm on team "don't allow it but make a friendlier error message".

cmoesel:
> https://github.com/FHIR/sushi/pull/866#issuecomment-875597833

Cool.  Is it fair to request that as part of this PR?

ngfreiter:
> Cool. Is it fair to request that as part of this PR?

@cmoesel, this should be done in https://github.com/FHIR/sushi/pull/866/commits/c530718c560120d7e695428f287312fe732da364. I also added some processing to indentation so that something like this:
```
Profile: MyP
Parent: Patient
* name
  * . ^short = "foo"
```
would be covered, since that seems like a similar situation.

cmoesel (approved):
Looks good to me, and I like having the additional friendly error message when `.` is used in an indented path.  Thanks, @ngfreiter!

mint-thompson (approved):

PR #867, opened 2021-07-06
Update some packages which had security vulnerabilities

ngfreiter:
This updates some packages which had security issues described by `npm audit`. One vulnerability still exists, because it looks like it is coming from a long chain of dependencies, so I'm not sure what we can really do about that one. It is the `glob-parent` package that causes an issue, which is a dependency of the `cpy-cli` package.


cmoesel (approved):
This truly is amazing work.

BTW -- I also tried to see if there was a safe, simple, and repeatable way to get rid of that last medium-level vulnerability.  But I, like you, was unable to make it so.

mint-thompson (approved):

PR #868, opened 2021-07-07
Include hierarchy in FSH for ConceptRule

mint-thompson:
Completes task [CIMPL-785](https://standardhealthrecord.atlassian.net/browse/CIMPL-785).

The hierarchy on a `ConceptRule` is now used in the `toFSH()` method. Previously, hierarchy information was ignored by this method, which meant that incorrect FSH would be produced for any concept with a hierarchy. If a code in the hierarchy contains spaces, it is quoted so that the syntax is valid.

### DIFF
--- a/src/fshtypes/rules/ConceptRule.ts
+++ b/src/fshtypes/rules/ConceptRule.ts
@@ -17,7 +17,10 @@ export class ConceptRule extends Rule {
   }
 
   toFSH(): string {
-    let line = `* #${/\s/.test(this.code) ? `"${this.code}"` : this.code}`;
+    const quotedCodes = [...this.hierarchy, this.code].map(code =>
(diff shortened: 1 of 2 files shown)


ngfreiter (commented):
Generally this looks good to me, but I'm just wondering how this should work with the `--preprocessed` option. When the input looks like this:
```
CodeSystem: MyCS
* #physical
  * #"solid phase" "platinum" "element with atomic number 78"
```
the preprocessed output I am seeing looks like this: 
```
// Originally defined on lines 1 - 3
CodeSystem: MyCS
Id: MyCS
* #physical
* #"solid phase" "platinum" "element with atomic number 78"
```
Which seems slightly off, because shouldn't `#"solid phase"` have `#physical` before it in the hierarchy?

ngfreiter (approved):

cmoesel (approved):
image

PR #869, opened 2021-07-08
Load predefined resources in nested folders

### LINKED ISSUE #786: SUSHI does not recursively read input files
If the user has a folder structure like:
```
input
- fsh
    - fsh-file1.fsh
- resources
    - structuredefinitions
        - file1.json
    - codesystems
    - valuesets
```
Then SUSHI does not recursively read in `file1.json` as a predefined resource. This should probably be supported.

cmoesel:
The IG templates define a set of supported input folders (e.g., input/profiles, input/resources, etc.). If an author creates further subfolders (e.g., input/resources/nested), SUSHI will now load resources from them as well.  Since the IG publisher doesn't support those paths, however, it will emit a warning and exclude them from the IG JSON.

You can manually test w/ the attached project: [RecursiveInput.zip](https://github.com/FHIR/sushi/files/6785975/RecursiveInput.zip).

Fixes #786

### DIFF
--- a/src/fhirdefs/load.ts
+++ b/src/fhirdefs/load.ts
@@ -7,7 +7,7 @@ import tar from 'tar';
 import axios from 'axios';
 import junk from 'junk';
 import temp from 'temp';
-import { logger } from '../utils';
+import { logger, getFilesRecursive } from '../utils';
 import { Fhir as FHIRConverter } from 'fhir/fhir';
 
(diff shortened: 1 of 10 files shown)


mint-thompson (approved):

ngfreiter (approved):
Looks good to me.

PR #870, opened 2021-07-08
Add support for type checking when fixing Canonical types

ngfreiter:
This addresses [CIMPL-783](https://standardhealthrecord.atlassian.net/browse/CIMPL-783) by adding support for type checking when fixing `Canonical` values. The type checking checks the type of the SD pointed to by the `Canonical` keyword against the `targetProfile` on the element being assigned to. If the type of the SD pointed to is not a child of one of the options listed in the `targetProfile`, then an error is logged. Otherwise, no error is logged.

If the type of the SD pointed to cannot be determined, no type checking is done, and the assignment is just allowed. Generally this is set up to mimic how things work for `Reference`.

### DIFF
--- a/src/utils/Fishable.ts
+++ b/src/utils/Fishable.ts
@@ -18,6 +18,7 @@ export interface Metadata {
   id: string;
   name: string;
   sdType?: string;
+  ancestor?: string;
   url?: string;
   parent?: string;
   abstract?: boolean;
--- a/src/utils/Fishable.ts
+++ b/src/utils/Fishable.ts
@@ -18,7 +18,7 @@ export interface Metadata {
   id: string;
   name: string;
   sdType?: string;
-  ancestor?: string;
+  resourceType?: string;
   url?: string;
   parent?: string;
   abstract?: boolean;
--- a/src/utils/MasterFisher.ts
+++ b/src/utils/MasterFisher.ts
@@ -3,6 +3,7 @@ import { FSHTank } from '../import';
 import { FHIRDefinitions } from '../fhirdefs';
 import { Package } from '../export';
 import { logger } from './FSHLogger';
+import { Instance } from '../fshtypes';
 
 /**
  * The MasterFisher can fish from the tank, the FHIR definitions, and the package that is currently
@@ -69,6 +70,11 @@ export class MasterFisher implements Fishable {
         // If it came from the tank, we need to get the sdType because the tank doesn't know.
         if (fishable instanceof FSHTank) {
           result.sdType = this.findSdType(result, types, fishables);
+        }
+        if (fishable instanceof FSHTank && !result.resourceType) {
+          result.resourceType = this.fishForMetadata(
+            (fishable.fish(item, Type.Instance) as Instance)?.instanceOf
+          )?.sdType;
         }
         return result;
       }
--- a/src/utils/MasterFisher.ts
+++ b/src/utils/MasterFisher.ts
@@ -71,6 +71,9 @@ export class MasterFisher implements Fishable {
         if (fishable instanceof FSHTank) {
           result.sdType = this.findSdType(result, types, fishables);
         }
+        // When an Instance comes from the FSHTank, the FSHTank doesn't know its resourceType,
+        // only its InstanceOf. But here we have access to the other fishers, so we can try
+        // to figure that resourceType out here
         if (fishable instanceof FSHTank && !result.resourceType) {
           result.resourceType = this.fishForMetadata(
             (fishable.fish(item, Type.Instance) as Instance)?.instanceOf
--- a/src/errors/InvalidTypeError.ts
+++ b/src/errors/InvalidTypeError.ts
@@ -31,6 +31,8 @@ function allowedTypesToString(types: ElementDefinitionType[]): string {
   types.forEach(t => {
     if (isReferenceType(t.code)) {
       strings.push(`Reference(${(t.targetProfile ?? []).join(' | ')})`);
+    } else if (t.code === 'canonical') {
+      strings.push(`Canonical(${(t.targetProfile ?? []).join(' | ')})`);
     } else if (t.profile?.length > 0) {
       strings.push(...t.profile);
     } else {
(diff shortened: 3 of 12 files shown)


mint-thompson (changes requested):
This looks good! Could you add a test case for assigning a Canonical of a type that is a child of one of the allowed types?

Also, the following test code passes when put into `InstanceExporter.test.ts`, but if I'm understanding correctly, I don't think it should:
```ts
const specialObservation = new Instance('SpecialObservation');
specialObservation.instanceOf = 'Observation';
specialObservation.usage = 'Inline';
doc.instances.set(specialObservation.name, specialObservation);

const assignedRefRule = new AssignmentRule('instantiatesCanonical');
assignedRefRule.value = new FshCanonical('SpecialObservation');
carePlanInstance.rules.push(assignedRefRule);
      
const exported = exportInstance(carePlanInstance);
expect(exported.instantiatesCanonical).toBeDefined();
```

cmoesel:
@mint-thompson and @ngfreiter -- I think Mint's example above likely works because there is no such thing as a canonical URL for an `Observation` (which is also a slight flaw in Nick's test case). Canonical URLs identify definitional resources in FHIR -- so they really only apply to a subset of resource types; namely those types that have a top-level `url` property (which _is_ the canonical URL).  So, to really test this, you should be using one of those resources.

So, I think the test case should be updated to be more realistic in that way -- and I agree w/ Mint that it would be good to have a positive test case as well.

ngfreiter:
I think there are more than flaws in the test case here, but also some flaws in the implementation that I've got to fix. Because I do think @mint-thompson's case should be an error.

ngfreiter:
Ok I think https://github.com/FHIR/sushi/pull/870/commits/decc7618e57721feaf7ac4c81ac922c818890b08 should be better. To better explain what the issue was, my original implementation was checking the `sdType` of the thing being referenced, which doesn't make a lot of sense. For example in this case:
```
Instance: CIn
InstanceOf: CarePlan
* instantiatesCanonical = Canonical(P)

Profile: P
Parent: PlanDefinition
```
the first implementation would check that the `sdType` of `P` is `PlanDefinition`, and allow this. But this is wrong, because `P` itself is a `StructureDefinition`, which is not one of the allowed types of `instantiatesCanonical`. And then if the thing pointed to by the `Canonical` was an `Instance`, my implementation would find no `sdType` at all, and just allow everything through, which is again wrong.

The new approach is to check the `resourceType` of the thing pointed to by the `Canonical`. So if the thing pointed to is a `Profile`, the type to check is `StructureDefinition` itself, not the type of the `StructureDefinition`. The one tricky thing here, is that on an `Instance` that is not yet exported, we don't know the `resourceType`, only the `InstanceOf`, but since we check if any ancestors conform to the expected types, I think using the `InstanceOf` value leads to checking against the `resourceType`, since that will be an ancestor of the `InstanceOf`. Sorry if I'm explaining that poorly, let me know if it isn't making sense.

ngfreiter:
Oh and I added more tests per @mint-thompson's requests (and tried to make those tests more relevant, per @cmoesel's request).

mint-thompson (approved):
Based on the new test cases for `InstanceExporter`, I think this is doing what we need. Hooray for canonical URLs!

src/utils/Fishable.ts
```diff
@@ -18,6 +18,7 @@ export interface Metadata {
   id: string;
   name: string;
   sdType?: string;
+  ancestor?: string;
```
cmoesel, line 21: I don't love the term `ancestor` for this.  At first I thought `resourceType` would be more accurate, but now I see that is not quite correct in the case of `Instance`s.  But I still feel like `ancestor` is kind of a deceptive name.  For example, someone might (reasonably) think that the `ancestor` of `MyConditionProfile` is `Condition`, not `StructureDefinition`.  Should this just be `instanceOf`?  A profile is an instance of a `StructureDefinition` after all...

cmoesel (changes requested):
It seems to be correct and works well, but I'd like to try to find a better property name than `ancestor`.  I feel like it is misleading or at least potentially confusing.

src/utils/Fishable.ts
ngfreiter: Ok so https://github.com/FHIR/sushi/pull/870/commits/c39f752629eb57bdc5f31f3fa9e0f72aa7bbdd31 renames `ancestor` to `resourceType`. The one complication of this is that on an `Instance` defined in the `FSHTank`, we don't yet know the `resourceType`. So in this case, the `FSHTank` returns nothing. The `MasterFisher` however knows a bit more, so it can fish for the `InstanceOf` on the `Instance`, and use the `sdType` of the result, and that should be the `resourceType` of the `Instance`. This does clarify the keyword, but it comes at the cost of some additional complexity and additional `fish`ing operations. Let me know if you think that ends up tipping the balance to making it not worth it. In that case, I think we can just call the keyword `instanceOf`, even if that is a bit tough to swallow.

src/utils/MasterFisher.ts
```diff
@@ -70,6 +71,11 @@ export class MasterFisher implements Fishable {
         if (fishable instanceof FSHTank) {
           result.sdType = this.findSdType(result, types, fishables);
         }
+        if (fishable instanceof FSHTank && !result.resourceType) {
+          result.resourceType = this.fishForMetadata(
+            (fishable.fish(item, Type.Instance) as Instance)?.instanceOf
+          )?.sdType;
+        }
```
cmoesel, line 78: I remember why you're doing it this way because we discussed it yesterday.  The future me (in 6 months) is unlikely to remember that.  Could you add a brief comment regarding what is happening here (and why it is happening)?

cmoesel (dismissed):
Nice work, @ngfreiter.  I am approving this PR as-is, but I did leave one comment as a suggestion.  I don't want to "request changes" since I'll be out of the office for the next few days.

src/utils/MasterFisher.ts
ngfreiter: Yes this is mysterious, 4ba168f adds a demystifying comment.

mint-thompson (approved):

PR #876, opened 2021-07-12
Remove ig-data folder from test fixtures

mint-thompson:
Completes task [CIMPL-767](https://standardhealthrecord.atlassian.net/browse/CIMPL-767).

Since the ig-data folder is not longer part of the recommended project structure, it is removed from the project structure for test fixtures. So, most of the changed files in this PR are just the test fixtures being moved to a new path that doesn't include an ig-data folder. The other changed files are the test cases, which no longer need to add the ig-data folder when building paths. There are no application code changes in this PR. I'm pretty sure I caught all the cases where tests use the ig-data folder, but if I missed any, those should be changed as part of this PR.

The only test that is intentionally unchanged is the one for `Processing` that specifically tests for legacy flat tank support, along with making sure we are showing an error message when the ig-data folder is present.


cmoesel:
@mint-thompson -- it looks like there is a conflict that needs to be resolved.

ngfreiter (dismissed):
Looks good to me, outside of the conflict.

jafeltra (dismissed):
This looks great. I didn't notice any ig-datas that you missed! This is so satisfying and makes me very happy.

jafeltra (dismissed):
Looks good!

ngfreiter (dismissed):
Looks good to me!

jafeltra (approved):
Still approve!

ngfreiter (approved):

PR #878, opened 2021-07-14
Don't change min to 1 when a slice is assigned a value

### LINKED ISSUE #810: SUSHI incorrectly handles cardinality inference in sliced arrays
It seems like SUHSI incorrectly infers the cardinality of a sliced array when values are assigned within that sliced array. See this example below:
```
Alias: $loinc = http://loinc.org
Alias: $snomed = http://snomed.info/sct

Profile: RespRate
Parent: Observation
* code.coding ^slicing.discriminator[0].type = #pattern
* code.coding ^slicing.discriminator[0].path = "$this"
* code.coding ^slicing.rules = #open
* code.coding contains
    loinc 1..* and
    snomed 0..*
* code.coding[loinc] = $loinc#9279-1
* code.coding[loinc].system 1..
* code.coding[loinc].code 1..
* code.coding[snomed] = $snomed#86290005
* code.coding[snomed].system 1..
* code.coding[snomed].code 1..
```
SUSHI incorrectly makes the `code.coding[snomed]` slice have cardinality `1..*` because of this line:
```
* code.coding[snomed] = $snomed#86290005
```
If you break that line into multiple pieces:
```
* code.coding[snomed].system = $snomed
* code.coding[snomed].code = #86290005
```
The `code.coding[snomed]` slice stays optional. I'm not quite sure what is going on here, but we need to investigate this and figure out what the correct behavior is.

cmoesel:
We previously would bump min 0 to min 1 if you assigned a value directly to a slice path (and that path was also a discriminator for the slice).  We don't do that anymore because we should never make an optional slice required.

The funny thing is that we had a test that actually checked for this faulty behavior and failed when I fixed it.

Fixes #810


test/export/StructureDefinitionExporter.test.ts
```diff
@@ -6199,6 +6244,33 @@ describe('StructureDefinitionExporter R4', () => {
...
+      const regularBinding = {
+        valueSet: 'http://hl7.org/fhir/us/minimal/RegularObservationCodes',
+        strength: 'required'
+      };
+      expect(rootCategory.binding).toEqual(regularBinding);
+      expect(loggerSpy.getLastMessage('error')).toMatch(
+        /Cannot override required binding with extensible/s
+      );
+      expect(loggerSpy.getLastMessage('error')).toMatch(/File: Weaker\.fsh.*Line: 4\D*/s);
+    });
```
jafeltra, line 6272: This test also mentions that the cardinality of the slice is not affected, but I don't see where that is tested. Does it make sense to add an additional check for that? Or is that not necessary? I was thinking something along the lines of: 

```
const sliceElement = sd.findElement('Observation.category:Procedure');
expect(sliceElement.min).toBe(0);
expect(sliceElement.max).toBe('1');
```

jafeltra (commented):
This makes sense to me generally, but I had two small questions. Since @cmoesel is out of the office now, I think if the other reviewer agrees with my comments, I can add them. Otherwise, this looks good as it is!

test/export/StructureDefinitionExporter.test.ts
mint-thompson: This test looks like it is a duplicate of the test up on line 6107. So, I think the wrong thing got put in here somehow.

jafeltra: Ohhh good catch. I skimmed the other positive case tests, so I must have missed that one. I'm not exactly sure what the intended test should test here. Do you think it is safe to remove this test? Or did you have another idea of a case to test here instead @mint-thompson?

mint-thompson: Based on the test description, it sounds like this scenario is already covered by the test added on line 3966. I can't think of any other test cases that need to be added for this scenario. So, it should be safe to just remove this test.

jafeltra: After discussion, we will confirm the purpose of this test with @cmoesel when he returns.

cmoesel: Oops!  I intended to write this test, so I copy/pasted from another test.  But then I realized that what I wanted to test was already covered elsewhere, so I didn't need it.  But I guess I forgot to delete the copy/pasted test.  I'll remove it.

mint-thompson (approved):

jafeltra (approved):

Issue #872, opened 2021-07-09
SUSHI doesn't allow aliases to URLs w/ fragments


cmoesel:
```
Alias: ABC = http://some.sdo.org/CodeSystems.html#systemABC
```
results in:
> error mismatched input 'http://some.sdo.org/CodeSystems.html#systemABC' expecting SEQUENCE

See: https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/Using.20DICOM's.20FHIR.20value.20sets/near/245465153

PR #880, opened 2021-07-15
Allow aliases with fragments

ngfreiter:
Fixes https://github.com/FHIR/sushi/issues/872.

The issue here is that aliases with fragments get parsed by the grammar as a `CODE`. It turns out, I'm really not sure how to make the grammar such that it can distinguish between `CODE` and `SEQUENCE` without breaking a bunch of things. So the easier solution was just to allow a `CODE` as the value of an `Alias: ` definition. And then to just read the value in from the `CODE` keyword. Is this too hacky? Let me know if so.


src/import/FSHImporter.ts
```diff
@@ -178,10 +178,11 @@ export class FSHImporter extends FSHVisitor {
       // Collect the aliases and store in global map
       ctx.entity().forEach(e => {
         if (e.alias()) {
-          const [name, value] = e
-            .alias()
-            .SEQUENCE()
-            .map(s => s.getText());
+          const name = e.alias().SEQUENCE()[0]?.getText();
```
mint-thompson, line 181: As a minor point, I don't think the `?` is necessary, since `SEQUENCE()[0]` is the left side of the equals sign, which is always present.

mint-thompson (commented):
I don't think this is too hacky. we already deal with some similar cases where we want to accept a `SEQUENCE`, but a valid input might also match another token. The other option is to enter a mode for reading Aliases, similar to the strategy we have for Rulesets. That would keep the code very simple, but makes the lexer a little more complicated. So it's basically a choice of where you want the extra logic to live.

Overall, I slightly prefer the choice of implementing with lexical modes. That lets us avoid having to think about "wait, why do I accept a `CODE` for the value of an Alias?" when looking at our import code.

src/import/FSHImporter.ts
ngfreiter: No longer an issue in 744155d, since this is handled with a lexical mode. Let me know what you think of the new implementation. One thing that would've been nice is if I could just redefine `SEQUENCE` in the relevant `mode ALIAS` block, but apparently that is not possible. Otherwise I think the approach you suggested @mint-thompson does do a better job avoiding any potentially confusing hacks.

mint-thompson (dismissed):
good work!! hooray for lexer modes!!

antlr/src/main/antlr/FSH.g4
```diff
@@ -5,7 +5,7 @@ options { tokenVocab = FSHLexer; }
 doc:                entity* EOF;
 entity:             alias | profile | extension | invariant | instance | valueSet | codeSystem | ruleSet | paramRuleSet | mapping | logical | resource;
 
-alias:              KW_ALIAS SEQUENCE EQUAL SEQUENCE;
+alias:              KW_ALIAS ALIAS_SEQUENCE ALIAS_EQUAL ALIAS_SEQUENCE NEWLINE;
```
jafeltra, line 8: This requirement for a newline between Aliases is new, right? Is there a reason for introducing this new requirement here? Is it just so we can understand properly when to leave the new lexical mode?

antlr/src/main/antlr/FSHLexer.g4
```diff
@@ -120,3 +120,9 @@ LINE_COMMENT:       '//' ~[\r\n]* [\r\n] -> skip;
 mode RULESET_OR_INSERT;
 PARAM_RULESET_REFERENCE:      WS* NONWS+ (WS* ('(' ('\\)' | '\\\\' | ~[)])+ ')')) -> popMode;
 RULESET_REFERENCE:            WS* NONWS+ -> popMode;
+
+mode ALIAS;
+ALIAS_EQUAL:        '=';
+ALIAS_SEQUENCE:     NONWS+;
+NEWLINE:            [\r\n] -> popMode;
+ALIAS_WHITESPACE:   WS -> channel(HIDDEN);
```
jafeltra, line 128: I don't see where the `ALIAS_WHITESPACE` keyword is used. Is there a reason to have it defined so we can use it in the future?

jafeltra (commented):
This looks good, but I did have a couple small questions below and inline:

- When I run `npm run build:grammar`, three of the generated files have diffs. Is that happening for you as well? Or do I have something cached somewhere?

mint-thompson:
@jafeltra I'm also getting changes when generating the grammar source, so it's not just you.

antlr/src/main/antlr/FSH.g4
ngfreiter: Yeah I couldn't figure out another way to know when to exit the lexical mode.

antlr/src/main/antlr/FSHLexer.g4
ngfreiter: It serves the same purpose as the `WHITESPACE` token in the other mode. So it consumes whitespace and sends it to the hidden channel to be ignored.

ngfreiter: No longer relevant as of https://github.com/FHIR/sushi/pull/880/commits/c2b7e1413b6cfc3b0096eb0a61dfdfc6f670e421.

antlr/src/main/antlr/FSH.g4
ngfreiter: No longer relevant as of https://github.com/FHIR/sushi/pull/880/commits/c2b7e1413b6cfc3b0096eb0a61dfdfc6f670e421.

src/import/FSHImporter.ts
ngfreiter: Became an issue again in https://github.com/FHIR/sushi/pull/880/commits/c2b7e1413b6cfc3b0096eb0a61dfdfc6f670e421, but fixed.

ngfreiter:
Ok gang @jafeltra and @mint-thompson https://github.com/FHIR/sushi/pull/880/commits/c2b7e1413b6cfc3b0096eb0a61dfdfc6f670e421 goes back to the original approach, but adds a clarifying comment to hopefully indicate why this is needed in the importer. The grammar does still change and will need to be updated in the spec, but I guess we don't want to do that until this is released?

cmoesel:
I think we should update the grammar in the spec as soon as this is approved and merged -- because we need the draft spec done very soon.

Regarding the changes, I think maybe it would be good to add a comment in the grammar on this line:
```
alias: KW_ALIAS SEQUENCE EQUAL (SEQUENCE | CODE);
```
so that smart users don't read the grammar and mistakenly think that aliases can (or should) be used for codes.  A comment indicating that the CODE token is accepted because the lexer parses URLs w/ fragments as codes would help clarify that.

ngfreiter:
> I think we should update the grammar in the spec as soon as this is approved and merged -- because we need the draft spec done very soon.
> 
> Regarding the changes, I think maybe it would be good to add a comment in the grammar on this line:
> 
> ```
> alias: KW_ALIAS SEQUENCE EQUAL (SEQUENCE | CODE);
> ```
> 
> so that smart users don't read the grammar and mistakenly think that aliases can (or should) be used for codes. A comment indicating that the CODE token is accepted because the lexer parses URLs w/ fragments as codes would help clarify that.

@cmoesel, done in https://github.com/FHIR/sushi/pull/880/commits/c3e976b75abb0fbd997d75be2cca7b92be37dbb7.

jafeltra (approved):
Looks good to me!

mint-thompson (approved):
approved!! sorry about lexer modes.

PR #881, opened 2021-07-16
Adjust error logging strategy for RuleSets with ConceptRules

### LINKED ISSUE #877: Rule Sets with Value Set Rules Generates errors
Testing `insert` rules with value sets. SUSHI generates errors that indicate the RuleSet is being interpreted as a CodeSystem:
```
RuleSet: Hands
* $SCT#87683000  "Left handed (finding)"
```
Results in:
`error Do not include the system when listing concepts for a code system.`

https://fshschool.org/FSHOnline/#/share/3i4Mc5w

mint-thompson:
Fixes #877 and completes task [CIMPL-796](https://standardhealthrecord.atlassian.net/browse/CIMPL-796).

A RuleSet can contain ConceptRules and ValueSetConceptComponentRules. Additionally, it is possible for the same FSH to parse using the concept or vsConceptComponent parser rules. Therefore, it can be ambiguous when parsing the RuleSet which rule should actually be made. When one of these rules is applied, it is handled in different ways depending on whether the target entity is a ValueSet or a CodeSystem.

If a ConceptRule with a system is part of a RuleSet applied to a ValueSet, turn it into a ValueSetConceptComponent and apply it. If a ConceptRule with a system is part of a RuleSet applied to a CodeSystem, apply it, but log an error, because concepts listed on a CodeSystem should not include their system. These special cases only apply when the ConceptRule is part of a RuleSet.

When using the full list of regression FSH repos, there are no differences in the resulting IGs or log counts when making these changes.


ngfreiter (approved):
Looks good to me!

jafeltra (approved):
This looks great! Sorry for the very delayed review!

PR #883, opened 2021-07-19
Support code systems with fragments

### LINKED ISSUE #873: SUSHI doesn't correctly parse codes whose system has a fragment
```
* code = http://some.sdo.org/CodeSystems.html#systemABC#1234
```
produces JSON with:
```json
{
  "code": "systemABC#1234",
  "system": "http://some.sdo.org/CodeSystems.html"
}
```
but it should be:
```json
{
  "code": "1234",
  "system": "http://some.sdo.org/CodeSystems.html#systemABC"
}
```

Note that we do need to support `#` in codes too, however.  In that case, the code should be wrapped in `""`.  So we need to support this as well:
```
* code = http://some.sdo.org/CodeSystems.html#"systemABC#1234"
```
(which *should* result in the first JSON object above).

See: https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/Using.20DICOM's.20FHIR.20value.20sets/near/245465153

jafeltra:
This PR fixes #873 by updating how we split a system and code apart. Instead of splitting the system and the code on the first `#`, this PR now makes it so we do that split on the last unescaped and unquoted `#`. Finding the last match was harder than I expected using regular expressions, although I'm very willing to believe I missed some clever and simple regular expression trick. In order to get around this, I found all matches using `exec`, which will give the capture groups and index of the matched unescaped `#`, and then worked my way backwards through that list to find the unquoted `#`.

I added two tests that check the new functionality this adds. I also updated an existing test that I think is incompatible with also supporting system fragments.

If anyone has ideas to explore that would make this operation simpler, please let me know.

### DIFF
--- a/src/import/parseCodeLexeme.ts
+++ b/src/import/parseCodeLexeme.ts
@@ -9,10 +9,15 @@ import { FshCode } from '../fshtypes';
  * @returns {FshCode} the parse FshCode
  */
 export function parseCodeLexeme(conceptText: string): FshCode {
-  const allSplitPoints = [...conceptText.matchAll(/(^|[^\\])(\\\\)*#/g)];
-  const allSplitPointsArray = allSplitPoints.map(p => p);
+  const codeRegex = new RegExp(/(^|[^\\])(\\\\)*#/g);
+  const allSplitPointsArray = [];
+  let currentMatch = codeRegex.exec(conceptText);
+  while (currentMatch !== null) {
+    allSplitPointsArray.push(currentMatch);
+    currentMatch = codeRegex.exec(conceptText);
+  }
   let splitIndex = allSplitPointsArray.length - 1;
-  let splitPoint = allSplitPointsArray[splitIndex];
+  let splitPoint: any = allSplitPointsArray[splitIndex];
   let system: string, code: string;
   if (splitPoint == null) {
     system = '';
--- a/src/import/parseCodeLexeme.ts
+++ b/src/import/parseCodeLexeme.ts
@@ -9,7 +9,10 @@ import { FshCode } from '../fshtypes';
  * @returns {FshCode} the parse FshCode
  */
 export function parseCodeLexeme(conceptText: string): FshCode {
-  const splitPoint = conceptText.match(/(^|[^\\])(\\\\)*#/);
+  const allSplitPoints = [...conceptText.matchAll(/(^|[^\\])(\\\\)*#/g)];
+  const allSplitPointsArray = allSplitPoints.map(p => p);
+  let splitIndex = allSplitPointsArray.length - 1;
+  let splitPoint = allSplitPointsArray[splitIndex];
   let system: string, code: string;
   if (splitPoint == null) {
     system = '';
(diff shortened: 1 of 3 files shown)


mint-thompson (changes requested):
This looks pretty good so far. There are two cases I found that don't seem to play nicely with the new implementation. The first looks like this:
```
* code = http://example.org#something"
```
This expected result is a system of `http://example.org` and a code of `something"`. This works on FSH online, but does not work on this branch.

The second case is when the system ends with a `#` character:
```
* code = http://example.org##"some#thing"
```
The expected result is a system of `http://example.org#` and a code of `some#thing`, but this generates a parse error on this branch.

Additionally, there's a case I found that doesn't seem to work in the current SUSHI build, but might be in scope for this PR:
```
* code = http://example.org#"something
```
The expected result is a system of `http://example.org` and a code of `"something`, but on both the current build and this branch, results in a code of `somethin`, presumably because it's trimming the first and last characters.

As far as an alternate implementation, I did some tinkering with regular expressions, and there is an alternative that I think is worth considering. Consider these regular expressions:
```ts
const quotedCodeRegex = /(.*)#("([^"\\]|\\"|\\\\)+")/;
const unquotedCodeRegex = /(.*)#([^#\s]+)/;
const combinedCodeRegex = /(.*)#("([^"\\]|\\"|\\\\)+")|(.*)#([^#\s]+)/; // this is just the two above combined by a |
```

If I've worked these out correctly, then `quotedCodeRegex` will _not_ match when the code portion is unquoted. So, if that doesn't match, then `unquotedCodeRegex` should provide the desired result. Alternatively, you can just throw `combinedCodeRegex` at your problem: if the code is quoted, your system and code are in the first and second groups, respectively. If the code is unquoted, your system and code are in the fourth and fifth groups, respectively. However, I completely respect that throwing a more complicated regex at a problem does make the code harder for everyone to understand, so if this feels like it's getting too clever, I'm fine with sticking with an implementation that's closer to your current approach.

jafeltra:
@mint-thompson and @cmoesel I've gone back and forth on this a few times now, so I am fairly turned around, but I'd like your opinion on this. I have been getting confused with how codings should work with escaped `#` characters. Through all the edits this message has gone through, I think I've decided that we want systems to allow `\#` to escape a `#` character, because we did explicitly add that in #208. And I don't think we want codes to allow `\#` to escape a `#` character, because that is when you should use a `"quoted code"`.

However, now that I've come to that conclusions, I think that means the solution to #873  should be to just `\#` escape the fragment # symbol. In that case, I don't even know if there is a test to add in this PR, and it should work as expected on master. Does that seem reasonable? Is there a case I missing that we talked about @cmoesel  or that you have thought of @mint-thompson  that I've lost while going back and forth on this? If this is the case, I could try to fix that case that Mint found is broken on master if we think it is in scope.

cmoesel:
Well, dang.  We did add that in #208, didn't we?  Now I feel bad for filing that bug. _(sigh)_.

I think escaping `#` by doing `\#` is the most straight-forward approach and makes it pretty obvious what part of the URL should be the system (whereas it is less obvious without the escape).  So I think I'm OK with this.

jafeltra:
> Well, dang. We did add that in #208, didn't we? Now I feel bad for filing that bug. _(sigh)_.
> 
> I think escaping `#` by doing `\#` is the most straight-forward approach and makes it pretty obvious what part of the URL should be the system (whereas it is less obvious without the escape). So I think I'm OK with this.

This was a while ago, and we could always decide now that we think something else is a better approach, so don't feel bad about the issue! But I did re-stumble onto that PR during my confusion and it felt very relevant.

I am curious to hear what @mint-thompson thinks since they were also the one to make that PR. But if Mint agrees, I might put my vote in for closing this PR and the issue. But the bonus is that I know why the `* code = http://example.org#"something` case breaks, so I can submit a new PR for that fix.

mint-thompson:
Alas! Somehow, I had forgotten about #208 as well. So yes, I think it's fair to say that we have already have a way of specifying a literal `#` in a system.

jafeltra:
Since we seem to be in agreement, I think we should just use the `\#` pattern to escape `#` used in fragments in system URLs. I believe that [this test](https://github.com/FHIR/sushi/blob/master/test/import/FSHImporter.test.ts#L61-L84) covers this case, so I don't think there are any code changes we want to include here.

Therefore, I'm going to close this PR.

cmoesel:
OK. I will reply back on that SUSHI issue and Zulip comment.

cmoesel:
Ha.  Julia beat me to it!

jafeltra:
I only did 1/3 of the above statement, since I haven't officially closed the issue or responded on Zulip!

PR #888, opened 2021-07-21
Set . path as context

### LINKED ISSUE #887: SUSHI incorrectly handles indented rules below a dot path
```
Extension: TestExtension
* ^context.type = #element
* ^context.expression = "Patient"
* . 0..1
  * ^short = "A test extension."
  * ^definition = "Yep.  It is a test extension."
* value[x] only boolean
```

In SUSHI 2.0.0 beta.1, this works correctly.  In SUSHI 2.0.0 beta.2, this results in the following errors:
```
error Rule cannot be indented below rule which has no path. The rule will be processed as if it is not indented.
  File: /Users/cmoesel/dev/fsh/TestBadThings/input/fsh/patient.fsh
  Line: 5
error Rule cannot be indented below rule which has no path. The rule will be processed as if it is not indented.
  File: /Users/cmoesel/dev/fsh/TestBadThings/input/fsh/patient.fsh
  Line: 6
```
and then later on in the log:
```
error The element or path you referenced does not exist: short
  File: /Users/cmoesel/dev/fsh/TestBadThings/input/fsh/patient.fsh
  Line: 5
error The element or path you referenced does not exist: definition
  File: /Users/cmoesel/dev/fsh/TestBadThings/input/fsh/patient.fsh
  Line: 6
```

mint-thompson:
Fixes #887.

When a rule's path is ".", the rule applies to the root element. Allow these rules to set a context so that they may be followed by an indented caret rule with no path.

Basically, the check for "did we indent a rule with a . path" was happening a bit too early. Now, when we have a . path on a rule with no indent, we'll first check that we're not indented, and therefore everything is fine.

### DIFF
--- a/src/import/FSHImporter.ts
+++ b/src/import/FSHImporter.ts
@@ -2064,7 +2064,6 @@ export class FSHImporter extends FSHVisitor {
       return path;
     }
 
-    // Otherwise, get the context based on the indent level.
     if (path.length === 1 && path[0] === '.') {
       logger.error(
         "The special '.' path is only allowed in top-level rules. The rule will be processed as if it is not indented.",
@@ -2076,6 +2075,7 @@ export class FSHImporter extends FSHVisitor {
       return path;
     }
 
+    // Otherwise, get the context based on the indent level.
     const currentContext = this.pathContext[contextIndex - 1];
     if (currentContext.length === 0) {
       logger.error(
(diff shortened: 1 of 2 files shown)


src/import/FSHImporter.ts
```diff
@@ -2078,6 +2065,17 @@ export class FSHImporter extends FSHVisitor {
     }
 
     // Otherwise, get the context based on the indent level.
+    if (path.length === 1 && path[0] === '.') {
```
ngfreiter, line 2068: I think this comment needs to be moved down below this new block.

ngfreiter (approved):
👍 🆒

julianxcarter (approved):

PR #889, opened 2021-07-22
Context setting rules now required to begin at column 1

### LINKED ISSUE #874: SUSHI 2.0.0 Beta 2 allows illegal indentation
The space counting is relative to the column where the first keyword is placed. So if `Profile:` starts in column 3, then everything that follows is evaluated as if column 3 were the left margin,

Example:
```
     RuleSet: NameRules
     * family MS
     * given MS

   Profile: MyPatientProfile
Parent: Patient
   * name 
     * insert NameRules
   * deceased[x] only boolean
  // More profile rules
```
This should really generate errors, but does not.

julianxcarter:
Addresses [CIMPL-789](https://standardhealthrecord.atlassian.net/browse/CIMPL-789) and fixes #874, requiring the first rule in a context path sequence to be left-aligned, rather than in line with the entity.

Something like this should work when run against the master branch, but fail when run against this branch:
```
  CodeSystem: ZOO
  * #anteater "Anteater"
    * #northern "Northern tamandua"
      * ^property[0].valueString = "They are strong climbers."
```

### DIFF
--- a/src/import/FSHImporter.ts
+++ b/src/import/FSHImporter.ts
@@ -129,6 +129,7 @@ enum Flag {
 
 const FLAGS = ['MS', 'SU', '?!', 'TU', 'N', 'D'];
 const INDENT_WIDTH = 2;
+const DEFAULT_START_COLUMN = 1;
 
 /**
  * FSHImporter handles the parsing of FSH documents, constructing the data into FSH types.
@@ -144,7 +145,6 @@ export class FSHImporter extends FSHVisitor {
   paramRuleSets: Map<string, ParamRuleSet>;
   private topLevelParse: boolean;
   private pathContext: string[][];
-  private baseIndent: number;
 
   constructor() {
     super();
@@ -246,9 +246,8 @@ export class FSHImporter extends FSHVisitor {
   }
 
   visitEntity(ctx: pc.EntityContext): void {
-    // Reset the pathContext and baseIndent level for each entity
+    // Reset the pathContext for each entity
     this.pathContext = [];
-    this.baseIndent = this.extractStartStop(ctx).startColumn;
 
     if (ctx.profile()) {
       this.visitProfile(ctx.profile());
@@ -1752,8 +1751,6 @@ export class FSHImporter extends FSHVisitor {
     const tempDocument = new FSHDocument(this.currentFile);
     // save the currentDoc so it can be restored after parsing this RuleSet
     const parentDocument = this.currentDoc;
-    // save the baseIndent so it can be restored after parsing this RuleSet
-    const parentIndent = this.baseIndent;
     const parentContext = this.pathContext;
     this.currentDoc = tempDocument;
     // errors should be collected, not printed, when parsing generated documents
@@ -1769,8 +1766,6 @@ export class FSHImporter extends FSHVisitor {
     } finally {
       // be sure to restore parentDocument
       this.currentDoc = parentDocument;
-      // and to restore the parentIndent
-      this.baseIndent = parentIndent;
       this.pathContext = parentContext;
     }
     // if tempDocument has appliedRuleSets, merge them in
@@ -2048,7 +2043,7 @@ export class FSHImporter extends FSHVisitor {
    */
   private prependPathContext(path: string[], parentCtx: ParserRuleContext): string[] {
     const location = this.extractStartStop(parentCtx);
-    const currentIndent = location.startColumn - this.baseIndent;
+    const currentIndent = location.startColumn - DEFAULT_START_COLUMN;
     const contextIndex = currentIndent / INDENT_WIDTH;
 
     if (this.pathContext.length && path.length === 1 && path[0] === '.') {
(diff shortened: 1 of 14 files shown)


src/import/FSHImporter.ts
```diff
@@ -144,7 +144,7 @@ export class FSHImporter extends FSHVisitor {
   paramRuleSets: Map<string, ParamRuleSet>;
   private topLevelParse: boolean;
   private pathContext: string[][];
-  private baseIndent: number;
+  private baseIndent = 1;
```
ngfreiter, line 147: Do we still need `baseIndent` and all of its processing? As I remember this was only used to determine relative indent, so why is it needed if there are no more relative indents?

ngfreiter (changes requested):

src/import/FSHImporter.ts
julianxcarter: I suppose we don't, I'll remove this more thoroughly.

julianxcarter: Commit `94220bf` removes the `baseIndent` class variable and some of the references to it, but I did think there was value in having the default start column saved as a constant variable so I added that instead.

test/import/FSHErrorListener.test.ts
```diff
@@ -5,11 +5,12 @@ describe('FSHErrorListener', () => {
...
 
   it('should not log any errors for valid FSH', () => {
-    const input = `
+    /*
     Profile: MyPatient
     Parent: Patient
     * active = true
-    `;
+    */
+    const input = 'Profile: MyPatient\nParent: Patient\n* active = true';
```
mint-thompson, line 13: I think it might be useful to build the test inputs by joining strings together with `os.EOL`. That way, we can be sure that things work as expected with different line endings. It will also make it a little easier to read what the input is without having to include the comment before the input. I realize this is asking for a lot of modifications, but maybe a search/replace will get most of them?

Also, most of the test cases for FSHImporter that aren't modified here are raising the left-alignment error on line 2015, but they still pass since only one of the test files uses context paths for rules. But, it might still be good to fix these as well? I can imagine problems/annoyances arising due to getting an extra error in the logger for each rule. But, I also recognize that this is a lot of extra spots to fix.

mint-thompson (changes requested):
Implementation looks good! I have one small (large) request for the test cases.

test/import/FSHErrorListener.test.ts
cmoesel: I think I agree w/ @mint-thompson .  These are kind of difficult to read as single-liners, and while the comments help, it's also easy for those to come out of sync w/ the real code.  I feel like it would be easier to read as:
```
const input = [
  'Profile: MyPatient',
  'Parent: Patient',
  '* active = true'
].join(os.EOL);
```

Another approach would be to define a function that takes a string and normalizes out the leading spaces.  E.g.:
```
const input = leftAlign(`
    Profile: MyPatient
    Parent: Patient
    * active = true
    `);
```
The `leftAlign` function would find the number of spaces to the first non-space character (in the first non-blank line) and then remove that number of spaces from all lines (similar to our implementation of `"""`).

Like Mint, I feel bad because this is likely a lot of (tedious) work.  If these are the only changes you made in these test files, though, it might be easier just to revert the files to the original versions and edit those.

julianxcarter: No this is totally reasonable! I'll update the test cases to make use of `os.EOL` and also make the unit test changes for the rest of the tests in the FSHImporter test suite as well.

julianxcarter: I agree with the both of you, I kind of struggled with finding the most intuitive way to make these changes. I'll update the FSHImporter test suite so that they're a bit more readable and consistent, and also use `os.EOL` instead of the explicit newline character.

test/import/FSHErrorListener.test.ts
```diff
@@ -1,33 +1,36 @@
...
-    `;
+    `);
     importSingleText(input, 'MyPatient.fsh');
     expect(loggerSpy.getAllMessages('error')).toBeEmpty();
   });
 
   it('should report mismatched input errors from antlr', () => {
-    const input = `
+    const input = leftAlign(
+      leftAlign(`
```
mint-thompson, line 20: I don't think you meant to call `leftAlign` twice here.

test/import/FSHErrorListener.test.ts
```diff
@@ -37,173 +40,175 @@ describe('FSHErrorListener', () => {
...
     importSingleText(input2, 'Invalid.fsh');
     expect(loggerSpy.getLastMessage('error')).toMatch(
       /Assignment rules must include at least one space both before and after the '=' sign.*File: Invalid\.fsh.*Line: 4\D*/s
     );
   });
 
   it('should make sense of errors due to no space around = in an assignment rule', () => {
-    const input = `
+    const input = leftAlign(
+      leftAlign(`
```
mint-thompson, line 122: Or here, either.

mint-thompson (changes requested):
very good. just two little spots to clean up, then it should be ready to go.

test/import/FSHErrorListener.test.ts
julianxcarter: I did not! Thank you for catching this.

test/import/FSHErrorListener.test.ts
julianxcarter: Thank you for catching this as well!

mint-thompson (dismissed):

ngfreiter (approved):
👍

ngfreiter (approved):
Woo!

PR #890, opened 2021-07-22
Allow Profiles on Logical Models

### LINKED ISSUE #857: Support Profiles of Logical Models
We were under the impression that you cannot have a profile on a logical model (e.g., cannot have `"kind": "logical"` and `"derivation: constraint"`) -- but [it seems that maybe you can](https://chat.fhir.org/#narrow/stream/179177-conformance/topic/Type.20on.20First.20Element.20in.20Differential.20for.20Logical.20Models/near/243972570).

mint-thompson:
Fixes #857 and completes task [CIMPL-788](https://standardhealthrecord.atlassian.net/browse/CIMPL-788).

A profile may specify a logical model as its parent. The StructureDefinition resulting from this will have kind "logical" and derivation "constraint".

This one turned out to be pretty simple, I guess? Hopefully I didn't miss something important.


jafeltra (dismissed):
This looks good to me! I created a profile of a Logical Model and ran it through the IG publisher and it didn't have any errors.

The one related task that I think we will need is an update to how FSH Online determines if a SD is a profile, since the `kind` can now be `'logical'` as well, but that isn't related to this PR!

cmoesel (changes requested):
I'm glad this didn't turn out to be too tricky!  I did find a couple of other things we probably want to address.

1. We sort things according to their type in FHIRDefinitions.  More specifically, in `FHIRDefinitions.ts`, we have this around lines 111-121:
```ts
      } else if (definition.kind === 'resource') {
        if (definition.derivation === 'constraint') {
          if (!this.isSupplementalFHIRDefinitions) {
            addDefinitionToMap(definition, this.profiles);
          }
        } else {
          addDefinitionToMap(definition, this.resources);
        }
      } else if (definition.kind === 'logical') {
        addDefinitionToMap(definition, this.logicals);
      }
```
I think we want to check derivation on logical too in order to add them to the right map (`this.logicals` or `this.profiles`).

2. We don't support creating _instances_ of logical models.  It basically doesn't work because we don't FSH for logicals when resolving the `InstanceOf` - but this is intentional.  So... I think we also should not support creating _instances_ of logical model profiles -- but right now, we're not blocking that.  You can use this example for testing:
```
Logical: MyLogical
* foo 0..1 boolean "A foo"

Profile: MyLogicalProfile
Parent: MyLogical
* foo 1..1 MS

Instance: MyLogicalInstance
InstanceOf: MyLogical
* foo = true


Instance: MyLogicalProfileInstance
InstanceOf: MyLogicalProfile
* foo = true
```

cmoesel (approved):
Hurray!  It's great!

jafeltra (approved):
Looks good!

PR #892, opened 2021-07-26
Support quoted code without closing quote

jafeltra:
While testing out PR #883, @mint-thompson noticed the following case wasn't working properly:

```
* code = http://example.org#"something
```
produced the following JSON:

```
{
  "code": "somethin",
  "system": "http://example.org"
}
```
This PR fixes this case so the same FSH now correctly produces the following JSON:

```
{
  "code": "\"something",
  "system": "http://example.org"
}
```


ngfreiter (approved):
This looks good to me!

mint-thompson (approved):

PR #893, opened 2021-07-27
Avoid duplicate entries in Instance's meta.profile due to versioned profiles

### LINKED ISSUE #891: Sliced meta.profile with version number forcing instances to have an extra meta.profile repeat without version number.
If there is a profile slice requirement for meta.profile where the canonical requires a version number, and FSH generated instance of the resource will have 2 meta.profile instances one with and one without the version number. If the latest version of the profile changed with breaking changes, this can cause a validation issue.

Occurs with the CARIN BlueButton IG.

FSHOnline example: https://fshschool.org/FSHOnline/#/share/3eM13Rq 
CARIN BB Jira ticket: https://jira.hl7.org/browse/FHIR-32428
CARIN BB Example: https://build.fhir.org/ig/HL7/carin-bb/Organization-Payer1.json.html

mint-thompson:
Fixes #891 and completes task [CIMPL-800](https://standardhealthrecord.atlassian.net/browse/CIMPL-800).

When creating an Instance of a profiled resource, the profile's URL will be put into the Instance's `meta.profile` array. However, if that URL is already present in the array with a version, remove the unversioned URL from `meta.profile`. It's a little weird to explain, but the example in the issue demonstrates it clearly.

### DIFF
--- a/src/export/InstanceExporter.ts
+++ b/src/export/InstanceExporter.ts
@@ -294,12 +294,6 @@ export class InstanceExporter implements Fishable {
       instanceDef._instanceMeta.sdType = instanceOfStructureDefinition.type;
     }
 
-    // Add the SD we are making an instance of to meta.profile, as long as SD is not a base FHIR resource
-    // If we end up adding more metadata, we should wrap this in a setMetadata function
-    if (isResource && instanceOfStructureDefinition.derivation === 'constraint') {
-      instanceDef.meta = { profile: [instanceOfStructureDefinition.url] };
-    }
-
     // Set Assigned values based on the FSH rules and the Structure Definition
     instanceDef = this.setAssignedValues(fshDefinition, instanceDef, instanceOfStructureDefinition);
     instanceDef.validateId(fshDefinition.sourceInfo);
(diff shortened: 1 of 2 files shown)


src/export/InstanceExporter.ts
```diff
@@ -314,6 +314,17 @@ export class InstanceExporter implements Fishable {
     // Once all rules are set, we should ensure that we did not add a duplicate profile URL anywhere
     if (instanceDef.meta?.profile) instanceDef.meta.profile = uniq(instanceDef.meta.profile);
 
+    // If we added a versioned profile that's the same as the InstanceOf profile, remove the InstanceOf profile
+    const versionedInstanceOf =
+      instanceDef.meta?.profile?.filter(profile =>
+        profile.startsWith(`${instanceOfStructureDefinition.url}|`)
+      ) ?? [];
```
cmoesel, line 321: Since this is really just checking to see if at least one item in the array meets the criteria, I think it would be better to use `some`:
```ts
const hasVersionedInstanceOf =
      instanceDef.meta?.profile?.some(profile =>
        profile.startsWith(`${instanceOfStructureDefinition.url}|`)
      );
```

src/export/InstanceExporter.ts
```diff
@@ -314,6 +314,17 @@ export class InstanceExporter implements Fishable {
...
+    if (versionedInstanceOf.length > 0) {
+      instanceDef.meta.profile = instanceDef.meta.profile.filter(
+        profile => profile !== instanceOfStructureDefinition.url
+      );
```
cmoesel, line 325: Maybe this is an edge case, but... what if the author wanted `meta.profile` to list both and explicitly stated so:
```
* meta.profile[+] = "http://myig.org/fhir/StructureDefinition/FooProfile"
* meta.profile[+] = "http://myig.org/fhir/StructureDefinition/FooProfile|4"
```
This would filter out the author-provided URL without a version.  I think, if it is possible, it would be better to just _stop_ the auto-population of `meta.profile` if we detect a versioned (or unversioned) URL for the `InstanceOf` already in there.  (But I'm not sure where that code lives or how feasible it is).

cmoesel (changes requested):
Fixes the problem as as stated (thanks!), but I've left one minor coding suggestion and a comment about an edge case I'd like to at least consider (if feasible).

src/export/InstanceExporter.ts
mint-thompson: This seems reasonable to me. Most authors won't be explicitly adding the unversioned profile, so I think it's safe to guess that any authors who do so definitely want to keep it around.

src/export/InstanceExporter.ts
```diff
@@ -322,6 +316,28 @@ export class InstanceExporter implements Fishable {
     cleanResource(instanceDef);
     this.pkg.instances.push(instanceDef);
 
+    // should we add the instanceOf to meta.profile?
```
cmoesel, line 319: I wonder if we should move this up a few lines so it is _before_ the call to `this.validateRequiredElements`.  My concern is that if a profile has `meta.profile 1..*` and the author did not explicitly set a `meta.profile`, then won't `validateRequiredElements` complain about the missing `meta.profile`?  If so, that would be confusing to authors since they would see `meta.profile` populated in the output.  It seems to me that this code would be at home on line 310, after the call to `this.setAssignedValues`.

cmoesel (commented):
The new code looks good!  I just have one question / comment about the placement of it.

src/export/InstanceExporter.ts
mint-thompson: I agree, the automatically provided profile should help contribute to a minimum cardinality. I've added that in the latest change set.

cmoesel (approved):
Yes!  Works great!  Thanks, @mint-thompson!

ngfreiter (approved):
Looks good to me 👍

PR #895, opened 2021-07-28
Bump version to v2.0.0-beta.3

jafeltra:
v2.0.0-beta.3, here we come!

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,6 +1,6 @@
 {
   "name": "fsh-sushi",
-  "version": "2.0.0-beta.2",
+  "version": "2.0.0-beta.3",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


cmoesel:
I'm running a full regression on this vs beta.2.  I'll report back w/ results.

cmoesel (approved):
The regression results are in, and there were some differences, but all of them are explained by recent changes in the code.  The main differences were:

1. Some IGs stopped generating some examples and emitted errors like this:
    > error Ignoring FSH definition for Practitioner/durant since it duplicates existing pre-defined resource. To use the FSH definition, remove the conflicting file from "input". If you do want the FSH definition to be ignored, please comment the definition out to remove this error.
    
    After looking into it, I realized it is because the JSON file in the `input` folder was nested in `input/examples/Practitioner`.  Earlier versions of SUSHI did not read this in, but beta.3 now recursively reads `input`.

2. Some `ConceptMap` instances dropped the `"sourceCanonical"` property, but it turns out that is because we now catch this error:
    > error The type "Canonical(CodeSystem)" does not match any of the allowed types: Canonical(http://hl7.org/fhir/StructureDefinition/ValueSet)
    
    So this is also expected.

So... I think we're good to go!

PR #896, opened 2021-07-30
FshCode code and display are fshified for string output

mint-thompson:
Completes task [CIMPL-771](https://standardhealthrecord.atlassian.net/browse/CIMPL-771).

A `FshCode`'s code can contain basically any non-whitespace characters, as long as it doesn't start or end with whitespace characters. If it contains any spaces or any characters that need to be escaped in FSH, fshify and quote the code. A `FshCode`'s display is basically free text, so it should also be fshified for output. The same applies to a `FshQuantity`'s display property. Fshifying strings in these locations also makes `FshRatio` and `ValueSetConceptComponentRule` produce correct output, since those classes make use of the `toString` methods provided by `FshCode` and `FshQuantity`.


src/fshtypes/FshCode.ts
```diff
@@ -1,14 +1,20 @@
...
 export class FshCode extends FshEntity {
   constructor(public code: string, public system?: string, public display?: string) {
     super();
   }
 
   toString(): string {
-    const str = `${this.system ?? ''}#${/\s/.test(this.code) ? `"${this.code}"` : this.code}`;
-    return this.display ? `${str} "${this.display}"` : str;
+    let codeToShow = this.code;
+    const fshifiedCode = fshifyString(this.code);
```
ngfreiter, line 12: I'm having trouble understanding why the code needs to be `fshified` and placed in the quotes unless it has a space. I've found that I can do stuff like this:
```
Profile: O
Parent: Observation
* code = #strange\"code
```
and now the preprocessed output comes out as:
```
Profile: O
Parent: Observation
Id: O
* code = #"strange\\\"code"
```
which also looks correct, but doesn't seem necessary. Are there cases I'm missing where the code does not have a space, but does still need to be placed in quotes?

mint-thompson: I could have sworn I ran into a case with just `\` and `"` that needed to be quoted, but now I can't seem to reproduce it. So, I think you are correct that the code only needs to be `fshified` and quoted if it contains spaces.

ngfreiter (approved):
Looks good to me, thanks!

julianxcarter (approved):
This looks good!

Issue #886, opened 2021-07-20
SUSHI should support RuleSets/Insert rules in Mappings


cmoesel:
Mark suggests that you should be able to make RuleSets of mapping statements and the balloted spec will indicate support for it.  So... we need to make it happen!

PR #897, opened 2021-08-02
Supporting MappingRules on RuleSets

ngfreiter:
This fixes https://github.com/FHIR/sushi/issues/886 by adding support for MappingRules on RuleSets. It seems like this was done on the export side, and InsertRules were supported on the import side on a Mapping, the one missing piece was that the grammar didn't allow for a MappingRule to be placed on a RuleSet. This fixes that, and adds a test.


julianxcarter (approved):
Looks good!

mint-thompson (approved):
Hooray for more rules in rule sets! Thank you for adding this.

PR #898, opened 2021-08-03
Elements can be constrained with canonicals that contain versions

### LINKED ISSUE #863: Constraining Canonicals does not support canonicals w/ versions
You can now do:
```
* action.definition[x] only Canonical(Questionnaire)
```
and
```
* action.definition[x] only Canonical(http://hl7.org/fhir/StructureDefinition/Questionnaire)
```
BUT you cannot do:
```
// broken
* action.definition[x] only Canonical(Questionnaire|4.0.1)
```
or
```
// broken
* action.definition[x] only Canonical(http://hl7.org/fhir/StructureDefinition/Questionnaire|4.0.1)
```

We need to support this too.

mint-thompson:
Fixes #863 and completes task [CIMPL-791](https://standardhealthrecord.atlassian.net/browse/CIMPL-791).

The version provided with the canonical type on the type constraint is appended to the `targetProfile` on the element. When values are assigned to the element, the version is _not_ checked. For example, if an element is constrained to `Canonical(Patient|4.0.1)`, then any `Patient` instance can have its canonical assigned to the element, regardless of the version of the `Patient` profile it is an instance of.

Basically: authors are allowed to specify a version in the type constraint, and that version will appear in the FHIR resource, but the version does _not_ affect SUSHI's type checks.


ngfreiter (approved):
Looks good to me!

julianxcarter (approved):

Issue #884, opened 2021-07-20
SUSHI init causes errors and warnings in IG Publisher


ngfreiter:
See this [thread](https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/New.20project.20errors) for details.

Running `sushi --init` creates a project which then causes some errors and warnings in the Publisher. We should clean up `--init` so that those warnings are avoided if possible.

PR #899, opened 2021-08-04
Update SUSHI --init to avoid publisher errors

ngfreiter:
Fixes https://github.com/FHIR/sushi/issues/884.

`sushi --init` was causing errors and warnings in the IG Publisher. These turned out to be related to:
* Not having any examples of the `MyPatient` profile
* Not having any description on the `MyPatient` profile
* Not having a `publisher` set

So to fix this, this PR updates the `patient.fsh` file to have an example and a description. It also updates the default `sushi-config.yaml` file to contain default values for `publisher.name` and `publisher.url`, and updates the `init()` function so that the user can specify custom values for these fields.

# Testing
To test this, try doing `sushi --init` on `master`, and build with the publisher with `_genonce`. The QA page, `output/qa.html`, will have several errors and warnings. Then do the same on this branch, and the QA page should look clean, like this: 
image

The QA page should definitely be clean if you use default values, and should likely be clean if you use custom values, but note that certain custom values could lead to warnings. But either way make sure to test both approaches. Ensure that the publisher value you use is inserted in the footer of pages as shown here:
image


mint-thompson (dismissed):
Tried it out on my windows machine, and everything appears to be in good order. Thank you!

src/utils/Processing.ts
```diff
@@ -429,6 +429,19 @@ export async function init(): Promise<void> {
       configDoc.set(field, userValue);
     }
   });
+
+  // And for nested publisher fields
+  ['name', 'url'].forEach(field => {
+    const userValue = readlineSync.question(
+      `Publisher ${field.charAt(0).toUpperCase() + field.slice(1)} (Default: ${configDoc
```
jafeltra, line 436: I'm not sure how much this matters, but I think you can alternatively use lodash's `upperFirst` function to do the same as this string concatenation.

jafeltra (commented):
This looks good! I confirmed the errors are resolved when I use the defaults and are the expected errors when I use my own values (my url did not resolve). I just had one small comment that you can take or leave.

src/utils/Processing.ts
ngfreiter: Good point, done in https://github.com/FHIR/sushi/pull/899/commits/b06aaf24c6f8cf172712947988bba9e8246ed902.

jafeltra (approved):
Looks good to me!

mint-thompson (approved):

Issue #865, opened 2021-07-01
SUSHI should log an error for duplicate Instance names


cmoesel:
See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/sushi.20exporting.20less.20resources.20than.20it.20converted

ngfreiter:
This requires some investigation. I'm not sure what exactly is happening here, because in some cases (but clearly not all) we do log an error when `Instance`s have duplicate names.

PR #900, opened 2021-08-06
Detect duplicates across files

ngfreiter:
Fixes https://github.com/FHIR/sushi/issues/865.

It seems that our current duplicate detection for all entities only worked within the same file. This updates that so that now duplicate detection is done globally across all files, for all entities.

The bug is specific to instances, but this should fix it for all types of entities.

To test, just make a FSH project with two files, and two entities of the same name, one in each file. On this branch, you should get an error alerting you that there are two entities of the same name, and so one is being ignored. On master, no such error will be logged, even though the second defined entity is ignored.

### DIFF
--- a/src/import/FSHImporter.ts
+++ b/src/import/FSHImporter.ts
@@ -139,6 +139,7 @@ const DEFAULT_START_COLUMN = 1;
  * we must call the explicit visitX functions.
  */
 export class FSHImporter extends FSHVisitor {
+  private docs: FSHDocument[] = [];
   private currentFile: string;
   private currentDoc: FSHDocument;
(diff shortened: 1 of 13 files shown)


mint-thompson (changes requested):
This looks good! Could you add a test case showing that no error is produced when two entities of different types have the same name?

ngfreiter:
> This looks good! Could you add a test case showing that no error is produced when two entities of different types have the same name?

Good call! Done in https://github.com/FHIR/sushi/pull/900/commits/8dd8f2654bd9c75c88a07468a074422ce69c4d0e

julianxcarter (approved):
This looks good!

PR #903, opened 2021-08-11
Do not add duplicate concepts from insert rules

mint-thompson:
Completes task [CIMPL-726](https://standardhealthrecord.atlassian.net/browse/CIMPL-726).

When inserting a rule set into a CodeSystem, check to make sure that none of the rules in the rule set would add concepts that are already part of the CodeSystem. This check is done using a function that is also used when checking concepts that are added from concept rules directly on the CodeSystem.

### DIFF
--- a/src/fhirtypes/common.ts
+++ b/src/fhirtypes/common.ts
@@ -490,7 +490,17 @@ export function applyInsertRules(
             }
             ruleSetRuleClone.path = newPath;
           }
-          expandedRules.push(ruleSetRuleClone);
+          if (ruleSetRuleClone instanceof ConceptRule && fshDefinition instanceof FshCodeSystem) {
+            try {
+              if (fshDefinition.checkConcept(ruleSetRuleClone)) {
+                expandedRules.push(ruleSetRuleClone);
+              }
+            } catch (e) {
+              logger.error(e.message, ruleSetRuleClone.sourceInfo);
+            }
+          } else {
+            expandedRules.push(ruleSetRuleClone);
+          }
           if (firstRule) {
             // Once one rule has been applied, all future rules should inherit the index used on that rule
             // rather than continuing to increment the index with the [+] operator
--- a/src/fshtypes/FshCodeSystem.ts
+++ b/src/fshtypes/FshCodeSystem.ts
@@ -27,6 +27,12 @@ export class FshCodeSystem extends FshEntity {
   }
 
   addConcept(newConcept: ConceptRule) {
+    if (this.checkConcept(newConcept)) {
+      this.rules.push(newConcept);
+    }
+  }
+
+  checkConcept(newConcept: ConceptRule): boolean {
     const existingConcept = this.rules.find(
       rule => rule instanceof ConceptRule && rule.code == newConcept.code
     ) as ConceptRule;
@@ -39,7 +45,7 @@ export class FshCodeSystem extends FshEntity {
         newConcept.definition == null &&
         isEqual(existingConcept.hierarchy, newConcept.hierarchy)
       ) {
-        return;
+        return false;
       } else {
         throw new CodeSystemDuplicateCodeError(this.id, newConcept.code);
       }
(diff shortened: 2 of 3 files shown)


src/fhirtypes/common.ts
```diff
@@ -490,7 +490,17 @@ export function applyInsertRules(
             }
             ruleSetRuleClone.path = newPath;
           }
-          expandedRules.push(ruleSetRuleClone);
+          if (ruleSetRuleClone instanceof ConceptRule && fshDefinition instanceof FshCodeSystem) {
```
ngfreiter, line 493: I notice that on line 462 above, we do some similar sort of checks I think. Would it be possible to just do this check up there? I think it would give some minor efficiency improvement, because we then don't have to clone a rule that we are not going to use. That's probably not super relevant, but I also do just think it creates better organization.

ngfreiter (changes requested):
Looks good! Just one minor comment for your consideration.

src/fhirtypes/common.ts
mint-thompson: The check here is a little different, because this one determines whether the rule gets added to `expandedRules`. It is a little wasteful to not add the rule after cloning it, but I don't think it's ever particularly expensive to clone a `ConceptRule`. But, if we wanted to avoid that, it would be possible to have the `checkConcept` come as soon as possible inside the `allowedRule` then-block. Something like this should work. Do you think this is clearer?
```ts
if (isAllowedRule(fshDefinition, ruleSetRule)) {
  let includeRule = true;
  if (ruleSetRule instanceof ConceptRule && fshDefinition instanceof FshCodeSystem) {
    try {
      includeRule = fshDefinition.checkConcept(ruleSetRule);
    } catch (e) {
      includeRule = false;
      logger.error(e.message, ruleSetRule.sourceInfo);
    }
  }
  if (includeRule) {
    const ruleSetRuleClone = cloneDeep(ruleSetRule);
    if (context) {
      let newPath = context;
      if (ruleSetRuleClone?.path === '.') {
        logger.error(
          "The special '.' path is only allowed in top-level rules. The rule will be processed as if it is not indented.",
          ruleSetRule.sourceInfo
        );
        newPath = ruleSetRuleClone.path;
      } else if (ruleSetRuleClone.path) {
        newPath += `.${ruleSetRuleClone.path}`;
      }
      ruleSetRuleClone.path = newPath;
    }
    expandedRules.push(ruleSetRuleClone);
  }
  if (firstRule) {
    // Once one rule has been applied, all future rules should inherit the index used on that rule
    // rather than continuing to increment the index with the [+] operator
    context = context.replace(/\[\+\]/g, '[=]');
    firstRule = false;
  }
}
```

ngfreiter: Yeah I think what I had been imagining was if a rule fails the `checkConcept` check, we just log an error and `return` so that we move on to the next rule in the list. So then I think we don't have to go into the `isAllowedRule` block at all? Does that make sense or am I still missing something.

mint-thompson: That could work too, although I'm not a huge fan of return-for-early-exit inside of `forEach`. So that would look something like this:
```ts
if (ruleSetRule instanceof ConceptRule && fshDefinition instanceof FshCodeSystem) {
  try {
    if (!fshDefinition.checkConcept(ruleSetRule)) {
      return;
    }
  } catch (e) {
    logger.error(e.message, ruleSetRule.sourceInfo);
    return;
  }
}
const ruleSetRuleClone = cloneDeep(ruleSetRule);
if (context) {
  let newPath = context;
  if (ruleSetRuleClone?.path === '.') {
    logger.error(
      "The special '.' path is only allowed in top-level rules. The rule will be processed as if it is not indented.",
      ruleSetRule.sourceInfo
    );
    newPath = ruleSetRuleClone.path;
  } else if (ruleSetRuleClone.path) {
    newPath += `.${ruleSetRuleClone.path}`;
  }
  ruleSetRuleClone.path = newPath;
}
expandedRules.push(ruleSetRuleClone);
```

ngfreiter: Ah actually I think my thinking does not work, because the `ruleSetClone.path` needs to be set to `newPath` before we do this check. So I'm fine with it as is, thanks!

ngfreiter (approved):

jafeltra (approved):
This makes sense to me and works as expected!

PR #904, opened 2021-08-17
Use HTTPS to download packages

cmoesel:
HL7 has turned off HTTP for some package servers -- so now is a good time to switch all our downloads to use HTTPS.  After all, it is just like HTTP but with more S.  We now connect to:
* https://packages.fhir.org
* https://packages2.fhir.org
* https://build.fhir.org

To test this, delete or rename your `~/.fhir` cache then attempt to use SUSHI to build:
* a project based on FHIR 4.0.1 (R4) - should SUCCEED on master and this branch
* a project based on FHIR 4.1.0 (R4B) - should FAIL on master but SUCCEED on this branch
* a project based on FHIR 4.6.0 (R5) - should FAIL on master but SUCCEED on this branch
* a project with a dependency using `current` version - should SUCCEED on master and this branch

### DIFF
--- a/src/errors/CurrentPackageLoadError.ts
+++ b/src/errors/CurrentPackageLoadError.ts
@@ -4,7 +4,7 @@ export class CurrentPackageLoadError extends Error implements Annotated {
   specReferences = ['https://confluence.hl7.org/display/FHIR/NPM+Package+Specification'];
   constructor(public fullPackageName: string) {
     super(
-      `The package ${fullPackageName} is not available on http://build.fhir.org/ig/qas.json, so no current version can be loaded`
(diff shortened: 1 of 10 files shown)


ngfreiter (changes requested):
I'm getting this output when I try to download a package that is current over https:
image

Now I think there is probably something I could do to make that work, but if it's happening to me, I'm guessing it could happen to many other people on corporate networks, so is this something worth introducing? I also can't say I understand why it happens for `build.fhir.org`, but not for the package servers.

mint-thompson (dismissed):

cmoesel:
OK @ngfreiter - 7b573db contains the error message w/ additional details as we had discussed.

ngfreiter (approved):
👍

mint-thompson (approved):

PR #905, opened 2021-08-20
Try to detect name conflicts adjacent to ParentNotDefinedError

### LINKED ISSUE #858: Improve error message (if possible) when clashing names cause issues
See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Parent.20PractitionerRole.20not.20found/near/244117777

mint-thompson:
Fixes #858 and completes task [CIMPL-790](https://standardhealthrecord.atlassian.net/browse/CIMPL-790).

If the name of an entity in the tank (defined by the author) is the same as the name of an entity in the FHIR definitions, the entity in the tank will prevent the entity in the FHIR definitions from being found while fishing. This can result in somewhat confusing scenarios if, for example, a ValueSet in the tank has the same name as a Profile in the FHIR definitions, and another Profile tries to use that name as its parent. Try to detect that scenario, and when detected, let the author know that there may be conflicting names.

### DIFF
--- a/src/errors/ParentNotDefinedError.ts
+++ b/src/errors/ParentNotDefinedError.ts
@@ -4,16 +4,7 @@ import { SourceInfo } from '../fshtypes';
 
 export class ParentNotDefinedError extends Error implements Annotated, WithSource {
   specReferences = ['https://www.hl7.org/fhir/R4/profiling.html#resources'];
-  constructor(
-    public childName: string,
-    public parentName: string,
-    public sourceInfo: SourceInfo,
-    public conflict?: boolean
-  ) {
-    super(
-      `Parent ${parentName} not found for ${childName}${
-        conflict ? ' - Do you have conflicting names in your definitions?' : ''
-      }`
-    );
+  constructor(public childName: string, public parentName: string, public sourceInfo: SourceInfo) {
+    super(`Parent ${parentName} not found for ${childName}`);
   }
 }
--- a/src/errors/ParentNotDefinedError.ts
+++ b/src/errors/ParentNotDefinedError.ts
@@ -4,7 +4,16 @@ import { SourceInfo } from '../fshtypes';
 
 export class ParentNotDefinedError extends Error implements Annotated, WithSource {
   specReferences = ['https://www.hl7.org/fhir/R4/profiling.html#resources'];
-  constructor(public childName: string, public parentName: string, public sourceInfo: SourceInfo) {
-    super(`Parent ${parentName} not found for ${childName}`);
+  constructor(
+    public childName: string,
+    public parentName: string,
+    public sourceInfo: SourceInfo,
+    public conflict?: boolean
+  ) {
+    super(
+      `Parent ${parentName} not found for ${childName}${
+        conflict ? ' - Do you have conflicting names in your definitions?' : ''
+      }`
+    );
   }
 }
--- a/src/errors/ParentNameConflictError.ts
+++ b/src/errors/ParentNameConflictError.ts
@@ -1,0 +1,17 @@
+import { WithSource } from './WithSource';
+import { SourceInfo } from '../fshtypes';
+
+export class ParentNameConflictError extends Error implements WithSource {
+  constructor(
+    public childName: string,
+    public parentName: string,
+    public parentType: string,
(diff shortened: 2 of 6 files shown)


src/errors/ParentNotDefinedError.ts
```diff
@@ -4,7 +4,16 @@ import { SourceInfo } from '../fshtypes';
...
-    super(`Parent ${parentName} not found for ${childName}`);
+  constructor(
+    public childName: string,
+    public parentName: string,
+    public sourceInfo: SourceInfo,
+    public conflict?: boolean
+  ) {
+    super(
+      `Parent ${parentName} not found for ${childName}${
+        conflict ? ' - Do you have conflicting names in your definitions?' : ''
```
jafeltra, line 15: I'm wondering if there is any additional information we can provide here. I feel like when I replicate the scenario from your test, I see this error message and think "no I don't have any conflicting names, I just have one `PractitionerRole`". Do you think there's any value is asking if there are any names that conflict with FHIR resources? It might be too much to say the type of the conflicting resource we found (i.e. "ValueSet" in the example case), so maybe just mentioning that the conflict has to do with the FHIR definitions will help.

Or am I just overthinking this?

jafeltra (commented):
I just had one question about the message included in the error. Let me know what you think.

src/errors/ParentNotDefinedError.ts
mint-thompson: I don't think you're overthinking this. I'm not super-attached to this error message, it was just the best one I managed to come up with. The scenario that leads to this is sort of a weird one, so it's hard to describe succinctly. I think it would be useful to include information about the conflicting definition we found. I guess at that point, it's not even really a `ParentNotDefinedError` error. It's kind of the opposite, really: we have an extra definition with the parent's name, but that definition's type won't work as the parent. But it also has to specifically be that the extra definition comes from FSH, when the definition that the author probably intended comes from included FHIR definitions, so it's not quite an invalid parent error either. It really is its own special case. Maybe a useful message could be something like `Parent ${parentName} for ${childName} is defined as a ${parentType} in FSH, which can't be used as a parent. A FHIR definition also exists with this name: did you intend to use that?`

Double alternatively, if we want to get fancy, we could consider this a warning case instead of an error case, and go ahead and use the FHIR definition as long as it is a valid parent for whatever kind of Structure Definition we're exporting. It's legal (although not a great idea) for difference resource types to share a name. So the case in the issue would show that warning, but it would use the PractitionerRole profile as defined in FHIR as the parent.

jafeltra: > Maybe a useful message could be something like `Parent ${parentName} for ${childName} is defined as a ${parentType} in FSH, which can't be used as a parent. A FHIR definition also exists with this name: did you intend to use that?`

I like this idea, but is there ever a way to use the FHIR definition without changing the name of the FSH created definition? If not, I think I do like the second idea of logging a warning and using the FHIR definition instead. I'm trying to think of if there is a case where the user would be intending to use their own definition, but since the FSH definition is the wrong type and can't be used as a parent, I'm not really coming up with a reason why they would ever want to do that. So I think I like that second idea!

mint-thompson: The FHIR definition can be referenced by URL. So, it's possible to keep the FSH definition with the same name and avoid these errors. I think the original idea behind making this error message clearer was to try to guide the user toward using the URL as the workaround.

jafeltra: Ahh okay I always forget about using the URL itself. I would say that if we want to avoid fanciness (which might be wise), maybe we can just suggest using the URL in the error message. Maybe something like this?

```
Parent ${parentName} for ${childName} is defined as a ${parentType} in FSH, which can't be used as a parent. A FHIR definition also exists with this name. If you intended to use that, reference it by its URL.
```

julianxcarter (approved):
I think you two hit the nail right on the head with this error message, it clearly lets you know what's going on and also gives you a solid suggestion on how to move forward. This is great!

jafeltra (approved):
I think this looks good! I really like the error message now - I think the FSH type and the suggestion to use the URL are really helpful. Thanks for making those changes!

PR #906, opened 2021-08-20
Extension slices now always include a type field in the differential 

### LINKED ISSUE #871: Slicing on extension doesn't produce a proper differential
I have the following slice definition in a FSH file:
```
* careTeam MS
* careTeam.extension contains CareTeamClaimScope named careTeamClaimScope 1..1 MS
* careTeam.sequence MS
* careTeam.provider MS
* careTeam.provider only Reference(PASPractitioner or PASOrganization)
* careTeam.role MS
* careTeam ^slicing.discriminator.type = #value
* careTeam ^slicing.discriminator.path = "extension('http://hl7.org/fhir/us/davinci-pas/StructureDefinition/extension-careTeamClaimScope').valueBoolean"
* careTeam ^slicing.rules = #open
* careTeam ^slicing.description = "Slice based on whether the care team member belongs to the overall claim or to an individual claim item."
* careTeam contains OverallClaimMember 0..14 and ItemClaimMember 0..10
```
This produces the following differential:
```
     {
        "id": "Claim.careTeam.extension",
        "path": "Claim.careTeam.extension",
        "slicing": {
          "discriminator": [
            {
              "type": "value",
              "path": "url"
            }
          ],
          "ordered": false,
          "rules": "open"
        },
        "min": 1
      },
      {
        "id": "Claim.careTeam.extension:careTeamClaimScope",
        "path": "Claim.careTeam.extension",
        "sliceName": "careTeamClaimScope",
        "min": 1,
        "max": "1",
        "type": [
          {
            "code": "Extension",
            "profile": [
              "http://hl7.org/fhir/us/davinci-pas/StructureDefinition/extension-careTeamClaimScope"
            ]
          }
        ],
        "mustSupport": true
      },
      ...
        "id": "Claim.careTeam:OverallClaimMember",
        "path": "Claim.careTeam",
        "sliceName": "OverallClaimMember",
        "min": 0,
        "max": "14"
      },
      {
        "id": "Claim.careTeam:OverallClaimMember.extension:careTeamClaimScope",
        "path": "Claim.careTeam.extension",
        "sliceName": "careTeamClaimScope"
      },
```

That second declaration of the careTeam.extension causes the publisher validation code to fail because whenever it sees a path that ends in '.extension' with a sliceName, it assumes that there will be a type element as in the first declaration.

When I overrode the profile structuredefinition to include the type information, the publisher executed fine.

This *might* be a bug in the publisher FHIRPathEngine code, but it does seem like that that element in the differential isn't correct.

julianxcarter:
Fixes #871, ensuring that the type element on extension slices is present in the differential array whether it's changed or not. This must be done to comply with the IG Publisher, as it will occasionally throw errors if an extension slice inherited from a slice on the extension's parent element does not explicitly state its type in the differential.

### DIFF
--- a/src/fhirtypes/ElementDefinition.ts
+++ b/src/fhirtypes/ElementDefinition.ts
@@ -392,6 +392,14 @@ export class ElementDefinition {
           // @ts-ignore
           diff[prop] = cloneDeep(this[prop]);
         }
+      } else if (
+        prop &&
+        prop === 'type' &&
+        this.sliceName &&
+        this.id.slice(0, this.id.lastIndexOf(':')).endsWith('.extension')
+      ) {
+        // Due to a quirk in the IG Publisher, we'll want to include the type within the diff on extension slices
+        diff[prop] = cloneDeep(this[prop]);
       }
     }
     // Set the diff id, which may be different than snapshot id in the case of choices (e.g., value[x] -> valueString)
--- a/test/fhirtypes/ElementDefinition.test.ts
+++ b/test/fhirtypes/ElementDefinition.test.ts
@@ -88,6 +88,65 @@ describe('ElementDefinition', () => {
         new ElementDefinitionType('Period')
       ]);
     });
+
+    it('should include the type property within the differential array of extension slices', () => {
+      const baseElement = {
+        id: 'Profile.element.extension:extensionSlice',
+        sliceName: 'extensionSlice',
+        type: [
+          {
+            code: 'Extension',
+            profile: ['http://example.org/StructureDefinition/extensionSlice']
+          }
+        ]
+      };
+
+      const sliceElement = {
+        id: 'Profile.element:elementSlice.extension:extensionSlice',
+        sliceName: 'extensionSlice',
+        type: [
+          {
+            code: 'Extension',
+            profile: ['http://example.org/StructureDefinition/extensionSlice']
+          }
+        ],
+        _original: baseElement
+      };
+
+      const slicedElementDef = ElementDefinition.fromJSON(sliceElement);
+      const elementDiff = slicedElementDef.calculateDiff();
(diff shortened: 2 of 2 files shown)


mint-thompson (dismissed):
Looks good to me. I tested it out with the FSH in the linked issue, and once I added a line to force the nested extension to appear in the differential, it correctly included the type information.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -392,6 +392,14 @@ export class ElementDefinition {
           // @ts-ignore
           diff[prop] = cloneDeep(this[prop]);
         }
+      } else if (
+        prop &&
+        prop === 'type' &&
+        this.sliceName &&
+        this.id.slice(0, this.id.lastIndexOf(':')).endsWith('.extension')
```
ngfreiter, line 399: I'm getting regressions on this PR, and they appear to be tied to the fact that this check also catches "inline" extensions. Here is an example that can be used to recreate what I'm seeing:
```
Extension: Inline
* extension contains
    volume 1..1 MS
```
On this branch, you are going to get something like:
```json
{
  "id": "Extension.extension:volume",
  "path": "Extension.extension",
  "sliceName": "volume",
  "min": 1,
  "max": "1",
  "type": [
    {
      "code": "Extension"
    }
  ],
  "mustSupport": true
}
```
where that `type` information is really not needed, since the type is unchanged. On `master`, you get no such type info.

ngfreiter (changes requested):

src/fhirtypes/ElementDefinition.ts
julianxcarter: This is a good catch! Addressed in commit `1a4f635`

mint-thompson (approved):
✨

PR #908, opened 2021-08-24
Url property now set automatically on instance definitions

### LINKED ISSUE #791: SUSHI should automatically populate `url` for appropriate #definition instances
If an `Instance` has `Usage: #definition` and is a type that has a `url` field intended to represent the definition's canonical, then SUSHI should automatically set the `url` like it does for `Profile`, `Extension`, `CodeSystem`, and `ValueSet` definitions.  Otherwise the user must manually type in the full URL, which can be brittle if the overall canonical of the IG changes.

julianxcarter:
Fixes #791, making it so that SUSHI now sets the URL on instances with a `Usage: #definition` just as it does on other entities. You can see this in action with [this simple example](https://fshschool.org/FSHOnline/#/share/3mtrCzr). Running this in FSHOnline will not set the url automatically, but running this same example on this branch will set a url based on the canonical. An assignment rule explicitly setting the url will take precedence over the auto-generated URL.

### DIFF
--- a/src/export/InstanceExporter.ts
+++ b/src/export/InstanceExporter.ts
@@ -297,6 +297,9 @@ export class InstanceExporter implements Fishable {
     }
     if (fshDefinition.usage) {
       instanceDef._instanceMeta.usage = fshDefinition.usage;
+      if (fshDefinition.usage === 'Definition') {
+        instanceDef.url = `${this.tank.config.canonical}/${instanceOfStructureDefinition.type}/${fshDefinition.id}`;
+      }
     }
     if (isResource) {
       instanceDef.resourceType = instanceOfStructureDefinition.type; // ResourceType is determined by the StructureDefinition of the type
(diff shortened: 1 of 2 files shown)


cmoesel:
Quick review here, but... what if the instance doesn't actually support a `url` field?  E.g.,
```
Instance: MyPatient
InstanceOf: Patient
Usage: #definition
* name.given = "Bob"
```

While it doesn't make a ton of sense to mark `MyPatient` as a `#definition`, I don't think we prohibit this -- and as your code is now, I think it would tack on a `url` -- which would be invalid for a `Patient` instance.

julianxcarter:
I had not thought of that!

> Quick review here, but... what if the instance doesn't actually support a `url` field? E.g.,
> 
> ```
> Instance: MyPatient
> InstanceOf: Patient
> Usage: #definition
> * name.given = "Bob"
> ```
> 
> While it doesn't make a ton of sense to mark `MyPatient` as a `#definition`, I don't think we prohibit this -- and as your code is now, I think it would tack on a `url` -- which would be invalid for a `Patient` instance.

I had not thought of that! Commit `1eb4942` should correct this.

cmoesel:
Perfect!

ngfreiter (approved):
Looks good to me, nice job!

jafeltra (approved):
This looks good to me!

PR #912, opened 2021-08-30
SUSHI now loads lower case package names by default

### LINKED ISSUE #800: Support case-insensitive package names but prefer lowercase
See: https://chat.fhir.org/#narrow/stream/179166-implementers/topic/Error.20loading.20package.20in.20java.20validator/near/234287943

julianxcarter:
Fixes #800, making it so that SUSHI will always download and store a package with a lowercase package Id regardless of how the name was input. If a package Id with upper-case letters is included as a dependency, a warning is emitted letting the user know that the lower case version will be downloaded instead.


src/fhirdefs/load.ts
```diff
@@ -25,13 +25,20 @@ export async function loadDependency(
   FHIRDefs: FHIRDefinitions,
   cachePath: string = path.join(os.homedir(), '.fhir', 'packages')
 ): Promise<FHIRDefinitions> {
+  if (/(?=.*[A-Z])/.test(packageName)) {
```
ngfreiter, line 28: This is just checking that the `packageName` has any capital letters right? I think you can simplify it to:
```javascript
/[A-Z]/.test(packageName)
```
or am I missing something?

src/fhirdefs/load.ts
```diff
@@ -25,13 +25,20 @@ export async function loadDependency(
...
+    logger.warn(
+      `${packageName} contains upper-case characters, which is discouraged. SUSHI will download ${packageName.toLowerCase()}#${version} if the package is not found in local cache.`
```
ngfreiter, line 30: I think `uppercase` (all one word) is more common. Also I think this error could in theory be slightly inaccurate, since if they are looking for a `dev` package for example, we aren't going to download anything. I think it could be more accurate to just say "SUSHI will use `${packageName.toLowerCase()}` as the packages name."

src/fhirdefs/load.ts
```diff
@@ -25,13 +25,20 @@ export async function loadDependency(
...
+    packageName = packageName.toLowerCase();
+  }
   let fullPackageName = `${packageName}#${version}`;
   let loadPath = path.join(cachePath, fullPackageName, 'package');
   let loadedPackage: string;
 
   // First, try to load the package from the local cache
   logger.info(`Checking local cache for ${fullPackageName}...`);
   loadedPackage = loadFromPath(loadPath, fullPackageName, FHIRDefs);
+
```
ngfreiter, line 41: Typo?

ngfreiter (changes requested):

src/fhirdefs/load.ts
julianxcarter: No, you're not missing anything! The regex you suggested works just fine. Addressed in commit `855cd26`.

src/fhirdefs/load.ts
julianxcarter: I hadn't thought of that! I changed the warning message in commit `855cd26`.

src/fhirdefs/load.ts
julianxcarter: Yes! I had initially added some code there but I realized it wasn't needed so I yanked it out. Addressed in commit `855cd26`.

src/fhirdefs/load.ts
```diff
@@ -25,6 +25,12 @@ export async function loadDependency(
   FHIRDefs: FHIRDefinitions,
   cachePath: string = path.join(os.homedir(), '.fhir', 'packages')
 ): Promise<FHIRDefinitions> {
+  if (/[A-Z]/.test(packageName)) {
+    logger.warn(
+      `${packageName} contains uppercase characters, which is discouraged. SUSHI will use ${packageName.toLowerCase()} as the package name.`
+    );
+    packageName = packageName.toLowerCase();
+  }
```
ngfreiter, line 33: When I try to use a package with capitals, like:
```yaml
dependencies:
  hl7.fhir.us.Core: current
```
I'm seeing an issue down the line from SUSHI:
image

This seems to happen because although we lowercase the package name here, we don't do that in the configuration, and that is the value that gets used in the publisher. I'm wondering if it may be easier to just move this earlier in the chain, and do this check as a part of importing the configuration itself. What do you think?

ngfreiter (changes requested):

src/fhirdefs/load.ts
julianxcarter: It definitely makes more sense to handle this as the configuration file is being imported and parsed, rather than when each individual dependency is being loaded. Commit `327b977` moved the warning and lowercase-ification to when the configuration is imported.

jafeltra (approved):
The latest changes here make sense to me. I get through the full process of building the IG resource and no longer see the error that Nick pointed out in his last comment. I think this should be set!

cmoesel:
So... I hate to throw a monkey wrench in this, but here is an excerpt from the Zulip conversation linked from the issue (emphasis mine):

> When someone now specifies `KBV.Basis@1.1.3` as a dependency in sushi or IG publisher or wherever, I hope you will:
> 1. query the local package cache for `kbv.basis` **_regardless of how it's capitalized there_**
> 2. if not locally available yet, query the package server for lowercase `kbv.basis` and put a lowercase folder in the FHIR package cache (although that shouldn't matter if other tools implement 1. as well, but just for good measure)

Since other tools write to the FHIR cache, and since they may have written folders _with capital letters_ into the cache, then we need to do a case-insensitive search for it in the cache.  As the code is in this PR, we would always use lowercase folder names with `fs.exists` to see if the package already exists in the cache; but unfortunately it seems that `fs.exists` is case-sensitive, so we'd miss packages with differing case _on case-sensitive file systems._

By default, Windows and Mac use case-insensitive file systems, so for many users, this doesn't matter.  But for those who are using linux (or non-default configurations of Windows/Mac), it does matter. :-/

julianxcarter:
> So... I hate to throw a monkey wrench in this, but here is an excerpt from the Zulip conversation linked from the issue (emphasis mine):
> 
> > When someone now specifies `KBV.Basis@1.1.3` as a dependency in sushi or IG publisher or wherever, I hope you will:
> > 
> > 1. query the local package cache for `kbv.basis` **_regardless of how it's capitalized there_**
> > 2. if not locally available yet, query the package server for lowercase `kbv.basis` and put a lowercase folder in the FHIR package cache (although that shouldn't matter if other tools implement 1. as well, but just for good measure)
> 
> Since other tools write to the FHIR cache, and since they may have written folders _with capital letters_ into the cache, then we need to do a case-insensitive search for it in the cache. As the code is in this PR, we would always use lowercase folder names with `fs.exists` to see if the package already exists in the cache; but unfortunately it seems that `fs.exists` is case-sensitive, so we'd miss packages with differing case _on case-sensitive file systems._
> 
> By default, Windows and Mac use case-insensitive file systems, so for many users, this doesn't matter. But for those who are using linux (or non-default configurations of Windows/Mac), it does matter. :-/

Mmmm I had actually thought of this, but assumed that `fs.exists` was case insensitive because it appeared to be as I was testing it on _my_ computer, which happens to have a case-insensitive file system 🤦🏽‍♂️. This is a good catch @cmoesel, I'll update this to check for the non-lowercased version of a package as well!

cmoesel:
> I'll update this to check for the non-lowercased version of a package as well!

I think the goal is to allow for _any casing_ of it in the cache.  So I'm not saying to check the lowered version and the non-lowered version, because it's possible it's in a completely different case than the author wrote in their config file.  I'm not sure, but perhaps the best approach is to get a directory listing of the FHIR cache (e.g., `fs.readdirSync('~/.fhir/packages')`) and then do a `find` on the array of folder names using a case-insensitive string comparison.

julianxcarter:
Commit `bfc2213` makes it so that `loadFromPath` doesn't discriminate based on the casing of the package name located in the cache. In the process I reorganized the `loadDependency` to `loadFromPath` routine that we currently have, making it so that the `loadFromPath` function expects a path to the FHIR cache rather than a path to the package itself and infers the path to the `targetPackage`. I also reorganized our test fixtures in `testhelpers/testdefs` to more closely mirror the structure of packages found in the FHIR cache (giving them all a `package` subdirectory).

ngfreiter (approved):
👍 👍

PR #913, opened 2021-08-31
Warn when a concept is included in a ValueSet more than once

### LINKED ISSUE #832: A code can be included muliple times in a Value Set
SUSHI allows easily detectable GIGO errors.... if you include the same code in a VS > 1 time, it comes out > 1 time.   An easy check for duplicate code/code-system would save ballot comments.

See https://fshschool.org/FSHOnline/#/share/3g9tIzO

The duplicate code ends in 008

mint-thompson:
Fixes #832 and completes task [CIMPL-814](https://standardhealthrecord.atlassian.net/browse/CIMPL-814).

Including the same concept (with the same code, system, and version) in a ValueSet multiple times has no effect. It does not cause any validation errors, but it may indicate a mistake when defining the ValueSet. Warn the user when this is detected, and do not add the concept again. This only applies to inclusion: no changes are made to exclusion logic.

### DIFF
--- a/src/export/ValueSetExporter.ts
+++ b/src/export/ValueSetExporter.ts
@@ -79,31 +79,6 @@ export class ValueSetExporter {
             }
             return composeConcept;
           });
-          if (component.inclusion) {
-            // warn the user if they have already included a concept in this component
-            // concept, system, and version must all match to be considered equal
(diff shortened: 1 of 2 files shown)


julianxcarter (approved):
Looks good to me! Great job @mint-thompson

jafeltra (approved):
Looks good to me!

PR #916, opened 2021-09-14
Select correct choice slice when validating properties of Instances

### LINKED ISSUE #911: Ghost issue in data types when adding several extensions makes sushi give up
When we declare an extension that has a valueExpression, if there are other extensions of other types, sushi complains that the other extensions are missing the valueExtension.language. 
The conversion works, and goFSH seems to convert it back ok. If we try the roundtrip conversion everything is well. 
But sushi complains about errors - it seems these are ghost errors.
And Sushi reports the errors upon exit, which blocks the IG build.

To reproduce: 
Try to convert this from fsh to json. Conversion goes ok (and goFSH works which confirms the code is ok), but there are 2 errors:
Removing lines 21 and 22 and/or 23-25 remove the errors, even if those lines are ok.
https://fshschool.org/FSHOnline/#/share/3sU6cfX

jafeltra:
This PR fixes #911.

When we validate the required child elements of instances, we have special logic to handle the case when a child is a choice slice. We find all possible choice slices that the current property could match and then checks each choice slice to see if it is the slice we are currently validating. When we do that, we weren't updating which element we were using to validate cardinality properties. Therefore, when there were two or more choice slices assigned, the wrong child element could potentially be used to compare cardinality requirements.

In the example provided in the issue, the following FSH snippet is present:

```
* item[=].extension[0].url = "http://hl7.org/fhir/StructureDefinition/variable"
* item[=].extension[=].valueExpression.name = "scoreExt"
* item[=].extension[=].valueExpression.language = #text/fhirpath
* item[=].extension[=].valueExpression.expression = "'http://hl7.org/fhir/StructureDefinition/ordinalValue'"
* item[=].extension[+].url = "http://hl7.org/fhir/StructureDefinition/questionnaire-unit"
* item[=].extension[=].valueCoding.display = "{score}"
```

When checking the required properties of the `valueCoding` slice, we were mistakenly using the `valueExpression` element on the structure definition to compare against. We correctly identified that `valueCoding` is the slice we care about, but didn't update the element definition we use. This PR updates the element definition whenever we find the relevant slice. A test was added that uses the above snippet from the issue.

### DIFF
--- a/src/export/InstanceExporter.ts
+++ b/src/export/InstanceExporter.ts
@@ -144,7 +144,8 @@ export class InstanceExporter implements Fishable {
   ): void {
     // Get only direct children of the element
     const children = element.children(true);
-    children.forEach(child => {
+    children.forEach(c => {
+      let child = c;
       // Get the last part of the path, A.B.C => C
       const childPathEnd = child.path.split('.').slice(-1)[0];
       // Note that in most cases the _ prefixed element will not exist. But if it does exist, we validate
@@ -160,6 +161,8 @@ export class InstanceExporter implements Fishable {
         for (const choiceSlice of choiceSlices) {
           instanceChild = instance[choiceSlice.sliceName];
           if (instanceChild != null) {
+            // Once we find the the choiceSlice that matches, use it as the child
+            child = choiceSlice;
             break;
           }
         }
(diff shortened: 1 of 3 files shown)


ngfreiter (approved):

cmoesel (approved):
Nice sleuthing!

PR #917, opened 2021-09-15
Update dependencies

jafeltra:
This PR updates our dependencies with reported vulnerabilities using `npm audit fix`.

Note that there is still one dependency with a moderate severity vulnerability. Our dependency `cpy-cli` has a dependency of `cpy`. `cpy` has upgraded the package that has a vulnerability but has not yet released a new version. `cpy-cli` is waiting on this new release to upgrade, as noted [here](https://github.com/sindresorhus/cpy-cli/issues/32).


cmoesel (approved):
Downloaded locally.  Tests pass.  CLI works and properly downloads/unzips dependencies (checked since `tar` was involved).  `npm-audit` only reports `glob-parent`, which, as you noted, can't be resolved right now until cpy-cli updates.  Looks good!

Issue #827, opened 2021-06-05
SUSHI Runs Forever on FSH Example


markkramerus:
We include this example of a Capability Statement in the FSH IG:

https://github.com/HL7/fhir-shorthand/blob/master/input/images/CapabilityStatementTemplate.fsh

SUSHI 1.3.2 starts compiling and never* completes. It is a massive instance but it still shouldn't be this bad. It is probably the same efficiency problem we have seen on big bundle profiles with many slices. 

I also tried this on 2.0.0B1 and the results were not different.

*I killed it after about 20 minutes.

alex-vg:
Your link is presenting me a 404.

PR #919, opened 2021-09-20
Improve performance by decreasing cloning

ngfreiter:
Potentially fixes https://github.com/FHIR/sushi/issues/827. 

This PR helps to improve the performance of SUSHI on large IGs with a lot of instances. It does this mainly by avoiding cloning elements in those IGs so many times. In the `validateValueAtPath` function, we used to clone the relevant element and see if we could assign a value to it. Instead, we now just save off the current value assigned to that element, and place that value back on the element after processing. This results in some pretty significant speedup. In my testing the https://github.com/HL7/fhir-saner was the best benchmark. On `master`, this IG takes ~250 seconds, but on this branch it should only take ~100 seconds.

There is always more room for performance improvements, but I was having trouble making any similarly significant changes beyond this one, so I've called it here. But if in your review you feel there is more room for improvement, let me know. This should not result in any changes to any IGs.


cmoesel (approved):
Nice find. I've run a full regression and confirmed that the results are byte-for-byte the same.  I also pushed up the updated repo list since some repos were no longer valid.

I also compared running the regression on master and on this branch (more specifically, once w/ remote master vs local master, and once with remote PR branch vs local PR branch).  The master regression took 3331 seconds while this PR regression took 2456 secs -- coming out to a reduction of more than 25%.  Since this only affects instances, some repos would seem larger reductions than others.

Good work!

mint-thompson (approved):

PR #920, opened 2021-09-23
SUSHI now resolves systems in FSH Codes defined in an instance

### LINKED ISSUE #801: SUSHI should look for code systems defined using Instance #definition when resolving FSH codes
This should work, but it does not:
```
Instance: FooBar
InstanceOf: CodeSystem
Usage: #definition
* name = "FooBar"
* title = "Foo Bar"
* url = "http://codesystem.com/foobar"
* status = #active
* content = #complete
* concept[+].code = #foo
* concept[+].code = #bar


Instance: ObservationTest
InstanceOf: Observation
* status = #final
* code = FooBar#foo
```

julianxcarter:
Fixes #801, making it so that SUSHI looks for code systems defined as an `Instance` if a literal `CodeSystem` with the same name is not found.

### DIFF
--- a/src/fhirtypes/common.ts
+++ b/src/fhirtypes/common.ts
@@ -301,8 +301,11 @@ export function replaceReferences<T extends AssignmentRule | CaretValueRule>(
   } else if (value instanceof FshCode) {
     const [system, ...versionParts] = value.system?.split('|') ?? [];
     const version = versionParts.join('|');
-    const codeSystem = tank.fish(system, Type.CodeSystem);
-    const codeSystemMeta = fisher.fishForMetadata(codeSystem?.name, Type.CodeSystem);
+    const codeSystem = tank.fish(system, Type.CodeSystem) ?? tank.fish(system, Type.Instance);
+    const codeSystemMeta =
+      codeSystem instanceof FshCodeSystem
+        ? fisher.fishForMetadata(codeSystem?.name, Type.CodeSystem)
+        : fisher.fishForMetadata(codeSystem?.name, Type.Instance);
     if (codeSystem && codeSystemMeta) {
       clone = cloneDeep(rule);
       const assignedCode = getRuleValue(clone) as FshCode;
--- a/test/export/InstanceExporter.test.ts
+++ b/test/export/InstanceExporter.test.ts
@@ -2226,6 +2226,33 @@ describe('InstanceExporter', () => {
       ]);
     });
 
+    it('should assign a code to a top level element if the code system was defined as an instance of type definition', () => {
+      const visibleSystem = new Instance('Visible');
+      visibleSystem.instanceOf = 'CodeSystem';
+      const urlRule = new AssignmentRule('url');
+      urlRule.value = 'http://hl7.org/fhir/us/minimal/Instance/Visible';
+      const nameRule = new AssignmentRule('name');
+      nameRule.value = 'Visible';
+      visibleSystem.rules.push(urlRule, nameRule);
+      doc.instances.set(visibleSystem.name, visibleSystem);
+      exportInstance(visibleSystem);
+
+      const brightInstance = new Instance('BrightObservation');
+      brightInstance.instanceOf = 'Observation';
+      const assignedCodeRule = new AssignmentRule('code');
+      assignedCodeRule.value = new FshCode('bright', 'Visible');
+      brightInstance.rules.push(assignedCodeRule);
+      doc.instances.set(brightInstance.name, brightInstance);
+
+      const exported = exportInstance(brightInstance);
+      expect(exported.code.coding).toEqual([
+        {
+          code: 'bright',
+          system: 'http://hl7.org/fhir/us/minimal/Instance/Visible'
+        }
+      ]);
+    });
+
     it('should assign a code to a nested element while replacing the local code system name with its url', () => {
       const brightInstance = new Instance('BrightObservation');
       brightInstance.instanceOf = 'Observation';


src/fhirtypes/common.ts
```diff
@@ -301,8 +301,11 @@ export function replaceReferences<T extends AssignmentRule | CaretValueRule>(
   } else if (value instanceof FshCode) {
     const [system, ...versionParts] = value.system?.split('|') ?? [];
     const version = versionParts.join('|');
-    const codeSystem = tank.fish(system, Type.CodeSystem);
-    const codeSystemMeta = fisher.fishForMetadata(codeSystem?.name, Type.CodeSystem);
+    const codeSystem = tank.fish(system, Type.CodeSystem) ?? tank.fish(system, Type.Instance);
+    const codeSystemMeta =
+      codeSystem instanceof FshCodeSystem
+        ? fisher.fishForMetadata(codeSystem?.name, Type.CodeSystem)
+        : fisher.fishForMetadata(codeSystem?.name, Type.Instance);
```
mint-thompson, line 308: I think this ends up being too permissive in terms of what types of Instances can be used. For example, a CapabilityStatement can have a name and a url, so an Instance of a CapabilityStatement can be used without any errors appearing. So, there probably needs to be a check that that the Instance fished up is actually a CodeSystem.

mint-thompson (changes requested):

src/fhirtypes/common.ts
julianxcarter: This is an excellent point, addressed in commit `17993d4`.

src/fhirtypes/common.ts
```diff
@@ -301,8 +321,18 @@ export function replaceReferences<T extends AssignmentRule | CaretValueRule>(
...
+      const csInstance = tank.fish(system, Type.Instance);
+      if (csInstance && isInstance(csInstance) && csInstance.instanceOf === 'CodeSystem') {
+        codeSystem = csInstance;
+      }
+    }
+
+    const codeSystemMeta =
+      codeSystem instanceof FshCodeSystem
+        ? fisher.fishForMetadata(codeSystem?.name, Type.CodeSystem)
+        : fisher.fishForMetadata(codeSystem?.name, Type.Instance);
```
mint-thompson, line 335: If `codeSystem` is null at this point, then here, we just fish for nothing. It might be possible to restructure this so that right after determining that `codeSystem` is something we can use, we immediately fish for metadata in a situation where we know what type to fish for. Something like:
```ts
let codeSystemMeta: Metadata;
if (codeSystem) {
  codeSystemMeta = fisher.fishForMetadata(codeSystem.name, Type.CodeSystem;
} else {
  const csInstance = tank.fish(system, Type.Instance);
  if (csInstance && isInstance(csInstance) && csInstance.instanceOf === 'CodeSystem') {
    codeSystem = csInstance;
    codeSystemMeta = fisher.fishForMetadata(codeSystem.name, Type.Instance);
  }
}
```

mint-thompson (changes requested):
Thank you for making the change I requested. I have one additional suggestion for a code change. Additionally, do you think it would be useful to add a negative test showing that if the Instance is _not_ an instanceOf CodeSystem, that its URL is not used in `replaceReferences`?

src/fhirtypes/common.ts
julianxcarter: This is a great point, especially in light of our recent performance gains. Commit `2f25964` takes this suggestion, and also added in a test case ensuring that non-CodeSytem instances aren't being resolved as system urls in concepts.

src/fhirtypes/common.ts
```diff
@@ -301,8 +321,17 @@ export function replaceReferences<T extends AssignmentRule | CaretValueRule>(
...
     const version = versionParts.join('|');
-    const codeSystem = tank.fish(system, Type.CodeSystem);
-    const codeSystemMeta = fisher.fishForMetadata(codeSystem?.name, Type.CodeSystem);
+    let codeSystem = tank.fish(system, Type.CodeSystem);
+    let codeSystemMeta: Metadata;
+    if (codeSystem) {
+      codeSystemMeta = fisher.fishForMetadata(codeSystem.name, Type.CodeSystem);
+    } else {
+      const csInstance = tank.fish(system, Type.Instance);
+      if (csInstance && isInstance(csInstance) && csInstance.instanceOf === 'CodeSystem') {
```
ngfreiter, line 330: I notice this doesn't check that the `csInstance` is definitional (has `Usage: #definition`). This was something @cmoesel specified in the [original issue](https://github.com/FHIR/sushi/issues/801). I don't think it's likely someone would define a `CodeSystem` that is _not_ meant to be definitional, but maybe it is good to require they be explicit? Not sure, would like to hear @cmoesel's thoughts since he did the issue.

src/fhirtypes/common.ts
```diff
@@ -301,8 +321,17 @@ export function replaceReferences<T extends AssignmentRule | CaretValueRule>(
   } else if (value instanceof FshCode) {
     const [system, ...versionParts] = value.system?.split('|') ?? [];
     const version = versionParts.join('|');
-    const codeSystem = tank.fish(system, Type.CodeSystem);
-    const codeSystemMeta = fisher.fishForMetadata(codeSystem?.name, Type.CodeSystem);
+    let codeSystem = tank.fish(system, Type.CodeSystem);
```
ngfreiter, line 324: Again going to appeal to the higher power of @cmoesel here, creator of the issue. I was thinking that something like
```
Instance: FooBar
InstanceOf: CodeSystem
Usage: #definition
* name = "FooBar"
* title = "Foo Bar"
* url = "http://codesystem.com/foobar"
* status = #active
* content = #complete
* concept[+].code = #foo
* concept[+].code = #bar

ValueSet: V
* include codes from system FooBar
```
should also work, but currently it does not. The issue was specific resolving FSH codes, but it feels a little confusing to me to support that case and not this one. I'm wondering if it would be possible to make this happen inside the `tank` itself. What if when you fished for a `CodeSystem`, we looked through `CodeSystems` first, but then looked through any `Instances` with `InstanceOf: CodeSystem` set. Definitely this could be not possible for some reason I'm not seeing, but what do you think?

ngfreiter (changes requested):

src/fhirtypes/common.ts
julianxcarter: It's definitely possible. I had considered implementing this from the `fish` method instead of in `replaceReferences`, but wasn't sure if there was some reason we wouldn't want this to happen in all cases so I went for the slightly narrower option. I do agree that facially it makes sense for these instance-defined CodeSystems to resolve in all areas that a regular CodeSystem would.

src/fhirtypes/common.ts
julianxcarter: Excellent point. Commit `081fd08` requires that the instance in question has a `usage` of `Definition` to be resolved as a `CodeSystem`.

src/fhirtypes/common.ts
julianxcarter: Commit `081fd08` moves this functionality into the tank so this fix is a bit more wide-ranging... I don't see any obvious problems with this approach but if @cmoesel objects then it can be rolled back.

src/fhirtypes/common.ts
mint-thompson: Now that the fishing has been updated to account for instances of CodeSystem, I think it's no longer necessary to have different logic for fishing up the metadata, since `fishForMetadata` calls `fish`. So, I think this can safely be turned back to its original form.

julianxcarter: You're right! Commit `14bc28dc` returns that bit back to its original form.

mint-thompson (approved):
With the latest commit, all of my findings are addressed. Hooray!

ngfreiter (approved):

PR #921, opened 2021-09-24
Show error when inserting RuleSet with concepts at a path

mint-thompson:
Completes task [CIMPL-807](https://standardhealthrecord.atlassian.net/browse/CIMPL-807).

If a RuleSet is inserted at a path, that path is the context for the rules in the RuleSet. However, a ConceptRule should not have a path as context. So, if a RuleSet that contains a ConceptRule is inserted at a path, show an error so that the author knows that this should not be done. The ConceptRule should still be applied.

### DIFF
--- a/src/fhirtypes/common.ts
+++ b/src/fhirtypes/common.ts
@@ -491,6 +491,14 @@ export function applyInsertRules(
             ruleSetRuleClone.path = newPath;
           }
           if (ruleSetRuleClone instanceof ConceptRule && fshDefinition instanceof FshCodeSystem) {
+            // ConceptRules should not have a path context, so if one exists, show an error.
+            // The concept is still added to the CodeSystem.
+            if (context) {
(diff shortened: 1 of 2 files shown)


jafeltra (commented):
This looks good to me and logs the error at the appropriate time. My only question is if this should be a warning or an error? I can't remember exactly the distinction that we have made in the past, but it feels like since we appropriately adjust how the rule is added and nothing really breaks, maybe this is more of a warning. Is there a reason you were thinking error?

mint-thompson:
@jafeltra The reason why I think an error is deserved here is because if we think of how the RuleSet's rules would get applied, it becomes something that isn't syntactically valid FSH:
```
* somePath #someCode "Some code"
```
This would result in an error from the importer. So, I feel like the situation in this PR is equivalent.

julianxcarter (approved):

PR #923, opened 2021-09-30
Catch and log errors in regression processing

ngfreiter:
Have you or a loved one been harmed by the regression script crashing when it can't load a repo? Then this PR is for you!

This gets rid of crashes when a regression fails, and instead logs an error and moves on to the next repo. It also logs the number of errors at the bottom once a regression is complete.

To test this, just run a regression on a repos file containing some fake repos. You should see errors logged, and the regression should continue, where on `master`, this will crash the script. You should test that everything still works on real repos as well.

### DIFF
--- a/regression/run-regression.ts
+++ b/regression/run-regression.ts
@@ -10,6 +10,7 @@ import extract from 'extract-zip';
 import opener from 'opener';
 import { isEqual, union } from 'lodash';
 import { createTwoFilesPatch } from 'diff';
+import chalk from 'chalk';
 
 // Track temporary files so they are deleted when the process exits
 temp.track();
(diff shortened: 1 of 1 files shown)


cmoesel (approved):
> Have you or a loved one been harmed by the regression script crashing when it can't load a repo?

Yes.  And all this time I thought no one cared.

The code changes make perfect sense and it works exactly as described.  Sign me up!

cmoesel:
@ngfreiter -- since this is a fairly straightforward change, and since it is not to production code, I'm feeling like one review is probably enough.  If you agree, we can merge it.  If you disagree, we'll let @julianxcarter do his thing.

ngfreiter:
> @ngfreiter -- since this is a fairly straightforward change, and since it is not to production code, I'm feeling like one review is probably enough. If you agree, we can merge it. If you disagree, we'll let @julianxcarter do his thing.

I feel comfortable just merging it!

PR #930, opened 2021-10-11
Try to validate usage of codes from local code systems

### LINKED ISSUE #595: Sushi should warn/error when an invalid code in a locally defined CodeSystem is referenced
Currently, if an invalid code from a locally defined CodeSystem is referenced, no warning/error is generated in both Profiles and Examples.   Should generate an error in both, IMHO. Or at least a warning.

mint-thompson:
Fixes #595 and completes task [CIMPL-829](https://standardhealthrecord.atlassian.net/browse/CIMPL-829).

CodeSystems may be defined by
- a `CodeSystem` entity
- an `Instance` entity that sets `InstanceOf` to `CodeSystem`

Codes from CodeSystems may be used by
- the assigned value in an `AssignmentRule`
- the concept list in a `ValueSetConceptComponentRule`
- the filter value in a `ValueSetFilterComponentRule`

So, for each combination of definition and usage, if the system is locally defined, make sure that the codes being used are present in the system. When any undefined codes are detected, log an error listing the undefined codes. The undefined codes are still allowed to be used, so the actual FHIR output doesn't change.

### DIFF
--- a/src/fhirtypes/common.ts
+++ b/src/fhirtypes/common.ts
@@ -340,11 +340,35 @@ export function listUndefinedLocalCodes(
 ): void {
   let undefinedCodes: string[] = [];
   applyInsertRules(codeSystem, tank);
-  if (codeSystem instanceof FshCodeSystem) {
+  // if the CodeSystem content is complete, a code not present in this system should be listed as undefined.
+  // if the CodeSystem content is not complete, then do not list any code as undefined.
+  // in a FshCodeSystem, content is complete by default, so make sure it isn't set to something else.
+  // in an Instance, content does not have a default value, so make sure there is a rule that sets it to complete.
+  if (
+    codeSystem instanceof FshCodeSystem &&
+    !codeSystem.rules.some(
+      rule =>
+        rule instanceof CaretValueRule &&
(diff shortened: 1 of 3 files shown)


src/fhirtypes/common.ts
```diff
@@ -325,6 +332,42 @@ function getRuleValue(rule: AssignmentRule | CaretValueRule): AssignmentValueTyp
...
+  sourceEntity: FshEntity
+): void {
+  let undefinedCodes: string[] = [];
+  applyInsertRules(codeSystem, tank);
+  if (codeSystem instanceof FshCodeSystem) {
+    undefinedCodes = codes.filter(code => {
+      return !codeSystem.rules.some(rule => rule instanceof ConceptRule && rule.code === code);
+    });
+  } else if (codeSystem instanceof Instance) {
+    const conceptRulePath = /^(concept(\[(\d+|\+|=)\])?\.)+code$/;
```
cmoesel, line 348: I don't remember the order of operations, but I assume that at this point we've already normalized out the indented rules so they are in the full path format, right?  Just want to be sure, as this would not work for indented rules otherwise.

PR #933, opened 2021-10-18
Update antlr4 to 4.9.2

mint-thompson:
Completes task [CIMPL-832](https://standardhealthrecord.atlassian.net/browse/CIMPL-832).

The updated antlr4 library generates classes with a different export style. Update the imports of those classes to use the default import rather than a destructured import.

The directory structure of the antlr4 package adds two new directories. Add package path mapping to the TypeScript compiler configuration and the Jest configuration.


cmoesel:
FYI - they've just released antlr4 4.9.3.

jafeltra:
When I tried to run this, I got this error `Error: Cannot find module 'antlr4/error'`, but I think that was part of the original issue, so I'm thinking I must be missing something. What version of Node are you running when you use this branch? I tried Node 14 and Node 16 and I tried wiping my `node_modules` and reinstalling. Do you have any other ideas I can try?

cmoesel:
I'm getting the same as @jafeltra.  The `npm install` works fine, as does `npm run check`, but when I try to run it against a FSH project, I get the `Cannot find module 'antlr4/error'` error.  I tried using `ts-node` as well as using the build `dist/app.js`.  Running Node 14.19.0.

mint-thompson:
It's looking like I hadn't been properly cleaning up my `node_modules` on this branch, and this may have always been broken. I'll get to fixing it now.

jafeltra:
This works for me now (yay!) but do you know why the `tsconfig-paths` part is needed? It seems similar to the `paths` property in `tsconfig.json`. Is the difference just runtime resolution?

mint-thompson:
@jafeltra my understanding is that it's just for runtime resolution, yes. Node doesn't use any information from `tsconfig.json` when it starts up, so its module resolution isn't informed by anything in there. I took a look at some information on [subpath imports](https://nodejs.org/api/packages.html#subpath-patterns) in Node, but this type of configuration doesn't seem to work when using typescript.

jafeltra:
Thanks @mint-thompson! It's a bummer that ts-config doesn't just do that by default, but this seems to fix the problem for me when I run SUSHI using `ts-node` or with the built code with plain `node`. I haven't noticed anything else, but I'm going to run the regression script just to be sure!

jafeltra (approved):
Yay! It works and I ran a full regression and there were no changes!

cmoesel (approved):
This looks great!  Thanks for figuring out all the hoops you need to jump through.  While jumping through hoops is not ideal, I'm glad that these are at least small hoops.  And jumping through them is better than tripping on them.

I also discovered a BONUS!  I'm 95% sure that at one point ANTLR 4.9.x did _not_ work with Node 12 at all.  BUT... I tested it out with Node 12, Node 14, and Node 16 and it worked well in all cases!  So now we don't have to force our users to upgrade Node.  That's great!

cmoesel (changes requested):
Wait.  Maybe not!  I just tried `npm pack`ing it and then using its API in another project and got:
```
internal/modules/cjs/loader.js:818
  throw err;
  ^

Error: Cannot find module 'antlr4/error'
```

So does every user for the fsh-sushi module need to jump through those same hoops too?

cmoesel:
Ah.  I figured out a solution.  We need to register the tsconfig-paths in `src/index.ts` just like we do in `src/app.ts`.

I added this to the top of SUSHI's `src/index.ts` and it worked w/ the API after that:
```
import { register } from 'tsconfig-paths';
register({
  baseUrl: __dirname,
  paths: {
    'antlr4/*': ['../node_modules/antlr4/src/antlr4/*']
  }
});
```

@mint-thompson -- if you agree with this approach, can you make this change?

mint-thompson:
@cmoesel thank you for finding that issue! I'll fix that right away.

cmoesel (approved):
Thanks, @mint-thompson.  I dig it!

jafeltra (approved):

PR #934, opened 2021-10-18
No shortcut syntax for manual slices of choice elements

### LINKED ISSUE #927: SUSHI creates incorrect id/path when manually slicing choice[x] elements
If you attempt to manually do a type slice like so:
```
Profile: Test
Parent: MedicationRequest
* medication[x] ^slicing.discriminator.type = #type
* medication[x] ^slicing.discriminator.path = "$this"
* medication[x] contains test 0..1 MS
* medication[x][test] only Reference
```
SUSHI will creating an element with `id` and `path` `MedicationRequest.test`.  It should be `id`: `MedicationRequest.medication[x]:test` and `path`: `MedicationRequest.medication[x]`.

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/value.5Bx.5D.20type.20based.20slicing

julianxcarter:
Fixes #927, making it so that SUSHI no longer applies the "shortcut" syntax to the diffs of choice elements when the slice is manually created. The example in [this FSHOnline link](https://fshschool.org/FSHOnline/#/share/3jdSWzi) should demonstrate the problem, as we do not want the `medication[x]` portion of the `id` and `path` to be chopped off when the slice is manually created. Running the example against this branch should result in the `id` being `MedicationRequest.medication[x]:test` and the `path` being `MedicationRequest.medication[x]`.

### DIFF
--- a/src/fhirtypes/ElementDefinition.ts
+++ b/src/fhirtypes/ElementDefinition.ts
@@ -40,7 +40,7 @@ import {
   TypeNotFoundError,
   ValueAlreadyAssignedError,
   ValueConflictsWithClosedSlicingError,
-  WideningCardinalityError,
+  ConstrainingCardinalityError,
   InvalidChoiceTypeRulePathError,
   CannotResolvePathError,
   MismatchedBindingTypeError,
@@ -663,7 +663,7 @@ export class ElementDefinition {
    * @param {number} min - the minimum cardinality
    * @param {number|string} max - the maximum cardinality
    * @throws {InvalidCardinalityError} when min > max
-   * @throws {WideningCardinalityError} when new cardinality is wider than existing cardinality
+   * @throws {ConstrainingCardinalityError} when new cardinality is wider than existing cardinality
    * @throws {InvalidSumOfSliceMinsError} when the mins of slice elements > max of sliced element
    * @throws {NarrowingRootCardinalityError} when the new cardinality on an element is narrower than
    *   the cardinality on a connected element
@@ -683,12 +683,12 @@ export class ElementDefinition {
 
     // Check to ensure min >= existing min
     if (this.min != null && min < this.min) {
-      throw new WideningCardinalityError(this.min, this.max, min, max);
+      throw new ConstrainingCardinalityError(this.min, this.max, min, max);
     }
 
     // Check to ensure max <= existing max
     if (this.max != null && this.max !== '*' && (maxInt > parseInt(this.max) || isUnbounded)) {
-      throw new WideningCardinalityError(this.min, this.max, min, max);
+      throw new ConstrainingCardinalityError(this.min, this.max, min, max);
     }
 
     // Sliced elements and slices have special card rules described here:
--- a/src/errors/ConstrainingCardinalityError.ts
+++ b/src/errors/ConstrainingCardinalityError.ts
@@ -1,6 +1,6 @@
 import { Annotated } from './Annotated';
 
-export class WideningCardinalityError extends Error implements Annotated {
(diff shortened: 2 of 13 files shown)


src/fhirtypes/ElementDefinition.ts
```diff
@@ -417,7 +417,9 @@ export class ElementDefinition {
       .split('.')
       .map(p => {
         const i = p.indexOf('[x]:');
-        return i > -1 ? p.slice(i + 4) : p;
+        const isChoiceSlice =
+          i > -1 ? this.type?.some(type => p.slice(i + 4).endsWith(upperFirst(type.code))) : false;
+        return i > -1 && isChoiceSlice ? p.slice(i + 4) : p;
```
ngfreiter, line 422: Isn't just `isChoiceSlice` sufficient here, since when `isChoiceSlice` is `true`, `i > -1` is also always true, and when `isChoiceSlice` is `false`, it doesn't matter if `i >  -1`, since we will just return `p` either way.

ngfreiter (changes requested):
I am not sure I really understand what this FSH in the issue is trying to do:
```
Profile: Test
Parent: MedicationRequest
* medication[x] ^slicing.discriminator.type = #type
* medication[x] ^slicing.discriminator.path = "$this"
* medication[x] contains test 0..1 MS
* medication[x][test] only Reference
```
To me it seems like trying to manually slice the `medication[x]` element is an error. The element only has max cardinality of 1, so it isn't an array, and if you are doing a choice slice, shouldn't the slice be named `medicationReference`? I can't find a place in the FHIR spec that explicitly says that the slice must be named `medicationReference`, but it doesn't seem like a good idea to break that rule. @cmoesel what do you think?

src/fhirtypes/ElementDefinition.ts
```diff
@@ -417,7 +417,9 @@ export class ElementDefinition {
       .split('.')
       .map(p => {
         const i = p.indexOf('[x]:');
-        return i > -1 ? p.slice(i + 4) : p;
+        const isChoiceSlice =
+          i > -1 ? this.type?.some(type => p.slice(i + 4).endsWith(upperFirst(type.code))) : false;
```
cmoesel, line 421: Since this iterates through the `id` split on `.`, I don't think this would work for deeper paths.  For example, something like `Observation.value[x]:valueQuantity.system` wouldn't work right when `p` is `value[x]:valueQuantity` because `type.code` would be `uri` (since `this.type` refers to the type of the element related to the _full_ id).

BUT... I'm a little worried about having to lookup the element for the intermediate id in this logic and check its type because that could have some effect on performance if its not done carefully.  So if that's the way you need to do it, please try to ensure that lookup only happens when absolutely needed.

For example, if it is `value[x]:fooBar`, we don't need to do a lookup because the slicename does not start with `value`, so it can't be the shortcut.

cmoesel (changes requested):
Thanks for the PR!  I found one likely issue w/ the logic related to choices in the middle of the element path (not at the tail end).  It's probably worth writing a test and looking into it.

I also ran `npm run regression` and got several changes that were not expected (and not related to intermediate paths as far as I can tell).  Here is one example (but all the others are like it):
image

The above seems over-agressive, as I would expect that element to keep the shortcut syntax.

cmoesel:
@julianxcarter -- there is a current discussion on Zulip that may affect this issue.  In short, we may have to remove choice type renaming completely.  Stay tuned and and see: https://chat.fhir.org/#narrow/stream/179177-conformance/topic/Choice.20Type.20Renaming/near/258186065

PR #935, opened 2021-10-18
Add validation of slicing on SDs

### LINKED ISSUE #885: SUSHI should emit an error for missing slicing information
When slicing, the `slicing` object requires the following fields:
* `slicing.discriminator.type` (for each defined `slicing.discriminator`)
* `slicing.discriminator.path` (for each defined `slicing.discriminator`)
* `slicing.rules`

While SUSHI does not attempt to validate everything, this construct is used often enough that SUSHI should probably validate it and warn on missing required properties or invalid values (unknown code assigned to `slicing.discriminator.type` or `slicing.discriminator.rules`).

E.g., if an author has this:
```
* identifier ^slicing.discriminator.type = #value
* identifier ^slicing.discriminator.path = "type.coding.code"
* identifier contains
    AccessionIdentifier 0..1 MS
```
Then:
> Error: Slicing on identifier does not set the required ^slicing.rules property.

(Or something like that).

ngfreiter:
## Description

Fixes #885 by logging errors when the `slicing` element on an `ElementDefinition` is invalid.

The specific checks are:
- The `slicing` element can only be on an array, a former array, or a choice.
- The `slicing.rules` element is defined and set to one of the allowed values.
- If a `slicing.discriminator` is defined, then for each `discriminator`:
  - `type` is defined and set to one of the allowed values.
  - `path` is defined.

Note that some tests also had to be updated, since they were creating invalid slices. Also in some of these cases, it could be possible to do this checking on the rules themselves, instead of on the SD. However, I had trouble figuring out how to do this in a well-defined way, since we can reassign caret rules, and values for `slicing` can be inherited from parent elements (so we can't tell at the time of reading rules if the required elements are present). The SD approach ended up being simpler and allowing all the slicing validation to be together, which I preferred.
 
## Testing

It would be good to test:
- A `slicing` element can only be placed on elements where you expect it, otherwise an error is logged.
- If `slicing.rules` is invalid, an error is logged.
- If `slicing.rules` is missing, an error is logged.
- `slicing.rules` can be set to the valid values with no error
- If `slicing.discriminator.type` is missing, an error is logged.
- If `slicing.discriminator.type` is invalid, an error is logged.
- `slicing.discriminator.type` can be set to the valid values
- If `slicing.discriminator.path` is missing, an error is logged.
- No badness happens with a regression (I ran one and got nothing, but paranoia leads me to ask that at least one reviewer does one)

### DIFF
--- a/src/fhirdefs/load.ts
+++ b/src/fhirdefs/load.ts
@@ -234,6 +234,11 @@ export function loadCustomResources(resourceDir: string, defs: FHIRDefinitions):
             continue;
           }
         } catch (e) {
+          if (e.message.startsWith('Unknown resource type:')) {
+            // Skip unknown FHIR resource types. When we have instances of Logical Models,
+            // the resourceType will not be recognized as a known FHIR resourceType, but that's okay.
+            continue;
+          }
           logger.error(`Loading ${file} failed with the following error:\n${e.message}`);
           continue;
         }
--- a/src/fhirtypes/ImplementationGuide.ts
+++ b/src/fhirtypes/ImplementationGuide.ts
@@ -1,5 +1,11 @@
 import { ContactDetail, UsageContext } from './metaDataTypes';
-import { CodeableConcept, Reference, BackboneElement, DomainResource } from './dataTypes';
+import {
+  BackboneElement,
+  CodeableConcept,
+  DomainResource,
+  Extension,
+  Reference
+} from './dataTypes';
 
 export type ImplementationGuide = DomainResource & {
   resourceType: 'ImplementationGuide';
@@ -60,6 +66,7 @@ export type ImplementationGuideDefinitionResource = {
   exampleBoolean?: boolean;
   exampleCanonical?: string;
   groupingId?: string;
+  extension?: Extension[];
 };
 
 export type ImplementationGuideDefinitionPage = {
--- a/src/ig/IGExporter.ts
+++ b/src/ig/IGExporter.ts
@@ -846,7 +846,24 @@ export class IGExporter {
               resource => resource.reference?.reference == referenceKey
             );
 
-            if (configResource?.omit !== true) {
+            // For predefined examples of Logical Models, the user must provide an entry in config
+            // that specifies the reference as Binary/[id], the extension that specifies the resource format,
+            // and the exampleCanonical that references the LogicalModel the resource is an example of.
+            // In that case, we do not want to add our own entry for the predefined resource - we just
+            // want to use the resource entry from the sushi-config.yaml
+            const hasBinaryExampleReference = (this.config.resources ?? []).some(
+              resource =>
+                resource.reference?.reference === `Binary/${resourceJSON.id}` &&
+                resource.exampleCanonical ===
+                  `${this.config.canonical}/StructureDefinition/${resourceJSON.resourceType}` &&
+                resource.extension?.some(
+                  e =>
+                    e.url ===
+                    'http://hl7.org/fhir/StructureDefinition/implementationguide-resource-format'
+                )
+            );
+
+            if (configResource?.omit !== true && !hasBinaryExampleReference) {
               const existingIndex = this.ig.definition.resource.findIndex(
                 r => r.reference.reference === referenceKey
               );
--- a/test/fhirdefs/fixtures/customized-ig-with-resources/input/examples/Binary-LogicalModelExample.json
+++ b/test/fhirdefs/fixtures/customized-ig-with-resources/input/examples/Binary-LogicalModelExample.json
@@ -1,0 +1,4 @@
+{
+  "resourceType": "CustomLogicalModel",
+  "id": "example-logical-model"
+}
--- a/test/fhirdefs/fixtures/customized-ig-with-resources/input/examples/Binary-LogicalModelExample.xml
+++ b/test/fhirdefs/fixtures/customized-ig-with-resources/input/examples/Binary-LogicalModelExample.xml
@@ -1,0 +1,3 @@
+<CustomLogicalModel>
+  <id value="custom-logical-model" />
+</CustomLogicalModel>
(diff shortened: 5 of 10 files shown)


src/fhirtypes/ElementDefinition.ts
```diff
@@ -2433,11 +2486,19 @@ export type ElementDefinitionSlicing = {
...
+const ALLOWED_SLICING_RULES = ['closed', 'open', 'openAtEnd'];
+
 export type ElementDefinitionSlicingDiscriminator = {
   type: string;
   path: string;
 };
 
+// Cannot constrain ElementDefinitionSlicingDiscriminator to have these values as a type
+// since we want to process other string values, but log an error
+const ALLOWED_DISCRIMINATOR_PATHS = ['value', 'exists', 'pattern', 'type', 'profile'];
```
cmoesel, line 2500: Shouldn't this be `ALLOWED_DISCRIMINATOR_TYPES`?

src/fhirtypes/StructureDefinition.ts
```diff
@@ -82,6 +83,29 @@ export class StructureDefinition {
    */
   private _sdStructureDefinition: StructureDefinition;
 
+  private _validationErrors: ValidationError[];
+
+  get validationErrors(): ValidationError[] {
+    return this._validationErrors;
+  }
```
cmoesel, line 90: I have the same thoughts/concerns as I noted in the `ElementDefinition` implementation.

src/fhirtypes/StructureDefinition.ts
```diff
@@ -82,6 +83,29 @@ export class StructureDefinition {
...
+  valid(): boolean {
+    this._validationErrors = [];
+    this.elements
+      ?.filter(e => !e.valid())
+      .forEach(e => {
+        e.validationErrors?.forEach(err => {
+          this._validationErrors.push(
+            new ValidationError(
+              err.issue,
+              [`StructureDefinition.snapshot.where(id = '${e.id}').first()`, err.fhirPath].join('.')
```
cmoesel, line 101: Fancy.  You've created a real FHIRPath expression that resolves to the element of interest in the SD. While I find that neat, I am also a total nerd.  I wonder if our average user will find this useful or if they'll find it cryptic? 

Another approach might be to convert the id to a FSH path -- which I think could be done fairly easily by using the `diffId()` and then replacing `:sliceName` with `[sliceName]`.  Then you could do `^${err.fhirPath}` too -- so it looks like a full FSH path...

It's just an idea.  Again, my main concern is that many casual profilers may not be very familiar with FHIRPath (even if they _should_ be).  What do you think?

ngfreiter: Yeah FSH path makes sense, I'll make the change.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -275,6 +276,58 @@ export class ElementDefinition {
...
+    this._validationErrors = [];
+    if (this.slicing) {
+      this.validateSlicing(this.slicing);
+    }
+    return !this._validationErrors.length;
+  }
+
+  private validateRequired(value: AssignmentValueType, fhirPath: string): void {
+    if (!value) {
+      this._validationErrors.push(new ValidationError('Missing required element', fhirPath));
```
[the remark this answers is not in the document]
ngfreiter: Remedied in c677fcf.

cmoesel (approved):
It's good.  I appreciate your use of interesting verbs when confirming each fix in the comments.  Well done.

src/fhirtypes/StructureDefinition.ts
```diff
@@ -82,6 +83,21 @@ export class StructureDefinition {
...
   private _sdStructureDefinition: StructureDefinition;
 
+  validate(): ValidationError[] {
+    const validationErrors: ValidationError[] = [];
+    this.elements.forEach(e => {
+      e.validate().forEach(err => {
+        validationErrors.push(
+          new ValidationError(
+            err.issue,
+            `${e.diffId().replace(/(\S):(\S+)/g, (_, c1, c2) => `${c1}[${c2}]`)} ^${err.fshPath}`
```
mint-thompson, line 93: Somehow I've never seen this style of `replace` call before, but it's pretty nice for quick simple regexes with capture groups.

test/export/InstanceExporter.test.ts
```diff
@@ -2945,13 +2953,21 @@ describe('InstanceExporter', () => {
       // Parent: Patient
       // * identifier ^slicing.discriminator.type = #value
       // * identifier ^slicing.discriminator.path = "value"
+      // * identifier ^slicing.rules = #open
       // * identifier contains ParentSlice 1..1
       // * identifier[ParentSlice] ^slicing.discriminator.type = #value
       // * identifier[ParentSlice] ^slicing.discriminator.path = "value"
+      // * identifier ^slicing.rules = #open
```
mint-thompson, line 2960: Since this is with the other `ParentSlice` slicing rules, I think this comment is supposed to be:
```suggestion
      // * identifier[ParentSlice] ^slicing.rules = #open
```

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -138,6 +143,31 @@ describe('StructureDefinitionExporter R4', () => {
       expect(loggerSpy.getLastMessage('warn')).toMatch(warning);
       expect(loggerSpy.getLastMessage('warn')).toMatch(/File: Wrong\.fsh.*Line: 2 - 5\D*/s);
     });
+
+    it('should log error messages for validation errros on the StructureDefinition', () => {
```
mint-thompson, line 147: Minor typo here:
```suggestion
    it('should log error messages for validation errors on the StructureDefinition', () => {
```

mint-thompson (changes requested):
I really like your implementation here. I don't have any functional changes to request, just a few typo-level findings.

ngfreiter:
@mint-thompson, thanks for finding those typos, applied your suggestions in https://github.com/FHIR/sushi/pull/935/commits/0f8a7a779e6569bc996055c63015f959845ff581!

Issue #844, opened 2021-06-16
Consider configurable behavior for instance generation


cmoesel:
Authors have indicated that in some cases they:
* don't want generated instances to have the `id` value serialized
* don't want generated instances of profiles to automatically put the profile in `meta.profile`

We should consider making things like this configurable in SUSHI.

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Handling.20meta.2Eprofile.20in.20instances/near/242212579

citizenrich:
I'll add one here, it seems the resource type is always prepended to the filename, e.g. Patient-123 if it is Patient/123.

PR #937, opened 2021-10-20
Allow configurable Instance generation

ngfreiter:
## Description

Fixes https://github.com/FHIR/sushi/issues/844 by adding configurable instance generation.

This PR does this by adding a `instanceOptions` configuration property, which has `setMetaProfile` and `setId` as children. The `setMetaProfile` value controls whether or not `meta.profile` is set on generated instances. It can have the following values:
  - always: Set meta.profile for all Instances of profiles (default)
  - never: Do not set meta.profile on any Instances
  - inline-only: Set meta.profile for only Instances of profiles with Usage set to #inline
  - standalone-only: Set meta.profile for only Instances of profiles where Usage is any value other than #inline
 
The `setId` value controls whether or not `id` is set on generated instances. It can have the following values:
  - always: Set id for all Instances (default)
  - standalone-only: Set id for only Instances where Usage is any value other than #inline
The `setId` flag can't have an `inline-only` or `never` option, because `id`s are necessary when generating non-inline instances. The publisher depends on them.

## Testing
Test:
- `meta.profile` is set on inline and non-inline instances when `setMetaProfile` is not set
- `meta.profile` is set on the expected instances (inline vs non-inline) for each of the four values
- `id` is set on inline and non-inline instances when `setId` is not set
- `id` is set on the expected instances (inline vs non-inline) for the two values


ngfreiter:
I will have to document this on FSHSchool, and I plan to, just haven't had a chance yet.

PR #941, opened 2021-10-21
Support User Provided Examples of Logical Models

jafeltra:
This PR makes a couple minor changes in order to support user provided instances of logical models. Right now, SUSHI does not do any additional processing of these examples and we don't yet support authoring examples of logical models in FSH, but SUSHI will no longer crash if the examples are provided and it properly builds an IG resource with the examples.

In order to provide an example of a logical model, you can author a logical model in FSH like usual, and you can provide either a JSON or XML version of an example of that logical model in either the `input/resources` or `input/examples` folder with the name `Binary-{id}.json`. You also need to provide additional configuration for that resource in the `resources` property of `sushi-config.yaml`. You need to specify `name` for the resource so the IG renders the name when it renders the link to the example, the `exampleCanonical` should reference the Logical Model that the example is an instance of so the IG publisher knows how to link it, and you need to add an `extension` to specify the format of the resource. The config entry I've been using to test looks like this:

```
resources:
  Binary/my-id:
    extension:
      - url: http://hl7.org/fhir/StructureDefinition/implementationguide-resource-format
        valueCode: application/fhir+json
    name: Example of LM
    exampleCanonical: http://example.org/StructureDefinition/MyLM
```

I've also been testing with a very simple Logical Model and very simple example:

```
// In any FSH file
Logical: MyLM
Id: MyLM
Title: "My LM"
Description: "This is mine"
* important 1..1 SU boolean "Is this resource important"
```

and 

```
// In a file called Binary-my-id.json
{
  "resourceType": "MyLM",
  "id": "my-id",
  "important": true
}
```

If you run SUSHI and the publisher, you should see an Examples tab on the Logical Model page, which should link to the 
`my-id` example. The example's page should should the JSON or XML representation on the Narrative Content tab and should show a Binary with base64 encoding on the JSON/XML tabs.

You should also be able to replace the Binary-my-id.json with the XML version of this file, change the `valueCode` from` application/fhir+json` to `application/xml` and everything should work the same. Most of these steps were from this [Zulip thread](https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/Logical.20Model.20Examples), and the [sample-ig](https://github.com/FHIR/sample-ig) has an example [here](https://github.com/FHIR/sample-ig/blob/master/input/examples/Binary-logical-example.xml).

### DIFF
--- a/src/export/ValueSetExporter.ts
+++ b/src/export/ValueSetExporter.ts
@@ -5,7 +5,7 @@ import {
   StructureDefinition
 } from '../fhirtypes';
 import { FSHTank } from '../import/FSHTank';
-import { FshValueSet, FshCode, ValueSetFilterValue } from '../fshtypes';
+import { FshValueSet, FshCode, ValueSetFilterValue, FshCodeSystem, Instance } from '../fshtypes';
 import { logger } from '../utils/FSHLogger';
 import { ValueSetComposeError, InvalidUriError } from '../errors';
 import { Package } from '.';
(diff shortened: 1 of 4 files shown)


mint-thompson (changes requested):
This looks good! Could you add a test case showing what happens when there is an example of a logical model that is not properly declared in the configuration's resource list?

PR #944, opened 2021-10-25
Add additional details to SUSHI --init sushi-config.yaml

ngfreiter:
Update the `sushi-config.yaml` file included when running `sushi --init` to include:
* Include references to the value sets that relevant properties can take on
* Additional fields and documentation on those fields (mostly taken from the [exhaustive example](https://fshschool.org/docs/sushi/configuration/#exhaustive-example))

A few things to note:
* I moved the `menu.xml` section above the additional IG fields, which is logically a little annoying to me, but I think the benefit of having all of the uncommented fields at the top is worth it
* If reviewers think this is too busy, some of the commented fields can be removed. My personal preference is to give the user all the fields they may need, and then let them remove what they don't, instead of requiring them to go to the documentation to see what they may be missing.

### DIFF
--- a/src/errors/ValueSetFilterValueTypeError.ts
+++ b/src/errors/ValueSetFilterValueTypeError.ts
@@ -1,12 +1,25 @@
 import { VsOperator } from '../fshtypes';
 
+type expectedType = 'string' | 'regex' | 'boolean' | 'code';
 export class ValueSetFilterValueTypeError extends Error {
   constructor(
     public operator: VsOperator,
-    public expectedType: 'string' | 'regex' | 'boolean' | 'code'
+    public expectedTypes: expectedType[],
+    public allowedStrings: string[] = []
   ) {
     super(
-      `Not applying value set filter: operator "${operator}" requires a ${expectedType} value.`
+      `Not applying value set filter: operator "${operator}" requires ${getExpectedTypes(
+        expectedTypes,
+        allowedStrings
+      )}.`
     );
   }
+}
+
+function getExpectedTypes(types: expectedType[], allowedStrings: string[]): string {
+  const formattedAllowedStrings = allowedStrings.map(s => `"${s}"`).join(' or ');
+  const allowedStringsClause = formattedAllowedStrings
+    ? ` or a string value of ${formattedAllowedStrings}`
+    : '';
+  return `a ${types.join(' or ')} value${allowedStringsClause}`;
 }
--- a/src/export/MappingExporter.ts
+++ b/src/export/MappingExporter.ts
@@ -70,7 +70,9 @@ export class MappingExporter {
     const sourceStructDef = this.pkg.fish(
       fshDefinition.source,
       Type.Profile,
-      Type.Extension
+      Type.Extension,
+      Type.Resource,
+      Type.Logical
     ) as StructureDefinition;
     if (sourceStructDef) {
       const parent = this.fisher.fishForFHIR(
@@ -78,7 +80,8 @@ export class MappingExporter {
         Type.Resource,
         Type.Type,
         Type.Profile,
-        Type.Extension
+        Type.Extension,
+        Type.Logical
       );
       const matchingParentMapping = parent?.mapping.find(
         (m: StructureDefinitionMapping) => m.identity === fshDefinition.id
--- a/src/import/FSHImporter.ts
+++ b/src/import/FSHImporter.ts
@@ -1969,42 +1969,35 @@ export class FSHImporter extends FSHVisitor {
       throw new ValueSetFilterMissingValueError(operator);
     }
     const value = ctx.vsFilterValue() ? this.visitVsFilterValue(ctx.vsFilterValue()) : true;
+
+    // NOTE: We support string value for every operator, in addition to the specific typed values
+    // for some operators based on the filter value documentation:
+    // http://hl7.org/fhir/R4/valueset-definitions.html#ValueSet.compose.include.filter.value
+    // and the discussion on https://github.com/FHIR/sushi/issues/936
     switch (operator) {
       case VsOperator.EQUALS:
       case VsOperator.IN:
       case VsOperator.NOT_IN:
-        // NOTE: We believe that =, in, and not-in operators should ONLY support code values
-        // based on the filter value documentation:
-        // http://hl7.org/fhir/R4/valueset-definitions.html#ValueSet.compose.include.filter.value
-        // Both string and code are supported for now in order to maintain backwards compatibility.
-        if (typeof value === 'string') {
-          logger.warn(
-            `The match value of filter operator "${operator}" must be a code. ` +
-              'For string valued properties, use the regex filter operator with a regular expression value. ' +
-              'Support for using strings as filter values has been deprecated and will be removed in a future release.',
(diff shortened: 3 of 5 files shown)


src/utils/init-project/sushi-config.yaml
```diff
@@ -1,22 +1,195 @@
...
+releaseLabel: ci-build # ci-build | draft | qa-preview | ballot | trial-use | release | update | normative+trial-use
+# license: CC0-1.0 # https://www.hl7.org/fhir/valueset-spdx-license.html
+# jurisdiction: urn:iso:std:iso:3166#US "United States of America" # https://www.hl7.org/fhir/valueset-jurisdiction.html
 publisher:
   name: Example Publisher
   url: http://example.org/example-publisher
 
+# The dependencies property corresponds to IG.dependsOn. The key is the
+# package id and the value is the version (or dev/current). For advanced
+# use cases, the value can be an object with keys for `uri` and `version`.
```
jafeltra, line 24: This is a minor discrepency between the exhaustive example on FSH School and the example config in the SUSHI repo, but the value can be an object with keys for `uri`, `version`, and `id`. The SUSHI example config lists all three in the comment above. Do you think we should include `id` here as well? It's pretty clearly commented out below, so it probably doesn't matter much.

src/utils/init-project/sushi-config.yaml
```diff
@@ -1,22 +1,195 @@
...
+# than a list of code/value pairs (as in the ImplementationGuide
+# resource), the code is the YAML key. If a parameter allows repeating
+# values, the value in the YAML should be a sequence/array. For a
+# partial list of allowed parameters see:
+# https://confluence.hl7.org/display/FHIR/Implementation+Guide+Parameters
+#
+# parameters:
+#   excludettl: true
+#   validation: [allow-any-extensions, no-broken-links]
+
```
jafeltra, line 65: I don't think this matters much, but I don't get a blank line between the commented out `parameters` and the `menu.xml` box. Should I?

src/utils/init-project/sushi-config.yaml
```diff
@@ -1,22 +1,195 @@
...
 # ╭────────────────────────────────────────────menu.xml────────────────────────────────────────────╮
 # │  To use a provided input/includes/menu.xml file, delete the "menu" property below.             │
 # ╰────────────────────────────────────────────────────────────────────────────────────────────────╯
 menu:
```
jafeltra, line 69: Now that we're adding more documentation to each property, do you think we should include the blurb for menu that's in the exhaustive config?

jafeltra (commented):
This looks really good! I just had a couple minor comments, but nothing major. I think this is a really helpful starter config file now!

Edit: Also, I agree with you that we should just give the all the properties that are possible and they can remove things they don't want. I feel like if we give them something in between this and what we had before, then it might look like we have shown them everything that is possible, when in reality, there is more possible, but it's just listed in the documentation. Once they create their first few projects/config files, they may also stop using the --init option and will know to add the fields they want.

src/utils/init-project/sushi-config.yaml
```diff
@@ -1,22 +1,195 @@
-# ╭──────────────────────────────────────ImplementationGuide───────────────────────────────────────╮
-# │  The properties below are used to create the ImplementationGuide resource. For a list of       │
-# │  supported properties, see: https://fshschool.org/docs/sushi/configuration/                    │
+# ╭─────────────────────────Commonly Used ImplementationGuide Properties───────────────────────────╮
+# │  The properties below are used to create the ImplementationGuide resource. The most commonly   │
+# │  used properties are included. For a list of all supported properties and their functions,     │
+# │  see: https://fshschool.org/docs/sushi/configuration/.                                         │
 # ╰────────────────────────────────────────────────────────────────────────────────────────────────╯
 id: fhir.example
 canonical: http://example.org
```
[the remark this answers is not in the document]
ngfreiter: The `url` should not really be necessary assuming that most people want to just use the `canonical` to construct the `url`. I should've called it out, but in this one case, I thought it would be confusing to even include that property.

src/utils/init-project/sushi-config.yaml
```diff
@@ -1,22 +1,195 @@
...
-fhirVersion: 4.0.1
+fhirVersion: 4.0.1 # https://www.hl7.org/fhir/valueset-FHIR-version.html
 copyrightYear: 2020+
-releaseLabel: ci-build
+releaseLabel: ci-build # ci-build | draft | qa-preview | ballot | trial-use | release | update | normative+trial-use
+# license: CC0-1.0 # https://www.hl7.org/fhir/valueset-spdx-license.html
+# jurisdiction: urn:iso:std:iso:3166#US "United States of America" # https://www.hl7.org/fhir/valueset-jurisdiction.html
 publisher:
   name: Example Publisher
   url: http://example.org/example-publisher
```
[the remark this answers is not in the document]
ngfreiter: I'll add it commented yeah!

src/utils/init-project/sushi-config.yaml
ngfreiter: 🤷

src/utils/init-project/sushi-config.yaml
ngfreiter: Blurb in https://github.com/FHIR/sushi/pull/944/commits/648a5d0cf98d797958a3096f1e93d52720b262aa

src/utils/init-project/sushi-config.yaml
ngfreiter: Spaced out in https://github.com/FHIR/sushi/pull/944/commits/648a5d0cf98d797958a3096f1e93d52720b262aa

src/utils/init-project/sushi-config.yaml
ngfreiter: Un-discrepancied in https://github.com/FHIR/sushi/pull/944/commits/648a5d0cf98d797958a3096f1e93d52720b262aa

src/utils/init-project/sushi-config.yaml
```diff
@@ -1,22 +1,195 @@
...
+#
+# Those who need more control or want to add additional details to the contact values can use
+# contact directly and follow the format outlined in the ImplementationGuide resource and
+# ContactDetail.
+#
+# contact:
+#   - name: Bob Smith
+#     telecom:
+#         # system from: phone | fax | email | pager | url | sms | other
+#       - system: email
```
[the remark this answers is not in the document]
ngfreiter: Re-aligned in https://github.com/FHIR/sushi/pull/944/commits/648a5d0cf98d797958a3096f1e93d52720b262aa

src/utils/init-project/sushi-config.yaml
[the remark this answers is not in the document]
ngfreiter: Commented in https://github.com/FHIR/sushi/pull/944/commits/648a5d0cf98d797958a3096f1e93d52720b262aa

jafeltra (approved):
Looks great! Thanks for making those changes!

cmoesel (approved):
Thanks, @ngfreiter!  I think our users will find this to be very helpful!

PR #945, opened 2021-10-25
Update allowed VS Filter Operator values

### LINKED ISSUE #936: Warning about string-typed filter values seems problematic
There are a number of standard CodeSystem filters defined in the FHIR spec, such as https://hl7.org/fhir/snomedct.html#4.3.1.0.8.3, but when I use these in my FSH files to define ValueSets, I'm getting the very concerning warning:

> The match value of filter operator "=" must be a code. For string valued properties, use the regex filter operator with a regular expression value. Support for using strings as filter values has been deprecated and will be removed in a future release.

Do I need to worry that these standard filters are going to be unusable in some upcoming version of Sushi?

jafeltra:
This PR fixes #936 by updating the types we allow for each VS filter operation. [This comment](https://github.com/FHIR/sushi/issues/936#issuecomment-947826886) describes what we now support. The changes to the code include:

- Support for string values on every operator, which also removes the deprecation warning since we now support it 😁 
- Special checking that the string value provided to the `exist` operator is either `"true"` or `"false"`
- Updating the `ValueSetFilterValueTypeError` to support allowing multiple types of values and specific string values
- Tests for the new error cases and new allowed cases


mint-thompson (approved):
outstanding work. hooray for allowed data type flexibility!

PR #949, opened 2021-11-03
Find Instances of Profile, Extension, CodeSystem, and ValueSet when fishing

mint-thompson:
Completes task [CIMPL-837](https://standardhealthrecord.atlassian.net/browse/CIMPL-837).

When fishing in `FSHTank` or `Package`, an `Instance` or `InstanceDefinition` may be returned if a definitional instance of the entity exists.
- Profiles have `InstanceOf` "StructureDefinition" and `derivation` "constraint", and a `type` that is not "Extension".
- Extensions have `InstanceOf` "StructureDefinition", `derivation` "constraint", and `type` "Extension" OR `InstanceOf` "Extension" and `usage` "inline".
- CodeSystems have `InstanceOf` "CodeSystem".
- ValueSets have `InstanceOf` "ValueSet".

The special case of allowing inline instances for Extension is because Extension is an Element type, not a Resource type, so SUSHI gets unhappy when a definitional instance of `Extension` is created. Inline instances are not allowed for the other types.

Adding these new possible return types didn't change too much logic, but did require a few updates in `StructureDefinitionExporter`.


src/import/FSHTank.ts
```diff
@@ -171,6 +172,29 @@ export class FSHTank implements Fishable {
...
           );
+          if (!result) {
+            result = this.getAllInstances().find(
+              profileInstance =>
+                profileInstance.instanceOf === 'StructureDefinition' &&
+                profileInstance.usage === 'Definition' &&
+                (profileInstance.name === item ||
+                  profileInstance.id === item ||
+                  getUrlFromFshDefinition(profileInstance, this.config.canonical) === item) &&
+                profileInstance.rules.some(
```
ngfreiter, line 183: Should this also require that the `kind` is set to `resource`? Or is that requiring too much? Or could this also end up pulling in instances of datatypes if we don't check `kind`?

src/import/FSHTank.ts
```diff
@@ -179,6 +203,41 @@ export class FSHTank implements Fishable {
...
+                ) &&
+                extensionInstance.rules.some(
+                  rule =>
+                    rule instanceof AssignmentRule &&
+                    rule.path === 'type' &&
+                    rule.value === 'Extension'
+                )
+            );
+          }
+          if (!result) {
```
ngfreiter, line 230: I remember talking about this and it making more sense at the time, but now I am getting confused. We are searching for an `Extension`, which is to say, a `StructureDefinition` of type `Extension`. I don't think we would want to return something which is an `InstanceOf` `Extension`, since that is not going to be an SD.

src/import/FSHTank.ts
```diff
@@ -179,6 +203,41 @@ export class FSHTank implements Fishable {
...
+              extensionInstance =>
+                (extensionInstance.name === item ||
+                  extensionInstance.id === item ||
+                  getUrlFromFshDefinition(extensionInstance, this.config.canonical) === item) &&
+                extensionInstance.instanceOf === 'Extension' &&
+                extensionInstance.usage === 'Inline'
+            );
+          }
           break;
         case Type.Logical:
```
ngfreiter, line 242: Why exclude `Logical` and `Resource` here? In theory they would work similarly to `Profile` and `Extension` right?

ngfreiter (changes requested):
Sorry for the delayed review, but I had a few questions below, and a more major question here.

I get why we want this for `CodeSystem` and `ValueSet`, because I think we have seen those cases where someone defines those using an `Instance` and needs to reference them. Although I think that case is somewhat less relevant now that it is easier to assign values on a `CodeSystem`. But I'm wondering in what situations we need this type of support for `Profile` and `Extension`. I can't think of any cases, so is this just something we are doing more proactively? Also asking @cmoesel since he created this task I believe.

src/import/FSHTank.ts
mint-thompson: I specifically avoided that check, since this may be a Profile of `complex-type` such as Quantity.

src/import/FSHTank.ts
mint-thompson: I think I'm in the same situation as you: it made sense before, but thinking about it again, I don't think we want `InstanceOf` `Extension`. That instance would be a specific element, not a definition of the rules for an extension. I'll change this implementation accordingly.

src/import/FSHTank.ts
mint-thompson: I think they would work similarly, yes. I didn't include them because they weren't listed in the task description. I think we're in the process of permitting `InstanceOf` a given `Logical`... but that's different than being an `InstanceOf` `StructureDefinition`, with a similar sort of reasoning as for my incorrect implementation for handling `InstanceOf` `Extension`. So I'll defer to @cmoesel on this, since he created the task.

src/export/Package.ts
```diff
@@ -35,9 +35,29 @@ export class Package implements Fishable {
...
         case Type.Profile:
           def = this.profiles.find(p => p.id === item || p.name === item || p.url === item);
+          if (!def) {
+            def = this.instances.find(
+              i =>
+                (i.id === item || i._instanceMeta.name === item || i.url === item) &&
+                i._instanceMeta.usage === 'Definition' &&
+                i.resourceType === 'StructureDefinition' &&
+                i.derivation === 'constraint' &&
+                i.type !== 'Extension'
```
cmoesel, line 45: I don't know how much of a difference this will make in reality, but... since fishing is fairly common, I think it's best to make the find function as efficient as possible.  To that end, do you think it makes sense to rearrange the conditional clause so that the rarest bits are checked _first_?  For example, in the case above, I might check `i.derivation === 'constraint'` first under the assumption that most instances are _not_ profiles or extensions, so we bail out quickly rather than checking 3 separate fields for the id first.

That said, ids are generally unique, so I can also see why you might want to check those first.  So I'm not 100% positive about what is best -- but I _think_ that by putting derivation first, you'll only need one equality check in most cases (vs 3 equality checks).

src/export/Package.ts
```diff
@@ -35,9 +35,29 @@ export class Package implements Fishable {
...
         case Type.Extension:
           def = this.extensions.find(e => e.id === item || e.name === item || e.url === item);
+          if (!def) {
+            def = this.instances.find(
+              i =>
+                (i.id === item || i._instanceMeta.name === item || i.url === item) &&
+                i._instanceMeta.usage === 'Definition' &&
+                i.resourceType === 'StructureDefinition' &&
+                i.derivation === 'constraint' &&
+                i.type === 'Extension'
```
cmoesel, line 58: Similar comment as above.  It might be more efficient to check `i.derivation === 'constraint' && i.type === 'Extension'` first, since this seems it will eliminate 99% of cases right away.

src/export/Package.ts
```diff
@@ -47,9 +67,25 @@ export class Package implements Fishable {
           break;
         case Type.ValueSet:
           def = this.valueSets.find(vs => vs.id === item || vs.name === item || vs.url === item);
+          if (!def) {
+            def = this.instances.find(
+              i =>
+                (i.id === item || i._instanceMeta.name === item || i.url === item) &&
+                i._instanceMeta.usage === 'Definition' &&
+                i.resourceType === 'ValueSet'
```
cmoesel, line 75: Again, under the assumption that _most_ instances are not ValueSets, it might be more efficient to check `i.resourceType === 'VaueSet'` first.

src/export/Package.ts
```diff
@@ -47,9 +67,25 @@ export class Package implements Fishable {
...
+          }
           break;
         case Type.CodeSystem:
           def = this.codeSystems.find(cs => cs.id === item || cs.name === item || cs.url === item);
+          if (!def) {
+            def = this.instances.find(
+              i =>
+                (i.id === item || i._instanceMeta.name === item || i.url === item) &&
+                i._instanceMeta.usage === 'Definition' &&
+                i.resourceType === 'CodeSystem'
```
cmoesel, line 86: And under the assumption that _most_ instances are not CodeSystems, it might be more efficient to check `i.resourceType === 'CodeSystem'` first.

src/export/Package.ts
cmoesel: ... but this may be so trivial compared to all the other things that happen that maybe it doesn't matter.  I don't know.  If you think it doesn't make sense to change this (and the similar cases below) then I won't argue with you.

src/import/FSHTank.ts
cmoesel: I think this was an oversight on my part.  It makes sense to support instances of logicals and resources too.  I don't know how often it will come up, but if it does, we should support it.  It seems it would be inconsistent not to support it.

cmoesel (changes requested):
Nice work, @mint-thompson.  To be honest, I thought this might be more painful than it turned out to be.  One thing I thought might be a problem was cases where we fish the tank for something specific (e.g., a `Profile`) and _expect_ a `Profile` class instance to be returned.  If we now return an `Instance` where a `Profile` is expected, that could be a problem.  Did you confirm that all _usages_ of `fish(...)` are correctly handling whatever we give back to them?

As for @ngfreiter's question about supporting profiles and extensions as instances, I'd like to support this for consistency.  I don't know _why_ someone might choose to implement one using `Instance:`, but if they do, it seems we should support it (since we do so for `ValueSet` and `CodeSystem`).  If it turned out that it added lots of complexity, then I'd consider not doing it.  But if it really is as easy as what we see in this PR, then I think it's worth it.  I also think it's worth supporting for `Logical` and `Resource` -- which were only left out of the task description by an oversight.

mint-thompson:
@cmoesel I think there's just one spot where we fish in the tank and want a Profile, Extension, Logical, or Resource, and may get an Instance equivalent: in `StructureDefinitionExporter`, when fishing for FHIR, we check the type to see which exporter to use. In other places, we only check whether the result exists. I'm basing this statement on a search for calls to `Package.fish`. It seems like most other places call `Package.fishForFHIR` instead, where nothing should need to change.

cmoesel (approved):
Thanks, @mint-thompson!  This looks great!

I did a quick look myself to see where `FSHTank.fish` is used, because that is another case where we return an `Instance` when perhaps something else was expected.  It looks like we handle it pretty well wherever it is used.

The only thing I found that _maybe_ could be improved would be [FSHTank.fishForMetadata](https://github.com/FHIR/sushi/blob/1d660484c335fccf233f37c3d698357b1daac36c/src/import/FSHTank.ts#L396-L404). For definitions other than `Instance`, we set the `resourceType` as appropriate (e.g., `StructureDefinition`, `ValueSet`, or `CodeSystem`) -- but if it's an `Instance`, we don't even try to set `resourceType`.  I think we could _try_, but since `InstanceOf` could technically be a profile of `StructureDefinition`, `ValueSet`, or `CodeSystem` that is defined outside the tank, there is no guarantee we can actually figure out the correct `resourceType`.  So... I don't know if we want _best effort_ (which would be inconsistent) or if we want _no effort_ (which at least consistently does not set resourceType for Instances.  TBH -- I'm fine leaving as-is or implementing best effort.  I'll leave that to @mint-thompson to decide.

mint-thompson:
@cmoesel I'm investigating the scenario you're describing, and I think it already gets sort of handled? There's a small bit in [MasterFisher.fishForMetadata](https://github.com/FHIR/sushi/blob/1d660484c335fccf233f37c3d698357b1daac36c/src/utils/MasterFisher.ts#L77-L81) that tries to figure out the `resourceType` when it fishes up an `Instance` from the tank. Does this provide the behavior that you're asking about?

cmoesel:
@mint-thompson A-ha!  Yes!  That should do the trick since the only code that cares about the resourceType would be using the MasterFisher anyway.  Thanks for finding that!  My approval still stands!

ngfreiter (approved):
👍 👍

PR #951, opened 2021-11-12
Replace (rather than overwrite) packages in fhir cache

cmoesel:
This PR fixes a bug I noticed in regression that occurs if you have a `#current` package in your fhir cache with certain characteristics and SUSHI attempts to update it with a newer version.  In this case, you get an error like the following:
> error Failed to load dk.fhir.ig.dk-core#current: ENOTEMPTY: directory not empty, rename '/Users/cmoesel/.fhir/packages/dk.fhir.ig.dk-core#current/example' -> '/Users/cmoesel/.fhir/packages/dk.fhir.ig.dk-core#current/package/example'

This happens because there is an `example` folder in the root of the project's cache and in the `package` subfolder.  When our package "cleaning" function runs, it tries to copy `./example` to `./package/example`, but since `./package/example` already exists, it fails.  The solution is to ensure that when we update a package in the cache, we completely _replace_ its contents rather than _overwriting_ its contents in place.  This is actually better anyway since the previous approach could allow stale data to remain if a file existed in an older package but was deleted in a newer package (and thus remained in the cache as the older version).

To reproduce, use the currently released SUSHI or the sushi on master:
1. Download and extract the [KL-dk-master.zip] project archive(https://github.com/hl7dk/KL-dk/archive/refs/heads/master.zip).
2. `cd` into the extracted project and run SUSHI.
3. If you already had an older version of its dependency (dk-core) in your fhir cache, you might get the error right away. You've reproduced it easily!
4. If you did not get an error, then you either did not have dk-core in your cache or it was already up-to-date and/or clean. You will need to manually adjust it in your cache as detailed below.
5. Copy the `package/example` folder to the root of the dk-core#current project in the cache.
    a. `cp -r ~/.fhir/packages/dk.fhir.ig.dk-core\#current/package/example/ ~/.fhir/packages/dk.fhir.ig.dk-core\#current/example/`
6. Edit dk-core#current's `package.json` so the `date` is earlier than whatever date is currently in there.
    a. `vi ~/.fhir/packages/dk.fhir.ig.dk-core\#current/package/package.json`
7. Run SUSHI on the KL-dk-master and you should get the package load error.
8. Note that if you run it again, it will succeed now since the date is updated in the `package.json` so it does not attempt to download again.  To get the error again, do steps 5-7 again.

If you do the same using the SUSHI from this PR branch, then you won't get the error.  Make sure you do steps 5-7 above to ensure the proper error condition is in place.  After word, you should also notice that `~/.fhir/packages/dk.fhir.ig.dk-core\#current` has only a single `package` subfolder since we replaced the full contents.

### DIFF
--- a/src/fhirdefs/load.ts
+++ b/src/fhirdefs/load.ts
@@ -128,26 +128,31 @@ export async function loadDependency(
 
   // If the packageUrl is set, we must download the package from that url, and extract it to our local cache
   if (packageUrl) {
-    // Create a temporary file and write the package to there
-    temp.track();
-    const tempFile = temp.openSync();
-    const targetDirectory = path.join(cachePath, fullPackageName);
     logger.info(`Downloading ${fullPackageName}...`);
     const res = await axios.get(packageUrl, {
       responseType: 'arraybuffer'
     });
     if (res?.data) {
       logger.info(`Downloaded ${fullPackageName}`);
-      fs.ensureDirSync(targetDirectory);
+      // Create a temporary file and write the package to there
+      temp.track();
+      const tempFile = temp.openSync();
       fs.writeFileSync(tempFile.path, res.data);
-      // Extract the package from that temporary file location
+      // Extract the package to a temporary directory
+      const tempDirectory = temp.mkdirSync();
       tar.x({
-        cwd: targetDirectory,
+        cwd: tempDirectory,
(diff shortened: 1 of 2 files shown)


julianxcarter (approved):

mint-thompson (approved):
hooray!

PR #952, opened 2021-11-12
Fix slicing cardinality validation

ngfreiter:
This makes it so that if a single element is sliced, we allow it as long as it is inherited from a parent element. But if the FSH tries to apply a `contains` rule on a single element, we log an error.

For example, this should be an error:
```
Profile: V
Parent: Immunization
* vaccineCode ^slicing.discriminator.type = #pattern
* vaccineCode ^slicing.discriminator.path = "$this"
* vaccineCode ^slicing.rules = #open
* vaccineCode ^slicing.description = "Slicing example"
* vaccineCode contains slice1 0..1 and slice2 0..1
```

But if you test the similar situation in the https://github.com/hl7ch/ch-etoc IG on the ChEtocImmunizationSection, which inherits from the ChVacdImmunizationSection profile, it should log no error, because the slicing is introduced in a dependency, not in the FSH itself.

### DIFF
--- a/src/errors/InvalidElementForSlicingError.ts
+++ b/src/errors/InvalidElementForSlicingError.ts
@@ -1,5 +1,12 @@
-export class InvalidElementForSlicingError extends Error {
+import { Annotated } from '.';
+
+export class InvalidElementForSlicingError extends Error implements Annotated {
+  specReferences = [
+    'http://hl7.org/fhir/elementdefinition-definitions.html#ElementDefinition.slicing'
+  ];
   constructor(public path: string) {
-    super(`Cannot slice element '${path}' which is not an array or choice`);
+    super(
+      `Cannot slice element '${path}' since FHIR only allows slicing on choice elements (e.g., value[x]) or elements with max > 1`
+    );
   }
 }
--- a/src/errors/InvalidElementForSlicingError.ts
+++ b/src/errors/InvalidElementForSlicingError.ts
@@ -1,0 +1,5 @@
+export class InvalidElementForSlicingError extends Error {
+  constructor(public path: string) {
+    super(`Cannot slice element '${path}' which is not an array or choice`);
+  }
+}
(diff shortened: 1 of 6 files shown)


src/errors/InvalidElementForSlicingError.ts
```diff
@@ -0,0 +1,5 @@
+export class InvalidElementForSlicingError extends Error {
```
cmoesel, line 1: I know we haven't really used this functionality much, but... I think it would be good for this to `implements Annotated` and have:
```ts
specReferences = ['http://hl7.org/fhir/elementdefinition-definitions.html#ElementDefinition.slicing'];
```

src/errors/InvalidElementForSlicingError.ts
```diff
@@ -0,0 +1,5 @@
...
+  constructor(public path: string) {
+    super(`Cannot slice element '${path}' which is not an array or choice`);
```
cmoesel, line 3: I think we could clarify this message a bit more.  Either:

> Cannot slice element '${path}' since it is not an array or choice

or spelling it out even more

> Cannot slice element '${path}' since FHIR only allows slicing on choice elements (e.g., value[x]) or elements with max > 1

cmoesel (changes requested):
This works well.  I only have a couple of minor suggestions regarding the error itself.

src/errors/InvalidElementForSlicingError.ts
ngfreiter: https://github.com/FHIR/sushi/pull/952/commits/6e644f6f9393105bf1b2b6925f49a939c47cbcd7

src/errors/InvalidElementForSlicingError.ts
ngfreiter: https://github.com/FHIR/sushi/pull/952/commits/6e644f6f9393105bf1b2b6925f49a939c47cbcd7

cmoesel (approved):
Looks good!  @mint-thompson or @julianxcarter -- can we get a second review on this by mid-afternoon-ish?

julianxcarter (approved):
Works well!

Issue #932, opened 2021-10-18
SUSHI should include inherited constraints from base in the diff of slices


cmoesel:
Right now, when SUSHI creates a slice, it calls `captureOriginal` on the new slice so that the diff only contains things that are different from the base.  Because of the way that the validator works, however, we must repeat constraints from the base in the slice differential.  I think the easiest way to do this is to _not_ call `captureOriginal` when we create the new slice -- but instead use the same captured original as the base (but someone should confirm).  This means we still would not include things that were constrained on the base _after_ the slice was created, but I think that's probably an OK compromise in order to reduce complexity.

See: https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/How.20to.20slice.20on.20the.20value.20of.20an.20extension/near/257451464

jafeltra:
This was partially addressed in #955, and it has not been reported as an issue recently.

PR #955, opened 2021-11-15
Include inherited constraints on slices

ngfreiter:
See https://github.com/FHIR/sushi/issues/932 for an explanation.

See https://fshschool.org/FSHOnline/#/share/3wQHuPR for an example to recreate this. 

On `master`, when you run this through SUSHI and the IG Publisher, you will see an error like the following:

image

On this branch you should see no error.

I've noticed that this PR does cause some regressions, but I looked through them and did not see any issues that were not expected from this change. But I would appreciate if reviewers also take a look.


src/fhirtypes/ElementDefinition.ts
```diff
@@ -2181,10 +2181,15 @@ export class ElementDefinition {
...
+      // When an element is a slice, all constraints should be carried on into the diff
+      // so we do not clear the original
+      const isSlice = e.sliceName != null;
+      const eClone = e.clone(!isSlice);
       eClone.id = eClone.id.replace(targetElement.id, this.id);
       eClone.structDef = this.structDef;
-      eClone.captureOriginal();
+      if (!isSlice) {
+        eClone.captureOriginal();
+      }
```
cmoesel, line 2192: You've implemented this behavior in `cloneChildren`, which is used in places _other than_ unfolding a slice (e.g, unfolding content references uses it too).  While it's probably harmless (as I don't think a content reference can point to a slice), it's behavior that really only applies to a particular use case -- and _could_ be unexpected if we ended up using `cloneChildren` in future code for a different use case.

I think it would be more targeted and less surprising to add a `resetOriginal` or `recaptureOriginal` argument to the `cloneChildren` signature (defaulting to `true`, the previous behavior).  Then in the code that calls this function specifically for the purpose of unfolding a slice, pass in `false` to get the behavior we want there.

cmoesel (changes requested):
I think you missed part of the use case because although this retains the constraint diffs for a slice's _children_, any constraints on the slice itself still will not show up in the diff because `ElementDefinition.addSlice` recaptures the original.  So I think you need to apply this there too.

I've also included an individual comment about perhaps needing to narrow when this is used with cloneChildren.

Last, it would be pretty sweet if we could have a few unit tests. ;-)

src/fhirtypes/ElementDefinition.ts
ngfreiter: This comment makes sense, I'll update it.

In response to your top level comment, I can definitely add the logic to `addSlice`, and in face I originally had it there as well. But I couldn't come up with an example where I actually needed that logic in `addSlice`. So I was having a hard time justifying adding this to `addSlice`, which is a relatively significant change to how we handle slices, when I couldn't figure out why exactly we should do that. Although I guess maybe it makes sense just for consistency? Or do you have a case where we can point to an issue in the publisher where we definitely want to have this logic in `addSlice`?

cmoesel: > which is a relatively significant change to how we handle slices

Is it any more significant than what we're doing in the unfolding logic now?  I'm just trying to understand if there is something I'm missing.  Or do you just think of it as more significant because it _always_ affects slices, vs unfolding which only affects them when there are child elements?

> Although I guess maybe it makes sense just for consistency?

Yes, I think so.  Unless there is some weirdness that I'm not anticipating.  For example, we obviously would _not_ want the `slicing` to show up in the diff of every _slice_ (even though it is a constraint on the slice root).

> do you have a case where we can point to an issue in the publisher where we definitely want to have this logic in `addSlice`?

I'm not sure that I do.  The problem arises when a slice's invariant/discriminator relies on some constraint introduced in the slice root.  In theory, since FHIRPath has a `elementDefinition()` function that references the ED itself, then I guess it's possible for an author to create an invariant that relies on any arbitrary constraint in the ED (for example the binding).  I'm not sure what real world use case would ever require it, but... if they did, then they'd need all the constraints to be there (I guess).  When it comes down to it, though, I guess it is just for consistency, and because of [this exchange](https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/How.20to.20slice.20on.20the.20value.20of.20an.20extension/near/257451464) w/ Grahame:

> **Chris Moesel:** Are there any other constraints or ElementDefinition fields that we also should redeclare (i.e. copy) from the base when people create slices of an element?
> **Grahame Grieve:** I think pretty much everything

ngfreiter: Ok @cmoesel, I've added this in https://github.com/FHIR/sushi/pull/955/commits/70d8784e7f630393ca10ef7337b336fc9d50f0e7, but be aware it does cause a fair bit of regression diffs. I looked through and what I was seeing was all expected.

cmoesel: @ngfreiter -- I just noticed another thing when reviewing the code.  Technically, you're only copying the constraints on the element with the sliceName; but if that element has children, those children are technically part of the slice -- so shouldn't they also have copied constraints?

For example:
```
Profile: MyObservation
Parent: Observation
* component MS
* component.code MS
* component ^slicing.discriminator.type = #pattern
* component ^slicing.discriminator.path = "code"
* component ^slicing.description = "Slice based on the component.code pattern"
* component contains Foo 0..1
* component[Foo].code ^short = "Foo Code!"
```

When I run this, `Observation.component:Foo` gets the `MS` flag in the differential, but `Observation.component:Foo.code` does not get its `MS` flag in the differential.  So it seems like when we are unfolding children, if the elements to unfold are in a slice, we ought to be preserving/copying constraints there too.  BUT... if there are some unfolded children that didn't get additional constraints in the slice, then we don't want those to show up in the differential at all -- we only want the ones that changed in the slice too.  Which is probably tricky to implement.

Bleh.  This is getting more complicated than I'd like.  I wish the IG Publisher just handled this better.  I am running the regression now to see what kind of churn it makes -- but if you fix the issue I just raised in the previous paragraph, that will surely create even more churn.  Ugh.  Time to go home.

cmoesel (changes requested):
I commented on one of our threads w/ a finding.  Here are a few more comments regarding the regression I ran:

**1.**

I noticed that in `HL7/carin-bb` the first change is this:
image
I thought it was weird that a type `BackboneElement` would be in the differential at all (since it is not different).  I looked into it, and SUSHI is in fact putting it in the differential for the slice root, so I guess it makes sense that it is here, but still... I wonder why it's there at all?  Out of scope of this PR, but just bringing it up in case you know.

**2.**

In `hl7ch/ch-allergyintolerance`, a few extension slice elements are added to the differential, but they don't have the extension details in them.  They just has a sliceName.  It seems to me that either this is an issue (and the extension details should be there) or this is unnecessary and the element does not need to be in the differential at all (since there are no constraints, so it doesn't actually add any value).

image

**3.**

I cannot successfully build `HL7/davinci-pas` using this branch's code.  It failed in regression, so I just tried it manually and it still did not work. I get this:
```
info  Converting FSH to FHIR resources...

<--- Last few GCs --->

[3296:0x102dc4000]    75540 ms: Scavenge 2041.0 (2046.9) -> 2040.2 (2046.9) MB, 6.6 / 0.0 ms  (average mu = 0.202, current mu = 0.159) allocation failure
[3296:0x102dc4000]    75622 ms: Scavenge 2041.0 (2046.9) -> 2040.2 (2047.2) MB, 7.0 / 0.0 ms  (average mu = 0.202, current mu = 0.159) allocation failure
[3296:0x102dc4000]    75698 ms: Scavenge 2041.1 (2047.2) -> 2040.4 (2047.2) MB, 5.4 / 0.0 ms  (average mu = 0.202, current mu = 0.159) allocation failure


<--- JS stacktrace --->

==== JS stack trace =========================================

    0: ExitFrame [pc: 0x100a04099]
Security context: 0x315225a808d1 <JSObject>
    1: baseClone [0x3152b88bff79] [/Users/cmoesel/dev/fhir/sushi/node_modules/lodash/lodash.js:~2662] [pc=0x68101dcd111](this=0x315268bc28c9 <JSGlobal Object>,0x3152983744a9 <Object map = 0x3152110fd6a9>,5,0x31524b6c04b1 <undefined>,33,0x3152964dcb31 <JSArray[50]>,0x3152ec8792f9 <Stack map = 0x3152110f5779>)
    2: baseClone [0x3152b88bff79] [/Users/cmoesel/dev...

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
 1: 0x1011f7af5 node::Abort() (.cold.1) [/Users/cmoesel/.config/nvm/12.22.7/bin/node]
 2: 0x1000a5fd9 node::Abort() [/Users/cmoesel/.config/nvm/12.22.7/bin/node]
 3: 0x1000a613f node::OnFatalError(char const*, char const*) [/Users/cmoesel/.config/nvm/12.22.7/bin/node]
 4: 0x1001efb57 v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, bool) [/Users/cmoesel/.config/nvm/12.22.7/bin/node]
 5: 0x1001efaf7 v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, bool) [/Users/cmoesel/.config/nvm/12.22.7/bin/node]
 6: 0x100386445 v8::internal::Heap::FatalProcessOutOfMemory(char const*) [/Users/cmoesel/.config/nvm/12.22.7/bin/node]
 7: 0x100387c9a v8::internal::Heap::RecomputeLimits(v8::internal::GarbageCollector) [/Users/cmoesel/.config/nvm/12.22.7/bin/node]
 8: 0x10038471e v8::internal::Heap::PerformGarbageCollection(v8::internal::GarbageCollector, v8::GCCallbackFlags) [/Users/cmoesel/.config/nvm/12.22.7/bin/node]
 9: 0x1003824d0 v8::internal::Heap::CollectGarbage(v8::internal::AllocationSpace, v8::internal::GarbageCollectionReason, v8::GCCallbackFlags) [/Users/cmoesel/.config/nvm/12.22.7/bin/node]
10: 0x100381521 v8::internal::Heap::HandleGCRequest() [/Users/cmoesel/.config/nvm/12.22.7/bin/node]
11: 0x100346be1 v8::internal::StackGuard::HandleInterrupts() [/Users/cmoesel/.config/nvm/12.22.7/bin/node]
12: 0x1006ab4bc v8::internal::Runtime_StackGuard(int, unsigned long*, v8::internal::Isolate*) [/Users/cmoesel/.config/nvm/12.22.7/bin/node]
13: 0x100a04099 Builtins_CEntry_Return1_DontSaveFPRegs_ArgvOnStack_NoBuiltinExit [/Users/cmoesel/.config/nvm/12.22.7/bin/node]
14: 0x68101dcd111
15: 0x68101dcc8bd
fish: 'ts-node src/app.ts /Users/cmoes…' terminated by signal SIGABRT (Abort)
```
It built fine on the _master_ branch of sushi though...  Did this happen to you?

**UPDATE:**  This is also happening on `HL7/fhir-saner`, `openhie/covid-ig`, `AudaciousInquiry/fhir-saner`, `IHE/PCC.PCS`, `HL7/fhir-COVID19Library-ig`, `ahdis/ch-crl`, and `IHE/QRPH.QORE`.  If you also see this, then it's definitely a show-stopping bug.  If you don't see this, then we should try to assess what is different between your system and mine.

src/fhirtypes/ElementDefinition.ts
ngfreiter: We don't need to put those constraints on the unfolded children of the slice do we? Applying `MS` on `Observation.component.code` indicates that all slices of `Observation.component` have `.code` set as `MS`. Wouldn't it be becoming somewhat an abuse of the "differential" to be putting those constraints in children of slices? It feels like an abuse of the differential already to be putting this info in slices when it technically shouldn't be needed. But at least in that case we have a bug we can point to and say, this is why we had to add this stuff to the differential.

cmoesel: I guess I don't see how it is different.  Either the slice carries the constraints it inherited or it doesn't.  An element three levels deep in the slice is still part of the slice.  The fact that the current code picks up extensions in a slice is actually just a happy by-product of the fact that extensions happen to have a sliceName -- otherwise they wouldn't have been copied over either.

I agree that the differential shouldn't need to have these things.  But it also seems like if we're going to make slices carry copies of all the constraints, then we should implement it fully (for the whole slice).  But... I do fear what that would mean.

I'll go back to that Zulip thread and try to get some clarification.

ngfreiter: Yeah I guess this is just me continuing to disagree on the consistency front. To me the original fix (just https://github.com/FHIR/sushi/pull/955/commits/4a3509dff0988566b2275322e460eb8dc5e806e8) is us making an exception to how we do things to satisfy the publisher on something that it seems like the publisher should handle. But to me consistency doesn't justify expanding that exception and creating all this churn.

cmoesel: I hear you.  TBH, don't be terribly surprised if you win me over to your side and I ask you to make this work *only* for extensions. ;-)

cmoesel:
@ngfreiter - I think you’ve convinced me. This is a LOT of churn and redundant differentials just to address a limitation in the IG Publisher. I think that maybe we should only address the immediate issue at hand, which is that *extensions* need to be copied from the slice base to the slice. If you agree with me, then I say go for it (and sorry for making you jump through all these hoops). If you disagree, we can continue to discuss when I get back.

ngfreiter:
Ok @cmoesel, I think https://github.com/FHIR/sushi/pull/955/commits/fe022920631786229fa5eeceea88d03599df57ca should work as you described only on extenionsions.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -2176,15 +2176,25 @@ export class ElementDefinition {
...
    * Returns an array of an ElementDefinition's unfolded children.
    * @param {ElementDefinition} targetElement - The ElementDefinition being unfolded
+   * @param {boolean} recaptureSlices - Indicates whether or not slices should be recaptured
    * @returns {ElementDefinition[]} An array of the targetElement's children, with the IDs altered and
    * the original property re-captured.
    */
-  private cloneChildren(targetElement: ElementDefinition): ElementDefinition[] {
+  private cloneChildren(
+    targetElement: ElementDefinition,
+    recaptureSlices = true
```
cmoesel, line 2185: With the most recent change, this argument name is a little misleading.  Can we rename it to `recaptureSliceExtensions`?  (And don't forget to rename it in the JSDoc above too).

ngfreiter: Good point, fixed in https://github.com/FHIR/sushi/pull/955/commits/313f7337079fbc4b1bb726d656bcd875944c8187

cmoesel:
OK.  Cool.  I'm going to run regression on this again and see where we land!

jafeltra (approved):
The changes work as described for me! I haven't run the regressions (because it sounded like Chris was), but I can if he hasn't! Pending the outcome of the regressions, this looks good to me!

cmoesel (approved):
OK.  I've reviewed the changes and done full regression.  This affects about 6 IGs, one of which is the IG for which the original bug was reported.  All of the changes make sense and don't seem to introduce any new problems.

I also updated the Zulip thread to reflect our current approach:
> We've decided that for now we will address the immediate problem at hand -- which is that the validator needs extensions redeclared in slices in order to process discriminator paths that use the `extension()` function. Since slices conceptually inherit constraints from the base slice, we think its less disruptive now if we only redundantly declare the inherited constraints in cases where we know they cause problems otherwise. This will be in the next SUSHI release.

PR #957, opened 2021-11-16
Fix findElementByPath w/ duplicated slicenames

### LINKED ISSUE #956: SUSHI fails to find element slices when other elements have the same slice name
_This issue is based on my findings while investigation a bug reported on Zulip [here](https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Bug.3A.20No.20element.20found.20at.20path)._

If the same slice name is used on multiple elements, and some of these elements have paths that are sub-paths of others, SUSHI may fail to find some slice elements.  For example, given elements w/ these ids:
```
Observation.category:Foo
Observation.category:Foo.coding:Bar
Observation.category:Bar
```
if you have a rule like this in SUSHI:
```
* category[Bar] MS
```
then SUSHI will fail to find the element, reporting an error like:
> No element found at path category[Bar] for FlagRule in ObsProfile, skipping rule errors.

This is because the logic to find the path  mistakenly focuses in on the first path that has slicename `Bar` - even if the path isn't a complete match, but then later notices that the paths don't match and returns `undefined`.  SUSHI needs to be updated to consider the path throughout the element matching algorithm.

cmoesel:
Fixes #956

When multiple elements have the same slicename (but different paths), sometimes SUSHI fails to find slice elements. This PR fixes it by ensuring that SUSHI takes the full path into account when trying to find slice elements.

There is a unit test that shows this. If you copy it to the test suite in master, it will fail. But in this PR branch, it passes.  (Sorry I forgot to make a separate commit for just the test).

If you want to test manually, you'll have to follow a few steps:
* Download the test project here: https://github.com/patrick-werner/testDerivedAddress
* Run SUSHI on it.  There will be errors about missing snapshots.  This is expected.
* Unzip the de.basisprofil.r4-1.1.0.tgz file at the root of the project and copy it's contents to your FHIR cache in the appropriate place.  E.g., the `package` folder in the zip should replace `~/.fhir/packages/de.basisprofil.r4#1.1.0/package` folder in your FHIR cache.
* Run SUSHI again.

On master, you get errors about not being able to find `address[Postfach]`.  On this branch, you do not get those errors.

### DIFF
--- a/src/fhirtypes/StructureDefinition.ts
+++ b/src/fhirtypes/StructureDefinition.ts
@@ -270,7 +270,12 @@ export class StructureDefinition {
 
       // After getting matches based on the 'base' part, we now filter according to 'brackets'
       if (pathPart.brackets) {
-        const sliceElement = this.findMatchingSlice(pathPart, matchingElements, fisher);
+        const sliceElement = this.findMatchingSlice(
+          fhirPathString,
+          pathPart,
+          matchingElements,
+          fisher
+        );
         if (sliceElement) {
           matchingElements = [sliceElement, ...sliceElement.children()];
         } else {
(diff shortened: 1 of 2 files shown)


mint-thompson (approved):
Looks good! Thank you.

ngfreiter (approved):

PR #960, opened 2021-11-22
Check if extension is or is not a modifier extension when constraining types or assigning instances

### LINKED ISSUE #576: Log error when a non-modifier extension is used in a modifierExtension path
The IG Publisher issues an error if an extension is used in `modifierExtension` but that extension does not have `isModifier` set.  SUSHI should consider checking for this and issuing an error so that authors can fix it _before_ they get to the stage of running the publisher.

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/ModiferExtension

mint-thompson:
Fixes #576. (thanks @ngfreiter)

The title is a bit of a mouthful, but the idea is that:
- a "modifier extension" is one where the root `Extension` element's `isModifier` property is `true`,
- modifier extensions should only be used on the `modifierExtension` element, and
- only modifier extensions should be used on the `modifierExtension` element.

When constraining an element's type by OnlyRule or ContainsRule, if the type being constrained is `Extension`, check to see if the profiles being applied are modifier extensions. If so, and if the element is `modifierExtension`, everything is fine. But otherwise, show an error. Similarly, if the profiles being applied are not modifier extensions, and the element being constrained is `modifierExtension`, show an error.

When assigning instances of an extension in an Instance definition, check to see if those extension instances represent modifier extensions. If a modifier extension instance is assigned to an element other than `modifierExtension`, show an error. If a non-modifier extension instance is assigned to `modifierExtension`, show an error. If information about the extension instance's type is not available, assume it is okay and do not show an error.

Honestly, I think it's easier to understand when written as FSH, so hopefully the tests clearly demonstrate the scenarios I want to handle.

My implementation for this specifically does _not_ check rules where `extension.url` or `modifierExtension.url` is set directly by the rule. My feeling is that checking the modifier-or-not status of an extension is most appropriate when using the extension, and not some contained element of the extension. But, I am willing to add these checks if folks think that it makes sense and adds value.


ngfreiter (commented):
Should the description say "fixes https://github.com/FHIR/sushi/issues/576"?

src/fhirtypes/common.ts
```diff
@@ -218,6 +218,28 @@ export function setPropertyOnInstance(
         } else {
           if (typeof assignedValue === 'object') {
             Object.assign(current[key][index], assignedValue);
+            // check if we have used modifier extensions correctly
+            if ((key === 'extension' || key === 'modifierExtension') && assignedValue.url != null) {
```
ngfreiter, line 222: One thing I'm noticing about the way this works is that it only works when we are assigning the entire `Extension` as an `Instance`. Consider the example below:
```
Extension: Foo
* value[x] only boolean
* . ?!

Instance: FooInstance
InstanceOf: Foo
Usage: #inline
* valueBoolean = true

Instance: 😄
InstanceOf: Patient
* extension[Foo].valueBoolean = true

Instance: 😔
InstanceOf: Patient
* extension[0] = FooInstance
```
In this example, the 😄  and 😔  Instances generate the same JSON, but for the 😔  Instance we log an error, while for the 😄  we don't. My guess would be that the 😄  would actually be the more common way to accomplish this. Do you think we should also log an error in that case?

src/fhirtypes/common.ts
```diff
@@ -218,6 +218,28 @@ export function setPropertyOnInstance(
...
+                  extension.snapshot.element.find((el: ElementDefinition) => el.id === 'Extension')
+                    .isModifier === true;
+                if (key === 'extension' && isModifierExtension) {
+                  throw new Error(
+                    `Instance of modifier extension ${
+                      extension.name ?? extension.id
+                    } assigned to extension path. Modifier extensions should only be assigned to modifierExtension paths.`
+                  );
+                } else if (key === 'modifierExtension' && !isModifierExtension) {
+                  throw new Error(
```
ngfreiter, line 235: Should we make these errors just for consistency? I'm not sure that anywhere else we throw errors without defining our own error class to throw.

src/fhirtypes/common.ts
```diff
@@ -218,6 +218,28 @@ export function setPropertyOnInstance(
         } else {
           if (typeof assignedValue === 'object') {
             Object.assign(current[key][index], assignedValue);
+            // check if we have used modifier extensions correctly
+            if ((key === 'extension' || key === 'modifierExtension') && assignedValue.url != null) {
+              const extension = fisher.fishForFHIR(assignedValue.url, Type.Extension);
+              if (extension) {
+                const isModifierExtension =
+                  extension.snapshot.element.find((el: ElementDefinition) => el.id === 'Extension')
+                    .isModifier === true;
```
ngfreiter, line 227: I _think_ this same check is used in all three places, so maybe this could be a separate function?

ngfreiter (changes requested):
Generally looking pretty good to me, but I had a few comments below.

src/fhirtypes/common.ts
mint-thompson: Oof... yeah, I can't really argue against this. This is a little more intense, but I'll see what I can conjure up.

mint-thompson: Also, writing the modifier rule as `* . ?!` is a lot more concise than my `* . ^isModifier = true`, so thank you for that as well.

ngfreiter (approved):

julianxcarter (approved):

PR #962, opened 2021-11-22
Bring back lower case package names

ngfreiter:
This re-applies the #912 commit which adds support for loading case-insensitive packages in SUSHI.

The changes in c2da55a268fa26070025958ff72d657343c3d00c are directly applied from #912, so they should have already been reviewed. However, it may be good for reviewers to take a look at that still, since we did let this bug slip past us the first time.

Then c2da55a268fa26070025958ff72d657343c3d00c fixes the bug that slipped through, which is that SUSHI failed to load a package with a version number if that package with a version number did not already exist in the FHIR cache. To test that this now works, you can add a dependency like:

```yaml
dependencies:
  hl7.fhir.us.core: 3.1.1
```

to your IG, and delete `hl7.fhir.us.core#3.1.1` from your FHIR cache. On c2da55a268fa26070025958ff72d657343c3d00c, SUSHI will be unable to read in the definitions from that package, but on c2da55a268fa26070025958ff72d657343c3d00c, it should work fine.


jafeltra (approved):
This looks good to me! I tested it out on the previous commit and the latest commit and see that the error goes away. I looked through the changes in `load.ts` and `importConfiguration.ts` and the remaining changes make sense so 🤞

cmoesel (approved):
I agree with @jafeltra on all counts!

PR #965, opened 2021-11-23
Package loading: Fallback to packages2

### LINKED ISSUE #961: Always fallback to packages2.fhir.org when a package doesn't exist at packages.fhir.org
SUSHI currently only checks packages2.fhir.org for FHIR R4B and FHIR R5 builds.  Instead, we should check packages2.fhir.org _whenever_ a package isn't found in packages.fhir.org.

See: https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/fhir.2Edicom/near/262334652

cmoesel:
If a package can't be found in the normal packages repo, fallback to the packages2 repo.  This is based on the Zulip conversation here: https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/fhir.2Edicom/near/262334652

To test, create a simple SUSHI project with the following dependency in sushi-config.yaml:

```
dependencies:
  fhir.dicom:
    uri: http://fhir.org/packages/fhir.dicom
    version: 2021.4.20210910
```

On master, SUSHI will fail to load the dependency.  On this branch, it will successfully load it.

Fixes #961.

### DIFF
--- a/src/fhirdefs/load.ts
+++ b/src/fhirdefs/load.ts
@@ -114,15 +114,14 @@ export async function loadDependency(
   } else if (!loadedPackage) {
     packageUrl = `https://packages.fhir.org/${packageName}/${version}`;
 
-    // If this is an R4B or R5 package, then we may need to get it from packages2 if it is not in packages
-    if (/^hl7\.fhir\.r(4b|5)\./.test(packageName)) {
-      try {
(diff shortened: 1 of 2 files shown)


ngfreiter (approved):
📦 🥈

mint-thompson (approved):
packages was so good, there just had to be a packages2

PR #966, opened 2021-11-29
Load package.json to provide canonical

### LINKED ISSUE #948: Automatically infer canonical URI of dependency from package.json
SUSHI needs a URI for each dependency IG so that it may populate `ImplementationGuide.dependsOn`. However, certain packages do not have ImplementationGuide resources, so we cannot get a canonical URI. In this case, we require that the user manually specify the canonical in the `dependencies` object in `sushi-config.yaml`. We could instead try to infer the URI from the `package.json` which should be present in the package. This may present its own problems, and we should investigate and test some different packages before adding support.

See this thread for details: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/ImplementationGuide.3A.20ImplementationGuide.20instance.20misses.20url.

mint-thompson:
Fixes #948 and completes [CIMPL-845](https://standardhealthrecord.atlassian.net/browse/CIMPL-845).

When loading a dependency package, store the package.json in the FHIRDefinitions. When exporting an IG, if the dependency package does not contain its own ImplementationGuide, check the package.json to try to get the canonical URL for the package.

As a side note: the package `de.gkvsv.eRezeptAbrechnungsdaten#1.1.0-rc`, which was mentioned in the thread that led to this issue, has a package.json and no ImplementationGuide. But, the package.json doesn't have a canonical field, so SUSHI will still (correctly) produce the same error as before.


jafeltra (approved):
These changes make sense to me! I edited by hand the package.json of the dependency you referenced and saw the error resolve itself as well, so I think this should work well to fallback to package.json files with a `canonical` field.

jafeltra (approved):
Still approved!

cmoesel (approved):
Looks good to me!

PR #972, opened 2021-12-03
Do not rename choice types on StructureDefinitions

### LINKED ISSUE #954: SUSHI should no longer use "choice type renaming"
Up until now, SUSHI has utilized an approach to differential ids and paths that allows choice types to be simplified.  For example, `Observation.value[x]:valueQuantity` could be simplified to `Observation.valueQuantity` in the differential and the IG Publisher would convert it to the "long form" during the snapshot generation process.

We felt comfortable doing it this way because a number of core FHIR profiles use this approach (e.g., most of the vital signs profiles), it's referenced in part in the spec, and the mechanics for it are outlined in the [Type Slicing in FHIR R4](https://fire.ly/blog/type-slicing-in-fhir-r4/) blog post.  Since then, however, we've learned that this is no longer the preferred approach (discussion in many places, but most recently [here](https://chat.fhir.org/#narrow/stream/179177-conformance/topic/Choice.20Type.20Renaming) and [here](https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/Odd.20IG.20broken.20link.20build.20issue.20with.20choice.20data.20elements)).

As a result, we should update SUSHI to no longer generate profiles/extensions with choice type renaming.  Instead, we should prefer the long form.  E.g., rather than:
```json
{
  "id": "Observation.valueQuantity",
  "path": "Observation.valueQuantity",
  "sliceName": "valueQuantity"
}
```
it should now be:
```json
{
  "id": "Observation.value[x]:valueQuantity",
  "path": "Observation.value[x]",
  "sliceName": "valueQuantity"
}
```

If there are parts of our code, however, that are built to _read_ and _understand_ this short-form syntax, however, those should stay in place in case we need to process other profiles that still use this approach.  I'm not sure if this is a concern, though, since we primarily read from the snapshot, and the snapshot didn't allow the short-form syntax.

This will affect many IGs -- but should be _safe_, as it means the same thing (and results in the same snapshot).  That being said, we many want to consider grouping this in with the update to Node 14 so it gets introduced in a major release (SUSHI 3.0).

mint-thompson:
Fixes #954 and completes task [CIMPL-851](https://standardhealthrecord.atlassian.net/browse/CIMPL-851).

A specific choice type on a StructureDefinition should not have its id or path renamed. Instead, put a slice name on the choice element. Specific choice types on Instances are not changed.

There are, naturally, many regression changes in IGs due to this PR. In my review of these changes, they were all correct and expected.


cmoesel:
_NOTE: We want to save this for a SUSHI 3.0 release since it is a significant change._  I think there were one or two other things we wanted to hold for that release too.  So... let's not merge this into master just yet.  We may want to consider creating a new branch for the features we want to hold for 3.0...  BUT... let's still review this one NOW._

cmoesel (changes requested):
This didn't end up being so bad, eh?

I noticed that you kept the `ElementDefinition.diffId()` function in place, presumably because we use it in `common.ts`.  I think it is a little confusing now, however, since we don't really use it for the differential id anymore (as the name implies).  I'd suggest we move it from `ElementDefinition.ts` to `common.ts` (where it is used).  It will need a slight refactoring (and a better name), but I think it will fit in better there.

julianxcarter (dismissed):

src/fhirtypes/ElementDefinition.ts
```diff
@@ -2662,7 +2633,7 @@ const PROPS = [
   'mapping'
 ];
 
-const CHOICE_TYPE_SLICENAME_POSTFIXES = [
+export const CHOICE_TYPE_SLICENAME_POSTFIXES = [
```
cmoesel, line 2636: Given that this constant is only used by `getShortcutId`, I think it makes sense to move this to `common.ts` too.  That way it also doesn't need to be exported (since no one else uses it).

cmoesel (changes requested):
Thanks for moving and renaming the `diffId` function!  I left a comment about one more piece of code that probably should be moved with it.

cmoesel:
Unfortunately, I discovered a wrinkle in this when testing this with the mCODE IG.  The current mCODE IG runs with 0 errors when you run it against SUSHI master.  But... when you run it against this branch, you get 14 errors!

Here is an example of one of the errors:
> **Bundle.entry[7].resource.ofType(Procedure).extension[4].extension[1].value.ofType(Quantity) (l735/c33)**: Slicing  cannot be evaluated: Discriminator ($this) is based on type, but slice  Extension.extension:totalDoseDelivered.value[x]:valueQuantity in  http://hl7.org/fhir/us/mcode/StructureDefinition/mcode-radiotherapy-dose-delivered-to-volume  has multiple types: base64Binary, boolean, canonical, code, date,  dateTime, decimal, id, instant, integer, markdown, oid, positiveInt,  string, time, unsignedInt, uri, url, uuid, Address, Age, Annotation,  Attachment, CodeableConcept, Coding, ContactPoint, Count, Distance,  Duration, HumanName, Identifier, Money, Period, Quantity, Range, Ratio,  Reference, SampledData, Signature, Timing, ContactDetail, Contributor,  DataRequirement, Expression, ParameterDefinition, RelatedArtifact,  TriggerDefinition, UsageContext, Dosage, Meta

What?!  Surely a `value[x]:valueQuantity` is not all those types, is it?  But lo and behold, here is that element in the differential of that extension (`mcode-radiotherapy-dose-delivered-to-volume`) when run against master:

image

And here is that element in the differential when run against this branch:

image

So what's going on?  Well... if you look at that element in the JSON differential of the `mcode-radiotherapy-dose-delivered-to-volume` StructureDefinition, you'll find this:
```json
{
  "id" : "Extension.extension:totalDoseDelivered.value[x]:valueQuantity",
  "path" : "Extension.extension.value[x]",
  "sliceName" : "valueQuantity",
  "min" : 0,
  "max" : "1",
  "patternQuantity" : {
    "system" : "http://unitsofmeasure.org",
    "code" : "cGy"
  }
}
```
Notice there is no `type` constraint to say it is only `Quantity`.  Now... in SUSHI, we _do_ set that constraint, BUT... since `value[x]` is already constrained to only `Quantity`, SUSHI sees that is the same in the slice, so it removes it from the differential slice (since it's not different than its slice base).  Unfortunately, however, the snapshot generator in the IG Publisher does not actually look at the slice base in the profile when generating the slice in the snapshot.  _Instead_, it looks at the slice base in the core resource and uses that.  And _that_ has a bajillion types in `value[x]`.  _Sigh_.

So... this is basically the same problem as we were trying to address in PR #955.  When we did #955 at first, it created such a mess (copying _all_ constraints from the slice base), that we decided to scale it back and only do it for extension declarations, since that was the real problem we were trying to solve at the moment. :-/

I suspect that when we used the shortcut syntax, the snapshot generator knew to apply the type constraint to the slices (since it is implied by the shortcut syntax).  So that's why it was not a problem before.

So... I think we basically need to ensure that if a type is constrained in a slice base, that the type constraint is output in the differential for each of its slices.  We can probably try to do this via some magic in the code that creates slices and captures originals, or in the code that actually generates the JSON differential elements.  But I think we need to do it _somewhere_.  Or convince Grahame to fix the snapshot generator (or fix our misunderstanding if the snapshot generator is actually somehow right).

I've made a simple SUSHI project to reproduce this problem.  Make two copies of it.  Run one against master (or SUSHI 2.2.3) and then run the IG Publisher on it.  Then run the other against this PR branch and then run the IG Publisher on it.  Then compare the profiles in the IG HTML views (and generated JSON if you wish).

[TypeSlicingExample.zip](https://github.com/FHIR/sushi/files/7730984/TypeSlicingExample.zip)

cmoesel:
NOTE: I reported the above behavior in Zulip here: https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/Bug.20in.20Snapshot.20Generation.20of.20Type.20Slices.3F

cmoesel (approved):
This is a thing of beauty, my friend.  I've reviewed your most recent changes and re-run the full regression.  There are lots of changes, but all of them are expected. In addition, the errors that this PR originally introduced to the mCODE IG are now all resolved.  THANKS!

julianxcarter (approved):

PR #974, opened 2021-12-08
Fix Soft Indexing Path Only Rules w/ Indentation

### LINKED ISSUE #973: SUSHI incorrectly handling [+] in combination w/ indenting
See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Sushi.20Bug.20with.20.5B.2B.5D.20notation.3F

cmoesel:
Fixes #973

SUSHI was not handling the context correctly when there were multiple path rules with soft-indexing.  The first issue was with indented rules like this:
```
* item[+]
  * item[+]
    * linkId = "MyLink"
```
This would result in creating or splitting extra items.  Reproduce this issue w/ FSHOnline here: https://fshschool.org/FSHOnline/#/share/3olm9LN

The 2nd issue was that path-only rules were ignored if they weren't used.  For example:
```
* item[+]
* item[+]
  * linkId = "MyLink"
```
Someone might *think* that the first rule creates an empty item and so the item w/ the linkId is at index 1.  But that's not how SUSHI works.  SUSHI ignores the first rule as a no-op, so just creates the one item at index 0 w/ the linkId.  I think this is actually correct since FSH says that path rules are only used to establish context for indented rules that follow them.  But since it can be confusing to users, I've added a WARNING when this happens.

The tests reproduce the issues, but you can easily reproduce them manually as well.  For the first one, you can just use the same FSH as in the linked FSH Online example.

### DIFF
--- a/regression/repos-all.txt
+++ b/regression/repos-all.txt
@@ -98,7 +98,6 @@ ehealthsuisse/ch-orf#master
 DavidPyke/HotBeverage#master
 # Added Wed Jun 16 2021 16:00:22 GMT-0400 (Eastern Daylight Time)
 HL7/fhir-radiation-dose-summary-ig#main
-HL7/pacio-adi#main
 hl7-it/terminology#master
 JohnMoehrke/testBundle#main
 hl7dk/kl-ffb-messaging#main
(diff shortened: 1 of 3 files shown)


mint-thompson (approved):
So, now `[+]` only becomes `[=]` after it is used in a non-path rule. Hooray!

julianxcarter (approved):
This makes a lot of sense! I think the warning on unused path rules is a smart choice.

PR #981, opened 2021-12-13
Convert dependency package ids to lowercase when extracted from IG resource

### LINKED ISSUE #980: SUSHI problems resolving mixed case dependency extracted from IG resource
SUSHI 2.2.2 apparently introduced a bug related to loading dependencies from the IG resource when those dependencies are not all lowercase.

See https://github.com/IHE/ITI.PIXm/issues/109 and https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Sushi.20problem.20with.20IHE.2EITI.2EPIXm.20uppercase.20package.20naming.

cmoesel:
Fixes #980.

A previous code change converted dependency package ids to lowercase when extracted from sushi-config.yaml.  This does the same for dependencies when extracted from an IG resource.

To test, run SUSHI against the following simple project:
[UpperDepsInIG.zip](https://github.com/FHIR/sushi/files/7705437/UpperDepsInIG.zip)

On master (or SUSHI 2.2.2), you'll get this error:
> error Failed to load IHE.ITI.PIXm#3.0.0: The package IHE.ITI.PIXm#3.0.0 could not be loaded locally or from the FHIR package registry.

On this PR, you'll get the following warning, but the build will still succeed:
> warn  Dependency IHE.ITI.PIXm contains uppercase characters, which is discouraged. SUSHI will use ihe.iti.pixm as the package name.

### DIFF
--- a/src/import/loadConfigurationFromIgResource.ts
+++ b/src/import/loadConfigurationFromIgResource.ts
@@ -58,6 +58,7 @@ export function loadConfigurationFromIgResource(igRoot: string): Configuration |
   });
   // Extract the configuration from the resource
   if (igResource && igResource.url && !multipleIgs) {
+    logger.info(`Extracting FSHOnly configuration from ${igPath}...`);
     const config = {
       canonical: igResource.url.replace(/\/ImplementationGuide.*/, ''),
       version: igResource.version,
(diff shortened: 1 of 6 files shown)


julianxcarter (approved):

jafeltra (approved):
This makes sense to me!

This does make me wonder a little if the warning is still valid. Is this IG doing anything wrong by including upper case letters? Is that documented somewhere or it a best practice we are encouraging? Either way, it's not something we should fix here since this aligns it with the other warning. Just figured I'd ask since I was thinking about it.

PR #985, opened 2021-12-15
Apply insert rules on Instances when fishing for metadata

### LINKED ISSUE #958: Cannot get Canonical for Instances when url is assigned via a RuleSet
Reported by Brian Kaney on Zulip: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Canonical.20with.20rule-set

Suppose a rule-set that sets the URL and version
```
RuleSet: LibraryMetadata(id)
* url = "http://fhir/ig/Library/{id}"
* version = "0.1.0"
```

When I insert it in a resource, called MyLibrary and try to make a canonical reference, it fails:
```
* library = Canonical(MyLibrary)
```

with:
> error Cannot use canonical URL of MyLibrary because it does not exist. Be sure that MyLibrary exists and it has a URL.

It seems that unless url is explicitly set (e.g. not via a rule-set), the canonical doesn't work...

mint-thompson:
Fixes #958 and completes task [CIMPL-857](https://standardhealthrecord.atlassian.net/browse/CIMPL-857).

An Instance's url may be set by a rule from a RuleSet. Apply insert rules on the found instance so that these rules can be found in order to correctly determine the url.

It is not necessary to do this with other FSH types because Instances are exported after all other FSH types. So, the correct url can be found on the exported resource in the Package.

### DIFF
--- a/src/import/FSHTank.ts
+++ b/src/import/FSHTank.ts
@@ -15,7 +15,7 @@ import {
 } from '../fshtypes';
 import { AssignmentRule } from '../fshtypes/rules';
 import { Type, Metadata, Fishable } from '../utils/Fishable';
-import { getUrlFromFshDefinition } from '../fhirtypes/common';
+import { getUrlFromFshDefinition, applyInsertRules } from '../fhirtypes/common';
 import flatMap from 'lodash/flatMap';
 
 export class FSHTank implements Fishable {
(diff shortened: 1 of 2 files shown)


jafeltra (approved):
Makes sense to me!

cmoesel (approved):
I agree with @jafeltra.  It looks like a nice clean solution and it works.  I've run also run it through regression and confirmed it does not cause any changes.

Issue #986, opened 2021-12-15
Improve the message when sushi detects files in deep nested folders


costateixeira:
Sushi seems a bit harsh when it detects that there are resources in nested folders:
> The following files were not added to the ImplementationGuide JSON because they are nested too deep. While SUSHI supports these paths, the IG Publisher does not. To fix this, move these files so they are directly under a supported input folder (e.g., input/resources, input/profiles, etc.)

In fact, the IG can support any folder in any depth, as long as it is added as an IG parameter (path-resource). The templates add a few reference folders (input/resources, input/vocabulary, ...) but the author can add more also with sushi-config.

To accommodate for this, a better message is proposed below: 

> The following files were not added to the ImplementationGuide JSON because they are not in one of the supported input folders or are nested too deep in one of those folders. While SUSHI automatically supports resources in sub-folders, the IG Publisher does not, unless it is specifically told to do so - via the template or an IG parameter. To fix any issues you may encounter due to this, move these files so they are directly under a supported input folder (e.g., input/resources, input/profiles, etc.) or adjust your IG parameters or template accordingly.

Additionally, of course, if sushi-config has the path-resources parameter set, sushi should look into that path as well and not warn about those resources found there.

PR #987, opened 2021-12-15
#986 - improve the warning message

costateixeira:
Note that this does not solve the (missing?) feature that sushi should not warn about the files that are in folders explicitly added via sushi-config.yaml.


cmoesel (approved):

cmoesel:
I think one approval is sufficient for this one, so I am merging.

PR #991, opened 2021-12-20
Fix output paths

### LINKED ISSUE #990: Malicious IDs allow for path traversal
Since SUSHI uses resource ids and IG ids to name files, it is possible to provide an id w/ path traversal characters (e.g., `/../../` to write to an arbitrary location).  Since SUSHI only writes `JSON` files, however, note that this can only be used to write/overwrite files w/ a `.json` extension.

cmoesel:
Fixes #990.

Since SUSHI uses resource ids and IG ids to name files, it is possible to provide an id w/ path traversal characters (e.g., `/../../` to write to an arbitrary location).  Since SUSHI only writes `JSON` files, however, note that this can only be used to write/overwrite files w/ a `.json` extension.

You can test this by first checking out only 7c8a745 and confirming the tests do not pass.  Then checking out the rest of the branch and confirming the tests pass.

I am also attaching a simple project ([NaughtyPaths.zip](https://github.com/FHIR/sushi/files/7748077/NaughtyPaths.zip)) that exercises the bug.  If you unzip it and run SUSHI on it:
* the currently released SUSHI will write SURPRISE.json and BOO.json files to your project root (instead of `fsh-generated/resources`)
* this PR branch will write them to `fsh-generated/resources` as expected, with problematic characters converted to `-`.

### DIFF
--- a/src/fhirtypes/CodeSystem.ts
+++ b/src/fhirtypes/CodeSystem.ts
@@ -53,7 +53,7 @@ export class CodeSystem {
    * @returns {string} the filename
    */
   getFileName(): string {
-    return sanitize(`CodeSystem-${this.id.replace(/\/\\/g, '-')}.json`, { replacement: '-' });
+    return sanitize(`CodeSystem-${this.id}.json`, { replacement: '-' });
   }
 
   /**
--- a/src/fhirtypes/CodeSystem.ts
+++ b/src/fhirtypes/CodeSystem.ts
@@ -1,3 +1,4 @@
+import sanitize from 'sanitize-filename';
 import { cloneDeep } from 'lodash';
 import { Meta } from './specialTypes';
 import { Extension } from '../fshtypes';
@@ -52,7 +53,7 @@ export class CodeSystem {
    * @returns {string} the filename
    */
   getFileName(): string {
-    return `CodeSystem-${this.id}.json`;
+    return sanitize(`CodeSystem-${this.id.replace(/\/\\/g, '-')}.json`, { replacement: '-' });
   }
 
   /**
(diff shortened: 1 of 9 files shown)


cmoesel:
BTW -- I searched the whole code base for all cases where we write files to disk to ensure I had captured all the potential issues.  I also ran a full regression and confirmed this does not affect an IGs -- which means that no one was using invalid characters, which is good!

julianxcarter (approved):
Looks good!

src/fhirtypes/CodeSystem.ts
```diff
@@ -52,7 +53,7 @@ export class CodeSystem {
    * @returns {string} the filename
    */
   getFileName(): string {
-    return `CodeSystem-${this.id}.json`;
+    return sanitize(`CodeSystem-${this.id.replace(/\/\\/g, '-')}.json`, { replacement: '-' });
```
jafeltra, line 56: Do we still need the `replace` with the regular expression if it is inside `sanitize`?

jafeltra (commented):
Looks good - I just had one minor comment.

src/fhirtypes/CodeSystem.ts
cmoesel: Doh.  I had started by using a replace and then switched to sanitize and forgot to remove that!

jafeltra, line 56: Cool! Handy dandy suggestion to fix it!

```suggestion
    return sanitize(`CodeSystem-${this.id}.json`, { replacement: '-' });
```

jafeltra (approved):
Amazing work!

PR #992, opened 2021-12-20
Update dependencies

cmoesel:
Update dependencies to address npm audit.  This required changes in the eslint config and jest config as well.  I've removed `no-use-before-define` from `.eslintrc` because it wasn't working properly and it seemed ok to go w/ the defaults for that one.  Any real ordering issues should be caught by `tsc` anyway (in my experience).

We still get one report in audit (due to cpy-cli), but we've assessed this in the past and confirmed it should not be an issue for us.


jafeltra (approved):
Fancy upgrades ✨ 

Sad that `cpy-cli` still hasn't updated though.

PR #1000, opened 2022-01-04
All resource types supported as instance values on fixed value rules

### LINKED ISSUE #971: Cannot assign ValueSets to contained element in instances
If you create a value set using `ValueSet:`, then you _can_ assign it to `^contained` in a profile, but you _can't_ assign it to `contained` in an instance.  But... it _should_ work in both cases.  _(NOTE: We should also ensure it works for other things like `CodeSystem`, `Profile`, `Extension`, etc)._

FSH Online: https://fshschool.org/FSHOnline/#/share/3rLqOJ5

julianxcarter:
Fixes #971, allowing `ValueSets` (and all other resource types) to be included on Instances as the value in fixed value rules. Also changes the copyrightDate in our `sushi-config.yaml` fixtures to reflect the new year.

### DIFF
--- a/src/export/InstanceExporter.ts
+++ b/src/export/InstanceExporter.ts
@@ -318,6 +318,12 @@ export class InstanceExporter implements Fishable {
       if (fshDefinition) {
         this.exportInstance(fshDefinition);
         result = this.pkg.fish(item, Type.Instance) as InstanceDefinition;
+      } else {
+        // If we don't find any Instances with the name, fish for other resources
+        const fishedFHIR = this.fisher.fishForFHIR(item);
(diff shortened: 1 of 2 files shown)


cmoesel (approved):
This looks good to me.  I'm glad the solution was so simple! (Although I'm sure it took some time to investigate and arrive at it).  In fact, it was so simple that I thought it must be wrong at first -- but nope, it's right!

I'm not running a regression because @jafeltra is already doing that.  But assuming it passes regression then it's all good!

jafeltra (approved):
I ran the regression and everything looks good to me! I also tested this with a couple other resource types, and everything works as expected!

PR #1012, opened 2022-01-28
Replace file copying command line utility

### LINKED ISSUE #1004: Replace cpy-cli w/ something else
The `cpy-cli` library brings in a dependency w/ a high vulnerability.  We've been waiting a long time for `cpy-cli` to address this, but it seems like maybe it is no longer maintained.  So... we should switch to something else.

I think we only use it in the `package.json` scripts.

mint-thompson:
Fixes #1004 and completes task [CIMPL-875](https://standardhealthrecord.atlassian.net/browse/CIMPL-875).

The cpy-cli package has a deep dependency with a security vulnerability. It is replaced by the copyfiles package. This is used by the build script. The source path is quoted because the copyfiles documentation recommended quoting paths that contain globs.

To test, run `npm clean-install`, which will clear out your `node_modules` directory and install based on the updated package requirements. Then, run `npm build`, and check to see that the `dist/utils/init-project` directory contains all the files from `src/utils/init-project`.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1487,16 +1487,6 @@
         }
       }
     },
-    "@mrmlnc/readdir-enhanced": {
-      "version": "2.2.1",
-      "resolved": "https://registry.npmjs.org/@mrmlnc/readdir-enhanced/-/readdir-enhanced-2.2.1.tgz",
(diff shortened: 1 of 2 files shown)


cmoesel (approved):
It works well on Mac.  I had to lookup the docs to see what `-u 3` does. Now I get it (although I think it's kind of weird that you need it).  Anyway... Perfect.  Thanks, @mint-thompson!

julianxcarter (approved):

PR #1013, opened 2022-02-02
Don't clone original ElementDefinition's StructureDefinition

mint-thompson:
Completes task [CIMPL-890](https://standardhealthrecord.atlassian.net/browse/CIMPL-890).

When cloning an ElementDefinition, save and remove the referenced StructureDefinitions on the element and the element's original version. Then, after cloning the other properties, restore those references.

By not duplicating the (often large) object at `ElementDefinition._original.structDef` during the `clone` process, time and memory usage are reduced. This reduction is sufficient to avoid the previously-occurring errors reported in the issue description.

When running a full regression, only one IG showed changes: https://github.com/aehrc/primary-care-data-technical/archive/master.zip. Specifically, the `AUPrimaryCareMedication` profile's `Medication.ingredient.itemCodeableConcept` element is different. The current version of SUSHI produces this element:
```json
{
  "id": "Medication.ingredient.itemCodeableConcept",
  "path": "Medication.ingredient.itemCodeableConcept",
  "mustSupport": true
}
```

The version of SUSHI as of this PR produces this element:
```json
{
  "id": "Medication.ingredient.itemCodeableConcept",
  "path": "Medication.ingredient.itemCodeableConcept",
  "min": 0,
  "max": "1",
  "type": [
    {
      "code": "CodeableConcept"
    }
  ],
  "mustSupport": true
}
```
There is only one rule for this element:
```
* ingredient.itemCodeableConcept[itemCodeableConcept] MS
```
The actual JSON definition of the resource from the downloaded `hl7.fhir.au.base#current` dependency at https://build.fhir.org/ig/hl7au/au-fhir-base/StructureDefinition-au-medication.html does not contain any constraints on this choice. So, my best guess is that the output with the cardinality and type present is correct, as the individual choice-slice is only now present, and it receives the default cardinality for such an element.

### DIFF
--- a/src/fhirtypes/ElementDefinition.ts
+++ b/src/fhirtypes/ElementDefinition.ts
@@ -385,10 +385,11 @@ export class ElementDefinition {
   /**
    * ElementDefinition is capable of producing its own differential, based on differences from a stored "original".
    * This function captures the current state as the "original", so any further changes made would be captured in
-   * the generated differential.
+   * the generated differential. The structDef reference isn't used in the differential, so it can be removed.
    */
   captureOriginal(): void {
     this._original = this.clone();
+    this._original.structDef = undefined;
   }
 
   /**
@@ -2467,18 +2468,9 @@ export class ElementDefinition {
     // We don't want to clone the reference to the StructureDefinition, so temporarily save it and remove it
     const savedStructDef = this.structDef;
     this.structDef = null;
-    // We don't want to clone the reference to the StructureDefinition on the original, either
-    let originalStructDef: StructureDefinition;
-    if (this._original?.structDef != null) {
-      originalStructDef = this._original.structDef;
-      this._original.structDef = null;
-    }
     const clone = cloneDeep(this);
     // Set the reference to the StructureDefinition again
     this.structDef = clone.structDef = savedStructDef;
-    if (originalStructDef != null) {
-      this._original.structDef = clone._original.structDef = originalStructDef;
-    }
     // Clear original if applicable
     if (clearOriginal) {
       clone.clearOriginal();
--- a/src/fhirtypes/ElementDefinition.ts
+++ b/src/fhirtypes/ElementDefinition.ts
@@ -2467,9 +2467,18 @@ export class ElementDefinition {
     // We don't want to clone the reference to the StructureDefinition, so temporarily save it and remove it
     const savedStructDef = this.structDef;
     this.structDef = null;
+    // We don't want to clone the reference to the StructureDefinition on the original, either
+    let originalStructDef: StructureDefinition;
+    if (this._original?.structDef != null) {
+      originalStructDef = this._original.structDef;
+      this._original.structDef = null;
+    }
     const clone = cloneDeep(this);
     // Set the reference to the StructureDefinition again
     this.structDef = clone.structDef = savedStructDef;
+    if (originalStructDef != null) {
+      this._original.structDef = clone._original.structDef = originalStructDef;
+    }
     // Clear original if applicable
     if (clearOriginal) {
       clone.clearOriginal();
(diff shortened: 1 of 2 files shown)


cmoesel:
I need to look into that one difference, because I really don't understand why there would be _any_ difference at all.  That is perplexing.  BUT... this is a _brilliant_ find, @mint-thompson.  I can't believe we missed doing this in the `_original` (after so carefully doing it in `this`).

Looking at the code, I'm wondering if `_original` even needs the `structDef` at all?  It looks to me like we access properties of `_original`, but we never call any functions on it -- so I doubt that `_original` even needs it.  Do you think it might make sense to just delete it from `_original` when we set it in `captureOriginal()`?  Or do you think it's better to keep it (just in case we ever do something else w/ `_original` in the future)?

cmoesel:
@mint-thompson -- I just ran a mini-regression including the project you mention above by changing `regression/repos-select.txt` to contain only two entries:
```
HL7/fhir-mCODE-ig#master
aehrc/primary-care-data-technical#master
```

I ran the regression via `npm run regression`, which compares the current SUSHI branch (`cimpl-890-original-clone`) against the `master` SUSHI branch.  The regression reported back `No changes detected.`  I really couldn't figure out how your changes could mess anything up and my regression seems to confirm my suspicion: it _doesn't_ mess things up.

I'll try running a full regression overnight to make sure I'm not missing anything, but perhaps you should also try another regression (narrowing in on just that project).

That said, I can't explain how or why you got changes before.  The primary-care-data-technical project did have a few new commits during the day, but since we only download it once and then copy it for each run, that shouldn't affect anything.  It does depend on a current IG, which means that it's possible that the version downloaded during the sushi-gh-master run was different than the one downloaded in the sushi-local run, but... that dependency didn't change at all today, so that also seems unlikely.  So... I feel like your PR is probably 100% OK.  But there is still an element of mystery here!  (But I'm not sure the mystery is worth investigation, partially because I don't know how we would).

mint-thompson:
> Looking at the code, I'm wondering if `_original` even needs the `structDef` at all? It looks to me like we access properties of `_original`, but we never call any functions on it -- so I doubt that `_original` even needs it. Do you think it might make sense to just delete it from `_original` when we set it in `captureOriginal()`? Or do you think it's better to keep it (just in case we ever do something else w/ `_original` in the future)?

I think it's reasonable to just remove `structDef` from `_original` when capturing the original version. `_original` is only used for capturing properties that would be on the `ElementDefinition` JSON as listed in the `PROPS` array. Nothing ever attempts to read the value at `_original.structDef`, so we can keep it with the other write-only values. So, if it's never there in the first place, there's no need for steps in `clone` to remove and restore it.

I also ran the regression again, and this time, there were no differences. So, maybe my computer just received some cosmic rays during the previous regression check.

julianxcarter (commented):
1. This is iconic. 
2. I ran a full regression and did encounter the change that Mint initially mentioned in the  `aehrc/primary-care-data-technical` project. I also encountered some changes in the [`hl7ch/ch-etoc#main` project](https://github.com/hl7ch/ch-etoc) that were a bit more pervasive. I've attached the diff folder to this element so that we can all see. I haven't figured out exactly why this is happening, but I'm about to look further into it. 

[hl7ch-ch-etoc-main.zip](https://github.com/FHIR/sushi/files/7996953/hl7ch-ch-etoc-main.zip)

cmoesel:
@julianxcarter -- I'm wondering if maybe there is a caching issue and your first time running the regression, maybe it is using older cached versions of dependencies.  Could you try editing `regression/repos-select.txt` to include just those two libraries that caused diffs for you, and then run `npm run regression` to re-run the regression on them again?  I'm wondering if they'll come out clean on the second run for you.

julianxcarter:
@cmoesel They did come out clean on a second run with `repos-select` instead of `repos-all`. It's definitely a caching issue

julianxcarter (approved):

cmoesel (approved):
This is my favorite PR for a while.  Even with the weird caching issues (that are totally unrelated to this PR).

PR #1014, opened 2022-02-03
path-resource parameter support for Predefined Resources

### LINKED ISSUE #986: Improve the message when sushi detects files in deep nested folders

julianxcarter:
Fixes #986, making It so that the `path-resource` parameter in `sushi-config.yaml` or an included ImplementationGuide resource are recognized, and resources in those listed directories are added as predefinedResources. Running sushi on the included FSH project should show this in action, as the config file includes a `path-resource` parameter that causes the extension included in `input/predefined-resource` to be included as a predefined resource as sushi runs.

[PathResourceTest.zip](https://github.com/FHIR/sushi/files/7995282/PathResourceTest.zip)


src/fhirdefs/load.ts
```diff
@@ -203,8 +203,13 @@ export function cleanCachedPackage(packageDirectory: string): void {
...
  * @param {string} resourceDir - The path to the directory containing the resource subdirs
  * @param {FHIRDefinitions} defs - The FHIRDefinitions object to load definitions into
+ * @param {string[]} pathResourcePaths - optional, an array of paths to additional directories containing predefined resources
  */
-export function loadCustomResources(resourceDir: string, defs: FHIRDefinitions): void {
+export function loadCustomResources(
+  resourceDir: string,
+  defs: FHIRDefinitions,
+  pathResourcePaths: string[] = null
+): void {
```
cmoesel, line 212: I feel like it makes more sense to put this new parameter directly after the `resourceDir` parameter -- because it is very similar.  That might mean having to change a few more places where this function is called, but I assume that's not very many places anyway.  I also think that `additionalResourcePaths` might be a better name.

src/fhirdefs/load.ts
```diff
@@ -216,11 +221,13 @@ export function loadCustomResources(resourceDir: string, defs: FHIRDefinitions):
...
   ];
+  const predefinedResourcePaths = pathEnds.map(pathEnd => path.join(resourceDir, pathEnd));
+  if (pathResourcePaths) predefinedResourcePaths.push(...pathResourcePaths);
   const converter = new FHIRConverter();
   let invalidFileCount = 0;
-  for (const pathEnd of pathEnds) {
+  for (const dirPath of predefinedResourcePaths) {
     let foundSpreadsheets = false;
-    const dirPath = path.join(resourceDir, pathEnd);
+    // const dirPath = path.join(resourceDir, pathEnd);
```
cmoesel, line 230: You can remove this commented out line.

src/fhirdefs/load.ts
```diff
@@ -257,7 +264,7 @@ export function loadCustomResources(resourceDir: string, defs: FHIRDefinitions):
         // All resources are added to the predefined map, so that this map can later be used to
         // access predefined resources in the IG Exporter
         defs.addPredefinedResource(file, resourceJSON);
-        if (pathEnd !== 'examples') {
+        if (!dirPath.endsWith('examples')) {
```
cmoesel, line 267: I think that using [path.basename](https://nodejs.org/api/path.html#pathbasenamepath-ext) instead of endsWith would be closer to the original intent (and wouldn't match on things like `input/definitelynotexamples`).

src/import/loadConfigurationFromIgResource.ts
```diff
@@ -64,7 +64,8 @@ export function loadConfigurationFromIgResource(igRoot: string): Configuration |
       version: igResource.version,
       fhirVersion: igResource.fhirVersion ?? [],
       dependencies: igResource.dependsOn?.filter(dep => dep.packageId && dep.version),
-      FSHOnly: true
+      FSHOnly: true,
+      parameters: igResource.definition?.parameter ?? []
```
cmoesel, line 68: Another petty comment, but... I like keeping `FSHOnly` as the last thing.  This would also mirror the order that we document it on FSHSchool ([link](https://fshschool.org/docs/sushi/configuration/)).

cmoesel (commented):
This is looking good!  I haven't had the chance to test it yet, but I have a few minor comments on the code for your consideration.

mint-thompson (changes requested):

src/fhirdefs/load.ts
julianxcarter: Whoops! Removed in commit `a080a34`.

src/app.ts
```diff
@@ -152,7 +152,13 @@ async function app() {
   await loadExternalDependencies(defs, config);
 
   // Load custom resources. In current tank configuration (input/fsh), resources will be in input/
-  loadCustomResources(path.join(input, '..'), defs);
+  const pathResources = config.parameters
+    ?.filter(parameter => parameter.value && parameter.code === 'path-resource')
+    .map(code => code.value);
```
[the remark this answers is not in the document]
julianxcarter: That makes more sense. I changed this in commit a080a34.

src/fhirdefs/load.ts
julianxcarter: Changed in commit a080a34! That should make this more robust.

src/fhirdefs/load.ts
julianxcarter: I actually moved all of the `config.paramater` parsing into `load.ts` so we could get that extra bit of code coverage. Let me know what you think of the new parameter names and ordering.

src/import/loadConfigurationFromIgResource.ts
julianxcarter: Addressed in commit a080a34.

src/fhirdefs/load.ts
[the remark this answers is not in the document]
julianxcarter: I chose to use `path.basename(dirPath)` instead per @cmoesel's suggestion, which accomplishes the same thing.

src/fhirdefs/load.ts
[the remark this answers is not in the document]
julianxcarter: Removed in in commit a080a34.

src/app.ts
```diff
@@ -152,7 +152,13 @@ async function app() {
...
+  const pathResourceDirectories = pathResources
+    .map(directoryPath => path.join(originalInput, directoryPath))
+    .filter(directoryPath => fs.existsSync(directoryPath));
+  loadCustomResources(path.join(input, '..'), defs, pathResourceDirectories);
```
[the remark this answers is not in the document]
julianxcarter: That makes sense! I moved all of the config parsing into `load.ts` in commit a080a34, and updated tests accordingly. So now all of the changes in this PR have test coverage.

cmoesel (changes requested):
A couple of small things...

src/fhirdefs/load.ts
```diff
@@ -202,9 +203,17 @@ export function cleanCachedPackage(packageDirectory: string): void {
...
+ * @param {string} projectDir - User's specified project directory
+ * @param {ImplementationGuideDefinitionParameter[]} configParameters - optional, an array of config parameters in which to
+ *    determine if there are addiional resource paths for predefined resource
  * @param {FHIRDefinitions} defs - The FHIRDefinitions object to load definitions into
  */
-export function loadCustomResources(resourceDir: string, defs: FHIRDefinitions): void {
+export function loadCustomResources(
+  resourceDir: string,
+  configParameters: ImplementationGuideDefinitionParameter[] = null,
+  projectDir: string = null,
```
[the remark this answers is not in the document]
julianxcarter: I'm glad you noticed! This was actually an elaborate scheme to prompt you to tell me your preferred order (read: actually a mistake). The implementation's parameter order has been updated to match the doc's order in commit `629a766` 🙂

src/fhirdefs/load.ts
```diff
@@ -202,9 +203,17 @@ export function cleanCachedPackage(packageDirectory: string): void {
 /**
  * Loads custom resources defined in resourceDir into FHIRDefs
  * @param {string} resourceDir - The path to the directory containing the resource subdirs
+ * @param {string} projectDir - User's specified project directory
+ * @param {ImplementationGuideDefinitionParameter[]} configParameters - optional, an array of config parameters in which to
+ *    determine if there are addiional resource paths for predefined resource
```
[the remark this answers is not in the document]
julianxcarter: Fixed in commit 629a766

cmoesel (approved):
This looks good and works well!  Thanks, @julianxcarter!

mint-thompson (approved):

PR #1015, opened 2022-02-04
Make Up IG URL for dependencies without an IG

### LINKED ISSUE #1007: Make Up IG URL for dependencies without an IG
Currently, if you specify a dependency by id, but that dependency package does not contain an IG resource, SUSHI will log an error like this:

> Failed to add de.basisprofil.r4:1.1.0 to ImplementationGuide instance because SUSHI could not find the IG URL in the dependency IG. To specify the IG URL in your sushi-config.yaml, use the dependency details format:
>
> ```yaml
> dependencies:
>   de.basisprofil.r4:
>     uri: http://my-fhir-ig.org/ImplementationGuide/123
>     version: 1.1.0
> ```

It is up to the IG author to determine what the uri should be, but it's often unclear.

Grahame Grieve has recently looked at this issue and indicated that when a package does not have an IG resource, the following format should be used as the IG URL when specifying it as a dependency: `http://fhir.org/packages/${packageId}/ImplementationGuide/${packageId}`.  In the example above, this would mean that the correct uri to use is: `http://fhir.org/packages/de.basisprofil.r4/ImplementationGuide/de.basisprofil.r4`.

Now that there is a defined approach, SUSHI should be updated so that it no longer produces an error in this case, but rather, simply uses a URI as specified by Grahame.

See: https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/fhir.2Edicom.20has.20no.20ImplementationGuide.20resource/near/265772431

guhanthuran:
Also updated some dependencies with vulnerabilities in npm

Fixes #1007

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -4426,9 +4426,9 @@
       "integrity": "sha512-GRnmB5gPyJpAhTQdSZTSp9uaPSvl09KoYcMQtsB9rQoOmzs9dH6ffeccH+Z+cv6P68Hu5bC6JjRh4Ah/mHSNRw=="
     },
     "follow-redirects": {
-      "version": "1.14.6",
-      "resolved": "https://registry.npmjs.org/follow-redirects/-/follow-redirects-1.14.6.tgz",
-      "integrity": "sha512-fhUl5EwSJbbl8AR+uYL2KQDxLkdSjZGR36xy46AO7cOMTrCMON6Sa28FmAnC2tRTDbd/Uuzz3aJBv7EBN7JH8A=="
+      "version": "1.14.7",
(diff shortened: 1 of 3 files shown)


cmoesel:
This is all working well for me, but when I run `npm install`, it changes the `npm-shrinkwrap.json` file (by adding `"dev": true` to a few entries).  @guhanthuran -- what version of node and npm are you using?

guhanthuran:
@cmoesel I'm using v12.22.9 for node and 6.14.15 for npm

cmoesel:
Interesting.  I'm using Node 12.22.10 and NVM 6.14.16.  Let's see if the next reviewer has the same experience as I did or if maybe it is just me.

julianxcarter (approved):
Looks good to me! Nice work @guhanthuran

cmoesel (approved):
Looks good! I think our users will really appreciate this!

PR #1016, opened 2022-02-09
Canonical keyword support for Implementation Guide url

### LINKED ISSUE #963: Support Canonical keyword w/ package id and name
In some cases (e.g., writing a `CapabilityStatement`), the author needs to refer to the canonical URL for the IG itself.  Currently, that would have to be hard-coded, but the `Canonical` keyword should support passing in the package id or name (specified in sushi-config.yaml) to resolve it to the Canonical URL.

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Canonical.20of.20the.20IG.20itself.3F

julianxcarter:
Fixes #963, allowing the canonical of the `ImplementationGuide` itself to be generated by the `Canonical()` shortcut when the user supplied the `packageId`, `name`, or `id` of the package. If the configuration in the package does not include a `url` property, it is generated with the pattern `<package-canonical>/ImplementationGuide/<package-id>`.


mint-thompson (dismissed):
Hooray for predefined resources!

julianxcarter:
> This looks good! As a minor addition, could you add a test case for a configuration that contains a `url` to demonstrate that it will take precedence?

This is a great idea! I've added a unit test that shows this behavior.

mint-thompson (approved):

cmoesel:
Ah-ha.  So that's why it didn't work w/ the IG resource!  Good work!  I think, however, it should go even further.

There isn't actually a rule that the IG resource's URL _has_ to follow a particular format.  It could be different than `${canonical}/ImplementationGuide/${packageId}`.  So I think you also need to bring in the IG's `url` property.  And while you're at it, you might as well bring in `id` too (although that one is 0..1, so only bring it in if it is present) -- that way the logic is consistent across the board (since sushi-config.yaml usually has an id).

src/import/loadConfigurationFromIgResource.ts
```diff
@@ -69,7 +70,8 @@ export function loadConfigurationFromIgResource(igRoot: string): Configuration |
       parameters: igResource.definition?.parameter ?? [],
       FSHOnly: true
     };
-    config.dependencies?.forEach(dep => {
+    if (igResource.id) config.id = igResource.id;
```
cmoesel, line 73: This is a little nitpicky, but... I think most IGs will have an `id` and it looks a little weird to me coming last in the log (because we almost always put `id` first):
```
info  Extracted configuration:
info    canonical: "http://example.org/fshonly"
info    url: "http://example.org/fshonly/ImplementationGuide/tricked-you-because-you-thought-it-was-fhir.example.fshonly"
info    name: undefined
info    packageId: "fhir.example.fshonly"
info    version: "0.1.0"
info    fhirVersion[0]: "4.0.1"
info    dependencies: undefined
info    parameters[0]: {"code":"copyrightyear","value":"2022+"}
info    parameters[1]: {"code":"releaselabel","value":"ci-build"}
info    FSHOnly: true
info    id: "fhir.example.fshonly"
```

What do you think about putting:
```
id: igResource.id,
```
as the _first property_ in the original config object and then in this spot (after the config object) deleting it if it is null/undefined?
```ts
if (config.id == null) {
  delete config.id;
}
```
(Also, I guess prettier doesn't enforce it, but I think we usually use fully formed `if` statements, even when they're a single statement.  I _think_.)

cmoesel (commented):
One more nit-picky comment.  If you really disagree then you can push back.

src/import/loadConfigurationFromIgResource.ts
julianxcarter: That makes sense! Commit [fd89bcd](https://github.com/FHIR/sushi/pull/1016/commits/fd89bcd24176953ce2b4a340e1d502c57db79ef4) implements this suggestion.

cmoesel (approved):
This looks great, @julianxcarter!  Thank you!

mint-thompson (approved):

PR #1026, opened 2022-02-22
Replace package loading with FPL

jafeltra:
This PR supports using the [FHIR Package Loader](https://github.com/standardhealth/fhir-package-loader/) to download any FHIR packages and add the definitions to `FHIRDefinitions`.

This PR removes any of the functions and tests that were moved to the separate library. It adds in the new functions from that library and uses them in place of the removed functions.

In order to test this, you will need to locally link the FHIR Package Loader with SUSHI. I'm not sure of how link it in a non-local way before the library is published to NPM because the directory with the complied JS is not tracked in git and we don't want to track an `npm pack`ed filed. Because of that, the action aren't passing. To link it locally, you can either use `npm link` to create a link between SUSHI and the compiled JS (let me know if you want more detailed instructions about for option) or you can use `npm pack`. You can run `npm pack` in the FHIR Package Loader library, which will create a `.tgz` file. Then in SUSHI, you can add the following dependency to the `dependencies` property in `package.json`: `"fhir-package-loader": "file:<relative-path-to-npm-pack-result>.tgz"`. Then if you run `npm install`, it will use that version of the FHIR Package Loader and SUSHI should work as expected.

Once you link the fhir-package-loader library locally, test that SUSHI still works as expected. Ensure that the functionality works as expected, that the new library works how you'd expect it to with SUSHI, that the tests that were removed make sense to have been moved, and that I didn't leave anything new uncovered. I've run the full regression script and there aren't any changes.

**NOTE**: Don't merge this until the dependency is added to `package.json`. Once that happens, the actions should pass.


cmoesel:
I converted this to a `DRAFT` PR since it will need to be updated once the FPL is published (and shouldn't be merged until then).

src/fhirdefs/FHIRDefinitions.ts
```diff
@@ -1,35 +1,23 @@
+import { flatten } from 'lodash';
+import {
+  cloneJsonMapValues,
+  addDefinitionToMap,
```
cmoesel, line 4: I feel like this function is an internal implementation detail that should not be exported from `fhir-package-loader`.  I think we can rewrite the code in SUSHI so it does not use it and then eliminate this import (and FPL export).  See below for suggestion.

src/fhirdefs/FHIRDefinitions.ts
```diff
@@ -43,97 +31,27 @@ export class FHIRDefinitions implements Fishable {
...
-      addDefinitionToMap(definition, this.codeSystems);
-    } else if (
-      definition.resourceType === 'ImplementationGuide' &&
-      !this.isSupplementalFHIRDefinitions
-    ) {
-      addDefinitionToMap(definition, this.implementationGuides);
+    } else {
+      super.add(definition);
     }
   }
```
cmoesel, line 56: As noted above, I think that `addDefinitionToMap` shouldn't be exported from FPL (because it is an implementation detail).  Since the FPL `FHIRDefinition` adds the appropriate types to the appropriate maps already, then we can get away with using `super.add` instead (since the difference in this code is just _when_ to add, not _how_ to add).  I tested this implementation instead and it seemed to work (or at least pass the unit tests):
```ts
  add(definition: any): void {
    // For supplemental FHIR versions, we only care about resources and types,
    // but for normal packages, we care about everything.
    if (this.isSupplementalFHIRDefinitions) {
      if (
        definition.resourceType === 'StructureDefinition' &&
        (definition.kind === 'primitive-type' ||
          definition.kind === 'complex-type' ||
          definition.kind === 'datatype' ||
          (definition.kind === 'resource' && definition.derivation !== 'constraint'))
      ) {
        super.add(definition);
      }
    } else {
      super.add(definition);
    }
  }
```

cmoesel (changes requested):
I do love me some deleted code!

I've yet to review the actual FPL, but I have a few comments here regarding functions that I think we should _not_ import from FPL (along w/ proposed replacement code).  These comments also affect the FPL PR (since I'm suggesting you do not export these functions).

jafeltra:
I think all of these suggestions make sense. I realized that I actually already stopped exporting both`addDefinitionToMap` and `cloneJsonMapValues` from FPL after the latest changes I made there, but didn't even realize the implications on SUSHI. So I think I must have had a similar thought process and just stopped thinking too soon. I'll update to fix that, which I think will address all of the comments above.

src/fhirdefs/FHIRDefinitions.ts
```diff
@@ -1,35 +1,23 @@
+import { flatten } from 'lodash';
+import {
+  cloneJsonMapValues,
```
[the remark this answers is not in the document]
jafeltra: Whoops, I had already removed the export of `cloneJsonMapValues` from FPL, so this wasn't working with the latest version of FPL anyway 😅 

Import removed in [1e1e8a9](https://github.com/FHIR/sushi/pull/1026/commits/1e1e8a9c1da76bf964a68f525572a28c3c562363).

src/fhirdefs/FHIRDefinitions.ts
```diff
@@ -43,97 +31,27 @@ export class FHIRDefinitions implements Fishable {
...
-  allCodeSystems(): any[] {
-    return cloneJsonMapValues(this.codeSystems);
-  }
-
-  allImplementationGuides(): any[] {
-    return cloneJsonMapValues(this.implementationGuides);
-  }
-
   allPredefinedResources(): any[] {
     return cloneJsonMapValues(this.predefinedResources);
```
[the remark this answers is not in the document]
jafeltra: Added inline implementation in [1e1e8a9](https://github.com/FHIR/sushi/pull/1026/commits/1e1e8a9c1da76bf964a68f525572a28c3c562363).

src/fhirdefs/FHIRDefinitions.ts
jafeltra: Same as `cloneJsonMapValues`, I had already removed the export of this from FPL, so this was broken. Oops.

Removed the import in [1e1e8a9](https://github.com/FHIR/sushi/pull/1026/commits/1e1e8a9c1da76bf964a68f525572a28c3c562363).

src/fhirdefs/FHIRDefinitions.ts
jafeltra: Simplified the `if` logic and swapped out `addDefinitionToMap` with `super.add` in [1e1e8a9](https://github.com/FHIR/sushi/pull/1026/commits/1e1e8a9c1da76bf964a68f525572a28c3c562363).

I also confirmed that the definition that gets passed in to `super.add` will still fall into the same cases in FPL, so `addDefinitionToMap` still gets called with the correct map from FPL, so this implementation seems correct to my eye and the tests.

jafeltra:
@cmoesel I believe [1e1e8a9](https://github.com/FHIR/sushi/pull/1026/commits/1e1e8a9c1da76bf964a68f525572a28c3c562363) addresses your comments so far. I don't think there is any equivalent changes that need to be made to FPL since I had already made them in [FPL commit 8da46e1](https://github.com/standardhealth/fhir-package-loader/pull/1/commits/8da46e134d1cc2d1c93563b273ab48b5f91e564c).

src/fhirdefs/FHIRDefinitions.ts
```diff
@@ -1,35 +1,19 @@
...
-  private extensions: Map<string, any>;
-  private types: Map<string, any>;
-  private valueSets: Map<string, any>;
-  private codeSystems: Map<string, any>;
-  private implementationGuides: Map<string, any>;
+export class FHIRDefinitions extends BaseFHIRDefinitions implements Fishable {
   private predefinedResources: Map<string, any>;
   private supplementalFHIRDefinitions: Map<string, FHIRDefinitions>;
-  private packageJsons: Map<string, any>;
   packages: string[];
```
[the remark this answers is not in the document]
jafeltra: Hmm this property gave me a lot of trouble, and I thought it was necessary still because the way the `FHIRDefinitions` work between SUSHI and FPL are slightly different, and I thought this property was the key to that difference. But you're right. It isn't adding to that list anymore, and I thought it was. So I'm going to look more into this and see if it actually should be using this property. If not, I'll just update the tests to get rid of it there as well.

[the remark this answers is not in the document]
julianxcarter: I do think it could potentially be useful to have a property that just lets you know which packages are contained in a `FHIRDefinitions` instance. Maybe we can add the `packages` array to the base `FHIRDefinitions` class over on the FPL side, and then add a getter that returns the current `package` property and the package property of all the `childDefs` in an array 🤔. I'm not sure if the need for that would ever come up though.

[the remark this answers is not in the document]
jafeltra: Phew 😅😰 I made it.

Turns out `packages` was really only used in tests. I spent a bit trying to rework how SUSHI actually uses the FHIRDefinitions class as part of looking in to this because it would probably be a little cleaner and let us get cleaner access to the list of packages that it updates. However, it really wasn't working completely correctly, probably would have involved more reworking than we really need right now, and would impact a bunch of tests.

So instead I just kept track of an array in the tests as each package is "loaded" and check against that array, rather than depend on a property on FHIRDefinitions. That's all done in [6f38ef1](https://github.com/FHIR/sushi/pull/1026/commits/6f38ef12511fedb585b2aa6d2125d9404955a6ea).

jafeltra:
I believe everything I was aware of is addressed here now. [6f38ef1](https://github.com/FHIR/sushi/pull/1026/commits/6f38ef12511fedb585b2aa6d2125d9404955a6ea) updates how the tests check for loaded packages, which came from @julianxcarter's previous comment, and [e42230e](https://github.com/FHIR/sushi/pull/1026/commits/e42230e800e258168301a32f066a5262fefdcdba) and [167ab32](https://github.com/FHIR/sushi/pull/1026/commits/167ab3288a015cf6c25017c6d0305b18488d7e5e) are related to changes @cmoesel commented about in FPL (there's more explanation about them on that PR).

Let me know if anyone notices anything else!

julianxcarter (dismissed):

test/fhirdefs/load.test.ts
```diff
@@ -1,427 +1,28 @@
...
-  });
-
-  it('should load base FHIR extensions', () => {
-    expect(defs.allExtensions().some(r => r.id === 'patient-mothersMaidenName')).toBeTruthy();
-  });
-
-  it('should load base FHIR value sets', () => {
-    expect(defs.allValueSets().some(r => r.id === 'allergyintolerance-clinical')).toBeTruthy();
-  });
+let loadedPackages: string[] = [];
```
cmoesel, line 8: ESLint says:
>  error  'loadedPackages' is never reassigned. Use 'const' instead  _prefer-const_

test/fhirdefs/load.test.ts
```diff
@@ -1,427 +1,28 @@
...
-        return {};
+jest.mock('fhir-package-loader', () => {
+  const original = jest.requireActual('fhir-package-loader');
+  return {
+    ...original,
+    loadDependency: jest.fn(
+      async (packageName: string, version: string, FHIRDefs: FHIRDefinitions) => {
+        // the mock loader can find R2, R3, and R5
+        if (/hl7\.fhir\.r(2|3|5).core/.test(packageName)) {
+          loadedPackages.push(`${packageName}#${version}`);
```
cmoesel, line 18: As far as I can tell, these tests don't ever actually read the `loadedPackages` variable?  Should they?  If yes, then I guess the tests need to be updated to do that.  If no, then I think `loadedPackages` can be removed from this file entirely.

test/utils/Processing.test.ts
```diff
@@ -435,47 +456,47 @@ describe('Processing', () => {
...
+        '4.0.1'
+      );
+      await testExtPackage(
+        'hl7.fhir.extensions.r4',
+        'hl7.fhir.r4.core',
+        '4.0.1',
+        'hl7.fhir.r5.core',
+        '4.5.0'
+      ),
+        await testExtPackage(
```
cmoesel, line 493: This looks like it's indented a little extra?

cmoesel (changes requested):
This is looking good.  It's nice to see so much deleted code (even if it just moved to somewhere else).  I found a couple of tiny little things to address.

test/fhirdefs/load.test.ts
jafeltra: Oops I didn't realize this was never used (and that it was using `let` above). But you're right, I took out the check where this would have been used (and didn't realize that was the only place that this was used in this file). The problem was that in the old places it would have been used ([here](https://github.com/FHIR/sushi/pull/1026/commits/6f38ef12511fedb585b2aa6d2125d9404955a6ea#diff-72b4539a412a1bf5daa36f8e26f20b222a0cdc6ef59eecb42038162acc2f2bf5L174), [here](https://github.com/FHIR/sushi/pull/1026/commits/6f38ef12511fedb585b2aa6d2125d9404955a6ea#diff-72b4539a412a1bf5daa36f8e26f20b222a0cdc6ef59eecb42038162acc2f2bf5L191), and [here](https://github.com/FHIR/sushi/pull/1026/commits/6f38ef12511fedb585b2aa6d2125d9404955a6ea#diff-72b4539a412a1bf5daa36f8e26f20b222a0cdc6ef59eecb42038162acc2f2bf5L206)), it was testing that the package was actually added to `supplementalFHIRPackages` and not added to the old `packages` list. But now that the `packages` list doesn't exist on `FHIRDefinitions` and the package is added to the static `loadedPackages` list any time a package is "loaded", there didn't seem to be anything to test for the `pacakges`/`loadedPackages` side of things. Instead, the tests just check that the package is added to `supplementalFHIRPackages`.

I can remove the `loadedPackages` property entirely. But if you disagree or think there's something I can test that I'm missing, let me know.

test/utils/Processing.test.ts
jafeltra: I guess prettier let me down. I'll fix!

jafeltra: *gasp* I apologize to Prettier for throwing it under the 🚌 . There was a comma instead of a semicolon!

jafeltra: Fixed in [377ba4f](https://github.com/FHIR/sushi/pull/1026/commits/377ba4f0f70bbf9b2226773fb692d3662cde7211).

test/fhirdefs/load.test.ts
jafeltra: Removed in [377ba4f](https://github.com/FHIR/sushi/pull/1026/commits/377ba4f0f70bbf9b2226773fb692d3662cde7211) but I can update if you have other idea from the above novel I wrote.

test/fhirdefs/load.test.ts
jafeltra: Fixed! [377ba4f](https://github.com/FHIR/sushi/pull/1026/commits/377ba4f0f70bbf9b2226773fb692d3662cde7211).

cmoesel (commented):
This is looking good, @jafeltra!  Did you by chance do a full regression after integrating the real FPL module?

jafeltra:
> Did you by chance do a full regression after integrating the real FPL module?

Ummm... I did not. I forgot. I can now though!

cmoesel (approved):
I ran the regression.  It came out clean!  Well done!

julianxcarter (approved):

PR #1027, opened 2022-02-23
Avoid call stack overflow when assigning large base64 values

### LINKED ISSUE #998: Large file causes maximum call stack size exceeded error
See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/.22Big.20data.22

mint-thompson:
Fixes #998 and completes task [CIMPL-876](https://standardhealthrecord.atlassian.net/browse/CIMPL-876).

When assigning a string to a base64Binary-type element, the value is checked to make sure it is meets the requirements of that type. The value must be composed of valid base64 chunks. Checking this one chunk at a time avoids the call stack overflow that may occur when checking the entire value all at once.

### DIFF
--- a/src/fhirtypes/ElementDefinition.ts
+++ b/src/fhirtypes/ElementDefinition.ts
@@ -1848,6 +1848,16 @@ export class ElementDefinition {
     }
   }
 
+  private isValidBase64(value: string): boolean {
+    const base64Part = /(\s*([0-9a-zA-Z\+\/=]){4}\s*)/y;
+    while (base64Part.lastIndex < value.length) {
+      if (!base64Part.test(value)) {
+        return false;
+      }
+    }
+    return true;
+  }
+
   /**
    * Assigns a string to this element.
    * @see {@link assignValue}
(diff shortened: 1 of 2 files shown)


src/fhirtypes/ElementDefinition.ts
```diff
@@ -1848,6 +1848,16 @@ export class ElementDefinition {
     }
   }
 
+  private isValidBase64(value: string): boolean {
+    const base64Part = /(\s*([0-9a-zA-Z\+\/=]){4}\s*)/y;
+    while (base64Part.lastIndex < value.length) {
```
cmoesel, line 1853: Today I learned something.  Thank you, @mint-thompson!

cmoesel (approved):
Great investigative powers, @mint-thompson -- and equally great solution!

I tested this on the graviate-health IG's commit [d2e9f4680daa6e53b778fb0229912b0d3918eaaf](https://github.com/hl7-eu/gravitate-health/commit/d2e9f4680daa6e53b778fb0229912b0d3918eaaf) -- which contains the FSH file that caused the initial error that was reported.  On `master`, SUSHI reported the `Maximum call stack size exceeded`, but on this PR's branch, it did not cause an error and the resulting JSON file looked correct!

jafeltra (approved):
Agreed! I learned something and this is super nifty! I also double checked and the `y` flag is supported by browsers, so there won't be any issues with FSH Online!

PR #1033, opened 2022-02-25
Update FSH version to 2.0.0

jafeltra:
FSH 2.0.0 is here!

This updates it so the latest FSH version is visible in the `--version` text.


guhanthuran (commented):
Good job!!! :)

PR #1034, opened 2022-03-01
Allow underscore-prefixed primitives to satisfy cardinality requirements

### LINKED ISSUE #1021: SUSHI should allow extensions on primitives to satisfy cardinality
Reported by Julian Sass on zulip [here](https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Error.20extension.20on.20element.20w.2F.20min.201.20cardinality):

SUSHI compiler throws an error for instances that use an extension like data-absent-reason on elements that have a min 1 cardinality defined in a profile. E.g.:

```
Profile: MyProcedure
Parent: Procedure
* performed[x] 1..1

Instance: my-procedure
InstanceOf: MyProcedure
* status = #not-done
* subject = Reference(Patient/example)
* performedDateTime.extension.url = "http://hl7.org/fhir/StructureDefinition/data-absent-reason"
* performedDateTime.extension.valueCode = #not-performed
```

produces
> error Element Procedure.performed[x] has minimum cardinality 1 but occurs 0 time(s).

I think the error should not be triggered.

https://fshschool.org/FSHOnline/#/share/34xsyMF

mint-thompson:
Fixes #1021 and completes task [CIMPL-902](https://standardhealthrecord.atlassian.net/browse/CIMPL-902).

This issue turns out to have been specific to primitive-type choice element slices. When checking choice elements for a required element, use the underscore-prefixed version if it exists. This is consistent with the existing logic for cases that are not choice elements.

### DIFF
--- a/src/export/InstanceExporter.ts
+++ b/src/export/InstanceExporter.ts
@@ -219,7 +219,8 @@ export class InstanceExporter implements Fishable {
           .forEach(ce => possibleChoiceSlices.push(...ce.children(true)));
         const choiceSlices = possibleChoiceSlices.filter(c => c.path === child.path && c.sliceName);
         for (const choiceSlice of choiceSlices) {
-          instanceChild = instance[choiceSlice.sliceName];
+          // as above, we use the _ prefixed element if it exists
+          instanceChild = instance[`_${choiceSlice.sliceName}`] ?? instance[choiceSlice.sliceName];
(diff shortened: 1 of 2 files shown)


jafeltra (approved):
This makes perfect sense! ✨

cmoesel (approved):
Yep.  This does the trick!

PR #164, opened 2020-01-29
Revert default status back to active

cmoesel:
Until we've discussed default status w/ the community, keep default status as active since this is what it was in previous versions of SUSHI.

Fixes #162

PR #1035, opened 2022-03-03
Update requirement for status and name

jafeltra:
We only require `name` and `status` properties in the `sushi-config.yaml` file when you are creating an IG resource. That means when you have `FSHOnly: true` in your configuration, you should not have any logged errors saying those values are required but missing. This PR updates the import of the configuration to only require those properties if `FSHOnly` is not set to true. This follows the pattern we already had for `copyrightYear` and `releaseLabel` [here](https://github.com/FHIR/sushi/blob/master/src/import/importConfiguration.ts#L687-L692). It also updates the tests to check that no errors are logged.

This should fix [GoFSH Issue #164](https://github.com/FHIR/GoFSH/issues/164) by loosening the requirement for `status`. GoFSH does not _need_ to provide `status` if it cannot find it, but SUSHI should not be expecting to find it on `sushi-config` files that are FSHOnly. When this change is merged into SUSHI and SUSHI is released, that issue will be resolved.

### DIFF
--- a/src/import/importConfiguration.ts
+++ b/src/import/importConfiguration.ts
@@ -132,10 +132,10 @@ export function importConfiguration(yaml: YAMLConfiguration | string, file: stri
     modifierExtension: yaml.modifierExtension,
     url: yaml.url ?? `${yaml.canonical}/ImplementationGuide/${yaml.id}`,
     version: normalizeToString(yaml.version), // minimum config property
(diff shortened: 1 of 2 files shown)


mint-thompson (approved):
Thank you!

cmoesel (approved):
That makes sense to me.  Thanks!

Issue #969, opened 2021-12-02
SUSHI self-update and/or out-of-date indicator


cmoesel:
A user suggested that it would be nice to have a SUSHI flag for upgrading sushi (e.g., `sushi -u`).  I'm not sure if this would work well or not, but we can look into it.

Another idea is to check if the local SUSHI is the latest version (using an npm api maybe) and print a log at the end if it is not -- along with instructions for upgrading.

cmoesel:
For now, let's detect if SUSHI is out of date and suggest updating it.  We can get package information about SUSHI by issuing an HTTP GET to https://registry.npmjs.org/fsh-sushi.

In the results, we can look at `dist-tag.latest` and compare against the currently running one.  If it's not the latest stable release, we can issue a log like:
```
You are using SUSHI version 2.2.6, but the latest stable release is version 2.2.7.
To install the latest release, run:
  npm install -g fsh-sushi
```
I like wrapping things like this in boxes so they are prominent.  It might be worth considering.  We should also experiment w/ when/where we want to report this (at the beginning of the run? or at the end of the run?)

PR #1036, opened 2022-03-07
Sushi self-update and xhtml outer div check

### LINKED ISSUE #305: Need a way to set xhtml values.
If I want to set the narrative of a resource example or anywhere else that the Narrative datatype is used, there does not seem to be a way to set a fixed xhtml value.

* section.text.div = "No adverse event happened.  Not sure why I'm sending this report."

doesn't work as it generates an error:
error Cannot fix string value: No drugs taken. Value does not match element type: xhtml

guhanthuran:
Fixes #305 and #969

For #305:
Changed the `checkXhtml` function to assure that the outer div is checked and that it both starts and ends with the appropriate `div` tag

For #969:
Added three functions: `getLocalSushiVersion()`, `getLatestSushiVersion()`, `checkSushiVersion()`. 
`getLocalSushiVersion()` checks local sushi version, outputs version as string
`getLatestSushiVersion()` makes an axios call to check the current sushi version, outputs version as string
`checkSushiVersion()`calls both of the above functions to then compare the two strings and output a warning if different

`checkSushiVersion()` is called at the end of the app.ts main process

Will probably need more tests.


mint-thompson (changes requested):
Thank you for addressing these two issues! I'm wondering about the warning given on the xhtml check. It's pretty deep into `ElementDefinition`, and the rule that assigned the value is several scopes away. Unfortunately, that means that the warning message gives no indication of where to look for the incorrect xhtml. While the FSH is long gone, maybe it could include the Element's `id` to help the user figure out where to go?

src/utils/Processing.ts
```diff
@@ -549,3 +549,33 @@ export function getFilesRecursive(dir: string): string[] {
...
+  const latest = await getLatestSushiVersion();
+  const current = getLocalSushiVersion();
+
+  if (latest != null && current !== 'unknown' && latest !== current) {
+    logger.info(
+      `You are using SUSHI version ${current}, but the latest stable release is version ${latest}.\n` +
+        'To install the latest release, run:\n' +
+        '  npm install -g fsh-sushi'
+    );
+  }
```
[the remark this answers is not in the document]
cmoesel: @mint-thompson TBH, I'm not sure what case would ever lead to not knowing the current version.  I think we're probably just doing defensive programming here to keep `tsc` happy -- so there is limited value in coding up a case for when current is unknown.

We pull the version from the `package.json` that is inside the npm module.  Can you think of a use case (aside from straight-up corrupted node_modules) in which we wouldn't be able to figure out the version?  It seems to me that any such case would probably result in many other errors too.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1903,6 +1903,13 @@ export class ElementDefinition {
 
   private checkXhtml(value: string): boolean {
     try {
+      if (this.path.endsWith('div')) {
+        const startsWithDiv = /^\s*<\s*div[\s>]/.test(value);
+        const endsWithDiv = /<\s*\/\s*div\s*>\s*$/.test(value);
```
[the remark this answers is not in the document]
cmoesel: I thought the same thing too, @mint-thompson -- but then I thought that since regexes are already hard enough, that this approach (keeping separate start/end regexes) is a little easier to read, understand, and maintain than combining them together.  What do you think?

src/utils/Processing.ts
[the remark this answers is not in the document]
mint-thompson: Yes, you are right: a case where the version is missing basically represents a scenario where something tremendously bizarre has happened that is likely causing bigger problems. As much as I like seeing the test coverage number go up, this test would probably not provide a lot of actual value.

src/app.ts
```diff
@@ -197,13 +199,15 @@ async function app() {
   console.log();
   printResults(outPackage);
 
+  console.log();
+  await checkSushiVersion();
```
julianxcarter, line 203:
```suggestion
 const sushiVersions = await checkSushiVersion();
```

julianxcarter (changes requested):
The code functions well! I do think it's a bit awkward logging an info message after the final "Sushi Results" box has already been printed, especially since (on my machine at least) node logs a `NODE_TLS_REJECT_UNAUTHORIZED` warning before the message is printed due to the axios call. I would be more amenable to either printing the log statement at the beginning of a run, or integrating the message into the "Sushi Results" box (probably my favorite solution).

We could append the message to the "Sushi Results" box by having `checkSushiVersion` return the `latest` and `current` variables, and then pass those values into the `printResults()` function and having it handle the out of date warning. I did some light refactoring and this is what it would look like:

Screen Shot 2022-03-18 at 12 15 32 PM
 
I pushed up those changes [to a separate branch](https://github.com/FHIR/sushi/tree/ig_and_xhtml_results_box) if you'd like to take a look or implement any of it!

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1903,6 +1903,13 @@ export class ElementDefinition {
...
+        if (!startsWithDiv || !endsWithDiv) {
+          logger.warn('xhtml div elements should start and end with <div> tags');
```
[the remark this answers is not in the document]
guhanthuran: Added test.

src/utils/Processing.ts
```diff
@@ -549,3 +549,33 @@ export function getFilesRecursive(dir: string): string[] {
     return [];
   }
 }
+
+export function getLocalSushiVersion(): string {
+  const packageJSONPath = path.join(__dirname, '..', '..', 'package.json');
+  if (fs.existsSync(packageJSONPath)) {
+    return fs.readJSONSync(packageJSONPath)?.version;
+  }
+  return 'unknown';
```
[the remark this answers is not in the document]
guhanthuran: Made this change!

src/app.ts
```diff
@@ -197,13 +199,15 @@ async function app() {
...
+
   process.exit(stats.numError);
 }
 
 function getVersion(): string {
-  const packageJSONPath = path.join(__dirname, '..', 'package.json');
-  if (fs.existsSync(packageJSONPath)) {
-    const sushiVersion = fs.readJSONSync(packageJSONPath)?.version;
+  const sushiVersion = getLocalSushiVersion();
+  if (sushiVersion !== 'unknown') {
```
mint-thompson, line 210: I think this should now be checking for `null` rather than `'undefined'`.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1897,7 +1897,7 @@ export class ElementDefinition {
       (type === 'integer64' && /^[-]?\d+$/.test(value))
     ) {
       this.assignFHIRValue(`"${value}"`, value, exactly, type);
-    } else if (type == 'xhtml' && this.checkXhtml(value)) {
+    } else if (this.checkXhtml(value) && type == 'xhtml') {
```
mint-thompson, line 1900: I think it would be better to check the value of `type` before calling `this.checkXhtml`. If you call `this.checkXhtml` first, you could get a warning from it even on an element of a different type.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1913,6 +1913,13 @@ export class ElementDefinition {
 
   private checkXhtml(value: string): boolean {
     try {
+      if (this.path.endsWith('div')) {
```
mint-thompson, line 1916: To be a little more careful here, it might be best to check that the path ends with `.div` in order to avoid checking an element where the last part of the path is more than just `div`.

test/fhirtypes/ElementDefinition.assignString.test.ts
```diff
@@ -1175,4 +1176,21 @@ describe('ElementDefinition', () => {
...
+  describe('#checkXhtml', () => {
+    beforeEach(() => {
+      loggerSpy.reset();
+    });
+    it("should warn and throw when path doesn't start and end with div", () => {
+      const narrativeDiv = patient.findElementByPath('text.div', fisher);
+      expect(() =>
+        narrativeDiv.assignValue('<piv xmlns="http://www.w3.org/1999/xhtml">Twas brillig</div>')
+      ).toThrow();
+    });
```
mint-thompson, line 1188: Since the test description says that it should warn, the test should check the logger for a warning message.

test/fhirtypes/ElementDefinition.assignString.test.ts
```diff
@@ -1175,4 +1176,21 @@ describe('ElementDefinition', () => {
...
+
+    it("shouldn't warn when path starts and ends with div", () => {
+      const narrativeDiv = patient.findElementByPath('text.div', fisher);
+      narrativeDiv.assignValue('<div xmlns="http://www.w3.org/1999/xhtml">Twas brillig</div>');
+      expect(loggerSpy.getAllLogs('info')).toHaveLength(0);
```
mint-thompson, line 1193: This should be checking the `warn` log, not the `info` log.

mint-thompson (changes requested):
Thank you for the updates! I have a few more suggestions.

src/app.ts
mint-thompson: Err, rather than `'unknown'`. Sorry about that.

guhanthuran: Done, should I also change the 'unknown' return in function getVersion() in app.ts or should I just keep it as-is?

src/fhirtypes/ElementDefinition.ts
guhanthuran: Done.

test/fhirtypes/ElementDefinition.assignString.test.ts
guhanthuran: Done.

src/fhirtypes/ElementDefinition.ts
guhanthuran: That makes sense - made the change.

test/fhirtypes/ElementDefinition.assignString.test.ts
guhanthuran: Julian and I actually went over this, we found that the logger is being called in that test (we can see that it's being tested in the coverage report), but because that specific case in ElementDefinition is designed to throw an exception, we can't actually check the logger in the test case itself. ( See `throw new MismatchedTypeError('string', value, type);` in function `assignString()` ) Is there a way to test the log when the program crashes? Not sure if there's an another test case out there that addresses this specific kind of situation.

cmoesel: I assumed that `expect(...).toThrow()` would catch the exception so you could check the log on the next line (after the line w/ `).toThrow();`.  Is that not the case?

If not, another common approach is to use a `try`/`catch` inside the test and put any additional assertions in the `catch`.  E.g.
```ts
let didThrow = false;
const narrativeDiv = patient.findElementByPath('text.div', fisher);
try {
  narrativeDiv.assignValue('<piv xmlns="http://www.w3.org/1999/xhtml">Twas brillig</div>');
} catch (error) {
  didThrow = true;
  // Check error here to see if it is the error you expect
  // Check loggerSpy here to see if it logged the message you expect
}
expect(didThrow).toBeTrue();
```
_NOTE: The code above is off the cuff.  I didn't try to compile it, so it might not be exactly right, but you get the idea._

guhanthuran: @cmoesel Thank you for the help, this worked!

mint-thompson (approved):
This is looking good! Thank you!

PR #1041, opened 2022-03-09
Try to find required elements without slice names when validating an Instance

### LINKED ISSUE #1028: SUSHI incorrectly flags cardinality violation in slice
It seems that SUSHI is incorrectly flagging a cardinality violation in [this Questionnaire instance](https://fshschool.org/FSHOnline/#/share/3C0Uxkl).  Note that the instance uses indices instead of slice names to assign the elements, and that seems to trigger the error (e.g., using slice names makes the error go away).  The error says:
> error Element Questionnaire.item.item:currentValue.answerOption.value[x] has minimum cardinality 1 but occurs 0 time(s).

This is a problem on multiple accounts:
1. The instance does not set _any_ `answerOption` in the `currentValue` slice (nor is one required by the profile).
2. When authors use indices instead of slice names, I think SUSHI is supposed to _not_ try to validate slice cardinalities since it cannot identify when elements are (or aren't) in slices.

We should investigate why this error is popping up and make the appropriate changes to suppress it.

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Setting.20value.5Bx.5D.20to.20a.20string.20in.20an.20example/near/273079588

mint-thompson:
Fixes #1028 and completes task [CIMPL-906](https://standardhealthrecord.atlassian.net/browse/CIMPL-906).

When assigning values on a sliced element on an Instance, it is possible to refer to slices by name or by numeric index. If a numeric index is used in the rule on the Instance, but the slice being modified is a named slice on the InstanceOf definition, the slice name will not be tracked correctly on child elements on that Instance. If this is found, warn the user that it is preferable to use slice names in rules when possible, as it will make the FSH easier to understand. Additionally, this will help ensure consistent behavior when slice ordering rules are overhauled such that numeric-indexed slices no longer overwrite named slices.

An additional small bug is fixed as part of this change set: when setting implied properties on an instance, falsy values should be set. I ran into this when investigating the scenario reported in the issue, and it was easier to investigate with that bug fixed.

As noted in the comment in [InstanceExporter.ts](https://github.com/FHIR/sushi/pull/1041/files#diff-7ccabd62381ce8ebf05457c24cc3884eeefb1ec67ff259b82ca7f2c14c4f51cfR229-R233), this logic should be removed once slicing logic gets overhauled such that it is no longer possible to access a named slice without referring to it by name.

When running this change against the full set of regression repos, all output was the same. So, this will hopefully help authors working on changes in-progress use named slices more often.


src/export/InstanceExporter.ts
```diff
@@ -226,6 +226,29 @@ export class InstanceExporter implements Fishable {
             break;
           }
         }
+        // if we don't have instanceChild yet, it may be due to a rule that referes to a named slice somewhere before the end of the path using a numeric index.
```
cmoesel, line 229:
```suggestion
        // if we don't have instanceChild yet, it may be due to a rule that refers to a named slice somewhere before the end of the path using a numeric index.
```

src/export/InstanceExporter.ts
```diff
@@ -226,6 +226,29 @@ export class InstanceExporter implements Fishable {
...
+        // eventually, this logic can be removed when array indexing no longer allows the author to refer to
+        // a named slice using a numeric index.
+        if (instanceChild == null) {
+          const namelessChoiceSlices =
+            element
+              .findConnectedSliceElement()
+              ?.children(true)
+              .filter(c => c.path === child.path && c.sliceName) ?? [];
+          for (const choiceSlice of namelessChoiceSlices) {
+            instanceChild = instance[choiceSlice.sliceName];
```
cmoesel, line 241: This is kind of some complex stuff to wrap one's head around.  I noticed that this code is specialized to work w/ choices (which makes sense since it is in a block that checks for a path ending with `[x]`).

But here is my question: does this issue only happen for choice slices?  The example in the filed bug happens to be on a path w/ a choice slice, but I don't know if that is coincidence or if the choice slice is actually part of the recipe for reproducing.

cmoesel: I did actually mess with the example some and try to reproduce the issue w/ a non-choice element -- and I _couldn't_.  So I guess it is just specific to choices.  But I still don't 100% understand why.

Our instance implementation is super tricky.

cmoesel (changes requested):
This is a very weird problem with a slightly weird solution (which is no fault of your own).

I suggested a fix for a minor typo and asked a question to try to help me understand.

julianxcarter (approved):
This looks good to me aside from the typo that Chris pointed out!

PR #1042, opened 2022-03-09
Do not warn on invalid names that are overwritten with caret rules + CODEOWNDERS

### LINKED ISSUE #1024: Don't warn on invalid names when there is a valid override
In some cases, authors may intentionally use an invalid name with the `Profile:` keyword (or `Extension:`, etc) -- and then override it using a caret rule.  E.g.:
```
Profile: PatientAccessOrg-v0.0.1
* ^name = "PatientAccessOrg"

Profile: PatientAccessOrg-v1.0.1
* ^name = "PatientAccessOrg"
```

Currently, SUSHI will emit a warning that the name `PatientAccessOrg-v0.0.1` uses invalid characters -- but ideally it should be smart enough to see that a sensible name has been provided via the `^name` assignment.

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Warning.20about.20name.20despite.20override.3F

julianxcarter:
Fixes #1024, making it so that a user is not warned for an invalid entity name when a Caret Rule is used to overwrite that name with a valid one. Also adds a `CODEOWNERS` file so we no longer need to manually add each other as reviewers on our PRs.

### DIFF
--- a/src/fhirtypes/mixins.ts
+++ b/src/fhirtypes/mixins.ts
@@ -1,4 +1,5 @@
-import { SourceInfo } from '../fshtypes/FshEntity';
+import { CaretValueRule } from '../fshtypes/rules';
+import { FshValueSet, FshStructure, FshCodeSystem, SourceInfo } from '../fshtypes';
 import { logger } from '../utils';
 import { FHIRId, idRegex } from './primitiveTypes';
 
@@ -16,18 +17,22 @@ export class HasName {
    * @see {@link http://hl7.org/fhir/R4/structuredefinition-definitions.html#StructureDefinition.name}
    * @see {@link http://hl7.org/fhir/R4/valueset-definitions.html#ValueSet.name}
    * @see {@link http://hl7.org/fhir/R4/codesystem-definitions.html#CodeSystem.name}
-   * @param {string} name - The name to check against the name invariant
+   * @param {FshStructure | FshCodeSystem | FshValueSet} fshDefinition - The entity who's name is being checked
    * @param {SourceInfo} sourceInfo - The FSH file and location that specified the name
    */
-  setName(name: string, sourceInfo: SourceInfo) {
-    this.name = name;
-    if (!nameRegex.test(name)) {
-      logger.warn(
-        `The name "${name}" may not be suitable for machine processing applications such as code generation. Valid names start with an ` +
-          "upper-case ASCII letter ('A'..'Z') followed by any combination of upper- or lower-case ASCII letters ('A'..'Z', and " +
-          "'a'..'z'), numerals ('0'..'9') and '_', with a length limit of 255 characters.",
-        sourceInfo
-      );
+  setName(fshDefinition: FshStructure | FshCodeSystem | FshValueSet, sourceInfo: SourceInfo) {
+    this.name = fshDefinition.name;
+    const nameRule = fshDefinition.rules.filter(rule => rule instanceof CaretValueRule && rule.caretPath === 'name').pop() as CaretValueRule;
+    if (!nameRegex.test(this.name)) {
+      // There's no need to warn on an invalid name if a Caret Rule overwrites the entity name
+      if (!nameRule || !nameRegex.test(nameRule.value as string)) {
+        logger.warn(
+          `The name "${this.name}" may not be suitable for machine processing applications such as code generation. Valid names start with an ` +
+            "upper-case ASCII letter ('A'..'Z') followed by any combination of upper- or lower-case ASCII letters ('A'..'Z', and " +
+            "'a'..'z'), numerals ('0'..'9') and '_', with a length limit of 255 characters.",
+          sourceInfo
+        );
+      }
     }
   }
 }
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+*   @jafeltra @cmoesel @julianxcarter @mint-thompson @guhanthuran
--- a/src/export/CodeSystemExporter.ts
+++ b/src/export/CodeSystemExporter.ts
@@ -16,7 +16,7 @@ export class CodeSystemExporter {
   ) {}
 
   private setMetadata(codeSystem: CodeSystem, fshDefinition: FshCodeSystem): void {
-    codeSystem.setName(fshDefinition.name, fshDefinition.sourceInfo);
+    codeSystem.setName(fshDefinition, fshDefinition.sourceInfo);
     codeSystem.setId(fshDefinition.id, fshDefinition.sourceInfo);
     if (fshDefinition.title) codeSystem.title = fshDefinition.title;
     if (fshDefinition.description) codeSystem.description = fshDefinition.description;
--- a/src/export/StructureDefinitionExporter.ts
+++ b/src/export/StructureDefinitionExporter.ts
@@ -298,7 +298,7 @@ export class StructureDefinitionExporter implements Fishable {
     // keep url since it was already defined when the StructureDefinition was initially created
     delete structDef.identifier;
     structDef.version = this.tank.config.version; // can be overridden using a rule
-    structDef.setName(fshDefinition.name, fshDefinition.sourceInfo);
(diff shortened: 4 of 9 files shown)


cmoesel (changes requested):
Nice work!  I have a few high-level suggestions:

1. You wrote tests to confirm that it does not warn if a caret rule overrides the name with a valid name.  That's good!  I think it would also be good to have at least one test that confirms that if both the original name and caret rule name override are invalid, you _do_ get a warning.

2. What if the standard name is valid, BUT they also include a caret rule override that is actually an _invalid_ name?  In that case, I think they should get the warning too (but right now they would not since we only look at the caret name if the first name fails).

3. Not to double the effort of this PR, but now that I'm looking at it... maybe we should be doing the same thing with `Id:` / `^id =` too?  What do you think?

julianxcarter:
> Nice work! I have a few high-level suggestions:
> 
> 1. You wrote tests to confirm that it does not warn if a caret rule overrides the name with a valid name.  That's good!  I think it would also be good to have at least one test that confirms that if both the original name and caret rule name override are invalid, you _do_ get a warning.
> 2. What if the standard name is valid, BUT they also include a caret rule override that is actually an _invalid_ name?  In that case, I think they should get the warning too (but right now they would not since we only look at the caret name if the first name fails).
> 3. Not to double the effort of this PR, but now that I'm looking at it... maybe we should be doing the same thing with `Id:` / `^id =` too?  What do you think?

These are all great point! Commit [`286323f`](https://github.com/FHIR/sushi/pull/1042/commits/286323f1c49bfa083aaa698bb0a3e3959be08898) adds new test cases and adds a warning for invalid Caret Rules, and commit [`d782e94](https://github.com/FHIR/sushi/pull/1042/commits/d782e9431ae539e2564043915559b8d7e3df991e) expands the same behavior to the `setId()` function.

src/fhirtypes/mixins.ts
```diff
@@ -53,27 +65,40 @@ export class HasId {
...
-        logger.warn(
-          `The string "${this.id}" represents a valid FHIR name but not a valid FHIR id. FHIR ids cannot contain "_" and can be at most 64 characters. The id will be exported as "${sanitizedId}". Avoid this warning by specifying a valid id directly using the "Id" keyword.`,
-          sourceInfo
-        );
-        this.id = sanitizedId;
+    const idRule = fshDefinition.rules
+      .filter(rule => rule instanceof CaretValueRule && rule.caretPath === 'id')
+      .pop() as CaretValueRule;
+    if (!validId) {
+      if (idRule && idRegex.test(idRule.value as string)) {
```
cmoesel, line 77: I think a similar simplification can probably happen here as I suggested for name.

src/fhirtypes/mixins.ts
```diff
@@ -53,27 +65,40 @@ export class HasId {
...
+        this.id = idRule.value as string;
```
cmoesel, line 78: Also... to maintain the same behavior as before, I don't think we should actually assign `this.id` from the rule.  Leave `this.id` as the original `Id:`.  This allows people to still fish on that id (even if it is overridden by a rule).

cmoesel (changes requested):
Thanks, @julianxcarter!  I've left a few more comments for your consideration.

src/fhirtypes/mixins.ts
```diff
@@ -16,17 +17,29 @@ export class HasName {
...
+   * @param {FshStructure | FshCodeSystem | FshValueSet} fshDefinition - The entity who's name is being set
    */
-  setName(name: string, sourceInfo: SourceInfo) {
-    this.name = name;
-    if (!nameRegex.test(name)) {
+  setName(fshDefinition: FshStructure | FshCodeSystem | FshValueSet) {
+    this.name = fshDefinition.name;
+    const nameRule = fshDefinition.rules
+      .filter(rule => rule instanceof CaretValueRule && rule.caretPath === 'name')
+      .pop() as CaretValueRule;
```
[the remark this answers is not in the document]
julianxcarter: Had no idea about this function. Thanks!

julianxcarter:
@cmoesel Thanks for the feedback! commit [`a1bf3de1](https://github.com/FHIR/sushi/pull/1042/commits/a1bf3dec76187bf7fbf0f2decbcaffbadebdc9d2) takes your suggestions into account and simplifies both of the if-statements. One thing I was a bit unsure of was whether to check if the ID set with the caret rule should be normalized if the id set in the metadata part of the entity doesn't pass the `nameRegex`. I opted not to do that because I thought editing something that was explicitly set with a rule was a bit heavy-handed, but I can add that in if you think it's appropriate.

src/fhirtypes/mixins.ts
```diff
@@ -53,27 +58,32 @@ export class HasId {
...
-  validateId(sourceInfo: SourceInfo) {
-    let validId = idRegex.test(this.id);
+  validateId(fshDefinition: FshStructure | FshCodeSystem | FshValueSet) {
+    const idRule = _.findLast(
+      fshDefinition.rules,
+      rule => rule instanceof CaretValueRule && rule.caretPath === 'id'
+    ) as CaretValueRule;
+    const idToCheck = idRule ? (idRule.value as string) : this.id;
+    let validId = idRegex.test(idToCheck);
     if (!validId && nameRegex.test(this.id)) {
```
cmoesel, line 70: If both the `Id:` and `^id` are invalid, but the `Id:` matches the name regex, then this will sanitize the `Id:` and log the warning that we did so.  BUT, the unsanitized `^id` is the one that will still be used in the formal resource -- so sanitizing the `Id:` is kind of pointless in that case.

I think we only want to do this if the `Id:` is invalid and was not overridden by a caret rule (i.e., it's the id that will be used in the resource definition).  So I think the condition should be:
```ts
if (!validId && idRule == null && nameRegex.test(this.id)) {
```

cmoesel (changes requested):
I agree that we don't want to sanitize the `^id` since it was explicitly set.  But I also think we probably don't want to (or need to) sanitize the `Id:` if the `^id` is explicitly set.  See my individual comment for details.

This *could* cause a change in some IGs though (if they referenced an `Id:` by its sanitized value and we no longer sanitize it).  This seems like an _extreme_ edge case, but we probably should run regression after you make that change, just in case.

src/fhirtypes/mixins.ts
julianxcarter: Yes this makes even more sense... I kind of talked myself into a tizzy when looking at this function. Commit [`5323159`](https://github.com/FHIR/sushi/pull/1042/commits/532315959432afc53a72d5a962cd21544af38fef) makes it so that the Id isn't sanitized at all if a caret rule overwrites it.

cmoesel (approved):
Thanks, @julianxcarter! This looks great!

mint-thompson (changes requested):
This looks good! I have two small changes to request, just to be extra careful about finding caret rules that set the name or id.

src/fhirtypes/mixins.ts
```diff
@@ -16,17 +18,21 @@ export class HasName {
...
    */
-  setName(name: string, sourceInfo: SourceInfo) {
-    this.name = name;
-    if (!nameRegex.test(name)) {
+  setName(fshDefinition: FshStructure | FshCodeSystem | FshValueSet) {
+    this.name = fshDefinition.name;
+    const nameRule = _.findLast(
+      fshDefinition.rules,
+      rule => rule instanceof CaretValueRule && rule.caretPath === 'name'
+    ) as CaretValueRule;
```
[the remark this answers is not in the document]
julianxcarter: This is a really good idea! I added that check in commit [`ac14e2f`](https://github.com/FHIR/sushi/pull/1042/commits/ac14e2f84da781aa5432b7413f88ee2c67ec1c0c)

mint-thompson:
Thank you for the review @cmoesel ! I've made updates based on your suggestions.

mint-thompson (approved):
Looks good! Thank you!

cmoesel (approved):
Yup!

PR #1046, opened 2022-03-15
Handle choice paths on Instance assignment rules

### LINKED ISSUE #1032: SUSHI handles value[x] assignments incorrectly in Instance
If a profile constrains `value[x]` to a single type, and an author assigns `value[x]` in an instance of that profile, then SUSHI incorrectly serialized the property as `value[x]` rather than using its type-specific name.  For example:
```
Profile: BooleanObservation
Parent: Observation
* value[x] only boolean

Instance: MyObservation
InstanceOf: BooleanObservation
* status = #final
* code = http://example#somecode
* value[x] = true
```
produces the following instance:
```json
{
  "id": "MyObservation",
  "meta": {
    "profile": [
      "http://example.org/StructureDefinition/BooleanObservation"
    ]
  },
  "status": "final",
  "code": {
    "coding": [
      {
        "code": "somecode",
        "system": "http://example"
      }
    ]
  },
  "value[x]": true
}
```
Note that the last property should be `"valueBoolean": true`.

Either SUSHI needs to disallow assignment to `value[x]` in Instances or it needs to handle it correctly. I'd prefer to handle it correctly, but if that proves to be complex, it may be reasonable to disallow it.

mint-thompson:
Fixes #1032 and completes task [CIMPL-910](https://standardhealthrecord.atlassian.net/browse/CIMPL-910).

When assigning values to choice elements on an Instance, type-specific elements should be used in the rules. If the choice element is constrained to a single type, it is possible to change the path to refer to the type-specific element. Warn the user that the path is being changed and that it is preferable to use the type-specific element name in the rule. If the choice element is constrained to multiple types, it is not possible to change the path. Show an error indicating that the value will not be assigned. These warnings and errors may apply regardless of where the choice element appears in the rule path.

Basically: you can't have an `[x]` element on an Instance, so show a warning and fix it in cases where we know the type, and show an error in cases where we don't know the type.

See also  http://hl7.org/fhir/R4/formats.html#choice


julianxcarter (approved):
Works great!

test/export/InstanceExporter.test.ts
```diff
@@ -1330,6 +1330,75 @@ describe('InstanceExporter', () => {
...
+      codeTestCode.value = new FshCode('testcode');
+      const valueSomeValue = new AssignmentRule('value[x]');
+      valueSomeValue.value = 'some value';
+      testInstance.rules.push(statusFinal, codeTestCode, valueSomeValue);
+      const exported = exportInstance(testInstance);
+      expect(exported.valueString).toBe('some value');
+      expect(exported['value[x]']).toBeUndefined();
+      expect(loggerSpy.getAllMessages('error')).toHaveLength(0);
+      expect(loggerSpy.getLastMessage('warn')).toMatch(
+        "When assigning values on instances, use the choice element's type. Rule path changed to valueString"
```
cmoesel, line 1366: In the context of this PR I know exactly what you're talking about.  But I wonder if authors might be confused (until they look up the line numbers).  I think it might be more obvious what is going on if we also refer to the old path.  E.g.:
> When assigning values on instances, use the choice element's type. Rule path changed from 'value[x]' to 'valueString'.

cmoesel (changes requested):
Thanks, @mint-thompson!  Looks good, but I have a couple of minor suggestions to consider.

mint-thompson:
@cmoesel thank you for the suggestions! I've made updates accordingly.

cmoesel (approved):
Nice work, @mint-thompson!

julianxcarter (approved):

PR #1049, opened 2022-03-21
Add content reference elements on Logical and Resource definitions

### LINKED ISSUE #1009: FSH / SUSHI should support contentReference elements in logical models and custom resources
Elements that have a contentReference do not need to declare a type (and I'm not sure about cardinality -- need to research that).  FSH / SUSHI, however, does not allow you to create a new element _without_ a type.  We need to figure out how to best support this in the grammar and SUSHI.

Zulip: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Logical.20models.20with.20content.20references/near/268304544

mint-thompson:
Fixes #1009 and completes task [CIMPL-880](https://standardhealthrecord.atlassian.net/browse/CIMPL-880).

An element definition usually has one or more types, but sometimes, it gets defined with a `contentReference` instead. When this is the case, it won't have any types defined (https://www.hl7.org/fhir/elementdefinition-definitions.html#ElementDefinition.contentReference). This PR adds grammar for defining such a rule and creates these elements on the Logical or Resource. 

The logic for handling the element with a content reference is pretty simple, but the grammar for it is something that I'd like to discuss with y'all. I went with a rule that is clearly not a typical `AddElementRule`, with the idea of wanting it to be clear from reading whether you're looking at an element with types or an element with a content reference. Some alternatives I considered were:
- No grammar change, but if the rule declares one type and that type is `contentReference`, it's a content reference element. This has the upside of not changing the grammar, but two downsides. One, that it's slightly magical behavior in that this specific type means something different. Two, that you would then need a separate rule in order to set the actual URL for the content reference. I felt like these two downsides together made this not the best choice.
- New grammar rule for content references, but with the `KW_CONTENTREFERENCE (SEQUENCE | CODE)` part of the rule in the same position as where `targetType (KW_OR targetType)*` would be. This one is similar, and has the advantage of everything being in basically the same order. But, it also means that you have to read further into the rule to see if you're dealing with a typed element or a content reference element. Additionally, I felt it was harder to read with the URL being immediately followed by the `short`. But, this is not a particularly strong objection, and if folks prefer this style, it's not a lot of work to change it. Using one of the rules from the `FSHImporter` tests as an example, here's the two different styles:
  - `* oranges contentReference http://example.org/citrus/orange 1..* MS "oranges" "oranges are a citrus"`
  - `* oranges 1..* MS contentReference http://example.org/citrus/orange "oranges" "oranges are a citrus"`


jafeltra (commented):
I looked through the code changes here and they all make sense to me. The tests test the appropriate pieces of code and the solution seems reasonable. I know you are waiting further feedback, but I had started taking a look, so I figured I'd leave a comment. I can re-review if any changes to the approach are made!

jafeltra (dismissed):
I really like the consistent order!

src/fshtypes/rules/AddElementRule.ts
```diff
@@ -21,9 +22,13 @@ export class AddElementRule extends FlagCarryingRule {
...
     const flagPart = this.flags.length ? ` ${this.flags.join(' ')}` : '';
-    const typePart = typeString(this.types);
     const shortPart = this.short ? ` "${fshifyString(this.short)}"` : '';
     const definitionPart = this.definition ? ` "${fshifyString(this.definition)}"` : shortPart;
-    return `* ${this.path} ${cardPart}${flagPart} ${typePart}${shortPart}${definitionPart}`;
+    if (this.types.length > 0) {
+      const typePart = typeString(this.types);
+      return `* ${this.path} ${cardPart}${flagPart} ${typePart}${shortPart}${definitionPart}`;
+    } else {
+      return `* ${this.path} contentReference ${this.contentReference} ${cardPart}${flagPart}${shortPart}${definitionPart}`;
```
cmoesel, line 31: Whoops!  This needs to be updated to reflect the new ordering.

src/import/FSHImporter.ts
```diff
@@ -970,6 +974,35 @@ export class FSHImporter extends FSHVisitor {
...
+    return addElementRule;
+  }
+
+  visitAddCRElementRule(ctx: pc.AddCRElementRuleContext): AddElementRule {
+    const addElementRule = this.parseNewElement(ctx);
+    if (ctx.SEQUENCE()) {
+      addElementRule.contentReference = ctx.SEQUENCE().getText();
+    } else if (ctx.CODE()) {
+      addElementRule.contentReference = ctx.CODE().getText();
+    }
```
cmoesel, line 999: Since `contentReference` is a URL, I think it probably makes sense to use the alias-aware function (I forget its exact name) to extract the value.

test/fshtypes/rules/AddElementRule.test.ts
```diff
@@ -21,98 +21,148 @@ describe('AddElementRule', () => {
...
-        }
-      );
-      rule.trialUse = true;
-      rule.mustSupport = true;
-      rule.short = 'Fresh barley';
-      rule.definition = 'Barley, one of the "ancient grains."';
-      const expectedFSH =
-        '* barley 0..* MS TU Quantity or Reference(Specimen) "Fresh barley" "Barley, one of the \\"ancient grains.\\""';
-      expect(rule.toFSH()).toEqual(expectedFSH);
+    describe('#contentReference', () => {
```
cmoesel, line 121: All these tests need to be updated to reflect the new order as well.

cmoesel (changes requested):
It's nice to plug this gap!  I found a few spots where the old order is still used -- and also suggest making this work with aliases since `contentReference` is technically a URL.

cmoesel (changes requested):
Everything is looking pretty good and I tried some manual tests as well as running the IG Publisher.  In those tests, I did discover that the IG Publisher requires the contentReference to be of the format: `{canonicalURLToSD}#{elementId}`.  Although core resources can get away with just `#{elementId}`, LMs and custom resources must use the canonical URL as well.

My one comment is to update the test cases to use realistic contentReference formats that could feasibly satisfy the IG Publisher.

jafeltra:
The latest changes look good to me as well! I didn't notice anything else besides Chris's last comment!

mint-thompson:
The test cases have now been updated with hopefully better-looking content references.

jafeltra (approved):
The changes all look consistent to me, and I tried using one of the updated examples with the IG Publisher and it didn't complain.

cmoesel (approved):
Excellent.  So smoothly integrated; it's almost like it was always this way!

PR #1050, opened 2022-03-21
Find concepts added by CaretValueRules in CodeSystems

### LINKED ISSUE #1039: SUSHI not recognizing CodeSystem codes defined using ^concept
If an author defines codes in a CodeSystem using `^concept` like the following:
```
* ^concept[2].code = #101
* ^concept[2].display = "display 2"
* ^concept[2].definition = "definition 2"
```
then SUSHI does not recognize those codes when used in a ValueSet.  This is reproduced in FSH Online here: https://fshschool.org/FSHOnline/#/share/3vLYBUq

Although this is not the recommended way to create CodeSystems, it still should be supported.

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/sushi.20suddenly.20failing.20on.20cs.2Fvs.20codes.20post.20upgrade/near/274386821

mint-thompson:
Fixes #1039 and completes task [CIMPL-912](https://standardhealthrecord.atlassian.net/browse/CIMPL-912).

When defining a CodeSystem, concepts may be added by ConceptRules or CaretValueRules. ValueSets that refer to the CodeSystem can add those concepts, regardless of what rules were used to add them.


src/fhirtypes/common.ts
```diff
@@ -340,6 +340,7 @@ export function listUndefinedLocalCodes(
 ): void {
   let undefinedCodes: string[] = [];
   applyInsertRules(codeSystem, tank);
+  const conceptRulePath = /^(concept(\[(\d+|\+|=)\])?\.)+code$/;
```
cmoesel, line 343: The grammar is a lot more lenient about whitespace than this regex is.  Maybe we should at least allow for variable whitespace inside the `[]`?  Just in case someone tries to align things like this:
```
* concept[ 8].code = #eight
* concept[ 9].code = #nine
* concept[10].code = #ten
```

cmoesel: Ha.  Now I see that you merely just moved this regex up.  But you were the last to touch it now.  So now it's your responsibility! ;-)

cmoesel (changes requested):
I'm glad this didn't turn out to be too bad!  I just made one comment to try to improve flexibility a little.

cmoesel (approved):
Looks good!  I did a simple manual test using this definition and it failed on master but passed on this branch:
```
CodeSystem: CS1
* ^concept[0].code = #foo
* ^concept[0].display = "Foo"
* ^concept[0].definition = "It's a Foo"
* ^concept[0].concept[0].code = #bar
* ^concept[0].concept[0].display = "Bar"
* ^concept[0].concept[0].definition = "It's a Bar"

ValueSet: VS1
* CS1#foo
* CS1#bar
```

jafeltra (approved):
This makes sense to me! I tried the example on master and this branch and the error goes away and the ValueSet looks right!

PR #1056, opened 2022-04-05
Prefer constraining to Reference type

### LINKED ISSUE #1044: SUSHI converting Reference to CodeableReference in R4B Snapshot1 extensions
If you have an R4B Snapshot1 project (e.g., `fhirVersion: 4.3.0-snapshot1`) and add this definition:
```
Extension: MyExtension
* value[x] only Reference(Resource)
```
SUSHI will (incorrectly) create a differential with this element:
```json
{
  "id": "Extension.value[x]",
  "path": "Extension.value[x]",
  "type": [
   {
      "code": "CodeableReference",
      "targetProfile": [
        "http://hl7.org/fhir/StructureDefinition/Resource"
      ]
    }
  ]
}
```

If you first restrict to just `Reference` and _then_ to `Reference(Resource)`, it works correctly though.

`¯\_(ツ)_/¯`

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/References.20for.204B.20Snapshot.20.231/near/274743533

jafeltra:
Fixes #1044.

When constraining an element type using the `Reference()` keyword, SUSHI will now prefer the `Reference` type over the `CodeableReference` type if both are allowed on the element.

If you try the simple example in the issue, you should see that the `Extension.value[x]` element now has the following type:

```
{
  "id": "Extension.value[x]",
  "path": "Extension.value[x]",
  "type": [
    {
      "code": "Reference",
      "targetProfile": [
        "http://hl7.org/fhir/StructureDefinition/Resource"
      ]
    }
  ]
}
```

Because the R4B Snapshot1 and R5 Extensions support both Reference and CodeableReference, SUSHI was previously finding the CodeableReference type first and just using that. It now looks for Reference first and only uses CodeableReference if it cannot find Reference.


src/fhirtypes/ElementDefinition.ts
```diff
@@ -1083,11 +1083,20 @@ export class ElementDefinition {
         // References always have a code 'Reference' w/ the referenced type's defining URL set as
         // one of the targetProfiles.  If the targetProfile property is null, that means any
         // reference is allowed.
+        // When 'Reference' keyword is used, prefer to match on the 'Reference' type over the
+        // 'CodeableConcept' type if they both exist on the element.
```
julianxcarter, line 1087: Think you mean `'CodeableReference`' in this comment 🙂.

julianxcarter (commented):
This looks good to me! I have the smallest of comments in the world but otherwise this is flawless.

src/fhirtypes/ElementDefinition.ts
jafeltra: Updated in [11a10a4](https://github.com/FHIR/sushi/pull/1056/commits/11a10a4d828fe0b790e6dff158a36c538f505df6)!

mint-thompson (approved):

julianxcarter (approved):

PR #1057, opened 2022-04-06
Warn on unused config properties in FSHOnly projects

### LINKED ISSUE #1051: SUSHI should warn on unused configuration properties in FSHOnly mode
When `sushi-config.yaml` is set to `FSHOnly: true`, many of the config properties are ignored (since SUSHI will not create an ImplementationGuide JSON file).  So... if an author does set these properties in `sushi-config.yaml`, it's very likely a mistake.

We should determine what properties are used in FSHOnly mode and what properties are extraneous -- and warn on the use of any extraneous ones when `FSHOnly` is `true`.

cmoesel:
In FSHOnly projects, only a small subset of config properties are used: `canonical`, `fhirVersion`, `version`, `dependencies`, `instanceOptions`, and `applyExtensionMetadataToRoot`.  If SUSHI encounters any other properties in a FSHOnly project, it should warn the user, since this is likely a mistake.

We already had code that warned on certain properties, but it was only a subset and missed some important ones.  Rather than use a ban-list, we use an allow-list now.

Fixes #1051


src/import/importConfiguration.ts
```diff
@@ -48,24 +48,17 @@ import {
...
-  'copyrightYear',
-  'releaseLabel',
-  'releaselabel'
+// Allowed properties for FSH Only projects (all other properties are irrelevant)
+const ALLOWED_FSH_ONLY_PROPERTIES = [
+  ...MINIMAL_CONFIG_PROPERTIES,
+  'version',
+  'dependencies',
+  'instanceOptions',
+  'applyExtensionMetadataToRoot'
```
jafeltra, line 61: I wonder if we should include `FSHOnly` in this list. When I tried it, I got this message:

```
 The following properties are unused and only relevant for IG creation: ... FSHOnly ...
```

which seems a little confusing to me. It also means I couldn't ever get that warning to go away since I can't remove `FSHOnly`.

jafeltra (commented):
This looks good! I just had one small comment below.

src/import/importConfiguration.ts
cmoesel: Oh. Yeah. That seems obvious. I wonder why that property didn't pop up in the warning given in the test? I'll fix it.

cmoesel: Fixed in 5c3934b.

mint-thompson (approved):

PR #1058, opened 2022-04-06
Merge existing values when assigning References

jafeltra:
Addresses [CIMPL-901](https://standardhealthrecord.atlassian.net/browse/CIMPL-901).

Previously, when assigning values to a `Reference` element that was `0..1` or `1..1` and then using the `Reference()` keyword, the previous values were overwritten. The [issue](https://standardhealthrecord.atlassian.net/browse/CIMPL-901) provides a good example to demonstrate this. Another very simple example is below:

```
Instance: ExamplePatient
InstanceOf: Patient

Instance: ExampleObservation
InstanceOf: Observation
* subject.identifier.use = #official
* subject.identifier.system = "https://example.org"
* subject.identifier.value = "12345"
* subject = Reference(ExamplePatient)
```

On master, this instance will produce the following `subject`:

```
"subject": {
  "reference": "Patient/ExamplePatient"
}
```

On this branch, it is correct and will produce this `subject`:

```
"subject": {
  "reference": "Patient/ExamplePatient",
  "identifier": {
    "use": "official",
    "system": "https://example.org",
    "value": "12345"
  }
}
```

To fix this, I made a pretty generic update that merges the existing object with the new value. This is similar to what happens when the reference can be an array (cardinality is `0..*` or `1..*`) and no tests broke, but if anyone thinks of a case that is negatively affected by this, please let me know.

### DIFF
--- a/src/fhirtypes/common.ts
+++ b/src/fhirtypes/common.ts
@@ -230,15 +230,12 @@ export function setPropertyOnInstance(
         } else {
           // Check if the instance already has the element being defined
           if (current[key] != null && typeof current[key] === 'object') {
-            // Check if the instance already has a quantity element
-            // Quantity elements are the only FHIR types with both a code and value
-            if (current[key].hasOwnProperty('value') && assignedValue.hasOwnProperty('code')) {
-              // Ensure that the existing value is not being overwritten
-              assignedValue = {
-                value: current[key].value,
-                ...assignedValue
-              };
-            }
+            // If the instance already has the element, we should merge it
+            // Cases where this applies:
+            // - Quantity elements that set a value and then set a code with the FSH code syntax
+            // - Reference elements that set other properties of reference (like identifier) directly
+            //   and set reference with the FSH Reference() keyword
+            Object.assign(assignedValue, current[key]);
           }
           current[key] = assignedValue;
         }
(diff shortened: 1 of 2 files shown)


mint-thompson (changes requested):
There is an admittedly somewhat unusual case that seems to behave incorrectly with this code change. Consider the following FSH:
```
* subject.reference = "Patient/Alice"
* subject = Reference(Bob)
```
Since the objects now get merged in such a way that prioritizes old values, it means that the resulting value of `subject` will be `{ reference: "Patient/Alice" }`. I think it would be correct in this case for the resulting value to be `{ reference: "Patient/Bob" }`, since that rule came later.

jafeltra:
@mint-thompson that's a great catch! Thank you!

[67829da](https://github.com/FHIR/sushi/pull/1058/commits/67829dabeff2448867bbc2694b4a159eda0e7dd3) adds a test for that case, and I updated the code so that it only reassigns `current[key]` if it isn't assigned with `Object.assign` already. This is definitely how it should work since `Object.assign` does, in fact, assign lol. This is also how the case for arrays above works, but I had missed that the reassignment only happens in the `else` block.

mint-thompson (approved):
Thank you for fixing that case! It seems that this change also fixes the case of writing something like this, which is also good:
```
* valueQuantity.unit = #milligrams
* valueQuantity = 14 'mg'
```

The set `unit` is now preserved.

jafeltra:
Oh wow @mint-thompson good point! I thought that used to work, but I'm glad it at least does now!

julianxcarter (approved):
Today I learned about `Object.assign()`. Neat!

PR #1059, opened 2022-04-06
1045 sushi does not install dependencies proxy

guhanthuran:
- Created axiosUtils.ts and created axiosGet function to address posible proxy
- Changed all axios.get calls to axiosGet within /src
- Wrote tests for axiosGet in axiosUtils.test.ts

To test:
- Make sure you have docker installed
- Set up a proxy server through docker: `docker run --name squid -d -p 3128:3128 datadog/squid`
- Set up environment variables `HTTP_PROXY` and `HTTPS_PROXY` as it was done in the issue.
- Use `env|grep HTTP_PROXY` and `env|grep HTTPS_PROXY` to verify that you're using a proxy
- `curl google.com` just to make sure you're connected
- Delete local .fhir cache
- Run sushi on a local sushi project
- If packages download correctly, it is working correctly. If it displays an error regarding the package download failing, then it is breaking.
- `unset HTTP_PROXY` and `unset HTTPS_PROXY` will let you stop using the proxy. Again you can use `env|grep` to verify the two proxies that are not running anymore
- Turn off the docker container: `docker stop squid`


regression/find-repos.ts
```diff
@@ -55,9 +57,18 @@ async function getReposFromGitHub(): Promise<GHRepo[]> {
...
+        // https://github.com/axios/axios/issues/3459
+        proxyAgent = new (HttpsProxyAgent as any)(httpsProxy);
+      }
       const res = await axios.get(
         `https://api.github.com/orgs/HL7/repos?sort=full_name&per_page=100&page=${page}`,
-        options
+        {
+          httpAgent: proxyAgent,
+          headers: options
+        }
```
cmoesel, line 71: I think this should use `axiosGet` too.  I suspect it doesn't use it now because it needs to pass in that extra option (the authorization header).  Maybe instead of the 2nd argument to `axiosGet` being `responseType?: any`, it should be `options?: any`.  Then where you need the response type, you pass in `{ responseType: 'arraybuffer' }` and here, if you need the authorization header, you pass in ``{ headers: { Authorization: `token ${process.env.GITHUB_API_KEY}` } }``.  Then that would be more flexible if we ever have to pass in other options as well.

cmoesel (changes requested):
I like consolidating the axios get into a utility function.  In fact, I like it so much, I'd like to see it also work for the axios get in the regression script!

src/utils/axiosUtils.ts
```diff
@@ -0,0 +1,28 @@
...
+ */
+export async function axiosGet(url: string, responseType?: any): Promise<AxiosResponse<any>> {
+  const httpsProxy = process.env.HTTPS_PROXY;
+  const axiosOptions: AxiosRequestConfig = {};
+  if (httpsProxy) {
+    // https://github.com/axios/axios/issues/3459
+    axiosOptions.httpsAgent = new (HttpsProxyAgent as any)(httpsProxy);
+    axiosOptions.proxy = false;
+  }
+  if (typeof responseType !== 'undefined') {
```
jafeltra, line 20: I think it would be a little cleaner to just have this line be `if (responseType) {`. I think this might be a JavaScript convention, and if not, I think it's at least something we tend to do. Since `undefined` is falsy, this if statement will evaluate to true as long as `responseType` is defined.

src/utils/axiosUtils.ts
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * This function is called to better handle axios.get calls with logic that allows
+ * the use of proxies. Not needed in tests unless specifically testing with proxies.
+ * Check https://github.com/axios/axios/issues/3459#issuecomment-766171276 for more info.
+ * @param url {string} - string representation of url to get
+ * @param responseType {any} - optional parameter to change the data type needed from get
+ * (ex. arraybuffer). In default it returns JSON
+ */
+export async function axiosGet(url: string, responseType?: any): Promise<AxiosResponse<any>> {
+  const httpsProxy = process.env.HTTPS_PROXY;
```
jafeltra, line 13: This is just a question for my own understanding, but do you only need to look at the `HTTPS_PROXY` and not the `HTTP_PROXY` because we only make requests to `https` urls?

jafeltra (commented):
This looks really good! I just had one suggestion in addition to Chris's comment.

src/utils/axiosUtils.ts
guhanthuran: Yep, we don't use the `HTTP_PROXY` with the `axios.get` functionality. Just to make sure, I tested only using `HTTP_PROXY` (keeping `HTTPS_PROXY` unset) and all the tests held up.

src/utils/axiosUtils.ts
guhanthuran: Made this change, thanks!

regression/find-repos.ts
guhanthuran: Pushed these changes, thank you.

guhanthuran:
Check latest changes, it should handle the new 'options' logic. Please test it with proxy again (with the directions above), since I am having socket issues on my end.

regression/find-repos.ts
```diff
@@ -51,11 +52,24 @@ async function getReposFromGitHub(): Promise<GHRepo[]> {
   const repos: GHRepo[] = [];
   try {
     for (let page = 1; true; page++) {
-      const options: any = {};
+      let options: any = {};
       if (process.env.GITHUB_API_KEY) {
-        options.headers = { Authorization: `token ${process.env.GITHUB_API_KEY}` };
+        options = { headers: { Authorization: `token ${process.env.GITHUB_API_KEY}` } };
```
cmoesel, line 57: I don't think these lines actually had to change at all.  You can keep them as they were, which is preferred, since we like to use `const` instead of `let` whenever we can.

regression/find-repos.ts
```diff
@@ -51,11 +52,24 @@ async function getReposFromGitHub(): Promise<GHRepo[]> {
...
+      //   // https://github.com/axios/axios/issues/3459
+      //   proxyAgent = new (HttpsProxyAgent as any)(httpsProxy);
+      // }
+      // const res = await axios.get(
+      //   `https://api.github.com/orgs/HL7/repos?sort=full_name&per_page=100&page=${page}`,
+      //   {
+      //     httpAgent: proxyAgent,
+      //     headers: options
+      //   }
+      // );
```
cmoesel, line 71: Please delete this commented out code.

src/utils/axiosUtils.ts
```diff
@@ -0,0 +1,28 @@
...
+ * @param url {string} - string representation of url to get
+ * @param responseType {any} - optional parameter to change the data type needed from get
+ * (ex. arraybuffer). In default it returns JSON
+ */
+export async function axiosGet(url: string, options?: any): Promise<AxiosResponse<any>> {
+  const httpsProxy = process.env.HTTPS_PROXY;
+  let axiosOptions: AxiosRequestConfig = {};
+  if (options) {
+    axiosOptions = options;
+  }
```
cmoesel, line 17: I think this can actually be simplified to the following:
```ts
const axiosOptions: AxiosRequestConfig = options ?? {};
```
The `??` operator is a neat little TypeScript trick.  Basically, `A ?? B` means "if A is defined, use A, otherwise use B".

You might have to also change the `options` argument in the signature to `options?: AxiosRequestConfig` -- but if it allows that (without complaining) that's probably better anyway.

cmoesel (changes requested):
Cool.  Exactly what I was thinking.  Nice job.  I just have a few minor suggestions that will make the code just a bit cleaner.

regression/find-repos.ts
guhanthuran: Done.

guhanthuran:
Resolved Chris' changes.

src/utils/axiosUtils.ts
jafeltra: I think it would be good to do the last thing that @cmoesel mentioned here. I believe you should be able to update the argument to the function to specify the `AxiosRequestConfig` type (so it's `options?: AxiosRequestConfig` like Chris mentioned). This also should mean that you don't need to specify the type when assigning a value to `axiosOptions` because it will get the same type as the option in the function.

I think this is probably preferable so we don't pass in options that aren't supported.

jafeltra (commented):
I only left one small additional comment, which was actually a comment on one of Chris's comments. Everything else is looking good to me!

src/utils/axiosUtils.ts
guhanthuran: Done!

jafeltra (approved):
Looks good to me!

PR #1064, opened 2022-04-11
Changed structure definition to throw correctly when adding extensions to Bundle type

guhanthuran:
- Made changes to validateValueAtPath to make sure that extension can handle a null bracket. 
- Added test to check this case


jafeltra (approved):
This makes a lot of sense! The example from Zulip works as expected for me and the test covers the case there as well!

cmoesel (approved):
Yep.  This looks good to me too!

PR #1065, opened 2022-04-15
Register plugins from configuration

mint-thompson:
Completes [CIMPL-946](https://standardhealthrecord.atlassian.net/browse/CIMPL-946).

A new configuration attribute is added to allow users to register plugins. Plugins can be managed directly on the filesystem or installed from npm. For npm plugins, a version must be included. Plugins should export an initialize function, which will be called when the plugin is loaded. The initialize function is passed a function that allows hooks to be registered. Additionally, any parameters provided in the configuration will also be passed to the initialize function.

The available hooks at which functions may be registered are:
- beforeFillTank
- beforeLoadCustomResources
- afterLoadCustomResources
- afterExportFHIR

The code to call registered functions at these points is not yet implemented.

For testing this, locally-managed plugins should be placed at the FSH project's `input/plugins` path. There currently aren't any SUSHI plugins on npm, so using that feature will usually result in the plugin failing to initialize properly. The package should still get installed to `input/plugins/node_modules/${package-name}`, even if it will not initialize.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -30,7 +30,8 @@ import {
   getRandomPun,
   setIgnoredWarnings,
   getLocalSushiVersion,
-  checkSushiVersion
+  checkSushiVersion,
+  PluginManager
 } from './utils';
 
 const FSH_VERSION = '2.0.0';
@@ -143,6 +144,10 @@ async function app() {
       process.exit(0);
     }
     config = readConfig(originalInput);
+    await PluginManager.processPluginConfiguration(
+      config.plugins,
+      path.join(input, '..', 'plugin')
+    );
     tank = fillTank(rawFSH, config);
   } catch {
     program.outputHelp();
--- a/src/import/importConfiguration.ts
+++ b/src/import/importConfiguration.ts
@@ -153,6 +153,7 @@ export function importConfiguration(yaml: YAMLConfiguration | string, file: stri
     resources: parseResources(yaml.resources, file),
     pages: parsePages(yaml.pages, file),
     parameters: parseParameters(yaml, yaml.FSHOnly, file),
+    plugins: yaml.plugins,
     templates: parseTemplates(yaml.templates, file),
     template: yaml.template,
     menu: parseMenu(yaml.menu),
--- a/src/utils/PluginManager.ts
+++ b/src/utils/PluginManager.ts
@@ -1,0 +1,123 @@
+import path from 'path';
+import fs from 'fs-extra';
+import { execFile } from 'child_process';
+import util from 'util';
+import { InvalidHookError, MissingPluginError, QuestionableNpmNameError } from '../errors';
+import { PluginConfiguration } from '../fshtypes';
+import { logger } from './FSHLogger';
+
+export const AVAILABLE_HOOKS = [
+  'beforeFillTank',
+  'beforeLoadCustomResources',
+  'afterLoadCustomResources',
+  'afterExportFHIR'
+];
+
+export class PluginManager {
+  static hooks = new Map<string, ((...args: any) => any)[]>();
+
+  static async processPluginConfiguration(
+    plugins: PluginConfiguration[],
+    pluginBasePath: string
+  ): Promise<void> {
+    for (const plugin of plugins) {
+      // if the plugin name is given as name@version, it's something we may need to load from npm
+      // otherwise, we load it from the filesystem
+      // when we load from the filesystem, it might be at pluginBasePath, or at pluginBasePath/node_modules
+      const versionSeparator = plugin.name.lastIndexOf('@');
+      let pluginName: string;
+      let pluginVersion: string;
+      if (versionSeparator > 0) {
+        pluginName = plugin.name.slice(0, versionSeparator);
+        pluginVersion = plugin.name.slice(versionSeparator + 1);
+      } else {
+        pluginName = plugin.name;
+      }
+      try {
+        if (pluginVersion != null) {
+          // this is a plugin that can be acquired from npm.
+          // it might already be installed, so check for that first.
+          const existingInstallationPath = path.join(pluginBasePath, 'node_modules', pluginName);
+          if (fs.existsSync(path.join(existingInstallationPath, 'package.json'))) {
+            // check the version number in package.json
+            const packageContents = fs.readJsonSync(
+              path.join(existingInstallationPath, 'package.json')
+            );
+            if (packageContents.version === pluginVersion) {
+              PluginManager.loadFromFilesystem([existingInstallationPath], pluginName, plugin.args);
+            } else {
+              await PluginManager.installFromNpm(pluginBasePath, plugin.name);
+              PluginManager.loadFromFilesystem([existingInstallationPath], pluginName, plugin.args);
+            }
+          } else {
+            await PluginManager.installFromNpm(pluginBasePath, plugin.name);
+            PluginManager.loadFromFilesystem([existingInstallationPath], pluginName, plugin.args);
+          }
+        } else {
+          // this is a plugin managed on the local filesystem.
+          // it may be at the base path, or in the node_modules directory
+          PluginManager.loadFromFilesystem(
+            [
+              path.join(pluginBasePath, pluginName),
+              path.join(pluginBasePath, 'node_modules', pluginName)
+            ],
+            pluginName,
+            plugin.args
+          );
+        }
+      } catch (err) {
+        logger.error(`Couldn't load plugin ${plugin.name}: ${err.message}`);
+      }
+    }
+  }
+
+  // a small test for name safety, based on typical naming patterns for npm packages
+  static isNameNpmSafe(name: string): boolean {
+    return /^[\w\d@/._-]+$/.test(name);
+  }
+
+  static async installFromNpm(installPath: string, installTarget: string) {
+    if (PluginManager.isNameNpmSafe(installTarget)) {
+      fs.ensureDirSync(installPath);
+      const installationResult = await util.promisify(execFile)(
+        'npm',
+        ['install', '--prefix', '.', installTarget],
+        {
+          cwd: installPath,
+          shell: true
+        }
+      );
+      if (installationResult.stderr.length > 0) {
+        throw new Error(installationResult.stderr);
+      } else {
+        logger.info(`Installed plugin ${installTarget}`);
+      }
+    } else {
+      throw new QuestionableNpmNameError();
+    }
+  }
+
+  static loadFromFilesystem(basePaths: string[], name: string, pluginArgs: any[] = []) {
+    for (const basePath of basePaths) {
+      const pluginPath = path.resolve(basePath);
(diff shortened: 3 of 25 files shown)


src/utils/PluginManager.ts
```diff
@@ -0,0 +1,123 @@
...
+            if (packageContents.version === pluginVersion) {
+              PluginManager.loadFromFilesystem([existingInstallationPath], pluginName, plugin.args);
+            } else {
+              await PluginManager.installFromNpm(pluginBasePath, plugin.name);
+              PluginManager.loadFromFilesystem([existingInstallationPath], pluginName, plugin.args);
+            }
+          } else {
+            await PluginManager.installFromNpm(pluginBasePath, plugin.name);
+            PluginManager.loadFromFilesystem([existingInstallationPath], pluginName, plugin.args);
+          }
```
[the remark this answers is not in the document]
cmoesel: I suppose it could be simplified even further if you just want the block above to only handle doing the NPM install.  Then you could take the block below (lines 57 - 66) out of the else block and let them load the npm module you just installed.  The main difference would be that a project could override the npm installed version w/ a local version instead.  Which maybe is a good thing?  I'm not sure.

src/utils/PluginManager.ts
```diff
@@ -0,0 +1,123 @@
...
+        }
+      } catch (err) {
+        logger.error(`Couldn't load plugin ${plugin.name}: ${err.message}`);
+      }
+    }
+  }
+
+  // a small test for name safety, based on typical naming patterns for npm packages
+  static isNameNpmSafe(name: string): boolean {
+    return /^[\w\d@/._-]+$/.test(name);
```
cmoesel, line 76: Is this to avoid someone making a name like `foo@1.0.0 && sudo rm -rf /*`?

src/utils/PluginManager.ts
```diff
@@ -0,0 +1,123 @@
...
+      throw new QuestionableNpmNameError();
+    }
+  }
+
+  static loadFromFilesystem(basePaths: string[], name: string, pluginArgs: any[] = []) {
+    for (const basePath of basePaths) {
+      const pluginPath = path.resolve(basePath);
+      if (fs.existsSync(pluginPath) && fs.statSync(pluginPath).isDirectory()) {
+        // eslint-disable-next-line @typescript-eslint/no-var-requires
+        require(pluginPath).initialize(PluginManager.registerHook, ...pluginArgs);
```
cmoesel, line 105: If the plugin does not export an `initialize` function, I think I think it might be nice to provide a friendlier error here saying that the plugin does not have an initialize function.

src/import/importConfiguration.ts
```diff
@@ -153,6 +153,7 @@ export function importConfiguration(yaml: YAMLConfiguration | string, file: stri
     resources: parseResources(yaml.resources, file),
     pages: parsePages(yaml.pages, file),
     parameters: parseParameters(yaml, yaml.FSHOnly, file),
+    plugins: yaml.plugins,
```
cmoesel, line 156: For some YAML fields we allow for "simple" declaration and complex declaration.  Do you think there is any value in supporting:
```yaml
plugins:
  - foo@1.0.0
```
as a simplified version of:
```yaml
plugins:
  - name: foo@1.0.0
```

Or do you think it would be better to be consistent?

cmoesel (changes requested):
This is really good, @mint-thompson.  Very well thought-out, designed, and implemented.  I really like the approach that you've taken here.

I started to test it and ran into one issue right away.  SUSHI exits with a usage message if you run it on a project that has a configuration _without_ `plugins` in the YAML.  It seems that SUSHI is _expecting_ a `plugins` config param now -- but it should be optional.

I've left a few individual comments, but I have one more general comment: I think it would be good to issue a warning if SUSHI finds any folders in the plugins directory that _aren't_ specified in the config.  It seems like this might be a common mistake (adding the plugin to the folder but forgetting the config) -- so I'd like to detect it and warn on it.

Again, this is super good, Mint.  Nice job!

src/utils/PluginManager.ts
mint-thompson: That's correct: since this is user input, I don't want to hand it to a shell without checking it first.

src/import/importConfiguration.ts
mint-thompson: I kind of want to stick with only allowing the complex declaration, or at least requiring the complex declaration if the declaration includes arguments. Otherwise, I feel like there might be too many variations:
```yaml
plugins:
  - dog@1.0.0 # simple, name only
  - name: cat@1.0.3 # complex, name only
  - horse@3.1.1 # simple with args
    - potoooooooo
    - true
  - name: eel@4.1.1 # complex with args
    args:
    - electric
    - 3
```
Do four variations seem like too many?

cmoesel: Ha! Yes! Four variations is too many! I guess I'm just trying to be consistent w/ how we treat other things in the config.  Here are some examples from the [exhaustive config](https://github.com/FHIR/sushi/pull/1065#discussion_r854394955) to consider:

```yaml
dependencies:
  hl7.fhir.us.core: 3.1.0
  hl7.fhir.us.mcode:
    id: mcode
    uri: http://hl7.org/fhir/us/mcode/ImplementationGuide/hl7.fhir.us.mcode
    version: 1.0.0

resources:
  Patient/my-example-patient:
    name: My Example Patient
    description: An example Patient
    exampleBoolean: true
  Patient/bad-example: omit

pages:
  index.md:
    title: Example Home
  implementation.xml:
  examples.xml:
    title: Examples Overview
    simpleExamples.xml:
    complexExamples.xml:
```

I'm noticing that none of our top-level properties are arrays.  They're all objects.  So I almost wonder if we should be considering this instead:

```yaml
plugins:
  dog: # simple name (local)
  cat: 1.0.3 # simple name and version (NPM)
  horse: # complex name and args (local)
    args:
    - potoooooooo
    - true
  eel: # complex name, version, args (NPM)
    version: 4.1.1
    args:
    - electric
    - 3
```

Ah... but now that I've typed that, I remember -- _order is significant_.  That's why it needs to be a sequence.  So... is this silly?

```yaml
plugins:
  - dog # simple name (local)
  - cat: 1.0.3 # simple name and version (NPM)
  - horse: # complex name and args (local)
      args:
      - potoooooooo
      - true
  - eel: # complex name, version, args (NPM)
      version: 4.1.1
      args:
      - electric
      - 3
```

I am probably overthinking it.  But since `dependencies` is `name: version`, it feels like maybe we should try to be at least a little consistent with that?  I dunno.

mint-thompson:
@cmoesel I've updated the configuration parsing to allow for plugin configurations similar to those that you posted. The example configuration file shows a few of the options. I think this should be more in line with the other configuration elements.

src/app.ts
```diff
@@ -143,6 +144,12 @@ async function app() {
       process.exit(0);
     }
     config = readConfig(originalInput);
+    if (config.plugins) {
+      await PluginManager.processPluginConfiguration(
+        config.plugins,
+        path.join(input, '..', 'plugin')
```
cmoesel, line 150: The PR description says to put plugins in `input/plugins`, but this is looking at `input/plugin`.  I expect we want the plural.

I also wonder if `input/plugins` is the right path anyway.  Plugins are more software than input; as I think of `input` as containing _data_.  In addition, I think the name needs to be SUSHI specific (to distinguish from plugins for, say, the IG Publisher).

So I'd suggest we consider something like `${projectRoot}/sushi-plugins/` instead.

src/utils/PluginManager.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  static loadFromFilesystem(basePaths: string[], name: string, pluginArgs: any[] = []) {
+    for (const basePath of basePaths) {
+      const pluginPath = path.resolve(basePath);
+      if (fs.existsSync(pluginPath) && fs.statSync(pluginPath).isDirectory()) {
+        // eslint-disable-next-line @typescript-eslint/no-var-requires
+        const pluginModule = require(pluginPath);
+        // does the module actually have an initialize function?
+        if (pluginModule.initialize != null && typeof pluginModule.initialize === 'function') {
+          pluginModule.initialize(PluginManager.registerHook, ...pluginArgs);
```
cmoesel, line 100: I wonder if the `initialize` function should be asynchronous, just in case the plugin needs to do anything fancy?

cmoesel, line 100: I wonder if the `initialize` function should be asynchronous, just in case the plugin needs to do anything fancy?  I tested w/ an async initialize function and since this does not `await` it, it never got resolved.

cmoesel (changes requested):
This is looking good!  I noticed a couple of things to look at, which I commented on individually.

I also tried installing an NPM dependency.  Specifically GoFSH, like this:
```yaml
plugins:
  - gofsh: 1.0.0
```

I think it is supposed to at least successfully load it, but I got this instead:
```
error Couldn't load plugin gofsh: npm WARN deprecated mkdirp@0.3.0: Legacy versions of mkdirp are no longer supported. Please update to mkdirp 1.x. (Note that the API surface has changed to use Promises in 1.x.)
npm WARN saveError ENOENT: no such file or directory, open '/Users/cmoesel/dev/fsh/TestPlugins/input/plugin/package.json'
npm notice created a lockfile as package-lock.json. You should commit this file.
npm WARN enoent ENOENT: no such file or directory, open '/Users/cmoesel/dev/fsh/TestPlugins/input/plugin/package.json'
npm WARN plugin No description
npm WARN plugin No repository field.
npm WARN plugin No README data
npm WARN plugin No license field.
```
It also added a `package-lock.json` to my `input/packages`.  I'm not sure if that's expected or not.

mint-thompson:
@cmoesel It seems like my code that checks for problems during installation is matching too many situations. I'll update it to only say that installation failed if npm emits an error, not just a warning. As for the `package-lock.json`, I'm not sure what's going on there, as that doesn't happen for me.

cmoesel (approved):
Thanks for the updates, @mint-thompson!  Everything works well for me now!  I know @julianxcarter reported that he is still getting a `package-lock.json`, but I wasn't able to reproduce this using Node 12, 14, or 16.  Julian -- is it possible that `package-lock.json` was already there before you ran SUSHI? Or that you're running an older version of Mint's code?

I am approving, but I do have one suggestion.  I noticed that the logging pauses for a while as it's installing a plugin from NPM.  I think it would be helpful to log a message indicating that SUSHI is installing the XYZ plugin from NPM.  Then it's more clear to users why there is a delay.

julianxcarter:
> Thanks for the updates, @mint-thompson! Everything works well for me now! I know @julianxcarter reported that he is still getting a `package-lock.json`, but I wasn't able to reproduce this using Node 12, 14, or 16. Julian -- is it possible that `package-lock.json` was already there before you ran SUSHI? Or that you're running an older version of Mint's code?

I was 😅. Nothing to see here!

cmoesel (approved):
Still good!

PR #1066, opened 2022-04-18
Warn when a slice element is accessed via numeric index

### LINKED ISSUE #1030: Warn when users access slices by numeric indices (or soft indexing)
SUSHI auto-populates _required_ slices into instances before processing any rules.  When users then write rules for the same array by number (or soft-index), then they may be accessing these pre-populated slice entries (intentionally or unintentionally).

It is much safer to write rules that call out the slices by name rather than using indices (which are subject to the order in which SUSHI has written the slices into the instance). So... SUSHI should warn users when they do this.

This ultimately helps lead us toward a situation in which we can more safely change how SUSHI handles slices and soft indexing.

See #1028 for one relevant example where this is a problem.

julianxcarter:
Fixes #1030, logging a warning when users access a slice using a numeric index. This warning is only logged when an array has it's slicing rules set to `closed` (in which case all elements in the array are slices by definition), or if the numeric index is determined to reference a slice that will be pre-loaded onto an instance as an implied property. A slice will be loaded automatically if it's cardinality indicates that it is required _and_ it contains a `pattern[x]` or `fixed[x]` value on itself or any of its children. 

This [FSH Online link](https://fshschool.org/FSHOnline/#/share/3ryTLXM) includes test cases.

### DIFF
--- a/src/export/InstanceExporter.ts
+++ b/src/export/InstanceExporter.ts
@@ -119,7 +119,8 @@ export class InstanceExporter implements Fishable {
           rule.path,
           rule.value,
           this.fisher,
-          inlineResourceTypes
+          inlineResourceTypes,
+          rule.sourceInfo
         );
         // Record each valid rule in a map
         // Choice elements on an instance must use a specific type, so if the path still has an unchosen choice element,
--- a/src/fhirtypes/ElementDefinition.ts
+++ b/src/fhirtypes/ElementDefinition.ts
@@ -663,22 +663,30 @@ export class ElementDefinition {
   }
 
   /**
+   * Returns an array of slices that will be pre-loaded.
+   * A slice is pre-loaded if if has a min of 1 and contains a fixed or pattern value on itself or it's descendents
+   * @returns {ElementDefinition[]} - Array of slices to be pre-loaded
+   */
+  getPreloadedSlices(): ElementDefinition[] {
+    return this.getSlices().filter(
+      slice =>
+        slice.min > 0 &&
+        slice
+          .getAssignableDescendents()
+          .some((element: ElementDefinition) =>
+            Object.keys(element).find(k => k.startsWith('fixed') || k.startsWith('pattern'))
+          )
+    );
+  }
+
+  /**
    * Determines if an array index references a slice that will be preloaded.
    * A slice is pre-loaded if if has a min of 1 and contains a fixed or pattern value on itself or it's descendents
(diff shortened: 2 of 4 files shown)


mint-thompson (changes requested):
It looks like the logic for `getPreloadedSlices` is similar to the logic in `setImpliedPropertiesOnInstance`. This makes sense to me, but there is a scenario which is not handled by `getPreloadedSlices`: when the Profile has a complex type as a fixed/pattern value on a slice, but the Instance of that Profile refers to an element within that complex type. This is an example:
```
Profile: UnusualQuestionnaire
Parent: Questionnaire
* contact ^slicing.discriminator[0].path = "name"
* contact ^slicing.discriminator[0].type = #value
* contact ^slicing.rules = #open
* contact contains bones 1..1
* contact[bones] = Cash4Bones

Instance: UnusualInstance
InstanceOf: UnusualQuestionnaire
* status = #in-progress
* item[0].linkId = "link id"
* item[0].type = #display
* contact[0].telecom[0].use = #work

Instance: Cash4Bones
InstanceOf: ContactDetail
Usage: #inline
* name = "someone suspicious"
* telecom[0].value = "1 800 cash 4 bones"
```

In `UnusualInstance`, the `contact[bones]` slice is referred to numerically, so a warning should be emitted. The warning _does_ get emitted correctly if the values on `contact[bones]` slices are defined individually, instead of by using the inline instance.

guhanthuran (approved):
Looks good to me - after implementing Mint's changes I think this is ready to go.

julianxcarter:
> It looks like the logic for `getPreloadedSlices` is similar to the logic in `setImpliedPropertiesOnInstance`. This makes sense to me, but there is a scenario which is not handled by `getPreloadedSlices`: when the Profile has a complex type as a fixed/pattern value on a slice, but the Instance of that Profile refers to an element within that complex type.

Thank you for catching this, I hadn't thought of the possibility that a sliced element _itself_ could have a `fixed[x]` or `pattern[x]` value, and not just it's children. Commit [b98a8a9](https://github.com/FHIR/sushi/pull/1066/commits/b98a8a9d906e2f253a14374a91af044895a4d7a2) addresses this.

mint-thompson (approved):
Thank you! Hopefully this will encourage better authoring around slices.

guhanthuran (approved):
Looks good to me!

Issue #1017, opened 2022-02-09
SUSHI does not load #current correctly when most recent build is a non-main branch


cmoesel:
SUSHI uses https://build.fhir.org/ig/qas.json to find #current packages. It filters the entries by package id, sorts by date, and chooses the most recent one.  It then uses the repo org and name to construct a URL to download the package.

The problem is that the qas.json contains build entries from all branches (not just master or main), but SUSHI _always_ grabs the package from the main build location (regardless of branch).  This is problematic for a few reasons:
* SUSHI is using the date of _whatever_ branch, but the package of the main branch -- so the package may not be as recent as SUSHI thinks it is.
* It's even possible that although a build exists for the non-main branch, there is no build on the master/main branch. This happened today with VRDR (see comment below).
* Should #current even reference non-main builds / branches?  Probably not.

So... SUSHI loading should probably be updated to filter qas.json entries for only those on `master` or `main` branches.

cmoesel:
Here is a message I sent a Saul K after he reported that a dependency on `hl7.fhir.us.vrdr#current` was resulting in a 404 error from SUSHI:

The way SUSHI handles #current is to look at https://build.fhir.org/ig/qas.json and filter down to the entries w/ matching packageId.  In the case of VRDR. that's 3 entries:
* saulakravitz/vrdr/branches/FSHversion/
* HL7/vrdr/branches/FHIR-33348-Death-Certificate-composition/
* HL7/vrdr/branches/master

It then finds the one w/ the most recent date.  In this case, that was saulakravitz/vrdr/branches/FSHversion/.  After that, it attempts to grab the package.tgz by constructing a URL based on the repo org (saulakravitz) and name (vrdr) -- resulting in: https://build.fhir.org/ig/saulakravitz/vrdr/package.tgz (which currently 404s).

There is a bug here -- because your package is actually in the FSHversion branch, but SUSHI looks for the main build (which is probably the master branch).  SUSHI needs to fix this, although I'm not 100% sure of what the right solution is.  Either:
* Grab the package.tgz from the correct branch location on the build server, OR
* Ignore non-master/non-main branches when filtering for entries on qas.json.

I'm inclined to think we should be doing the latter.

PR #1068, opened 2022-04-20
Only load current packages from main or master branches

jafeltra:
This PR fixes SUSHI issue #1017.  It supports only downloading current packages from `main` or `master` branches by filtering out the `qas.json` entries that are not from those branches.


mint-thompson (approved):

cmoesel (approved):
Looks good to me and I confirmed with a full regression as well.  Thanks!

PR #1071, opened 2022-04-25
Avoid assigning duplicate values on sliced elements

### LINKED ISSUE #1008: SUSHI duplicates pattern values when profiles re-declare patterns and slices are involved
In some cases, when a profile specifies the same value in patterns at different levels, SUSHI applies both patterns even though they are trying to enforce the same thing -- and as a result, SUSHI duplicates values in the pattern.  After some exploration, this seems to only happen when slices are involved (and more specifically, slices at each pattern application.

This was reported [here](https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/.5Bsushi.5D.20pattern.20in.20Instance.20applied.20twice/near/268375871), but I've created a simplified example [here](https://fshschool.org/FSHOnline/#/share/33ueh2O).

The simplified example is:
```
Profile: MyProfile
Parent: Observation
* component ^slicing.discriminator.type = #pattern
* component ^slicing.discriminator.path = "code"
* component contains MySlice 1..1
* component[MySlice].code = http://foo#bar
* component[MySlice].code.coding 1..*
* component[MySlice].code.coding = http://foo#bar
* component[MySlice].code.coding ^slicing.discriminator.type = #pattern
* component[MySlice].code.coding ^slicing.discriminator.path = "$this"
* component[MySlice].code.coding ^slicing.rules = #open
* component[MySlice].code.coding contains MySubSlice 1..1
* component[MySlice].code.coding[MySubSlice] = http://foo#bar

Instance: MyInstance
InstanceOf: MyProfile
* code = http://any.com#code
* status = #final
```

And results in this instance:
```json
{
  "id": "MyInstance",
  "meta": {
    "profile": [
      "http://example.org/StructureDefinition/MyProfile"
    ]
  },
  "component": [
    {
      "code": {
        "coding": [
          {
            "code": "bar",
            "system": "http://foo"
          },
          {
            "code": "bar",
            "system": "http://foo"
          }
        ]
      }
    }
  ],
  "code": {
    "coding": [
      {
        "code": "code",
        "system": "http://any.com"
      }
    ]
  },
  "status": "final"
}
```
_Note that the component.code.coding is repeated in the instance._

mint-thompson:
Fixes #1008 and completes task [CIMPL-881](https://standardhealthrecord.atlassian.net/browse/CIMPL-881).

When there are required slices on an element, the sliced element's minimum cardinality is at least the sum of the slice minimums. If the minimum cardinality is greater than the sum of slice minimums, additional elements should be added on Instances in order to satisfy the minimum cardinality. However, if the slices provide enough elements to satisfy the element's minimum cardinality, there is no need to add extra elements. Previously, when assigning implied values on an Instance, the "must-satisfy-minimum" elements were added before any named slices, resulting in more elements being present than expected. By adding the named slices first, those unnecessary elements are not added.

The strategy for ensuring that named slices are added before trying to pad out the sliced element is to reorder the rules on the Instance's `InstanceOf` profile so that the following conditions hold:
- if there is a rule on a sliced element and on a named slice of that element, the named slice rule should come first. The first new test addresses this case.
- if there is a rule on an element and a descendant of that element, the descendant rule should come first. The second new test addresses this case.

Because the rule path ordering should be preserved apart from these conditions, my solution was to build a tree out of the element paths. This is more reliable than using the built-in sort with a comparator, because the correct order of two paths cannot always be determined by paths alone if one path's ancestor is in the set of paths, and that ancestor is the other path's sibling. For example, `Resource.cookie` and `Resource.cake.frosting` cannot be ordered without knowing whether `Resource.cake` is also in the set of paths, and where it would exist in the ordering. The tree implementation handles this scenario without any trouble. Also, if you read this all the way to the end, please treat yourself to a cookie and/or cake in any order.

### DIFF
--- a/src/fhirtypes/common.ts
+++ b/src/fhirtypes/common.ts
@@ -1,4 +1,4 @@
-import { isEmpty, cloneDeep, upperFirst } from 'lodash';
+import { isEmpty, cloneDeep, upperFirst, remove } from 'lodash';
 import {
   StructureDefinition,
   PathPart,
@@ -33,7 +33,7 @@ import {
 } from '../fshtypes';
 import { FSHTank } from '../import';
 import { Type, Fishable } from '../utils/Fishable';
-import { logger } from '../utils';
+import { logger, parseFSHPath } from '../utils';
 
 export function splitOnPathPeriods(path: string): string[] {
   return path.split(/\.(?![^\[]*\])/g); // match a period that isn't within square brackets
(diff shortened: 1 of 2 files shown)


jafeltra (approved):
Wow this is a really nice solution! The code makes sense to me and the example in the issue works as expected now.

mint-thompson:
With the new changes, things should be in a reasonably good state, I think? I'm not going to make any guarantees about nested arrays, but things look acceptable for the cases described in the issue, and one more that I thought of when I looked at the test coverage report.

jafeltra (approved):
This seems like it makes sense to me (I say hesitantly after approving this the last time). The code for `assignComplexValue` looks really good and the comments are very helpful to understand what is going on, which will hopefully help future-us. The test cases and the initial examples in the issue look good to me as well.

guhanthuran (approved):
Looks good to me - the method you use to place the values in the tree is neat, it took me a bit to understand it but it makes a lot of sense to use it with the Depth-First traversal.

PR #1078, opened 2022-05-05
Support for parents with versioned canonical URLs as their baseDefinition

### LINKED ISSUE #1067: Unable to get type lineage for types that contain a version
The `StructureDefinition.baseDefinition` element is used to find a type lineage, which is necessary in order to constrain a type. This element's type is `canonical(StructureDefinition)`, which means it may contain a version at the end. For example, `https://example.org/StructureDefinition/MyProfile|1.0.1` is a valid canonical. However, when fishing for that string, no results will be found, as the maps that contain FHIR resources will not have the version in their keys. Separate the key before fishing so that a resource can be found, then confirm that the version is correct.

See https://github.com/michaelonken/sushi_problem for an example of this issue: the `https://fhir.kbv.de/StructureDefinition/KBV_PR_Base_Observation_Body_Weight` profile's `baseDefinition` is `http://fhir.de/StructureDefinition/observation-de-vitalsign-koerpergewicht|0.9.13`, which will not be found even if the `de.basisprofil.r4` package at version 0.9.13 is included as a dependency.

guhanthuran:
Fixes #1067 

When the error is detected, throw a more specific type of error that specifies that SUSHI has found a different version type than the one provided. 

To test:
Take a look at the example IG provided in the issue. If the error implemented in this PR is shown when you run it, that means the bug has successfully been fixed.

### DIFF
--- a/test/fhirtypes/ElementDefinition.constrainType.test.ts
+++ b/test/fhirtypes/ElementDefinition.constrainType.test.ts
@@ -1,4 +1,4 @@
-import { TestFisher } from '../testhelpers';
+import { TestFisher, loggerSpy } from '../testhelpers';
 import { loadFromPath } from '../../src/fhirdefs/load';
 import { FHIRDefinitions } from '../../src/fhirdefs/FHIRDefinitions';
 import { StructureDefinition } from '../../src/fhirtypes/StructureDefinition';
@@ -31,6 +31,7 @@ describe('ElementDefinition', () => {
   });
 
   beforeEach(() => {
+    loggerSpy.reset();
     observation = fisher.fishForStructureDefinition('Observation');
     planDefinition = fisher.fishForStructureDefinition('PlanDefinition');
     extension = fisher.fishForStructureDefinition('Extension');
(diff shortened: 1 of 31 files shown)


test/fhirtypes/ElementDefinition.constrainType.test.ts
```diff
@@ -338,6 +339,46 @@ describe('ElementDefinition', () => {
       );
     });
 
+    it('should allow us to constrain a reference to a profile with a parent with a canonical including the version', () => {
```
cmoesel, line 342: I know we came up w/ this test name together, but I found it confusing when I was reviewing.  I think this might be better:
```suggestion
    it('should allow us to constrain a reference to a profile whose parent is specified using a versioned canonical URL', () => {
```

test/fhirtypes/ElementDefinition.constrainType.test.ts
```diff
@@ -338,6 +339,46 @@ describe('ElementDefinition', () => {
...
+      expect(hasMember.type).toHaveLength(1);
+      expect(hasMember.type[0]).toEqual(
+        new ElementDefinitionType('Reference').withTargetProfiles(
+          'https://fhir.kbv.de/StructureDefinition/KBV_PR_Base_Observation_Body_Weight'
+        )
+      );
+      expect(loggerSpy.getAllMessages('error')).toHaveLength(0);
+    });
+
+    it('should log an error when referencing a profile with a parent with a wrong version', () => {
```
cmoesel, line 361: And maybe this is a little better?
```suggestion
    it('should log an error when constraining a reference to a profile whose parent is specified using the wrong version', () => {
```

cmoesel (changes requested):
Thanks, @guhanthuran!  This seems to be working well!  I left a few small suggestions, but they should be easy to implement!

I think the PR description on how to test is a little confusing though, so I'd like to clarify for the 2nd reviewer:

> If you run the released SUSHI (or master) against that test project, you should get an error about not matching any allowed types.
> 
> Initially if you run this PR's SUSHI against that test project, you'll still get that error! That's because the referenced profile declares a profile from another IG as its parent -- but that IG isn't a dependency, so SUSHI can't find it.  To fix it, you need to uncomment the `de.basisprofil.r4: 0.9.13` dependency in sushi-config.yaml.  Then this PR SUSHI will succeed (but the master SUSHI will still fail).
> 
> As for the new error message added in this PR, you can reproduce that by changing the version of `de.basisprofil.r4` in the dependencies to `0.9.12` -- that way there will be a mismatch in versions.  When you run this PR's SUSHI, you'll get the message about the mismatch, but the type constraint will still succeed (since SUSHI does best-effort and accepts the mismatched version).

jafeltra (commented):
Thanks for the additional information @cmoesel. That helped me a lot!

I didn't notice anything else beyond the few things that Chris pointed out. Once those are set, I'm good to approve!

cmoesel (approved):
Assuming all the CI tests pass, this looks good to me!

@jafeltra -- can you re-review as well?

jafeltra (approved):
Looks good!

PR #1082, opened 2022-05-11
Honor ENV variable for NPM_CONFIG_REGISTRY when checking sushi version

bkaney:
For large organizations it is common to have a local mirror of npmjs.org, and even go so far as disallowing installing NPM packages directly or blocking requests to registry.npmjs.org. This patch will honor the ENV setting to allow sushi to work in these situations -- preventing the sushi command from always failing.

This problem emerged after https://github.com/FHIR/sushi/pull/1036 was merged.

### DIFF
--- a/src/utils/Processing.ts
+++ b/src/utils/Processing.ts
@@ -560,13 +560,14 @@ export function getLocalSushiVersion(): string {
 }
 
 async function getLatestSushiVersionFallback(): Promise<string> {
+  logger.info('Attempting fallback to determine version of sushi.');
   try {
     const res = await axiosGet('https://registry.npmjs.org/fsh-sushi');
     const latestVer = res.data['dist-tags'].latest;
-    if (latestVer == null) {
-      logger.warn('Unable to determine the latest version of sushi.');
-    } else {
+    if (latestVer.match(/^[0-9\.]*$/)) {
       return latestVer;
+    } else {
+      logger.warn('Unable to determine the latest version of sushi.');
     }
   } catch (e) {
     logger.warn(`Unable to determine the latest version of sushi: ${e.message}`);
@@ -579,18 +580,11 @@ export async function getLatestSushiVersion(): Promise<string | undefined> {
   const getRegistryCmd = 'npm view fsh-sushi version';
   try {
     const execResult = execSync(getRegistryCmd)?.toString()?.replace(/\s*$/, '');
-
     if (execResult.match(/^[0-9\.]*$/)) {
       latestVer = execResult;
-    } else {
-      logger.warn(
-        'Unable to determine the latest version of sushi: ' +
-          `command "${getRegistryCmd}" returned "${execResult}", ` +
-          'which does not look like a semver.'
-      );
     }
   } catch (e) {
-    logger.warn(`Unable to determine the latest version of sushi: ${e.message}`);
+    logger.info(`Unable to determine the latest version of sushi: ${e.message}`);
   }
 
   if (latestVer === undefined) {
--- a/src/utils/Processing.ts
+++ b/src/utils/Processing.ts
@@ -559,29 +559,51 @@ export function getLocalSushiVersion(): string {
   return null;
 }
 
-export function getLatestSushiVersion(): string | undefined {
+async function getLatestSushiVersionFallback(): Promise<string> {
+  try {
+    const res = await axiosGet('https://registry.npmjs.org/fsh-sushi');
+    const latestVer = res.data['dist-tags'].latest;
+    if (latestVer == null) {
+      logger.warn('Unable to determine the latest version of sushi.');
+    } else {
+      return latestVer;
+    }
+  } catch (e) {
+    logger.warn(`Unable to determine the latest version of sushi: ${e.message}`);
+  }
+}
+
+export async function getLatestSushiVersion(): Promise<string | undefined> {
   let latestVer = undefined;
 
   const getRegistryCmd = 'npm view fsh-sushi version';
   try {
-    const execResult = execSync(getRegistryCmd)
-      ?.toString()
-      ?.replace(/\s*$/, '');
+    const execResult = execSync(getRegistryCmd)?.toString()?.replace(/\s*$/, '');
 
     if (execResult.match(/^[0-9\.]*$/)) {
       latestVer = execResult;
     } else {
-      throw new Error(`command "${getRegistryCmd}" returned "${execResult}", which does not look like a semver.`);
+      logger.warn(
+        'Unable to determine the latest version of sushi: ' +
+          `command "${getRegistryCmd}" returned "${execResult}", ` +
+          'which does not look like a semver.'
+      );
     }
   } catch (e) {
     logger.warn(`Unable to determine the latest version of sushi: ${e.message}`);
-  } finally {
-    return latestVer;
   }
+
+  if (latestVer === undefined) {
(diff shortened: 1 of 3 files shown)


bkaney:
FYI - I plan update this. I just realized a better approach would be to use `npm get registry` to obtain the URL - this way it would honor the ENV or `~/.npmrc` settings, etc.

cmoesel:
Thanks, @bkaney!  We'll gladly take this once it is ready.

Be sure to see #1085 for some related changes.  One thing we did in that PR is we demoted the failure to reach NPM from _error_ to _warning_.

src/utils/Processing.ts
```diff
@@ -560,7 +560,11 @@ export function getLocalSushiVersion(): string {
 
 export async function getLatestSushiVersion(): Promise<string> {
   try {
-    const res = await axiosGet('https://registry.npmjs.org/fsh-sushi');
+    const registryRoot = process.env['NPM_CONFIG_REGISTRY']
+      ? process.env['NPM_CONFIG_REGISTRY']
+      : 'https://registry.npmjs.org';
```
cmoesel, line 565: I expect this will end up changing with your update, but note that I think this can be simplified to:
```suggestion
    const registryRoot = process.env['NPM_CONFIG_REGISTRY'] ?? 'https://registry.npmjs.org';
```

bkaney:
@cmoesel  -- this last commit is what I think is better - using the output of `npm get registry`. What do you think of this approach?

cmoesel:
Thanks, @bkaney.  I always like to think twice before introducing code that does an `exec` call.  I was hoping there would be a way to do it through an existing library (that doesn't call `exec` itself).  I just spent way too much time learning that:
* NPM gets the registry by looking at env vars, command flags, and files in four different possible locations (some dependent on where the `npm` binary lives).
* The `npm` module used to expose an API prior to version 8.0.0 (but we probably don't want to forever depend on an old version).
* While [npm-registry-fetch](https://github.com/npm/npm-registry-fetch) and [pacote](https://github.com/npm/pacote/) are good for fetching from the registry, they don't have the built-in logic to detect the registry (I tried).
* [@npmcli/config](https://github.com/npm/config/) doesn't load your current NPM config for you (I tried).

So it seems that if we want a robust solution (and we do), doing a simple `exec` on npm is the best way to go.  The only remaining question is... if we're going to do an exec on `npm` anyway, should we just cut to the chase and do `npm view fsh-sushi version`?  And maybe fallback to axios (w/ default registry) if calling npm fails? Or is that just silly?

Also, did you happen to test this when invoked via the IG Publisher to see if calling `npm` in that context still works?  IIRC, it may do some funky stuff w/ PATH.  I just want to be sure.

bkaney:
Well it depends on what you think the importance for this information is. 

That fix that doesn't make the build fail if the orig implementation is what I would say is the minimum required (e.g. : https://github.com/FHIR/sushi/pull/1085/files#diff-3fef64f89103af27309d9b93d97ae52e1f0858ea40e887534c83a9829a69bc47).

But I would be happy to:
1. Replace with `npm view ...` (which if we are going to do any of this I agree makes the most sense)
2. Test with IG Publisher
3. Test on Windows

Let me know if this would be helpful.

cmoesel:
Hi @bkaney -- Thanks for the response. I think I'm fine w/ using `npm view...` if we fail silently and fallback to the axios call to get the version from `https://registry.npmjs.org`.  If that sounds reasonable to you, then let's do it!  If you can test w/ IG Publisher and Windows, I can test with IG Publisher and Mac.

We're likely going to put out a SUSHI release today w/ the other fix.  If you continue w/ this PR, it will make it into the next release after that.

bkaney:
Hey @cmoesel  - here is a reimplementation. I tested on mac/linux including IG Publisher. It will take a bit of time for me to test on windows with IG Publisher, because I don't really ever use windows and will have to learn how to setup java, ruby, etc on that platform.

Does someone on your team have windows?

cmoesel:
Hey @bkaney -- thanks for continuing on this.  I can ask @mint-thompson to test this on Windows.  But first, I was really hoping to keep the GET request to the npmjs registry as a fallback for if the `npm` call fails.  There may be systems where `npm` is not on the path or the user cannot execute `npm` for some other reason.  In these cases, I'd like to try the NPMJS API call, as that stands a chance of succeeding when `npm view` does not.  Does that make sense?

cmoesel:
Hi @bkaney.  It was nice to see you in Cleveland a few weeks ago!  I just wanted to check in to see if you were planning to reintegrate the GET call to npmjs just in case the NPM exec call fails...  If so, great!  If not, do you mind if I make that update before merging in this PR?

bkaney:
> Hi @bkaney. It was nice to see you in Cleveland a few weeks ago! I just wanted to check in to see if you were planning to reintegrate the GET call to npmjs just in case the NPM exec call fails... If so, great! If not, do you mind if I make that update before merging in this PR?

I have a local WIP for this, let me try to get it pushed this evening for you to review.

bkaney:
@cmoesel  - Here is an implementation that will fallback to use axios, and I rebased so there should be no conflicts.

cmoesel:
Thanks, @bkaney!  Things have been pretty busy here, so I haven't had a chance to review yet, but we will! I see all of those nice unit tests!  Thanks!

cmoesel:
@bkaney -- thanks for putting up with all my requests!  Functionally, this is looking good and I appreciate the tests!  One thing I'm not sure about though... If `npm` fails for whatever reason, SUSHI will log an error saying it can't determine the latest SUSHI version; but since we have a fallback, this isn't necessarily true (because the fallback method might work fine).  It ends up being confusing since SUSHI says it failed to get the latest version, but then later might tell you that your local version is not the latest version.

I'm a little torn, because as long as SUSHI finds the latest version _some way_, then the user should not care (and maybe shouldn't be bothered by error/warning messages).  On the other hand, if it does fail all the way through, it's helpful to have those messages for debugging.

I think we should either:
* Change the npm-based error and warning's log-level to debug and only log visibly to the user if the fallback also fails.
* Change the npm-based error's log-level to warning and update both NPM log messages to indicate that SUSHI failed to get the latest version using the local npm executable, but will attempt to get the latest version by querying the npm registry directly.
* Store the error causes in variables and if the primary and fallback both fail, log all relevant errors/warnings.

What do you think?

bkaney:
@cmoesel - I like the first bullet, only warn-level, visible log if everything fails. Otherwise put in debug log.

cmoesel:
@bkaney -- did you want to make that last change or shall I?

cmoesel (approved):
Thanks for all the great work, @bkaney.  I apologize for going back and forth with you on this, but I think we landed in a good place.  It all looks good to me!  I tested this in the following ways:
* Confirmed that it doesn't warn me about an old SUSHI version when it shouldn't.
* Changed `package.json`'s `version` to be a lower value and confirmed that it _does_ warn me about an old SUSHI version.
* Temporarily replaced my `npm` executable w/ a garbage executable and confirmed that SUSHI logs that it failed running the npm command but then successfully uses the fallback.
* Completely disconnected from internet and confirmed that both npm command and fallback fail and SUSHI warns that it couldn't check latest version.

We usually require 2 reviews to merge, so I'll ask someone else from my team to do the 2nd review.

julianxcarter (approved):
This looks good and works well! Thanks for making this change @bkaney!

PR #1085, opened 2022-05-11
Support for --require-latest

### LINKED ISSUE #1083: Support flag to require latest version of SUSHI
During the FHIR IG publication workflow, it is very important that the latest version of SUSHI be used. To support this, SUSHI should introduce an optional flag that indicates that SUSHI should exit with an error if it isn't the latest version (or if it can't determine the latest version).  The default value should be false, but FHIR publication engineers can set the flag when doing a publication.

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/FSH.20install.20problem/near/281720829

### LINKED ISSUE #1084: SUSHI should only warn when it cannot determine latest version (unless --require-latest flag is set)
Currently, SUSHI issues an error if it can't reach NPM to determine if it is running the latest version.  This is likely too strong, especially for authors who may be intentionally running a different version of SUSHI.  This should be demoted to be a warning.

If SUSHI is run w/ the `--require-latest` flag, however, SUSHI should _still_ exit w/ an error.

cmoesel:
This PR contains two changes to how SUSHI handles version checking:
* If SUSHI cannot determine the latest version (failure to contact NPM), issue a _warning_ instead of an _error_.  Also add a message to the summary in this case so that it's easier to see that there was an issue.
* If SUSHI is run w/ the `--require-latest` flag then exit with an error if SUSHI is not the latest (or cannot determine the latest).

Some of this has modified unit tests, but the code in `app.ts` does not have unit tests. To manually test this:

First:
* Run SUSHI(`ts-node src/app.ts path-to-project`) and confirm it runs successfully w/ no errors or warnings about the SUSHI version.
* Run SUSHI w/ the `--require-latest` flag and confirm that SUSHI still runs successfully.

Then:
* Change the `package.json` in this branch to say `2.4.0` so it appears to be out of date.
* Run SUSHI and confirm there are no warnings/errors, but there _is_ a message in the box indicating that SUSHI is out of date.
* Run SUSHI with the `--require-latest` flag and confirm that an error is logged and SUSHI exits.

Finally:
* Turn off wifi and unplug network cables so you have no internet.
* Run SUSHI and confirm that there is a warning that it could not detect the latest version (individual log as well as summary at end).
* Run SUSHI with the `--require-latest` flag and confirm that the warning _and_ an error is logged and SUSHI exits.
* Don't forget to turn back on wifi and/or plug back in network cables.

Fixes #1083 
Fixes #1084

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -50,6 +50,11 @@ async function app() {
     .option('-d, --debug', 'output extra debugging information')
     .option('-p, --preprocessed', 'output FSH produced by preprocessing steps')
     .option('-s, --snapshot', 'generate snapshot in Structure Definition output', false)
+    .option(
+      '-r, --require-latest',
+      'exit with error if this is not the latest version of SUSHI',
+      false
+    )
     .option('-i, --init', 'initialize a SUSHI project')
     .version(getVersion(), '-v, --version', 'print SUSHI version')
     .on('--help', () => {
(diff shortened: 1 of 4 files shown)


src/app.ts
```diff
@@ -71,6 +76,41 @@ async function app() {
...
+  if (program.snapshot) {
+    logger.info('  --snapshot');
+  }
+  if (program.requireLatest) {
+    logger.info('  --require-latest');
+  }
+  if (program.out) {
+    logger.info(`  --out ${path.resolve(program.out)}`);
+  }
+  logger.info(`  ${path.resolve(input)}`);
```
cmoesel, line 97: I moved this up as I think we should print out this information _first_ before doing _any_ processing.

jafeltra (approved):
The changes work as described and I couldn't think of any other cases that you haven't covered.

julianxcarter (approved):
Works exactly as described!

PR #1093, opened 2022-05-20
Invoke registered functions from plugins

mint-thompson:
Completes the implementation portion of [INT-1177](https://standardhealthrecord.atlassian.net/browse/INT-1177).

The four hook points provided by PluginManager will call each registered function. The functions are called in the order that they were registered. If a function throws an error, the functions registered after it will still be called.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -150,6 +150,7 @@ async function app() {
         path.join(originalInput, 'sushi-plugins')
       );
     }
+    await PluginManager.callHook('beforeFillTank', rawFSH, config, logger);
     tank = fillTank(rawFSH, config);
   } catch {
     program.outputHelp();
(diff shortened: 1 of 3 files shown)


jafeltra (approved):
This looks good to me! I created a simple plugin and confirmed it is called at the various hook locations!

src/app.ts
```diff
@@ -161,7 +162,23 @@ async function app() {
   await loadExternalDependencies(defs, config);
 
   // Load custom resources. In current tank configuration (input/fsh), resources will be in input/
+  await PluginManager.callHook(
+    'beforeLoadCustomResources',
+    path.join(input, '..'),
+    originalInput,
```
cmoesel, line 168: Is it necessary to pass both the project directory _and_ the input directory?  Now that we've normalized to the recommended FHIR IG project structure, isn't it always the case that the inputs can be found at `${projectRoot}/input`?  Or are there exceptions that I'm not remembering?

cmoesel (approved):
Approving this PR as-is since we've decided to put this work to the side for now... Once it is merged into the `plugins` branch, we'll put branch protection on it in case we need it in the future.

PR #1096, opened 2022-06-08
Avoid bogus error when constraining types on choice elements involving semi-matching slice names

### LINKED ISSUE #1061: SUSHI emits bogus error when complex extensions have sub-extensions w/ similar names
If sub-extensions have names that are substrings of each other, it seems SUSHI might handle them incorrectly, resulting in a bogus error.  See [this Zulip conversation](https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Error.20on.20creation.20of.20complex.20Extensions/near/276847477) for an example.  While there is a workaround, we should figure out what's happening here and fix it in SUSHI.

mint-thompson:
Fixes #1061 and completes task [CIMPL-947](https://standardhealthrecord.atlassian.net/browse/CIMPL-947).

When checking for choices that are made obsolete as the result of type constraints, you have to be careful when checking slice names. If the element being constrained has no slice name, then any slice of it could contain a matching choice element. If the element being constrained has a slice name, then only that slice and its reslices can contain a matching choice element.

### DIFF
--- a/src/fhirtypes/StructureDefinition.ts
+++ b/src/fhirtypes/StructureDefinition.ts
@@ -804,9 +804,15 @@ export class StructureDefinition {
 
   findObsoleteChoices(baseElement: ElementDefinition, oldTypes: ElementDefinitionType[]): string[] {
     // first, find all the elements representing choices for the same choice element
-    const parentId = baseElement.parent().id;
+    const parentSlice = baseElement.parent()?.sliceName;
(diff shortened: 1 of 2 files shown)


cmoesel (approved):
Beauty.  I don't know how many times we've now found and fixed code that checked something like `e.id.startsWith(parentId)`.  Thanks for finding and fixing another one!

julianxcarter (approved):

PR #1101, opened 2022-06-14
Restores ability to run sushi without arguments

### LINKED ISSUE #1092: SUSHI fails w/ error when running without any arguments
In SUSHI 1.4.1, you can run just `sushi` and it assumes the current working directory as the path.  In SUSHI 2.5.0, however, this capability is broken.  If you run just `sushi`, you get:
> info  Running SUSHI v2.5.0 (implements FHIR Shorthand specification v2.0.0)
> info  Arguments:
> error SUSHI encountered the following unexpected error: The "path" argument must be of type string. Received undefined

If you run `sushi .` instead, it works as expected.  We should fix SUSHI so it works without any arguments again.

julianxcarter:
Fixes #1092, allowing sushi to be run without arguments (with the current directory assumed to be the FSH project). 

To test, checkout this branch and run `npm pack`. Then run `sudo npm install -g fsh-sushi-2.5.0.tgz` and attempt to use the sushi command without any arguments from the root of a FSH project.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -94,7 +94,7 @@ async function app() {
   if (program.out) {
     logger.info(`  --out ${path.resolve(program.out)}`);
   }
-  logger.info(`  ${path.resolve(input)}`);
+  logger.info(`  ${path.resolve(input || '.')}`);
 
   const sushiVersions = await checkSushiVersion();
   if (


mint-thompson (approved):
Worked perfectly on my windows system. Thank you!

cmoesel (approved):
Whodathunkit?   A logger message.  Go figure.

Nice find!

PR #1103, opened 2022-06-15
Add JSON diff to regression

mint-thompson:
Completes task [CIMPL-900](https://standardhealthrecord.atlassian.net/browse/CIMPL-900).

The JSON diff is provided alongside the text-based diff in order to help analyze potential regressions. This can help in cases where object cases are reordered, since this will appear as a change in the text-based diff, but will not appear in the JSON diff.

There are no test cases for this change, since this is only a change to the regression testing output. To test this, you can run the regression script between any two versions or branches that you would expect to show different output. For example, I found the following provided useful JSON diff output: `npm run regression regression/repos-select.txt local gh:next`.


cmoesel (approved):
Splendid!

PR #1104, opened 2022-06-16
Allow name/description in Instances

### LINKED ISSUE #1003: Consider populating title/description/name(?) in instances when possible
Authors have to use instances for any conformance resources other than SD, CS, and VS.  These conformance resources often have a `title` and `description`.  We should consider auto-populating these items from keywords (e.g., `Title:`, `Description:`) when the corresponding elements exist in the target instance type.

So far, we've avoided doing so since not all instances have these elements -- and we wanted to be consistent with our behavior across instances.  But in doing so, we're inconsistent w/ our behavior on other conformance types.

We may want to consider making this a configurable option (which means we would need to decide what the default behavior should be).  We also may want to look at the `name` element and see if we should autopopulate that (although the "name" authors give after `Instance:` is already used as the `id`).

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/instance.20title/near/266818927

guhanthuran:
Fixes #1003

Allows users to put a specified Title and/or Description in their instances. 
To test, use this: https://fshschool.org/FSHOnline/#/share/32OCOPB

If the Instance has the title and description filled out then it is working.


mint-thompson (changes requested):
There is one other place where `_instanceMeta.title` and `_instanceMeta.description` are checked: in `IGExporter.addResources`, these values can be used when populating the resource list. So, that function should be updated to look for `title` and `description` directly on the `InstanceDefinition`.

src/export/InstanceExporter.ts
```diff
@@ -432,14 +432,13 @@ export class InstanceExporter implements Fishable {
     }
 
     const instanceOfStructureDefinition = StructureDefinition.fromJSON(json);
-
     let instanceDef = new InstanceDefinition();
     instanceDef._instanceMeta.name = fshDefinition.id; // This is name of the instance in the FSH
     if (fshDefinition.title) {
-      instanceDef._instanceMeta.title = fshDefinition.title;
+      instanceDef.title = fshDefinition.title;
```
cmoesel, line 438: I have a few concerns about this implementation:
* The title should only be assigned on instances that have usage `definition`.
* The title should only be assigned on conformance instances that support a `title` element.  For example, the `Patient` resource does not have a title element, but this code would set one anyway.
* Existing code still depends on `instanceDef._instanceMeta.title`, but that's been removed here -- so that existing code won't work right anymore.

test/export/InstanceExporter.test.ts
```diff
@@ -910,6 +910,29 @@ describe('InstanceExporter', () => {
...
+      const statusDraft = new AssignmentRule('status');
+      statusDraft.value = new FshCode('draft');
+      const nameDemo = new AssignmentRule('name');
+      nameDemo.value = new FshCode('DemoQuestionnaire');
+      goalInstance.rules.push(statusDraft, nameDemo);
+      const exported = exportInstance(goalInstance);
+      expect(exported.title).toMatch('My Demo Questionnaire');
+      expect(exported.description).toMatch("My Demo Questionnaire's description");
+    });
+
```
cmoesel, line 935: You should also have tests that check:
* That it does not populate `title` on an instance with usage `example`.
* That is does not populate `title` on an instance that doesn't have a `title` element (like `Patient`).

cmoesel (changes requested):
This is a bit too agressive in setting `title` and `description`.  See individual comments for details.

test/export/InstanceExporter.test.ts
guhanthuran: Done

src/export/InstanceExporter.ts
guhanthuran: Fixed.

cmoesel:
I will run regression on this to see how many changes it introduces, but I'm thinking we might want to merge this into `next` instead of `master` -- so it goes into the SUSHI 3.0 release.

cmoesel:
Actually, I'm going to wait to change the target until after Julia's rebase of next gets in.  Stand by.

test/export/InstanceExporter.test.ts
```diff
@@ -4998,6 +4998,75 @@ describe('InstanceExporter', () => {
         /CaretValueRule.*Instance.*File: Caret\.fsh.*Line: 1 - 3.*Applied in File: Insert\.fsh.*Applied on Line: 5 - 7/s
       );
     });
+
+    it('should populate title and description when specified for instances with #defintion', () => {
```
cmoesel, line 5002:
```suggestion
    it('should populate title and description when specified for instances with #definition', () => {
```

test/export/InstanceExporter.test.ts
```diff
@@ -4998,6 +4998,75 @@ describe('InstanceExporter', () => {
...
+      statusDraft.value = new FshCode('draft');
+      const nameDemo = new AssignmentRule('name');
+      nameDemo.value = new FshCode('DemoQuestionnaire');
+      goalInstance.rules.push(statusDraft, nameDemo);
+      const exported = exportInstance(goalInstance);
+      expect(exported.title).toMatch('My Demo Questionnaire');
+      expect(exported.description).toMatch("My Demo Questionnaire's description");
+    });
+
+    it("should not populate title and description when specified for instances that aren't #defintion", () => {
```
cmoesel, line 5025:
```suggestion
    it("should not populate title and description when specified for instances that aren't #definition", () => {
```

test/export/InstanceExporter.test.ts
```diff
@@ -4998,6 +4998,75 @@ describe('InstanceExporter', () => {
...
+      // Instance: DemoQuestionnaire
+      // InstanceOf: Patient
+      // Usage: #definition
+      // Title: "My Demo Questionnaire"
+      // Description: "My Demo Questionnaire's description"
+      //* name = "DemoQuestionnaire"
+      //* status = #draft
+      const goalInstance = new Instance('DemoQuestionnaire');
+      goalInstance.instanceOf = 'Patient';
+      goalInstance.usage = 'Example';
```
cmoesel, line 5058:
```suggestion
      goalInstance.usage = 'Definition';
```

cmoesel (changes requested):
I think the code still looks good, but the tests need a little tweaking.

PR #1108, opened 2022-06-22
Check complex type values when validating assignments

mint-thompson:
Completes task [CIMPL-969](https://standardhealthrecord.atlassian.net/browse/CIMPL-969).

When validating the a complex type value used in an AssignmentRule, the value may be an InstanceDefinition. Do not assign the value if it would conflict with existing values within the complex type. Assignment of complex values is all-or-nothing: if a conflict exists on part of the complex value, no parts of the complex value are assigned. An error is reported on the line where the Instance is assigned, and the error message refers to the specific value within the Instance that has a conflict.

The test for the "not the first type" is there as a guard against some prior logic involving `type[0].code` that caused problems for me during development. That logic has been removed, but I kept the test in just in case something similar comes up later.

There is some duplicated code for restoring values in `checkAssignInlineInstance`, but I felt that this implementation made more sense than adding new parameters for the existing key and value.

### DIFF
--- a/src/fhirtypes/ElementDefinition.ts
+++ b/src/fhirtypes/ElementDefinition.ts
@@ -1642,7 +1642,8 @@ export class ElementDefinition {
           stringVal,
           value.toJSON(),
           exactly,
-          value._instanceMeta.sdType ?? value.resourceType
+          value._instanceMeta.sdType ?? value.resourceType,
+          fisher
         );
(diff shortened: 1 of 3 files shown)


julianxcarter (approved):

jafeltra (approved):
Neato!

Issue #1106, opened 2022-06-17
FSHLogger ErrorsAndWarnings location type should be any


alexgwalley:
In FSHLogger.ts, the current class of ErrorsAndWarnings is this:
```
export class ErrorsAndWarnings {
  public errors: { message: string; location?: string; input?: string }[] = [];
  public warnings: { message: string; location?: string; input?: string }[] = [];
  public shouldTrack = false;

  reset(): void {
    this.errors = [];
    this.warnings = [];
    this.shouldTrack = false;
  }
}
```

however, in varying tests, the location is assumed to be other types. For example:

Here is is assumed to be a 4 element array:

![location is array](https://user-images.githubusercontent.com/10475866/174309383-b2186c18-e4d7-4c6b-8b0b-bfefcdf7448a.PNG)

Here is is assumed to be an object with 4 elements:

![location is object](https://user-images.githubusercontent.com/10475866/174309409-23fa6ecc-d8a3-4218-965b-d4ec68bc27a8.PNG)

Therefore, I believe that the class should be:

```
export class ErrorsAndWarnings {
  public errors: { message: string; location?: any; input?: string }[] = [];
  public warnings: { message: string; location?: any; input?: string }[] = [];
  public shouldTrack = false;

  reset(): void {
    this.errors = [];
    this.warnings = [];
    this.shouldTrack = false;
  }
}
```

where location type is any.

Would you like me to submit a PR for this?

cmoesel:
Well, this is embarrassing.  That said, I'm surprised that TS isn't complaining and that things still seem to generally work.  If you have a PR ready to go, feel free to submit it and we'll do a little more digging. I'd like to understand how/why we're getting away with this -- and if perhaps we ought to be more consistent about things (in which case, we'd want the `location` type to be more specific rather then `any`).  But a PR with `any` is a good start.

Issue #1105, opened 2022-06-16
Fsh-Sushi API not working with Webpack


alexgwalley:
### The Issue
I am trying to pack the sushi with webpack and then run, however I get this error:

> webpack 5.73.0 compiled successfully in 8115 ms
> node:fs:208
>   let mode = stats[1];
>                   ^
> 

> TypeError: Cannot read properties of undefined (reading '1')
>     at isFileType (node:fs:208:19)
>     at Object.readFileSync (node:fs:456:16)
>     at C:\Users\awalley\SushiTesting\dist\extension.js:117148:19
>     at Array.map (<anonymous>)
>     at C:\Users\awalley\SushiTesting\dist\extension.js:117147:30
>     at Object.<anonymous> (C:\Users\awalley\SushiTesting\dist\extension.js:117152:2)   
>     at __webpack_require__ (C:\Users\awalley\SushiTesting\dist\extension.js:153931:42) 
>     at Object.<anonymous> (C:\Users\awalley\SushiTesting\dist\extension.js:98848:16)   
>     at __webpack_require__ (C:\Users\awalley\SushiTesting\dist\extension.js:153931:42) 
>     at Object.<anonymous> (C:\Users\awalley\SushiTesting\dist\extension.js:96599:25)

As you can see, webpack compiles it correctly, but my guess is that is is having a difficult time requiring/importing everything.
When debugging, it breaks at this file read:

![image](https://user-images.githubusercontent.com/10475866/174090420-18f16067-1d2e-4bc8-b817-a3e421d4ee26.png)

Looking for the 808 file descriptor in the dist/extension.js file, it brings me here:

![image](https://user-images.githubusercontent.com/10475866/174091507-8d6e3689-aac5-4b30-9f03-35883f3215c6.png)

**UglifyJS** is used by the **html-minifer** dependency as seen in node_modules/fsh-sushi/node-shrinkwrap.json

### To Reproduce
clone the repo: https://github.com/alexgwalley/SushiTesting.git, run npm i, then run the npm script "run". This packs and trys to run everything. When you debug the extension.js, if you check both boxes "caught Exceptions" and "uncaught exceptions" it should break here.

I am not quite sure how to fix this error, but any help is appreciated! Thanks

alexgwalley:
[...]

PR #1113, opened 2022-06-30
Fix Minifier Dependency and Location Type

alexgwalley:
Change html-minifier dependency to html-minifier-terser so that it plays nicely with webpack.
Also change Location type to Any to avoid inconsistencies.

Fixes Issues: [https://github.com/FHIR/sushi/issues/1105](1105) and [https://github.com/FHIR/sushi/issues/1106](1106)


cmoesel:
Hi @alexgwalley -- I just wanted to let you know that we'll be reviewing this in our current sprint (so some time in the next week or so).  Thanks for the contribution!

cmoesel:
Actually, it looks like CI is failing.  The shrinkwrap file seems like it was not merged correctly.  I also noticed that you added `html-minifier-terser` to the `devDependencies`, but it really out to replace `html-minifier` in the production `dependencies`, right?  Could you fix those things?

Also, is there a reason you went with 7.0.0-beta.0?  Or would the most recent stable version (6.1.0) work OK?  I think we'd prefer a stable release rather than the beta, unless the beta is necessary to fix the issue you're running into.

npm-shrinkwrap.json
```diff
@@ -3237,13 +14377,51 @@
...
+          "integrity": "sha512-fgAN3jGAh+RoxUGZHTSOLJIqUc2wmoBwGR4tbpNAKmmovFoWq0OdRkb0VkldReO2a2iBT/OEulG9XSUc10r3zg==",
+          "dev": true,
+          "requires": {
+            "lower-case": "^2.0.2",
+            "tslib": "^2.0.3"
+          }
+        }
+<<<<<<< HEAD
+=======
+=======
```
cmoesel, line 14415: Here's some of the evidence of a merge gone wrong...  It might be easier to replace the `npm-shrinkwrap.json` and `package.json` in this branch with the most current ones from master, then re-do the dependency changes from scratch.  That shouldn't cause any conflicts.

package.json
```diff
@@ -71,6 +71,7 @@
     "eslint": "^8.5.0",
     "eslint-config-prettier": "^6.10.1",
     "extract-zip": "^2.0.1",
+    "html-minifier-terser": "^7.0.0-beta.0",
```
cmoesel, line 74: This should be in `dependencies`, not `devDependencies` -- and the old `html-minifier` module needs to be removed from the dependencies as well.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1924,7 +1924,7 @@ export class ElementDefinition {
    * @throws {ValueAlreadyAssignedError} when the value is already assigned to a different value
    * @throws {TypeNotFoundError} when the value does not match the type of the ElementDefinition
    */
-  private assignString(value: string, exactly = false): void {
+  private async assignString(value: string, exactly = false): Promise<void> {
```
cmoesel, line 1927: Does this really work?  Wouldn't you need to change the code that calls `assignString` so that it awaits the response now?  And then that would make that code `async`, which means anything that calls it needs to `await` too... and on and on.  In other words, I'm not sure that we can easily change this to `async` without causing a cascade of other changes.

According to [this issue](https://github.com/terser/html-minifier-terser/issues/78), the `minify` function was changed to async in v6.x.  Would using the most recent 5.x version (5.1.1) work for you?  Then `minify` could stay synchronous.

cmoesel (changes requested):
Noting changes in their respective places (already noted generally in a previous comment).  Also just noticed the change to async, which might be a problem...

src/fhirtypes/ElementDefinition.ts
alexgwalley: I moved html-minifier-terser to dependencies and downgraded to the synchronous version to avoid the overhaul issues you had mentioned.  The FSH conversion still works from my end of things.
Let me know if you need any other updates. Thanks!

src/utils/FSHLogger.ts
```diff
@@ -147,8 +147,8 @@ class LoggerStats {
 export const stats = new LoggerStats();
 
 export class ErrorsAndWarnings {
-  public errors: { message: string; location?: string; input?: string }[] = [];
-  public warnings: { message: string; location?: string; input?: string }[] = [];
+  public errors: { message: string; location?: any; input?: string }[] = [];
+  public warnings: { message: string; location?: any; input?: string }[] = [];
```
jafeltra, line 151: I looked into this type for `location` a bit because it seemed odd to me that the type would be so off based on the examples you pointed out in issue #1106. It looks like in practice, `location` will always be of the type `TextLocation`, which is defined [here](https://github.com/FHIR/sushi/blob/master/src/fshtypes/FshEntity.ts#L53-L58). This actually matches the type that is checked in the second test you called out in the issue. Instead of changing this type to `any`, I would suggest changing it to `TextLocation` here. Then I would also suggest updating the first test you called out in the issue to update the `location` property that is used and checked there. Instead of using

```
location: [1, 2, 3, 4]
```

you can transform that array to a text location like 
```
location: {
  startLine: 1,
  startColumm: 2,
  endLine: 3,
  endColumn: 4
}
```
If you replace the `location` in the test ([here](https://github.com/FHIR/sushi/blob/master/test/utils/FSHLogger.test.ts#L86-L103)) where it is used in the `logger.warn` call and in the expect statement, I think the test will be represent what happens in the code.

cmoesel:
Hey @alexgwalley -- just checking in.  Let us know if you have any trouble or disagreements regarding @jafeltra's suggestions.

alexgwalley:
Hey @cmoesel, I was out of the office for a few days but am looking into it today and will do.

npm-shrinkwrap.json
```diff
@@ -1,8 +1,11129 @@
 {
   "name": "fsh-sushi",
   "version": "2.6.0",
-  "lockfileVersion": 1,
```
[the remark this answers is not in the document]
alexgwalley: What version of Node.js and NPM do you use? I am unable to use the latest Node.js with npm v 6.0.0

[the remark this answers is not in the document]
jafeltra: I'm currently using Node `v14.18.0` and npm `v6.14.15`.

alexgwalley:
I used npm v 6.0.0 and ran npm i  to build the lock file, I also did the changes @jafeltra suggested with TextLocation, hopefully this works. Let me know if it doesn't or if I need to uses a different version

jafeltra (changes requested):
The updates look good to me! In addition to the two small things below, I noticed one more thing with the `npm-shrinkwrap.json` file. The `lockfileVersion` looks correct to me now, but I now get differences in the `npm-shrinkwrap.json` file when I run `npm i`. Would you mind trying to run `npm i` one more time to see if you get any changes? If so, can you add those as well?

jafeltra:
Hey Alex, I can definitely help out with the lockfile issue! Thanks for making the prettier changes! I'll take another look at this now.

jafeltra (approved):
This looks great! Thanks @alexgwalley for making all those updates. I pushed up the final changes I had to `npm-shrinkwrap.json` and all the other updates you made look good to me. We will just need @cmoesel to do one last review as well, and then this should be good to merge.

Once it it merged, it will be included in our next SUSHI release, which usually happens every couple weeks, but we can discuss as a team when our next release is to ensure you can access your change via npm.

cmoesel (approved):
Yes, this looks good!  Thank you, @alexgwalley!

PR #1114, opened 2022-06-30
Do not add an element that is defined by an ancestor

guhanthuran:
Fixes CIMPL-961.

When defining a Resource or Logical, it should not be possible to add an element with the same path as an inherited element. The following is an example of this:

```
Resource: MyResource
* extension 0..1 string "This should not be allowed"
```

This issue addresses this by adding a new error type: `ElementAlreadyDefinedError`. This error is thrown in StructureDefinition where we check if there already exists an element name that is defined by an ancestor.

To test: replicate the above example, if the error `ElementAlreadyDefinedError` is not thrown, then something is breaking.


src/errors/ElementAlreadyDefinedError.ts
```diff
@@ -0,0 +1,5 @@
+export class ElementAreadyDefinedError extends Error {
```
jafeltra, line 1: The name of this class has a typo which will affect places where it is imported as well.

```suggestion
export class ElementAlreadyDefinedError extends Error {
```

src/fhirtypes/StructureDefinition.ts
```diff
@@ -12,7 +12,8 @@ import {
   MissingSnapshotError,
   InvalidResourceTypeError,
   InvalidTypeAccessError,
-  ValidationError
+  ValidationError,
+  ElementAreadyDefinedError
```
jafeltra, line 16: This is one of the places that should be updated once the typo is fixed.

src/fhirtypes/StructureDefinition.ts
```diff
@@ -366,6 +367,10 @@ export class StructureDefinition {
    * @returns {ElementDefinition} the new ElementDefinition
    */
   newElement(name = '$UNKNOWN'): ElementDefinition {
+    //Check if there already exists an element that is defined by an ancestor
+    if (this.elements.find(e => e.id == `${this.id}.${name}`)) {
+      throw new ElementAreadyDefinedError(name);
```
jafeltra, line 372: When I saw this messaged logged, I thought it would be helpful to include the definition that it is a part of. If we did that, the error message could look something like this:

```
Cannot define element extension on MyResource because it has already been defined
```

After looking around at a few other errors thrown in this file, I can't find many that actually include the definition name, so I could be wrong, but when I read the error message without looking at other errors, I thought this would be helpful.

src/errors/ElementAlreadyDefinedError.ts
```diff
@@ -0,0 +1,5 @@
...
+  constructor(public elementID: string) {
```
jafeltra, line 2: This is a bit of a personal preference, but I think it's generally the pattern in the code - I would suggest making this variable name camel case and just have it as `elementId`.

test/export/StructureDefinition.ResourceExporter.test.ts
```diff
@@ -13,6 +13,7 @@ import {
   ContainsRule,
   FlagRule
 } from '../../src/fshtypes/rules';
+import { logger } from '../../src/utils';
```
jafeltra, line 16: It looks like this variable was added but never used. Can you remove it?

test/export/StructureDefinition.ResourceExporter.test.ts
```diff
@@ -269,12 +270,21 @@ describe('ResourceExporter', () => {
         /FHIR prohibits logical models and resources from constraining parent elements. Skipping.*at path 'language'.*File: ConstrainParent\.fsh.*Line:\D*/s
       );
     });
+  });
 
-    expect(exported.name).toBe('MyTestResource');
-    expect(exported.id).toBe('MyResource');
-    expect(exported.type).toBe('MyResource');
-    expect(exported.baseDefinition).toBe('http://hl7.org/fhir/StructureDefinition/DomainResource');
-    expect(exported.elements).toHaveLength(12); // 9 AlternateIdentification elements + 3 added elements
```
jafeltra, line 277: Is there a reason these 5 expect statements were removed? If there wasn't, can they be added back in? I think we want them to ensure the resource is being exported correctly.

test/export/StructureDefinition.ResourceExporter.test.ts
```diff
@@ -269,12 +270,21 @@ describe('ResourceExporter', () => {
...
+  it('should log an error when adding an element with the same path as an inherited element', () => {
+    const resource = new Resource('MyResource');
+    const addElementRule = new AddElementRule('extension');
+    addElementRule.min = 0;
+    addElementRule.max = '1';
+    resource.rules.push(addElementRule);
+    doc.resources.set(resource.name, resource);
+    exporter.export();
```
jafeltra, line 282: I think it would be good to also include a check within this test that a new `extension` element is not added. I think you can do this by grabbing the exported resource (there is an example of that in the test above this one) and checking the elements array. Without the error you added in, there would be two elements with the id `MyResource.extension`. You could check the total length of the `elements` array and check that there is only one element with that path. There may be other/additional ways to check that, but those are the two ideas I thought of.

jafeltra (changes requested):
This is looking good. I had a few things to change inline, as well as one more below that didn't have a good inline place:

- I think it would be good to include a test that the new error is thrown from `newElement`. This would be in addition to the test you added to check that the message is logged properly. It looks like there is a describe block for `newElement` in the StructureDefinition.test.ts file [here](https://github.com/FHIR/sushi/blob/master/test/fhirtypes/StructureDefinition.test.ts#L738). I think adding a test there would be good. Again, this is the first place I found that seemed appropriate, but if you find a better/additional place for this test, that works too.

src/errors/ElementAlreadyDefinedError.ts
cmoesel: I think it probably depends on which code you're looking at. ;-)  I'm pretty sure I'm guilty of doing it like `lowerPrefixUpperID` elsewhere myself.  Oops.  But yeah, do what @jafeltra says. ;-)

mint-thompson (changes requested):
The update in `StructureDefinition.newElement` looks correct to me. I think it could be useful to add some more test cases:
* a test where two rules add a new element with the same name
* a test that directly calls `StructureDefinition.newElement` and checks that it throws an `ElementAlreadyDefinedError`

test/export/StructureDefinition.ResourceExporter.test.ts
jafeltra: @guhanthuran can you add these lines back in? They should still be included because your PR does not change the test above the one you added. They should be added back in after line 270.

jafeltra:
@guhanthuran I re-reviewed this and agree with Mint's suggested test cases as well. The change to the test from this morning looks good to me! In addition to what Mint mentioned, there is still one outstanding comment from my initial review that should be fixed as well.

cmoesel:
@guhanthuran -- Can you add those tests so we can wrap this PR up? It's _so close_!

test/export/StructureDefinition.ResourceExporter.test.ts
```diff
@@ -13,7 +13,7 @@ import {
   ContainsRule,
   FlagRule
 } from '../../src/fshtypes/rules';
-
+import { StructureDefinition } from '../../src/fhirtypes';
```
jafeltra, line 16: Can you add a space between the imports and the tests?
```suggestion
import { StructureDefinition } from '../../src/fhirtypes';

```

test/export/StructureDefinition.ResourceExporter.test.ts
jafeltra: These lines are still not added back in. This test should still be testing those expect statements. The warning above this line in the diff view of the PR shows that the `exported` value isn't used because these expect statements aren't included in the test. I'm happy to help today if you want.

src/fhirtypes/StructureDefinition.ts
```diff
@@ -366,6 +367,10 @@ export class StructureDefinition {
    * @returns {ElementDefinition} the new ElementDefinition
    */
   newElement(name = '$UNKNOWN'): ElementDefinition {
+    //Check if there already exists an element that is defined by an ancestor
```
jafeltra, line 370: Sorry for the super nitpicky comment, but can you add a space after the `//`? I thought we had a lint rule for that, but seems like maybe not.
```suggestion
    // Check if there already exists an element that is defined by an ancestor
```

test/export/StructureDefinition.ResourceExporter.test.ts
```diff
@@ -269,12 +269,86 @@ describe('ResourceExporter', () => {
...
+      const addElementRule2 = new AddElementRule('elem1');
+      addElementRule2.min = 0;
+      addElementRule2.max = '1';
+      alternateIdentification.newElement(addElementRule.path);
+      alternateIdentification.newElement(addElementRule2.path);
+    } catch (e) {
+      expect(e.message).toMatch(
+        'Cannot define element elem1 on AlternateIdentification because it has already been defined'
+      );
+    }
```
jafeltra, line 351: I think there is a little cleaner of a way to test that the error is thrown. This is following another pattern we have used in other tests, so it might make sense to use this approach for consistence.

Instead of wrapping everything in a try/catch statement, you can wrap the second `newElement` in a function and use a `expect(...).toThrow(...)` instead of the try/catch. So in this case, you can do this:

```
expect(() => {
  alternateIdentification.newElement(addElementRule2.path);
}).toThrow(
  'Cannot define element elem1 on AlternateIdentification because it has already been defined'
);
```

Once you do that, you don't need the try/catch block anymore.

There are various examples of this in the code, but you can see one of them [here](https://github.com/FHIR/sushi/blob/master/test/fhirtypes/ElementDefinition.assignBoolean.test.ts#L59-L77).

test/export/StructureDefinition.ResourceExporter.test.ts
```diff
@@ -269,12 +269,86 @@ describe('ResourceExporter', () => {
...
+    );
+  });
+
+  it('should log an error when a rule with the same path is added by directly calling newElement', () => {
+    doc = new FSHDocument('fileName');
+    const input = new FSHTank([doc], minimalConfig);
+    pkg = new Package(input.config);
+    const fisher = new TestFisher(input, defs, pkg);
+    const alternateIdentification: StructureDefinition =
+      fisher.fishForStructureDefinition('AlternateIdentification');
```
jafeltra, line 336: I'm seeing a type error on this line. Since the `fishForStructureDefinition` returns `StructureDefinition | undefined`, it doesn't like that `alternateIdentification` has a type of `StructureDefinition`. In this case, we know that `AlternateIdentification` can be fished, so you can add `as StructureDefinition` to say you're sure it will return that type, rather than `undefined`. 


```suggestion
    const alternateIdentification: StructureDefinition =
      fisher.fishForStructureDefinition('AlternateIdentification') as StructureDefinition;
```

jafeltra (changes requested):
Thanks for adding the new tests. They're looking good. I left a few comments inline to update.

test/export/StructureDefinition.ResourceExporter.test.ts
guhanthuran: Done!

test/export/StructureDefinition.ResourceExporter.test.ts
guhanthuran: Done, thanks!

guhanthuran:
@jafeltra Added back the lines in the correct place.

test/export/StructureDefinition.ResourceExporter.test.ts
```diff
@@ -262,19 +263,98 @@ describe('ResourceExporter', () => {
 
     const exported = exporter.export().resources[0];
 
+    expect(exported.name).toBe('MyTestResource');
+    expect(exported.id).toBe('MyResource');
+    expect(exported.type).toBe('MyResource');
+    expect(exported.baseDefinition).toBe('http://hl7.org/fhir/StructureDefinition/DomainResource');
+    expect(exported.elements).toHaveLength(12);
+
```
jafeltra, line 271: I'm sorry to keep doing this, but I'd really like these lines to show up without any differences. To do that, you need to change two things:
- The comment is missing on the last expect statement. We need to keep this because it provides context about why we are expecting to have a length of `12`. It's a pretty arbitrary looking number, so we don't want to lose that context.
- These lines should get added _underneath_ the expect statement checking each log statement. Looking at the current branch as it is now, the lines should get added after line 272. This change is technically not required since it doesn't make any difference in the order of the expect statements, but I think it's good to avoid making unnecessary changes in the tests. Before pushing this change, if you want to check that the lines are actually the same as they are on master, you can run `git diff master test/export/StructureDefinition.ResourceExporter.test.ts`. When you run that, the only differences you should see are the additions of the import and the three tests you added. You should no longer see that any lines were removed.

guhanthuran: Done!

jafeltra (approved):
This looks good to me! Thank you for making the updates!

This just needs a re-review from @mint-thompson and then it will be good to merge.

mint-thompson (approved):
Thank you for adding those tests. Approved!

PR #1115, opened 2022-07-05
Fixes SUSHI adding instances of custom resources in the IG + outputs warning

### LINKED ISSUE #1090: SUSHI should not include instances of custom resources in the IG
The IG Publisher does not support custom resources.  SUSHI already suppresses custom resources from the generated IG JSON.  If an author creates an example of a custom resource, however, SUSHI _does_ still include that instance in the IG JSON.  This then causes problems in the IG Publisher (since it's referencing a custom resource).  SUSHI should probably suppress instances of custom resources too.

See: https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/bad.20resource.20identity/near/282797466

Example FSH from issue reported on Zulip:
```
Resource: FormTemplate
Parent: Resource
Id: formtemplate
Title: "FHIR2 Form Template"
Description: "FHIR2 Form Template"
* template 1..1 BackboneElement "The root of the doc"
  * templateStatus 1..1 string "the template status"


Instance: FHIR2FormTemplateExample
InstanceOf: FormTemplate
Title: "Example : Questionnaire"
Description: "Example : Questionnaire"
Usage: #example
* template.templateStatus = "draft"
```

guhanthuran:
Fixes #1090 

The IG Publisher does not support instances of custom resources. This PR fixes that issue and issues a warning. 

To test:
Test with this FSH code and see if the FormTemplate Instance displays under `resources` in the IG. If it does, then the code is not working.

```
Resource: FormTemplate
Parent: Resource
Id: formtemplate
Title: "FHIR2 Form Template"
Description: "FHIR2 Form Template"
* template 1..1 BackboneElement "The root of the doc"
  * templateStatus 1..1 string "the template status"

Instance: FHIR2FormTemplateExample
InstanceOf: FormTemplate
Title: "Example : Questionnaire"
Description: "Example : Questionnaire"
Usage: #example
* template.templateStatus = "draft"
```
Also try to test if a normal Instance (like a Patient Instance) still displays under `resources` in IG.

Thanks for reviewing! 💯


src/export/InstanceExporter.ts
```diff
@@ -26,6 +27,37 @@ export class InstanceExporter implements Fishable {
...
+
+      // prettier-ignore
+      const message = [
+          clr('\n╭─────────────────────────────────────────────────────────────────') + '' + clr('╮'),
+            clr('│') + ' Detected the following instance(s) of custom resources:         ' + clr('│'),
+            ...instancesOfCustomResourceLogs,
+            clr('│') + '                                                                 ' + clr('│'),
+            clr('│') + ' The IG Publisher does not support custom resources, these       ' + clr('│'),
+            clr('│') + ' instances will not be included in the IG.                       ' + clr('│'),
+            clr('╰─────────────────────────────────────────────────────────────────') + '' + clr('╯\n')
```
jafeltra, line 55: I like the fancy box!

One thing I'm not sure about is if we want to align the language we use here with the warning we give when we don't add custom resources. In that box, we refer to the resources we filter out as "non-conformant Resource definitions" and we just say they are not conformant to the FHIR specification. This makes me wonder if we should avoid referring to these as "custom resources" here and avoid referring directly to the IG publisher.

Before we workshop the language too much, this might be something we want to double check with @cmoesel on.

src/ig/IGExporter.ts
```diff
@@ -782,6 +782,11 @@ export class IGExporter {
     );
     instances
       .filter(instance => instance._instanceMeta.usage !== 'Inline')
+      .filter(
+        instance =>
+          // Filter out custom defintions that contain a type that has the same type as our instance type
```
jafeltra, line 787: I might just be getting confused, but this comment sounds like it might be describing something different than the code does. This is filtering out any instance that has a type that is one of the custom resources defined in the package, right? Maybe the comment could be something like 
```suggestion
          // Filter out instances that have a type that has the same type as a custom resource defined in the package
```

test/ig/IGExporter.IG.test.ts
```diff
@@ -1683,11 +1707,6 @@ describe('IGExporter', () => {
     });
 
     it('should not warn on deeply nested resources when implicated by the path-resource parameter', () => {
-      defs = new FHIRDefinitions();
-      loadFromPath(path.join(__dirname, '..', 'testhelpers', 'testdefs'), 'r4-definitions', defs);
-      fixtures = path.join(__dirname, 'fixtures', 'customized-ig-with-nested-resources');
-      loadCustomResources(path.join(fixtures, 'input'), fixtures, config.parameters, defs);
-
```
jafeltra, line 1690: Is there a reason these lines were deleted? I think we still want them to set this test up correctly.

jafeltra (changes requested):
This looks pretty good. I had a few comments below, including one we might want Chris's opinion on.

src/export/InstanceExporter.ts
julianxcarter: Another suggestion for the language here, I think we should explicitly say that the instances will not be included in the IG _resource_, rather that the IG itself. I think the warning as it currently exist implies that SUSHI doesn't include the instance in the `fsh-generated` folder at all, which it does.

test/export/InstanceExporter.test.ts
```diff
@@ -4779,6 +4780,41 @@ describe('InstanceExporter', () => {
...
+      sdExporter.export();
+      const exported = exporter.export().instances;
+      expect(exported.length).toBe(1);
+      expect(loggerSpy.getAllMessages('warn').length).toBe(2);
+      expect(loggerSpy.getLastMessage('warn')).toMatch(
+        /following instance\(s\) of custom resources:.*- FooInstance/s
+      );
+    });
+
+    it('should log a warning when exporting a multiple instances of custom resources', () => {
```
julianxcarter, line 4799:
```suggestion
    it('should log a warning when exporting multiple instances of custom resources', () => {
```
Small typo!

test/ig/IGExporter.IG.test.ts
guhanthuran: No, this was redundant code (it copied from setup that was established previously).

src/export/InstanceExporter.ts
guhanthuran: Done.

test/ig/IGExporter.IG.test.ts
jafeltra: Hm it looks like the last line is actually different and the `config.parameters` is passed in to `loadCustomResources`, and based on the test description, it looks like that might be what this is trying to test. But if that's the case, I'm not really sure why the test is still passing. Maybe @julianxcarter knows since it looks like he added this test. But if not, it's probably fine to keep these lines removed.

jafeltra: After looking at this test a little more, I do think we want to keep the lines that were removed. We want to be testing that passing in the config.parameters does properly add the nested resources without providing a warning. The problem is that the test here isn't exactly working as expected. We will need to include these four lines again, but we also need to redefine the `exporter` so that it has the `defs` that are loaded with the new parameter in `loadCustomResources`. We also probably only want `config.parameter` to be set in this test, so we probably will want to move where that happens from in the `beforeAll` into this test. That means we'd also need to redefine `pkg` to create a package with the appropriate config.

We can talk about if these are changes we want to make as part of this PR or a separate task when you're back!

guhanthuran: Sound good, undeleted the changes for now. Took a look at it again and since we want `config.parameter` set it makes sense that it's not dine in the `beforeAll`.

julianxcarter (approved):

src/ig/IGExporter.ts
jafeltra: I'm not sure where/how this happened, but I don't see this change anymore? Should it have been included?

jafeltra (approved):

julianxcarter (approved):

PR #1116, opened 2022-07-05
Don't slice single-type choices (unless already sliced)

### LINKED ISSUE #1088: Avoid type slicing when a choice is constrained to a single type
**For SUSHI 3.0**: do not create a type slice for elements when the choice has already been restricted to a single type.  For example, consider this FSH:
```
Profile: MyObservation
Parent: Observation
* value[x] only CodeableConcept
* valueCodeableConcept from http://example.org/foo
```

*Currently*, whenever SUSHI sees a type-specific path (like `valueCodeableConcept`), it uses the standard type slicing mechanism to represent it.  It does this even after `value[x]` has been constrained to a single type (as above).  So you end up w/ a differential something like this:
```json
"differential": {
    "element": [
      {
        "id": "Observation.value[x]",
        "path": "Observation.value[x]",
        "slicing": {
          "discriminator": [
            {
              "type": "type",
              "path": "$this"
            }
          ],
          "ordered": false,
          "rules": "open"
        },
        "type": [
          {
            "code": "CodeableConcept"
          }
        ]
      },
      {
        "id": "Observation.value[x]:CodeableConcept",
        "path": "Observation.value[x]",
        "sliceName": "CodeableConcept",
        "min": 0,
        "max": "1",
        "binding": {
          "strength": "required",
          "valueSet": "http://example.org/foo"
        }
      }
    ]
  }
```

Technically, this type slicing is totally unnecessary since `value[x]` can only be a single type -- therefore you don't need slices to disambiguate.  Since slicing adds complexity, we like to avoid it when we can.  To that end, SUSHI should detect when this happens (type slice on a choice that has already been constrained to a single type) and apply any constraints directly to the choice element instead.  This would effectively be the same as:
```
Profile: MyObservation
Parent: Observation
* value[x] only CodeableConcept
* value[x] from http://example.org/foo
```

Resulting in this much simpler differential:
```json
"differential": {
    "element": [
      {
        "id": "Observation.value[x]",
        "path": "Observation.value[x]",
        "type": [
          {
            "code": "CodeableConcept"
          }
        ],
        "binding": {
          "strength": "required",
          "valueSet": "http://example.org/foo"
        }
      }
    ]
  }
```

This was proposed on Zulip and received positive feedback.  See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Type.20Slices.20on.20Choices.20w.2F.20a.20Single.20Type/near/282241129

cmoesel:
If a user refers to a specific choice type (e.g., valueQuantity), but the choice (e.g., value[x]) only allows one type anyway, then don't create a type slicing and slice, because they're not necessary.

On the other hand, if the choice has already been sliced, then go ahead and create the slice for consistency.

To test, you can use the example in #1088:
```
Profile: MyObservation
Parent: Observation
* value[x] only CodeableConcept
* valueCodeableConcept from http://example.org/foo
```

On `next`, it will slice `value[x]` and create a `valueCodeableConcept` slice.  On this branch, it will apply the binding directly to `value[x]` instead.

If you're curious, I'm attaching the regression reports here (I filtered it to just html and log files): 
[regression-output.zip](https://github.com/FHIR/sushi/files/9049900/regression-output.zip)

Fixes #1088

### DIFF
--- a/src/export/InstanceExporter.ts
+++ b/src/export/InstanceExporter.ts
@@ -16,7 +16,7 @@ import {
 import { InstanceOfNotDefinedError } from '../errors/InstanceOfNotDefinedError';
 import { InstanceOfLogicalProfileError } from '../errors/InstanceOfLogicalProfileError';
 import { Package } from '.';
-import { cloneDeep, merge, uniq } from 'lodash';
+import { cloneDeep, merge, uniq, upperFirst } from 'lodash';
 import { AssignmentRule } from '../fshtypes/rules';
 
 export class InstanceExporter implements Fishable {
@@ -276,6 +276,17 @@ export class InstanceExporter implements Fishable {
                 `Element ${child.id} has its cardinality satisfied by a rule that does not include the slice name. Use slice names in rule paths when possible.`
               );
               child = choiceSlice;
+              break;
+            }
+          }
+        }
+        // If we still haven't found it, it's possible that a type slice just wasn't created. In that case, there would
+        // be a type in the choice element's type array that would be a match if it were type-sliced.
+        if (instanceChild == null) {
+          for (const type of child.type) {
+            const name = childPathEnd.replace(/\[x\]$/, upperFirst(type.code));
+            instanceChild = instance[`_${name}`] ?? instance[name];
+            if (instanceChild != null) {
               break;
             }
           }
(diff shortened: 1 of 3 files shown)


julianxcarter (approved):

jafeltra (approved):
The changes and tests look good to me. I confirmed the example in the issue works as described and the diffs in the regression look good to me too!

PR #1118, opened 2022-07-13
Order IG resources based on config

### LINKED ISSUE #1080: Revisit approach to ordering resources in IG JSON
See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/order.20of.20resources.20in.20the.20generated.20ig.2Ejson/near/261472343

mint-thompson:
Fixes #1080 and completes task [CIMPL-966](https://standardhealthrecord.atlassian.net/browse/CIMPL-966).

If all generated resources are present in the config resources section, use the order from that section when adding resources to the IG. Otherwise, if all generated resources are present in the config groups section, use the order from that section when adding resources to the IG. If neither config section contains all resources, add the resources sorted by title or id.


cmoesel (changes requested):
I think this largely solves what we were looking to solve, so thank you! I found a few spots where I think we should consider improving re-use of code (and/or reducing nearly duplicate code). Those are noted in individual comments.

I also wondered about one potential improvement.  I'm not positive it would be better, but... what if we just processed the resources as usual in whatever order we got them.  But then, afterword, applied a `sort(...)`, passing in the most relevant sort function (based on if all resources are in config or in groups).  Then instead of having three separate iteration loops for processing resources, there is only one, plus a sort.  I think the sort functions would be pretty straightforward since the config-based sorts would basically just be comparing based on the original indices/groups.  What do you think?

src/ig/IGExporter.ts
```diff
@@ -983,6 +973,7 @@ export class IGExporter {
                   name ??
                   resourceJSON.id;
               }
+              newResource._sortKey = resourceJSON.id;
```
cmoesel, line 976: It's possible that predefined resources are conformance resources. If so, they have a title and maybe a name, and I'd guess that's what the IG Publisher uses when creating the Table of Contents and Artifacts pages...  So, I wonder if this would be better as:
```ts
newResource._sortKey = resourceJSON.title ?? typeof resourceJSON.name === 'string' ? resourceJSON.name : resourceJSON.id;
```
That would more closely match what we do for profiles, extensions, etc.  What do you think?

cmoesel: Actually, I just had a crazy thought.  All along, I've been assuming that the IG Publisher uses data from the resource (e.g., title, name, id) to determine what to display in the Table of Contents and Artifacts pages.  BUT... what if it is actually using `ImplementationGuide.definition.resource.name`??? I mean, now that I think about it, that might be the whole reason that you have to put a name in the `ImplementationGuide.definition.resource` in the first place!  If so, then (a) I feel dumb, (b) I'm sorry, and (c) that means you can just sort `ImplementationGuide.definition.resource` by `ImplementationGuide.definition.resource.name` and get rid of the `_sortKey`!  I think it's worth investigating.

cmoesel (commented):
Thanks, @mint-thompson. I really like this implementation.  Very clean and very little duplicated code!  You've also used the lodash sort methods quite effectively!

I ran out of time before I could test this... but I had one thought and one possible `DOH!` moment (on my part).  See my individual comments.

julianxcarter (approved):

cmoesel (changes requested):
Nice work, @mint-thompson.  I've tested this out and noticed a couple of things:

1. Our sort is case-sensitive, so all capital letters go first, and then all lowercase letters.  I think it might be better to do case-insensitive sort, because this just looks a little off to me: 
image

2. The IG Publisher still seems to do it's own initial sort by type in the TOC, but _then_ honors our sort order.  I don't think there's anything we can do about this, but it is a little unfortunate. For example, this is the mCODE Table of Contents:
image

mint-thompson:
The latest update changes the default sort ordering to be case-insensitive, still based on the name. I noticed some issues where `sha1` values were being added to `npm-shrinkwrap.json`, so I updated `npm` and reinstalled all the dependencies. This caused some other changes in the shrinkwrap file, but hopefully nothing that causes any problems.

julianxcarter (approved):

cmoesel (approved):
Fantastic! I tested this on mCODE for natural ordering.  I also tested manual ordering using the `resources` config property and `groups` config property (w/ a smaller IG).  It's all working well!  Thanks @mint-thompson!

PR #1120, opened 2022-07-19
Allow cardinality constraints to be applied to connected elements

mint-thompson:
Completes task [CIMPL-986](https://standardhealthrecord.atlassian.net/browse/CIMPL-986).

When constraining cardinality on an element, if there are connected elements that are the child of a slice, it may be possible to apply the cardinality constraint to those elements. As long as no cardinality is widened, applying the constraints is allowed. This may mean that only the new min or max are applied. If the new cardinality has no overlap with the connected element's existing cardinality, the cardinality constraint can't be applied.

### DIFF
--- a/test/export/StructureDefinitionExporter.test.ts
+++ b/test/export/StructureDefinitionExporter.test.ts
@@ -6743,7 +6743,7 @@ describe('StructureDefinitionExporter R4', () => {
       expect(loggerSpy.getAllMessages('error')).toHaveLength(0);
     });
 
-    it('should not apply a CardRule that would make the cardinality of the child of a slice too small', () => {
+    it.skip('should not apply a CardRule that would make the cardinality of the child of a slice too small', () => {
       loggerSpy.reset();
       // * component[Lab].interpretation 0..4
       // * component.interpretation 1..5 // this rule is invalid!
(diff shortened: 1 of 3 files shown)


cmoesel (changes requested):
I think the general concept and approach is good, as it's friendlier than the spurious error we used to give and it's helpful in updating those cardinalities.  When I ran regression, however, some of the changes in one repo didn't make sense to me.

You can reproduce by running regression on `HL7/davinci-pas#master`.  You'll see a bunch of changes in `StructureDefinition-profile-claim-update.json` -- all of which constrain `Claim.supportingInfo:SomeSliceName.extension` to have max `1`.  Here is one example:
image

If you look at the FSH definition for that profile, however, you'll see that the only rules on `supportingInfo` are these:
```
* supportingInfo.extension contains InfoChanged named infoChanged 0..1
* supportingInfo.extension[infoChanged] ^short = "A code indicating how the piece of information has changed."
* supportingInfo.modifierExtension contains InfoCancelledFlag named infoCancelledFlag 0..1
* supportingInfo.modifierExtension[infoCancelledFlag] ^short = "Indicates that this piece of information is not to be used."
```
It seems like adding an extension with max `1` (for that specific extension only) is going through and updating the `max` of the non-specific extension element in every slice.  Which is not what it should be doing.

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -6743,27 +6743,142 @@ describe('StructureDefinitionExporter R4', () => {
...
+        'Cardinality on Observation.component.interpretation cannot be narrowed'
+      );
+      expect(loggerSpy.getLastMessage('error')).toMatch(/File: Incompatible\.fsh.*Line: 8\D*/s);
+    });
+
+    it('should not apply a CardRule that is incompatible with the existing cardinality on some of the children of slices', () => {
+      // * component contains Field 0..1
+      // * component[Lab].interpretation 1..5
+      // * component[Field].interpretation 0..2
+      // * component.interpretation 3..4
```
jafeltra, line 6847: How do you feel about this incredibly small fix? Sorry lol
```suggestion
      // * component.interpretation 6..10
```

jafeltra (dismissed):
This makes sense to me! I have one incredibly small change in a comment, so I'm approving anyway.

jafeltra:
Oh I also ran a full regression again, and there were no differences besides that one sliced element that has the `MS` flag properly applied (which was discussed in Slack).

cmoesel (dismissed):
This looks good and works well!  Thanks, @mint-thompson.  I'm approving this, but you probably should accept Julia's minor suggestion before we merge it.

cmoesel (approved):
Re-approval.  Good work!

PR #1124, opened 2022-08-01
Properly validates required elements on sliced elements

### LINKED ISSUE #1089: Incorrect validation of choice elements that are children of slices
Zulip thread: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/access.20slice.20by.20name.20validation.20error

When validating required elements on an Instance, the logic that gets the applicable choice element doesn't check to make sure that it is within the correct slice. For example, in this FSH:
```
Profile: MyCommunication
Parent: Communication
Id: my-communication
* payload ^slicing.discriminator.type = #exists
* payload ^slicing.discriminator.path = "content.url"
* payload ^slicing.rules = #closed
* payload contains
    inline 1..1 and
    link 0..*
* payload.content[x] only Attachment
* payload[inline].contentAttachment.url ..0
* payload[inline].contentAttachment.data 1..
* payload[inline].contentAttachment.contentType 1..
* payload[inline].contentAttachment.contentType = #text/plain
* payload[link].contentAttachment.url 1..1
* payload[link].contentAttachment.data ..0
* payload[link].contentAttachment.contentType 1..
* payload[link].contentAttachment.contentType = #image/png

Instance: communication-example1
InstanceOf: my-communication
Usage: #example
* status = #completed
* payload[inline].contentAttachment.contentType = #text/plain
* payload[inline].contentAttachment.data = "OR3sMFkRaXsOtHff85+zrL+DXU3s5nkfIenVwcokMUx1qh8="
* payload[link].contentAttachment.contentType = #image/png
* payload[link].contentAttachment.url = "Binary/1234"
```

When validating any element at path `Communication.payload.content[x]`, the two choice slices found are the two defined slices: id `Communication.payload:inline.content[x]:contentAttachment` and id `Communication.payload:link.content[x]:contentAttachment`. Since the first one will successfully resolve on the instance, it will always be used. Unfortunately, it is used for _any_ element with the path `Communication.payload.content[x]`, which is true for both elements with id `Communication.payload:inline.content[x]` and id `Communication.payload:link.content[x]`.

julianxcarter:
Fixes #1089, updating the logic that validates required elements on instances to try and use the correct slice during validation rather than always utilizing the first slice that is found on the instance. The FSH included on the linked issue is an adequate test case for this fix.


mint-thompson (changes requested):
This change works well with the provided case. But, it doesn't work when different choice types are available on different slices. The following example will generate errors:
```
Profile: MyCommunication
Parent: Communication
Id: my-communication
* payload ^slicing.discriminator.type = #exists
* payload ^slicing.discriminator.path = "content.url"
* payload ^slicing.rules = #closed
* payload contains
    inline 1..1 and
    link 0..* and
    extra 0..*
* payload[inline].content[x] only Attachment
* payload[inline].contentAttachment.url ..0
* payload[inline].contentAttachment.data 1..
* payload[inline].contentAttachment.contentType 1..
* payload[inline].contentAttachment.contentType = #text/plain
* payload[link].content[x] only Attachment
* payload[link].contentAttachment.url 1..1
* payload[link].contentAttachment.data ..0
* payload[link].contentAttachment.contentType 1..
* payload[link].contentAttachment.contentType = #image/png
* payload[extra].content[x] only string

Instance: communication-example1
InstanceOf: my-communication
Usage: #example
* status = #completed
* payload[inline].contentAttachment.contentType = #text/plain
* payload[inline].contentAttachment.data = "OR3sMFkRaXsOtHff85+zrL+DXU3s5nkfIenVwcokMUx1qh8="
* payload[link].contentAttachment.contentType = #image/png
* payload[link].contentAttachment.url = "Binary/1234"
* payload[extra].contentString = "Extra content"
```
Because both `contentAttachment` and `contentString` elements exist at the path `Communication.payload.content[x]`, the loop continues to run over all the `choiceSlices`. This means that the final value of `instanceChild` is based on the type of the last choice, which is `contentString`. Since that doesn't exist for the `inline` or `link` slices, and `content` is a required element of `payload`, an error is generated for each of those slices.

jafeltra (approved):
This makes sense to me. It looks like the test covers the additional case Mint had identified, and it runs as expected.

npm-shrinkwrap.json
```diff
@@ -7677,7 +7677,7 @@
     "relateurl": {
       "version": "0.2.7",
       "resolved": "https://registry.npmjs.org/relateurl/-/relateurl-0.2.7.tgz",
-      "integrity": "sha512-G08Dxvm4iDN3MLM0EsP62EDV9IuhXPR6blNz6Utcp7zyV3tr4HVNINt6MpaRWbxoOHT3Q7YN2P+jaHX8vUbgog=="
```
jafeltra, line 7680: Hm actually I just noticed that when I ran `npm i`, the changes you have here are reverted back to what is on `master`. @mint-thompson do you see this as well? If so, @julianxcarter can you try to run `npm i` again and see if you get changes?

jafeltra: After chatting in the lab, I learned that NPM changed the integrity checksum. Because your changes added back in a `sha1` integrity, you can update your npm to use the newer, `sha512` integrity, which should result in no diffs on this PR compared to master.

To do this, you can remove your `node_modules` folder, run `npm cache clear --force`, and then re-run `npm i`. That should update your npm and revert you back to the `sha512` integrity. I'm going based off of some discussions in the lab and [this stack overflow](https://stackoverflow.com/a/47834854).

jafeltra (changes requested):

npm-shrinkwrap.json
julianxcarter: Unfortunately this didn't work for me, but I did fix it by just copy and pasting the shrink-wrap file from `master` and ignoring the changes made here 😕.

jafeltra (approved):

npm-shrinkwrap.json
jafeltra: Ugh darn that's unfortunate. It seemed so promising 😕

src/export/InstanceExporter.ts
```diff
@@ -279,7 +279,11 @@ export class InstanceExporter implements Fishable {
           // against the correct choice slice
           if (
             instanceChild != null &&
-            (instance._sliceName ? choiceSlice.sliceName === instance._sliceName : true)
+            (instance._sliceName
+              ? parseFSHPath(choiceSlice.id).some(pathPart => {
+                  return pathPart.base.split(':')[1] === instance._sliceName;
```
mint-thompson, line 284: I think you can simplify this a little if you use `splitOnPathPeriods` instead of `parseFSHPath` here. Also, rather than checking the whole path with `some`, I think you always want to see the matching slice name on the second-to-last part of the path? The last part of the path is the choice element, so the part right before that should correspond to the element that `instance` currently represents. I'm not 100% sure about that, though. I tried to verify this by putting together a profile that had a choice element with multiple slice elements along its ancestry (for example, a path like `something[Blue].somethingElse[Red].value[x]`), but I haven't found a good starting resource type for such a profile yet.

mint-thompson (changes requested):
Thank you for the updates! I have one minor change to request, and a question about how the slice name is checked.

mint-thompson (approved):
Thank you!

jafeltra (approved):
Looks good to me!

PR #1127, opened 2022-08-10
3.0.0 Beta 1 Candidate: Update Node and Dev Tools

cmoesel:
This PR represents the proposed candidate for SUSHI 3.0.0-beta.1.  In addition to bumping the version number, I've also taken this opportunity to update the dev stack this is built on:

* Node 16 or 18 should now be used (with NPM 8.x).
  * Node 12/14 still work at runtime, but should not be used for development (and are not officially supported).
  * I kept the `@types/node` at 12.x for now so we don't actually use any Node 14/16 features.
* Typescript was bumped to 4.7.x.
* Jest, ES Lint, Prettier, etc, were also all bumped.
* The npm-shrinkwrap.json was upgraded to v2 and completely rebuilt.

I personally tested this on Node 12, 14, 16, and 18 -- and it seems to work for all of them.  But again, we're highly recommending 16 or 18 since Node 12 is about to go EOL and Node 14 is already in its maintenance phase.

To test this, you should upgrade your node to 16 or 18.  I recommend a version manager such as `nvm` for easily switching between versions.  Check to make sure you can build it, run the tests, etc.  Also test that you can successfully run SUSHI against FSH projects.

If we can approve and merge this PR on Wednesday, I will do an official SUSHI 3.0.0 Beta 1 release. Please also review the draft release notes here: https://github.com/FHIR/sushi/releases/tag/untagged-5a50947545b44dbd7267


julianxcarter (commented):
This looks good to me! I just have two comments: 

1. I think we should change the node version in the `.nvmrc` file to either version `16` or `18` now that our recommended version has changed. We can also get rid of the file if we don't think that it's necessary, but I don't think we should keep it in it's current form. 
2. While running our test suite under node version 18, [this unit test in `utils/Processing.test.ts`](https://github.com/FHIR/sushi/blob/bb343df99dbb6c03f0376166ba7753cf320de25e/test/utils/Processing.test.ts#L1348) fails due to a difference in how node 18 formats the "accessing undefined property" error. I'm not sure that this is something we could or should address now, but just wanted to note it.
Screen Shot 2022-08-10 at 3 05 09 PM

The release notes look great as well!

cmoesel:
@julianxcarter -- are you sure you pulled down this branch correctly?  I _did_ change the `.nvmrc` file, as evidenced in this PR's diff:
image

And I also had noticed the test issue and fixed that too, as can also be seen in this PR's diff:
image

So it seems to me that if you're running on this branch, you should see those things...

cmoesel:
@julianxcarter - I think maybe you were looking at the `next` branch instead of the `three-oh-beta-one` branch.  The link you put in your comment is to the commit at the head of the next branch.  Whereas [this link](https://github.com/FHIR/sushi/blob/8c7a40507886cd58b1279f3207b63fec319a0660/test/utils/Processing.test.ts#L1348-L1357) goes to that test on this PR's branch and you can see that the expectation is different.

julianxcarter:
Whoops 😅. Disregard!

julianxcarter (approved):

PR #1133, opened 2022-08-16
Create markdown pages for each artifact type

mint-thompson:
Completes task [CIMPL-933](https://standardhealthrecord.atlassian.net/browse/CIMPL-933).

Each artifact type that is added to the IG's `resources` section can be added to a generated markdown page. The generated page contains a basic table with links and information. Each table has columns for the resource's name and description. The table for Profiles has an extra column for the base definition. The name and description match the values in the IG's `resources` section.

These files are only generated if the new configuration option `createArtifactPages` is set to true. If it is not true, no pages are added.

When generating these pages, SUSHI tries to avoid overwriting author-created files at the paths that it would write to. If a file exists at the path it wants to write to, it checks the starting contents. If the file starts with the comment that SUSHI writes at the beginning of these files, then the file is probably an older version of the generated file, and it's okay to overwrite it. Otherwise, the file is not generated, and an error is emitted.

An example of file contents for profiles would be something like this:
```markdown
<!-- SUSHI generated file -->
| Profiles | Derived From | Description |
| --- | --- | --- |
| [MyPatient](http://hl7.org/fhir/sushi-test/StructureDefinition/MyPatient "StructureDefinition/MyPatient") | [Patient](http://hl7.org/fhir/StructureDefinition/Patient) | Description for MyPatient is here. |
| [SamplePatient](http://hl7.org/fhir/sushi-test/StructureDefinition/sample-patient "StructureDefinition/sample-patient") | [Patient](http://hl7.org/fhir/StructureDefinition/Patient) | Demographics and other administrative information about an individual or animal receiving care or other health-related services. |
| [This patient has a title](http://hl7.org/fhir/us/mcode/StructureDefinition/MyTitlePatient "StructureDefinition/MyTitlePatient") | [Patient](http://hl7.org/fhir/StructureDefinition/Patient) | This is the description for a patient with a title. |
```


src/ig/IGExporter.ts
```diff
@@ -444,6 +456,54 @@ export class IGExporter {
...
+                info.parentUrl
+              }) | ${info.description ?? ''} |`;
+            } else {
+              return `| [${info.name}](${info.url} "${info.referenceKey}") | ${
+                info.description ?? ''
+              } |`;
+            }
+          })
+          .join('\n');
+        outputFileSync(artifactPagePath, content);
```
jafeltra, line 502: I don't think there's much we write out anymore besides the FHIR resources, but I was feeling a little weird that we generated files to somewhere besides `fsh-generated`. But I tested moving the files, and the IG publisher doesn't pick up the files if I move them inside `fsh-generated/pagecontent/`, so I think the comment is a very clear indication that it isn't safe to change that file.

src/ig/IGExporter.ts
```diff
@@ -444,6 +456,54 @@ export class IGExporter {
...
+        // overwriting the file is okay only if the existing file looks like it was generated by this function
+        const existingContent = readFileSync(artifactPagePath, 'utf-8');
+        if (!existingContent.startsWith('<!-- Generated by SUSHI -->')) {
+          logger.error(
+            `A page named ${artifactType}.md already exists. No artifact page will be created at this path.`
+          );
+          continue;
+        }
+      }
+      let content = '<!-- Generated by SUSHI -->\n';
```
jafeltra, line 482: When I run the IG publisher with the generated file, it looks like the markdown isn't being properly rendered when I just have the comment and the table in the file. If I add a blank line between the comment and the table, it renders fine. It also is fine if there is a line of text before the table as well, which could mirror what mCODE creates by hand if we added a line saying "The following ___ have been defined for this implementation guide" (see mCODE example [here](https://build.fhir.org/ig/HL7/fhir-mCODE-ig/profiles.html)). I think either works, I have a slight preference for the text just to give some context to the page, but I have no strong feeling either way.

src/ig/IGExporter.ts
```diff
@@ -827,10 +876,52 @@ export class IGExporter {
...
+            name: newResource.name,
+            url: instance.url,
+            description: newResource.description,
+            referenceKey
+          });
         }
       });
   }
 
+  private addResource(
```
jafeltra, line 889: I might be missing something, but is there a reason you don't use this function in the `addPredefinedResouces` function as well? I know that function and `addResources` are very similar, so I was just curious if there to not.

jafeltra (commented):
This looks good! I had just one thing that probably should be updated and one question.

Also, once this is merged, we should also update FSH School and the example configurations we have there to include the new `createArtifactPages` property.

src/ig/IGExporter.ts
mint-thompson: It looks like `addPredefinedResources` has some extra logic for things like overriding existing values, getting values from meta extensions, etc. So, I don't think I can make use of `addResource` in that function.

jafeltra (approved):
Yay for beautifully formatted tables!

cmoesel:
Dang.  The supported paths under `fsh-generated` are enumerated in [fhir.base.template](https://github.com/HL7/ig-template-base/search?q=fsh-generated).  I wish that when I added them, I had the foresight to add the page paths. :-/

We could still potentially make that change, but I think it might take a while to ripple through the other templates.  Perhaps I'll ask Lloyd about it.  Lloyd was very much against us modifying content under `input/`...

cmoesel:
I have an active conversation with Lloyd about this topic.  I've made this PR DRAFT so we don't merge it until we've determined the right path forward.

cmoesel:
It sounds like the FHIR crew would prefer that features such as this go directly into the IG Publisher and/or templating system rather than in SUSHI.  So... I'm going to close this PR since it seems we can't/shouldn't put this in SUSHI.  @mint-thompson -- I apologize 1000 times for having you do this work when it turns out that we can't use it.  I'm sorry I did not realize this sooner.

PR #1136, opened 2022-08-22
Create markdown file with resource links

mint-thompson:
Completes task [CIMPL-931](https://standardhealthrecord.atlassian.net/browse/CIMPL-931).

When authoring an IG, it is frequently useful for the narrative text to include links to various resources within the IG. Generate file with easy-to-use links to the page for each resource in the IG.

The name of each link is the name of the IG resource. For example, a profile named `MyProfile` with id `my-profile` would appear like this in the resource links file: `[MyProfile]: StructureDefinition-my-profile.html`. If the IG resource lacks a name, SUSHI tries to build a usable name from the resource reference key.

As part of reviewing this, please try running SUSHI against a small test project and see what kind of `markdown-link-references.md` file is produced.

### DIFF
--- a/src/ig/IGExporter.ts
+++ b/src/ig/IGExporter.ts
@@ -102,6 +102,7 @@ export class IGExporter {
     this.sortResources();
     this.addConfiguredGroups();
     this.addIndex(outPath);
+    this.addLinkReferences(outPath);
     if (!this.config.pages?.length) {
       this.addOtherPageContent();
     } else {
@@ -420,6 +421,23 @@ export class IGExporter {
         generation
       });
     }
+  }
+
+  addLinkReferences(igPath: string): void {
+    // no need to make this file if there are no resources
+    if (!this.ig.definition?.resource.length) {
+      return;
+    }
+    const linkReferencesDir = path.join(igPath, 'fsh-generated', 'includes');
+    const linkReferencesExportPath = path.join(linkReferencesDir, 'markdown-link-references.md');
+    ensureDirSync(linkReferencesDir);
+    const content = this.ig.definition.resource.map(igResource => {
+      // a configured resource may lack a name
+      // in that case, try to build a useful name from the reference
+      const linkName = igResource.name ?? igResource.reference?.reference?.replace(/^[^\/]*\//, '');
+      return `[${linkName}]: ${igResource.reference?.reference?.replace('/', '-')}.html`;
+    });
+    outputFileSync(linkReferencesExportPath, content.join('\n'));
   }
 
   /**
--- a/src/ig/IGExporter.ts
+++ b/src/ig/IGExporter.ts
@@ -432,9 +432,15 @@ export class IGExporter {
     const linkReferencesExportPath = path.join(linkReferencesDir, 'markdown-link-references.md');
     ensureDirSync(linkReferencesDir);
     const content = this.ig.definition.resource.map(igResource => {
+      // FSH resources and predefined resources will have a _linkRef
       // a configured resource may lack a name
       // in that case, try to build a useful name from the reference
-      const linkName = igResource.name ?? igResource.reference?.reference?.replace(/^[^\/]*\//, '');
+      const linkName =
+        igResource._linkRef ??
+        igResource.name ??
+        igResource.reference?.reference?.replace(/^[^\/]*\//, '');
+      // delete the _linkRef now that we've used it
+      delete igResource._linkRef;
       return `[${linkName}]: ${igResource.reference?.reference?.replace('/', '-')}.html`;
     });
     outputFileSync(linkReferencesExportPath, content.join('\n'));
@@ -1001,6 +1007,7 @@ export class IGExporter {
               if (path.basename(dirPath) === 'examples') {
                 newResource.name =
                   configResource?.name ?? metaExtensionName ?? existingName ?? resourceJSON.id;
+                newResource._linkRef = resourceJSON.id;
                 // set exampleCanonical or exampleBoolean, preferring configured values
                 if (configResource?.exampleCanonical) {
                   newResource.exampleCanonical = configResource.exampleCanonical;
@@ -1036,6 +1043,7 @@ export class IGExporter {
                   title ??
                   name ??
                   resourceJSON.id;
+                newResource._linkRef = name ?? resourceJSON.id;
               }
               if (configResource?.extension?.length) {
                 newResource.extension = configResource.extension;
--- a/src/ig/IGExporter.ts
+++ b/src/ig/IGExporter.ts
@@ -856,10 +856,12 @@ export class IGExporter {
         configResource?.name ?? pkgResource._instanceMeta.title ?? pkgResource._instanceMeta.name;
       newResource.description =
         configResource?.description ?? pkgResource._instanceMeta.description;
+      newResource._linkRef = pkgResource.id;
     } else {
       newResource.name =
         configResource?.name ?? pkgResource.title ?? pkgResource.name ?? pkgResource.id;
       newResource.description = configResource?.description ?? pkgResource.description;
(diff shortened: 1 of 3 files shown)


jafeltra (dismissed):
This looks good to me! I tested it and the publisher uses the links as expected!

cmoesel (changes requested):
Thanks, @mint-thompson! I like the idea behind this, and I'm glad you were able to do it so succinctly, and even use the `fsh-generated/includes` folder!

I noticed that you're using the IG JSON's `definition.resource.name` as the "shorthand" link handle (with `definition.resource.reference.reference` as a fallback).  The `definition.resource.name` is typically the resource's _title_ (e.g., _My Patient Profile_) rather than its _name_ (e.g., _MyPatientProfile_).

Looking back at the original request for this feature, the expectation is that the FSH _names_ would be used. I think this probably makes sense, since most authors use the FSH _names_ to reference FSH items in FSH code (e.g., `* subject only Reference(MyPatientProfile)`) -- so using the same names to reference FSH items in the pages would be consistent.  Names also provide a better guarantee of uniqueness than titles do.  So, can we change this so that the link handles are the _name_ for profile, extension, VS, CS, resource, and logical?  And the _id_ for all other instances? I think this would be most consistent w/ the FSH authoring approach.

src/ig/IGExporter.ts
```diff
@@ -398,6 +399,29 @@ export class IGExporter {
...
+      // FSH resources and predefined resources will have a _linkRef
+      // a configured resource may lack a name
+      // in that case, try to build a useful name from the reference
+      const linkName =
+        igResource._linkRef ??
+        igResource.name ??
+        igResource.reference?.reference?.replace(/^[^\/]*\//, '');
+      // delete the _linkRef now that we've used it
+      delete igResource._linkRef;
+      return `[${linkName}]: ${igResource.reference?.reference?.replace('/', '-')}.html`;
```
cmoesel, line 420: Are there cases there `ig.reference` or `ig.reference.reference` might be undefined? If so, might it make sense to skip those links rather than write them out as `undefined.html`?

Or do you think that outputting `undefined.html` might be better since it's a more visible failure than just supressing the link?

Or is it an impossible condition, so it doesn't matter?  I have a feeling it might be an impossible condition and there will always be a reference.  But I'm looking for confirmation I guess.

cmoesel (commented):
I've provide one comment which may or may not need addressing (probably not).

I also wonder if maybe we should name the file something different so it doesn't clash with existing files that people have manually created.  Maybe even something like `generated-link-references`?

I tested and confirmed that the file looks like you described -- but I'd like to actually _try_ it so I understand how it works from the author-perspective. Could you include instructions on how to actually use this file?  We'll need those for the release notes and FSH School documentation anyway.

src/ig/IGExporter.ts
mint-thompson: I think it's an impossible condition, and I was using `?.` aggressively because these attributes are not guaranteed by the type definition. But based on the current implementation, I think that `igResource.reference.reference` is guaranteed to exist here.

mint-thompson:
To explain a bit how to use the generated links:

To use these links, the generated file must be included in the markdown file you are authoring. This can be done by including this line at the bottom of the file:
```
{% include markdown-link-references.md %}
```
Then, you can create links using the names of your FSH entities surrounded by square brackets. For example, if you had Profiles named `ThisPatient` and `ThatObservation`, your markdown file could look like this:
```
## Patients and Observations
This IG provides [ThisPatient] for patient information. When observing these patients, [ThatObservation] should be used.

{% include markdown-link-references.md %}
```

jafeltra (approved):
The latest updates still look good to me, so I'm re-approving. I think we still will want @cmoesel's approval eventually since he has outstanding changes requested.

I did test out using the links before, but this time I followed the instructions you wrote up, and everything worked as described, so I think those look good. I'm not exactly sure where we will want to include those instructions in FSH School, but they are definitely helpful so we should put them _somewhere_ lol. _Edit:_ I looked again, and maybe the additional text would fit in "Tips and Tricks" and just a basic mention of the new config property in the table of configuration options.

I also thought about the file name change Chris suggested, and I think because the file is generated to `fsh-generated`, there's probably no risk of us overwriting an existing file and the name is probably okay given the folder it is in. But I'll also let Chris decide as well.

cmoesel (changes requested):
This is neat!  It works well!

I'd still like to use a different file name.  The problem is that the IG Publisher will attempt to copy all the files from `/fsh-generated/includes` and `/input/includes` into `temp/pages/_includes`.  If there are files w/ the same name, then the IG Publisher crashes.

This is particularly a problem for us because several MITRE IGs already have an include file named `markdown_link_references.md` -- so if we generate a file of the same name, those IGs will start crashing the IG Publisher.  For example, use this PR's SUSHI to build the mCODE IG -- and then run `./_genonce.sh -no-sushi` on the results.  It will crash:
> Exception in thread "main" java.lang.Error: Error generating build: the file /Users/cmoesel/dev/HL7/fhir-mCODE-ig/temp/pages/_includes/markdown-link-references.md is being generated more than once (may differ by case)

As for what the name should be, maybe something as simple as `fsh-link-refs.md`?  I don't have a super strong opinion, but I think it needs to be different than what's already in use and ideally should somehow indicate it's related to FSH.

cmoesel:
@jafeltra - See above regarding the file name...

jafeltra:
> @jafeltra - See above regarding the file name...

Ahh I forgot about the clash once you were on the publisher step. My bad! I'm in agreement!

mint-thompson:
The latest commit updates the file name to hopefully avoid causing any problems for existing IGs. And, I've created a draft PR for documentation changes as well: https://github.com/FSHSchool/site/pull/70

jafeltra:
The new file name looks good to me.

I also just realized that I might have gotten this confused with the slicing PR the first time I reviewed, but for some reason, I had linked the creation of the file with a configuration property. (I just went back and re-read some of my old comments and I even mentioned it there oops 😳). But now I feel the need to ask. Should we _always_ create this file? Then projects like mCODE would have two markdown link files that have mostly the same content, but they could update to only manually create the links that SUSHI won't create in their own `markdown-link-references.md` file. Or should the file only be created based on a configuration property in `sushi-config.yaml`?

I realize this is a late breaking question because I had gotten confused, so don't feel like we need to include this by any means. This is just a question that I had while thinking through things again, so I'm curious to hear your opinion.

cmoesel:
> Should we always create this file?

When I discussed this with @markkramerus in the past, we decided to always include it.  Its presence is not harmful -- and the fewer special configuration options the better (as it's easier for our users).

cmoesel (dismissed):
Awesome!  Thank you!

jafeltra (dismissed):

jafeltra:
> When I discussed this with @markkramerus in the past, we decided to always include it. Its presence is not harmful -- and the fewer special configuration options the better (as it's easier for our users).

That makes sense to me! Thanks for the additional insight!

cmoesel:
@jafeltra and I were talking in our one-on-one and wondered if we should keep this on `master` or move it to `next`?  It is a new feature, and it does create a new file.  Might be good 3.0 material.  @mint-thompson, what do you think?  If you agree, don't officially move it yet -- because we'll rebase `next` super soon and it will be easier if you wait until after that.

cmoesel:
>  If you agree, don't officially move it yet -- because we'll rebase next super soon and it will be easier if you wait until after that.

OK, `next` has been rebased and force-pushed, so if you agree that this would be a good feature for SUSHI 3.0, then it's now safe to rebase it on `next` and change the target here.

cmoesel (approved):
Cool stuff, @mint-thompson!  Thank you for this!

jafeltra (approved):
Looks good to me!

PR #1140, opened 2022-08-30
Fix problem w/ regression extract and copy

cmoesel:
The extract and copy approach used when getting repo sources was causing some unintended diffs.  It seems the copy is not as reliable as it should be.  Instead, we now extract the zip twice so that each copy of the repo is exactly the same.

This also updates the repos-all.txt file.

To rest, first run `npm run regression` on `master`.  If it works on your system like it does on mine, then at least the mcode repo will show diffs related to random files (mainly Java classes).  If you then run `npm run regression` on this branch, you'll see the diff come out clean.


jafeltra (approved):
I haven't seen any files trying to be compared that weren't relevant files in the past, and I ran `npm run regression` again on `master` and still didn't see anything unexpected. But I looked through the code and ran `npm run regression` on this branch, and everything works as expected, so I'd say this looks good!

cmoesel:
> I haven't seen any files trying to be compared that weren't relevant files in the past, and I ran `npm run regression` again on `master` and still didn't see anything unexpected.

I wonder if it has to do with path length.  Like maybe I have my projects at a deeper path and the copy struggles with long paths?  I'm not sure.  My SUSHI project is at `/Users/cmoesel/dev/fhir/sushi`.  @jafeltra -- is your SUSHI project at a shorter path?

jafeltra:
> is your SUSHI project at a shorter path?

It's actually at a longer path 😬

julianxcarter (approved):

Issue #1022, opened 2022-02-14
SUSHI should not automatically set fields that the IG Publisher already sets


cmoesel:
There are some fields (e.g. version) that SUSHI automatically sets when blank, but ideally the IG Publisher should set them (and already has functionality to do so).  SUSHI should stop setting these and just rely on the IG Publisher to do it.

This is probably a breaking change, so should be lined up for SUSHI 3.0.0.

cmoesel:
See: https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/Profiles.20Versioning/near/271893352

cmoesel:
I haven't tested this yet, but the code all looks good! My only concern (in hindsight) is for people who use SUSHI _without_ the IG Publisher.  I don't know how many people do that, but I guess they would end up having to set the version and status themselves.

And now that I'm looking at it -- I'm seeing that `status` is `1..1`.  This means that profiles that SUSHI creates would be invalid if we don't include a status.  I'm not sure that's a good thing.  We may need to think about that.

I also posted this on the Zulip thread to get peoples thoughts before we merge it.  See: https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/Profiles.20Versioning/near/298581914

PR #1143, opened 2022-09-07
Removing status and version presets on exported artifacts

julianxcarter:
Addresses #1022, removing SUSHI's default values for the `version` (previously defaulted to the IG version), and `status` (previously defaulted to `active`) properties on all exported artifacts. The purpose of this change is to allow the IG Publisher to set these fields automatically instead, as it provides users the option of whether or not to apply these presets where SUSHI does not.

In a regression, the only differences present should be the absence of `status` and `version` fields on exported artifacts, unless there is a rule present in the FSH that explicitly sets these fields.

**The previous default for `version` provided by SUSHI was the `version` used in the IG, the IG Publisher uses the same default.**

**The previous default for `status` provided by SUSHI was `active`, the default used by IG Publisher is the `status` used in the IG.**


cmoesel:
@julianxcarter -- do you know if the IG publisher will automatically default `status` to the same value as in the IG JSON (which corresponds to the status value in our `sushi-config.yaml`)?

I see that there is an [IG parameter](https://confluence.hl7.org/display/FHIR/Implementation+Guide+Parameters) for `propagate-status`, but I can't tell if it's talking about the same status as we are or if it's talking about a different status.

cmoesel:
I haven't tested this yet, but the code all looks good! My only concern (in hindsight) is for people who use SUSHI _without_ the IG Publisher.  I don't know how many people do that, but I guess they would end up having to set the version and status themselves.

And now that I'm looking at it -- I'm seeing that `status` is `1..1`.  This means that profiles that SUSHI creates would be invalid if we don't include a status.  I'm not sure that's a good thing.  We may need to think about that.

I also posted this on the Zulip thread to get peoples thoughts before we merge it.  See: https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/Profiles.20Versioning/near/298581914

julianxcarter:
> And now that I'm looking at it -- I'm seeing that `status` is `1..1`. This means that profiles that SUSHI creates would be invalid if we don't include a status. I'm not sure that's a good thing. We may need to think about that.

In light of discussion on Zulip, I've changed all of our exporters to pull the `status` property from the in-memory configuration object. I've also updated the code in our configuration importers to always populate the status property (defaulting to `draft` if one isn't present on FSHOnly IGs or imported IG resources, this wasn't necessary in the base case as `status` is required in the sushi-config file). 

**Prior to this PR, the previous default for version provided by SUSHI was the version used in the IG, the IG Publisher uses the same default.**

**Prior to this PR, the previous default for status provided by SUSHI was `active`, status is now pulled from the IG resource (which itself now defaults to `draft` when a user does not include one in FSHOnly mode).**

src/export/CodeSystemExporter.ts
```diff
@@ -24,8 +24,8 @@ export class CodeSystemExporter {
     codeSystem.setId(fshDefinition);
     if (fshDefinition.title) codeSystem.title = fshDefinition.title;
     if (fshDefinition.description) codeSystem.description = fshDefinition.description;
-    // Version is set to value provided in config, will be overriden if reset by rules
-    codeSystem.version = this.tank.config.version;
+    delete codeSystem.version; // deleting to allow the IG Publisher default to take hold
+    codeSystem.status = this.pkg.config.status;
```
cmoesel, line 28: The previous code used `this.tank.config` and this uses `this.pkg.config`.  Is there a meaningful difference?  I see that we're referencing `this.tank.config` in the line below, so if there isn't a difference, it might be best to use `this.tank.config` here too (for consistency).

src/export/StructureDefinitionExporter.ts
```diff
@@ -310,7 +310,8 @@ export class StructureDefinitionExporter implements Fishable {
     } else {
       delete structDef.title;
     }
-    structDef.status = 'active'; // it's 1..1 so we have to set it to something; can be overridden w/ rule
+    structDef.status = this.pkg.config.status;
```
cmoesel, line 313: As noted in a previous comment: Unless there is a reason to use `this.pkg.config`, we should use `this.tank.config` since that's where we pull the config from when setting other config-based values in this same function.

src/export/ValueSetExporter.ts
```diff
@@ -32,8 +32,8 @@ export class ValueSetExporter {
     if (fshDefinition.description) {
       valueSet.description = fshDefinition.description;
     }
-    // Version is set to value provided in config, will be overriden if reset by rules
-    valueSet.version = this.tank.config.version;
+    delete valueSet.version; // deleting to allow the IG Publisher default to take hold
+    valueSet.status = this.pkg.config.status;
```
cmoesel, line 36: As noted in a previous comment: Unless there is a reason to use `this.pkg.config`, we should use `this.tank.config` since that's where we pull the config from when setting other config-based values in this same function.

cmoesel (changes requested):
Thanks, @julianxcarter!  I left a few comments that I _think_ need to be addressed -- but let me know if there is a reason you did what you did.

I'd also like to run a full regression on this, but haven't yet had a chance to.  So, I may have more comments after that.

src/export/CodeSystemExporter.ts
julianxcarter: There's no difference, the config from the `tank` is actually used to initialize the `pkg`. Honestly there's no reason that I did this, I just referenced the config from the package out of habit :). I'll change that line here and in the other exporters for consistency!

src/export/ValueSetExporter.ts
julianxcarter: Addressed In commit [25e2afa](https://github.com/FHIR/sushi/pull/1143/commits/25e2afad1a873bf611c84ecca118ffc500f0d74e).

src/export/StructureDefinitionExporter.ts
julianxcarter: Addressed In commit [25e2afa](https://github.com/FHIR/sushi/pull/1143/commits/25e2afad1a873bf611c84ecca118ffc500f0d74e).

src/export/CodeSystemExporter.ts
julianxcarter: Addressed In commit [25e2afa](https://github.com/FHIR/sushi/pull/1143/commits/25e2afad1a873bf611c84ecca118ffc500f0d74e).

cmoesel (approved):
This looks great, @julianxcarter!  I've discussed the `status` issue w/ Mark and we are in agreement to keep it as you've implemented here -- but we need to document the heck out of it when we write the release notes!  We want people to know about this change and how to address it (by changing the status of their IG).

guhanthuran (approved):
Looks good - ran tests and regression and it seems all green to me.

PR #1147, opened 2022-09-08
Cimpl 993 improve warning message

### LINKED ISSUE #1123: Unhelpful advice when Instance name contains underscore
Currently, when an Instance name contains underscores, SUSHI emits a warning message: `The string "My_Instance" represents a valid FHIR name but not a valid FHIR id. FHIR ids cannot contain "_" and can be at most 64 characters. The id will be exported as "My-Instance". Avoid this warning by specifying a valid id directly using the "Id" keyword.` However, the `Id` keyword cannot be used on Instances. Revise this message so that when this occurs on an Instance, they are advised to set the id using an assignment rule.

Zulip thread: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Instance.20names.20and.20underscores

guhanthuran:
Fixes #1123 

Fixes issues caused by incorrect naming in Instances, and made the text more specific. Changed the warning to an error.


guhanthuran:
Fixed the branching for this PR by rebasing - it should work fine now.

guhanthuran:
Rebasing didn't work, so I just reverted the changes...ignore the gigantic commit history 😅

src/fhirtypes/mixins.ts
```diff
@@ -81,10 +81,12 @@ export class HasId {
...
-      logger.error(
-        `The string "${idToCheck}" does not represent a valid FHIR id. FHIR ids may contain any combination of upper- or lower-case ASCII letters ('A'..'Z', and 'a'..'z'), numerals ('0'..'9'), '-' and '.', with a length limit of 64 characters.`,
-        idRule?.sourceInfo ?? fshDefinition.sourceInfo
-      );
+      let message = `The string "${idToCheck}" does not represent a valid FHIR id. FHIR ids only allow ASCII letters (A-Z, a-z), numbers (0-9), hyphens (-), and dots (.), with a length limit of 64 characters.`;
+      if (fshDefinition instanceof Instance) {
+        message +=
+          ' Avoid this warning by changing the Instance declaration to follow the FHIR id requirements.';
+      }
+      logger.error(message, idRule?.sourceInfo ?? fshDefinition.sourceInfo);
```
jafeltra, line 89: I don't think we want to change the current behavior from a warning to an error. I think we want to continue to keep logging a warning and fix the `id` as best we can, but we just want to update the wording of the warning. If we change this to an error and no longer sanitize the id, we are actually now generating FHIR instances that are not valid FHIR because they have the invalid `id` set in them. I also checked the Github issue reported for this, and it sounds like we just want to change the wording of the warning message.

In the current implementation, the warning in the case above this one is being reached; SUSHI doesn't output an error, so it doesn't reach this case. Instead of including this check for `if (fshDefinition instanceof Instance)` here, I think it might make more sense to include it in the warning case above and update the warning message if the `fshDefinition` is an instance. The updated message should also include the sentence about updating the id: `The id will be exported as "${sanitizedId}".` That way we can keep the logic for sanitizing the string and continue to update the instance that is output to be valid FHIR.

jafeltra (changes requested):
This is looking good. I have two comments, one I couldn't make inline so I included it below with a link to the line it's referring to.

- This isn't something you changed, but your changes made me think of it. Since we can pass in `Instance` into the `validateId` function and it technically isn't one of the three types we call out in the type declaration, we should add it in. That means [this line](https://github.com/FHIR/sushi/pull/1147/files#diff-36114c2c00bf2fa2fe103657004e79582f568b3ce9fbccbaa9138fa108bf8977R63) would look like `validateId(fshDefinition: FshStructure | FshCodeSystem | FshValueSet | Instance)`. It works without that now because the `Instance` type is compatible with the `FshStructure` type, but it doesn't hurt to add the type explicitly.

src/fhirtypes/mixins.ts
cmoesel: @jafeltra and @julianxcarter - Guhan and I worked on this together, and agreed on this implementation.  I understand your concerns, but here is our thinking on this one...

For profiles, extensions, value sets, and code systems, the value of the declaration is the `name` (not the `id`).  E.g., `Profile: My_Profile` specifies the SD's `name` as `My_Profile`.  Then, as a convenience feature, we default the `id` to the `name` if the `id` is not supplied.  _That_ is why we introduced the feature to sanitize the ID -- because in many cases, the author has not intentionally supplied an `id`; they just got the default value derived from the name. Rather than ding the author for that, we clean up the default value and warn them.

_Instances_, however, are another matter.  The value of the declaration is the `id`.  E.g., `Instance: My_Instance` specifies the instance's `id` as `My_Instance`.  In this case, the `id` is _never_ a default value derived from something else; it's _always_ a direct author-provided `id`. In that case, I think it may be appropriate to signal that as an error -- because there is no reason the author can't (or shouldn't) specify a valid ID on their own.  They really ought to fix it.

Does that reasoning make sense to you?  Does it help to explain the approach taken in this PR?

jafeltra: That makes sense to me! I'm good with keeping the implementation as is. I'll run the full regression to see how many projects this might affect.

jafeltra:
I finished running the regression in this branch, and the changes cause no regressions. So I think we're safe to keep this as merging into `master`.

jafeltra (approved):
Looks good to me! Thanks for making that change.

Can you also hit the "Update branch" button so it is fully up to date with `master` for when @julianxcarter re-reviews?

julianxcarter (approved):

PR #1150, opened 2022-09-15
Support for R5 ImplementationGuide JSON

jafeltra:
The R5 ballot version is coming soon, and with it come changes to the `ImplementationGuide` resource. The version this branch targets is http://hl7.org/fhir/2022Sep/implementationguide.html. This PR adds support for converting any properties that changed and adds support for new properties introduced in the new version. The differences are summarized in the "R4 diff" table on that page.

All new properties are supported in the `sushi-config.yaml`. Any choice elements are supported by specifying the choice you are using. For example, `source[x]` can be provided as `sourceString`, `sourceUrl`, or `sourceMarkdown` in the `sushi-config.yaml`. Any property that changed from R4 to R5 that can be inferred based on an old R4 property are converted automatically. For example, if `IG.definition.resource.exampleBoolean` or `IG.definition.resource.exampleCanonical`  were set in the R4 resource, SUSHI will now remove those properties and set `IG.definition.resource.isExample` as appropriate.

The draft release notes describe the changes made, and the commits do a fairly good job of describing each change. I'm also happy to provide more details as anyone reviews, because I'm currently creating an update to FSH School as well.


src/ig/IGExporter.ts
```diff
@@ -29,6 +30,17 @@ import { ConfigurationMenuItem } from '../fshtypes';
 import { logger, Type, getFilesRecursive } from '../utils';
 import { FHIRDefinitions } from '../fhirdefs';
 import { Configuration } from '../fshtypes';
+import { parseCodeLexeme } from '../import';
+
+function isR4(fhirVersion: string[]) {
+  let containsR4Version = false;
+  fhirVersion.forEach(v => {
+    if (/(^4\.0\.)|(^(4\.1\.|4\.3.\d+-))|(^4\.3.\d+$)/.test(v)) {
```
cmoesel, line 38: This seems overly complex.  Wouldn't it be sufficient just to check if it starts w/ 4.0., 4.1., or 4.3.?
```suggestion
    if (/^4\.[013]\./.test(v)) {
```

src/ig/IGExporter.ts
```diff
@@ -1294,6 +1317,148 @@ export class IGExporter {
...
+          d => d.packageId === dependency.packageId
+        );
+        if (configDependency.reason) {
+          dependency.reason = configDependency.reason;
+        }
+      });
+    }
+  }
+
+  updatePageName(page: ImplementationGuideDefinitionPage): void {
```
cmoesel, line 1413: Should this be named `updatePageNameForR5` for clarity (since you would never want to do this on an R4 IG)?

test/ig/IGExporter.IG.test.ts
```diff
@@ -2657,4 +2657,474 @@ describe('IGExporter', () => {
...
+      codeSystemDef.url = 'http://example.org/CodeSystem/sample-code-system';
+      pkg.codeSystems.push(codeSystemDef);
+
+      const exporter = new IGExporter(pkg, defs, fixtures);
+      // No need to regenerate the IG on every test -- generate it once and inspect what you
+      // need to in the tests
+      exporter.export(tempOut);
+    };
+
+    beforeAll(beforeAllR5Setup);
```
cmoesel, line 2805: Just curious... why did you break this into a separate function rather than just inlining it in `beforeAll`?

cmoesel (changes requested):
This code is beautiful.  Very well done.

Unfortunately, when I test this locally, it crashes with this error:
image

It's pretty easy to reproduce.  Just run `sushi --init` (using globally installed SUSHI) and change the `fhirVersion` in the resulting project to `5.0.0-ballot`.  Then run SUSHI from this branch and it will crash.

src/ig/IGExporter.ts
jafeltra: Whoops yeah it would! I had just quickly copied the regular expressions we use elsewhere and never looked back. Oops.

src/ig/IGExporter.ts
jafeltra: Oh yeah I like that. Do you think it's worth also adding a `ForR5` to the other two function (`addPageSource` and `defaultPageSourceUrl`) or do you think that it's implied since `source[x]` doesn't exist in R5?

test/ig/IGExporter.IG.test.ts
jafeltra: I'm not 100% convinced by myself here, but the reasoning was because I had two tests for the two different `versionAlgorithm[x]` type choices. Since it was a top level property, I couldn't just put all the different cases into one list, like I did for `resources` and `pages`. So when I made those tests, they had to re-export the IG with the updated config. But after doing that, any tests that followed those tests used the latest exported IG and not the one that was exported in the `beforeAll`. So to avoid side effects and to avoid requiring those tests to be at the end of the list of tests, I called the `beforeAllR5Setup` function at the end of those two tests.

I think the same thing essentially could be accomplished by pulling the two `versionAlgorithm[x]` tests out into their own `describe` block outside of the `#r5-ig-format` block, but I decided to go with this to keep the list of tests in one group. It was a coin flip decision, so I can always go for the separate describe blocks.

src/ig/IGExporter.ts
cmoesel: I thought about that. Since `source[x]` exists only in R5, it's probably not as important -- but it certainly wouldn't hurt!  I'll leave it up to you.

test/ig/IGExporter.IG.test.ts
cmoesel: Oh.  I see.  The other option would be to put some of that (including the export) in `before` instead of `beforeAll`.  It might be a little more expensive, but then you know you don't have to ever worry about one test impacting another.

I just noticed now that you only reset the logger spy in the `beforeAll` too, so your spy is collecting all the logs from all the tests -- which could be a problem if you decide to inspect logs for any of the tests.

src/ig/IGExporter.ts
jafeltra: I'll do it. In the name of consistency.

test/ig/IGExporter.IG.test.ts
jafeltra: Ahh yeah I did think of the `beforeEach` and was wondering if it was worth all the extra exports. I could go either way really. Do you think having the separate `beforeAllR5Setup` is too confusing? It's definitely something we don't normally do, so I would be fine re-exporting each time in a `beforeEach` if it makes it easier to understand in the future.

I might have convinced myself typing that out, but still curious what you think.

jafeltra:
[eb76ad8](https://github.com/FHIR/sushi/pull/1150/commits/eb76ad8ba9c8c9c14f1b73c864a8553d5cb69c1e) should fix the crash you were seeing. It also renames the functions to indicate they are for R5 IGs and simplifies the R4 regex.

test/ig/IGExporter.IG.test.ts
jafeltra: I swapped to `beforeEach` because it probably is simpler, and at the very least, it's what we normally do. So I think ease of understanding in the future is worth the extra export calls.

And the bonus is that it resets the logger spy before each, so we can inspect logs in the test cases in the future without worrying.

cmoesel (approved):
This looks really good.  The implementation is clean and easy to follow.  Not only that, but it seems to work. ;-)  I tested this manually and the results were as expected.

I wish I could test this with the IG Publisher, but alas, it is known to be broken.  Still, it seems that everything is in order here, so I'm approving in order to get out ahead of the IG Publisher fix!

mint-thompson (approved):
The generated IG JSON looked correct when I tested it on my system. Thank you!

PR #1152, opened 2022-09-21
Add command-line option to update dependencies

mint-thompson:
Completes task [CIMPL-995](https://standardhealthrecord.atlassian.net/browse/CIMPL-995).

When this option is provided, SUSHI will attempt to update the dependencies listed in the FSH project's configuration file. Dependencies listed with a version of "current" or "dev" will not be modified. The user is prompted before any updates are made and is shown a list of all of the updates that will be applied. If the configuration was not imported from a sushi-config file, this option does nothing.

If the user provides this flag and applies updates, the project's sushi-config file will be rewritten with the updated version information. This may also result in some formatting changes to the configuration file, but all other file contents (including comments) will be preserved.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -29,6 +29,7 @@ import {
   findInputDir,
   ensureOutputDir,
   readConfig,
+  updateExternalDependencies,
   loadExternalDependencies,
   fillTank,
   writeFHIRResources,
@@ -64,6 +65,7 @@ async function app() {
       false
     )
     .option('-i, --init', 'initialize a SUSHI project')
+    .option('-u, --update-dependencies', 'update FHIR packages in project configuration')
     .version(getVersion(), '-v, --version', 'print SUSHI version')
     .on('--help', () => {
       console.log('');
@@ -183,6 +185,14 @@ async function app() {
       logger.error(`An unexpected error occurred: ${e.message ?? e}`);
     }
     process.exit(1);
+  }
+
+  // Update dependencies
+  if (program.updateDependencies) {
+    const updateResult = await updateExternalDependencies(config);
+    if (!updateResult) {
+      process.exit(0);
+    }
   }
 
   // Load dependencies
--- a/src/app.ts
+++ b/src/app.ts
@@ -163,6 +163,13 @@ async function app() {
   let tank: FSHTank;
   let config: Configuration;
 
+  // Update dependencies
+  if (program.updateDependencies) {
+    config = readConfig(originalInput);
+    await updateExternalDependencies(config);
+    process.exit(0);
+  }
+
   try {
     let rawFSH: RawFSH[];
     if (!fs.existsSync(input)) {
@@ -173,7 +180,13 @@ async function app() {
     } else {
       rawFSH = getRawFSHes(input);
     }
-    if (rawFSH.length === 0 && !fs.existsSync(path.join(originalInput, 'sushi-config.yaml'))) {
+    if (
+      rawFSH.length === 0 &&
+      !(
+        fs.existsSync(path.join(originalInput, 'sushi-config.yaml')) ||
+        fs.existsSync(path.join(originalInput, 'sushi-config.yml'))
+      )
+    ) {
       logger.info('No FSH files or sushi-config.yaml present.');
       process.exit(0);
     }
@@ -185,14 +198,6 @@ async function app() {
       logger.error(`An unexpected error occurred: ${e.message ?? e}`);
     }
     process.exit(1);
-  }
-
-  // Update dependencies
-  if (program.updateDependencies) {
-    const updateResult = await updateExternalDependencies(config);
-    if (!updateResult) {
-      process.exit(0);
-    }
   }
 
   // Load dependencies
--- a/src/utils/Processing.ts
+++ b/src/utils/Processing.ts
@@ -185,10 +185,10 @@ export async function updateExternalDependencies(config: Configuration): Promise
         latestVersion = res?.data?.['dist-tags']?.latest;
       } catch (e) {
         try {
-          res = await axiosGet(`https://packages2.fhir.org/${dep.packageId}`);
+          res = await axiosGet(`https://packages2.fhir.org/packages/${dep.packageId}`);
           latestVersion = res?.data?.['dist-tags']?.latest;
         } catch (e) {
-          logger.info(`Could not get version info for package ${dep.packageId}`);
+          logger.warn(`Could not get version info for package ${dep.packageId}`);
           return;
         }
       }
@@ -198,14 +198,14 @@ export async function updateExternalDependencies(config: Configuration): Promise
           changedVersions.set(dep.packageId, dep.version);
         }
       } else {
-        logger.info(`Could not determine latest version for package ${dep.packageId}`);
+        logger.warn(`Could not determine latest version for package ${dep.packageId}`);
       }
     }
   });
   await Promise.all(promises);
   if (changedVersions.size > 0) {
     // before changing the file, check with the user
-    const continuationChoice = readlineSync.keyInYN(
+    const continuationChoice = readlineSync.keyInYNStrict(
       [
         'Updates to dependency versions detected:',
         ...Array.from(changedVersions.entries()).map(
@@ -216,12 +216,8 @@ export async function updateExternalDependencies(config: Configuration): Promise
         'Do you want to apply these updates?',
         '- [Y]es, apply updates to sushi-config.yaml',
         '- [N]o, quit without applying updates',
-        'Choose one [Y,N]: '
-      ].join('\n'),
-      {
-        limit: 'AQ',
-        cancel: false
-      }
+        'Choose one: '
+      ].join('\n')
     );
     if (continuationChoice === true) {
       const configText = fs.readFileSync(config.filePath, 'utf8');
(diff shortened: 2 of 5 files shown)


src/utils/Processing.ts
```diff
@@ -161,6 +162,75 @@ export function readConfig(input: string): Configuration {
...
 
+export async function updateExternalDependencies(config: Configuration): Promise<boolean> {
+  // only try to update if we got the config from sushi-config.yaml, and not from an IG
+  const changedVersions: Map<string, string> = new Map();
+  if (config.filePath != null && config.dependencies?.length > 0) {
+    const promises = config.dependencies.map(async dep => {
+      // current and dev have special meanings, so don't try to update those dependencies
+      if (dep.version != 'current' && dep.version != 'dev') {
+        try {
+          const res = await axiosGet(`https://packages.fhir.org/${dep.packageId}`);
```
jafeltra, line 173: I'm not sure if this is even something we want to do, but should we also check for packages at `packages2.fhir.org`?

jafeltra (commented):
This is a really neat feature. The implementation and tests make a lot of sense to me. I had one question inline.

Also, should this be one of the fancy, fun features we put into `next` so the major release has neat new features? Or is this something we want available in the `2.x` version for those who decide to not update right away?

cmoesel:
> Also, should this be one of the fancy, fun features we put into next so the major release has neat new features?

Yeah.  I hadn't thought about it much before, but I think this would be a nice feature to tie to 3.0.  I'm also thinking about maybe(?) changing the way we do commands on the CLI in SUSHI 3.0, so it's better to hold off on new commands until 3.0 for that reason too.

src/utils/Processing.ts
cmoesel: At first I was going to say that in theory, `packages2` should only be needed for looking up unreleased versions of FHIR (e.g., FHIR R5), so it's _probably_ OK not to support it here.

But now I'm thinking about it more -- and it's really that `packages` doesn't support R4B or R5 yet.  And I think that means it also doesn't support IGs that depend on R4B or R5.  So `packages2` probably hosts R4B, R5, and all IGs that depend on R4B or R5.  And I think maybe that's a big enough group that we should check both?

src/utils/Processing.ts
```diff
@@ -161,6 +162,75 @@ export function readConfig(input: string): Configuration {
...
+        [
+          'Updates to dependency versions detected:',
+          ...Array.from(changedVersions.entries()).map(
+            ([packageId, version]) => `- ${packageId}: ${version}`
+          ),
+          'SUSHI can apply these updates to your sushi-config.yaml file.',
+          'This may affect the formatting of your file.',
+          '- [A]pply updates to sushi-config.yaml',
+          '- [Q]uit without applying updates',
+          'Choose one [A,Q]: '
```
jafeltra, line 201: This is just genuinely a question. My head understands that `[A]pply` and `[Q]uit` are more obviously clear about what actions you are choosing when you type in an answer, but when I was using this feature, my muscle memory wasn't listening to my head, and I kept expecting to type `y` or `n` to choose to update the packages. I think it might be because there are only two options to really choose from that I kept getting confused with my instinct for yes or no options. Do you think it makes sense to just have the typical y/n choice here since we really only have two options? Or is the additional clarity worthwhile and my muscle memory will eventually catch up?

jafeltra (changes requested):
In addition to the things mentioned previously, I tested this a little more thoroughly and found a few other things that should either be updated or I had questions about.

- When I have a dependency that uses the object format instead of just the version, I lose the additional information. For example, if I have this:
  ```yaml
  dependencies:
      hl7.fhir.us.mcode:
        id: mcode
        uri: http://hl7.org/fhir/us/mcode/ImplementationGuide/hl7.fhir.us.mcode
        version: 1.0.0
  ```
  I get prompted to update the version to `2.0.0`, but when it re-writes the config file, the dependency is replaced with just
  ```yaml
  dependencies:
    hl7.fhir.us.mcode: 2.0.0
  ```
- There were some blank commented lines below the dependencies section that was rewritten that got lost on the rewrite. This might be okay since it isn't a substantive change in any way, but I wanted to mention it. For example, this
  ```yaml
  dependencies:
    hl7.fhir.us.core: 3.1.0
  #
  #
  # Comment with content
  ```
  became
  ```yaml
  dependencies:
    hl7.fhir.us.core: 3.1.0
  # Comment with content
  ```
  I'm not sure if this counts just as a formatting change, so I wanted to ask about it.
- This is another change that I think might just be a formatting change but want to double check. In our documentation on FSH School and in the config file we generate with the `--init` option, we specify that pages that don't include a `title` or subpages, just have the key with a blank value. For example, this is in both FSH School and in the example config file:
  ```yaml
  pages:
    index.md:
      title: Example Home
    implementation.xml:
   ```
  However, after the config file is re-written, that property gets rewritten with `null`s instead of the blank value, like this:
  ```yaml
  pages:
    index.md:
      title: Example Home
    implementation.xml: null
  ```
  This seems equivalent logically, but it does deviate from our documented norms. Is this something we're okay with?
  - There were a few places where blank lines between comment blocks were removed. I think that is definitely a formatting change and we're okay with it, right?

src/utils/Processing.ts
guhanthuran: I agree with Julia here - Y/N is something I'm also more used to using when accepting changes in the CLI. In regards to the changes here - maybe we can keep the A/Q options while also allowing the user to have the additional option of using Y in lieu of A and N in lieu of Q?

guhanthuran: Another note - A and Q are very close on the keyboard and can lead to a misinput.

jafeltra:
There were two other potential ideas/changes we had mentioned the other day, so I wanted to include them here to keep track of them. Let me know what you think about them, and we can decide if we want to actually include them in this PR or if we want them to be in a later iteration.

- Should this option only run the check for updated dependencies and give the option to update them, without running any other SUSHI functionality?
- Right now, any packages that need to be updated are presented to the user at once, and they can choose to accept or reject all of those changes. Should this option present each package to be updated one at a time? That way if there are three dependencies to be updated, you can choose to update two and not the third.

Happy to discuss either of these (and anything else on this PR) to decide if it should be included here!

mint-thompson:
> There were two other potential ideas/changes we had mentioned the other day, so I wanted to include them here to keep track of them. Let me know what you think about them, and we can decide if we want to actually include them in this PR or if we want them to be in a later iteration.
> 
> * Should this option only run the check for updated dependencies and give the option to update them, without running any other SUSHI functionality?
> * Right now, any packages that need to be updated are presented to the user at once, and they can choose to accept or reject all of those changes. Should this option present each package to be updated one at a time? That way if there are three dependencies to be updated, you can choose to update two and not the third.
> 
> Happy to discuss either of these (and anything else on this PR) to decide if it should be included here!

I think that it makes sense to add the option to just update dependencies and not run SUSHI when we do some refactoring of the CLI options into subcommands.

As for updating one-by-one, I don't think that is quite as necessary: since SUSHI provides a list of dependencies to be updated, a user can just refer to that list when manually updating the config. Then again, that could get a little messy if an IG has a few dozen dependencies and the user wants to update half of them. I'd like to know what other folks think about this one.

cmoesel:
FYI -- a rebased `next` was just force-pushed, so you'll probably need to rebase this PR again.

jafeltra (changes requested):
I reviewed this by looking through a different comparison because this branch still needs to be updated since the `next` rebased. I can plan to re-review after that update as well to make sure the changes all still look the same.

> I think that it makes sense to add the option to just update dependencies and not run SUSHI when we do some refactoring of the CLI options into subcommands.

I'm not completely sure I agree. If the intended behavior is for this to only run the update dependencies functionality and not run SUSHI, I think it would be good to include the change here so the feature works as intended. I don't know that it makes sense to add updating functionality into the task that is intended to rework the existing options into subcommands. I think it should be pretty straightfoward to just move where we call the function from earlier in `app.ts`. If it isn't that easy, then maybe we can have that be updated in a separate task. If you disagree (strongly or even a little), happy to chat about it.

Also, were you able to confirm if the remainder of the changes in the config file are considered styling changes? I did confirm that the dependencies were preserved correctly now! 🎉

src/app.ts
```diff
@@ -171,7 +180,13 @@ async function app() {
     } else {
       rawFSH = getRawFSHes(input);
     }
-    if (rawFSH.length === 0 && !fs.existsSync(path.join(originalInput, 'sushi-config.yaml'))) {
+    if (
+      rawFSH.length === 0 &&
+      !(
+        fs.existsSync(path.join(originalInput, 'sushi-config.yaml')) ||
+        fs.existsSync(path.join(originalInput, 'sushi-config.yml'))
```
jafeltra, line 187: Whoa. Interesting that we hadn't realized this until now. Nice!

jafeltra (commented):
This looks great! Thanks for making all the updates!

I remembered one more thing that we might want to include. The `README` file includes the `sushi --help` text, which now includes the `-u --update-dependencies` option. Do you mind adding in the new line [here](https://github.com/FHIR/sushi/blob/master/README.md?plain=1#L24-L44)?

jafeltra (approved):
Yay this looks great!!

guhanthuran:
Looks good to me!

PR #1153, opened 2022-09-21
Escape Characters in menu.xml

### LINKED ISSUE #1097: SUSHI does not properly encode menu item names
Including something like this in `sushi-config.yaml`'s `menu` causes problems:
```yaml
Specification:
    Measures: measures.html
    Plans & Interventions: interventions.html
    Conformance & Workflow: workflow.html
```
The problem is that `&` needs to be encoded to `&amp;` in the XML menu.

If we fix SUSHI to do this encoding, we need to be careful not to double-encode items that were already encoded by the user.

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Menu.20issue/near/285404188

jafeltra:
This PR supports encoding the five characters that need to escaped in XML. To do this, because there are only five and they are defined in XML, I just encoded each character when it is in any menu item in a `menu.xml` that we generated from the `sushi-config.yaml`. This also checks to make sure we don't double encode any character if the user has already encoded it. For example, if the user has written `Profile &amp; Extensions`, we want to just leave that `&` character and not try to re-encoded it.

Also, based on the suggestion on the issue, if SUSHI detects that there are any escaped characters provided by the user in the configuration, SUSHI will output a warning letting them know we will handle that now and they can replace their character with the un-encoded version, which will be easier to read in the config file.

To test this, you can include both encoded and unencoded characters in a menu.xml file, run SUSHI and check that it properly encodes characters that need to be encoded and warns about characters that are already encoded.

You can also run the IG Publisher to see how the menu works. If you run `master` or the current SUSHI release with a configuration that includes a character that needs to be escaped, the menu will not render correctly in the IG. If you run this branch with the same configuration, you should see that the menu generated by the IG Publisher renders and works as expected. I tested the IG Publisher with menu items with each of the characters that need to be encoded in XML, and it actually works fine with `<`, `>`, `"`, and `'` and only breaks when an `&` is used in the name. However, the IG Publisher correctly handles when they all are escaped as well and two places I read online said it's safer to just always escape all five characters, so that's what I went with.

This PR fixes #1097.

### DIFF
--- a/src/ig/IGExporter.ts
+++ b/src/ig/IGExporter.ts
@@ -644,7 +644,7 @@ export class IGExporter {
     const menuXMLDefaultPath = path.join(this.inputPath, 'input', 'includes', 'menu.xml');
     const menuXMLOutputPath = path.join(igPath, 'fsh-generated', 'includes', 'menu.xml');
 
-    // If user provided file and config, log a warning but prefer the config.
+    // If user provided file and config, log a warning but prefer the file.
     if (existsSync(menuXMLDefaultPath) && this.config.menu) {
       const filePathString = path.join(
         path.basename(this.inputPath),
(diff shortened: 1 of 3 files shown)


mint-thompson (approved):
Tried out this update with some of the targeted characters in menu items. Everything looks to be in good order. Thank you!

julianxcarter (approved):
Looks good to me, good work :).

PR #1158, opened 2022-10-06
Warn on empty strings

### LINKED ISSUE #1087: Warn on empty string
FHIR indicates that `string` values should have at least one non-whitespace character (see: https://build.fhir.org/datatypes#string).  SUSHI should warn users when they assign an empty string.

If this is complex to do for arbitrary assignments, we should at least do it for the major metadata keywords (`Title:`, `Description:`, etc) -- especially since the VS Code snippets produce empty strings for these by default.

See: https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/Miss.20leading.20QA.2Ehtml.20results/near/282548095

guhanthuran:
Fixes #1087 

Adds warnings whenever the user makes their Title or Description an empty string within their Instance, Code System, Structure Definition, or Value Set. 

Testers would most likely want to test situations where Title or Description within these types exist but are empty strings. 

Thanks for reviewing!

### DIFF
--- a/test/export/StructureDefinitionExporter.test.ts
+++ b/test/export/StructureDefinitionExporter.test.ts
@@ -82,7 +82,7 @@ describe('StructureDefinitionExporter R4', () => {
       expect(pkg.extensions.length).toBe(1);
     });
 
-    it('should warn when the structDef is a profile title and/or description is an empty string', () => {
+    it('should warn when the structDef is a profile and title and/or description is an empty string', () => {
       const profile = new Profile('Foo');
       profile.parent = 'Patient';
       profile.title = '';
--- a/test/export/StructureDefinitionExporter.test.ts
+++ b/test/export/StructureDefinitionExporter.test.ts
@@ -147,7 +147,7 @@ describe('StructureDefinitionExporter R4', () => {
       doc.resources.set(resource.name, resource);
       exporter.exportStructDef(resource);
 
-      //expect(loggerSpy.getAllMessages('warn').length).toBe(2);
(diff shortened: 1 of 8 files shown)


guhanthuran:
One question for reviewers: should we also add this functionality to `MappingExporter`?

cmoesel:
> One question for reviewers: should we also add this functionality to `MappingExporter`?

I don't think we need it on `MappingExporter`, but as discussed last week, it would be awesome to have tests for `Extension`, `Logical`, and `Resource`.  I think they're all covered by your work in the `StructureDefinitionExporter` -- but it would be good to have the tests to show it!

jafeltra (commented):
This is looking good to me! I don't think I have any other comments besides the tests that @cmoesel mentioned.

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -82,7 +82,7 @@ describe('StructureDefinitionExporter R4', () => {
       expect(pkg.extensions.length).toBe(1);
     });
 
-    it('should warn when title and/or description is an empty string', () => {
+    it('should warn when the structDef is a profile title and/or description is an empty string', () => {
```
cmoesel, line 85:
```suggestion
    it('should warn when the structDef is a profile and title and/or description is an empty string', () => {
```

guhanthuran: Nice catch, thanks!

cmoesel (approved):
Looks good, @guhanthuran!  Thank you for this!

PR #1163, opened 2022-10-26
Updated Command Structure

julianxcarter:
Updates SUSHI's command structure to make use of commands rather than flags. SUSHI now takes the following commands:

1. `build` which builds an IG. This takes a path to a FSH Project as an argument and also supports the following flags (`-o, --out <out>`, `-d, --debug`, `-p, --preprocessed`, `-s, --snapshot`, `-r, --require-latest`). Similar to before, `sushi build` can be run with either `.` as an input directory or with no input directory specified at all, in which case the input is assumed to be the current working directory. 
2. `updateDependencies` which updates FHIR dependencies in a user's configuration file. This takes a path to a FSH Project as an argument and also has the same default behavior for the path argument (`.` and no path argument both default to current working directory.
3. `init` which goes through our initialize project workflow. This takes no arguments.

All three commands also support the `-h`/`--help` flag independently, meaning you can run `sushi build -h` for instance and get the options that are specific to that command. 

SUSHI is still backwards compatible with our old command structure and can still be run the same way as before (e.g. with `sushi .`). This is done by inserting `build` into the arguments array when we detect that the third element is a file path. I think it's possible to add the options and a separate action to the `Command` object and handle it that way, but I didn't want to do that because then users would see those options when running `sushi --help`. Basically I thought that backwards compatibility was a goal, but continuing to highlight usage of the Old Ways™ was not what we wanted to do.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-*   @jafeltra @cmoesel @julianxcarter @mint-thompson @guhanthuran
+*   @jafeltra @cmoesel @mint-thompson @guhanthuran
--- a/src/app.ts
+++ b/src/app.ts
@@ -10,7 +10,7 @@ register({
 
 import path from 'path';
 import fs from 'fs-extra';
-import { Command } from 'commander';
+import { Command, OptionValues } from 'commander';
 import chalk from 'chalk';
 import process from 'process';
 import { pad, padStart, padEnd } from 'lodash';
@@ -43,6 +43,7 @@ import {
 } from './utils';
 
 const FSH_VERSION = '2.0.0';
+const SUPPORTED_COMMANDS = ['build', 'updateDependencies', 'init', '--help'];
 
 app().catch(e => {
   logger.error(`SUSHI encountered the following unexpected error: ${e.message}`);
@@ -50,11 +51,14 @@ app().catch(e => {
 });
 
 async function app() {
-  let input: string;
-
   const program = new Command()
     .name('sushi')
-    .usage('[path-to-fsh-project] [options]')
+    .version(getVersion(), '-v, --version', 'print SUSHI version');
+
+  program
+    .command('build')
+    .description('export a FSH project into a FHIR IG')
+    .argument('[path-to-fsh-project]')
     .option('-o, --out <out>', 'the path to the output folder')
     .option('-d, --debug', 'output extra debugging information')
     .option('-p, --preprocessed', 'output FSH produced by preprocessing steps')
@@ -64,9 +68,9 @@ async function app() {
       'exit with error if this is not the latest version of SUSHI',
       false
     )
-    .option('-i, --init', 'initialize a SUSHI project')
-    .option('-u, --update-dependencies', 'update FHIR packages in project configuration')
-    .version(getVersion(), '-v, --version', 'print SUSHI version')
+    .action(async function (projectPath, options) {
+      await runBuild(projectPath, options);
+    })
     .on('--help', () => {
       console.log('');
       console.log('Additional information:');
@@ -74,19 +78,44 @@ async function app() {
       console.log('    Default: "."');
       console.log('  -o, --out <out>');
       console.log('    Default: "fsh-generated"');
-    })
-    .arguments('[path-to-fsh-defs]')
-    .action(function (pathToFshDefs) {
-      input = pathToFshDefs;
-    })
-    .showHelpAfterError()
-    .parse(process.argv)
-    .opts();
+    });
 
-  if (program.init) {
-    await init();
-    process.exit(0);
+  program
+    .command('updateDependencies [path-to-fsh-project]')
+    .description('update FHIR packages in project configuration')
+    .action(async function (projectPath) {
+      const input = ensureInputDir(projectPath);
+      const config: Configuration = readConfig(input);
+      await updateExternalDependencies(config);
+      process.exit(0);
+    })
+    .on('--help', () => {
+      console.log('');
+      console.log('Additional information:');
+      console.log('  [path-to-fsh-project]');
+      console.log('    Default: "."');
+    });
+
+  program
+    .command('init')
+    .description('initialize a SUSHI project')
+    .action(async function () {
+      await init();
+      process.exit(0);
+    });
+
+  // Maintain backwards compatability with prior SUSHI command structure by defaulting to build
+  if (
+    process.argv.length >= 2 &&
+    !SUPPORTED_COMMANDS.includes(process.argv[2]) &&
+    fs.existsSync(process.argv[2])
+  ) {
+    process.argv = [...process.argv.slice(0, 2), 'build', ...process.argv.slice(2)];
   }
+  program.showHelpAfterError().parse(process.argv).opts();
+}
+
+async function runBuild(input: string, program: OptionValues) {
   if (program.debug) logger.level = 'debug';
 
   logger.info(`Running ${getVersion()}`);
--- a/src/app.ts
+++ b/src/app.ts
@@ -59,15 +59,15 @@ async function app() {
     .command('build', { isDefault: true })
     .description('build a SUSHI project')
     .argument('[path-to-fsh-project]')
-    .option('-o, --out <out>', 'the path to the output folder')
     .option('-d, --debug', 'output extra debugging information')
+    .option('-o, --out <out>', 'the path to the output folder')
     .option('-p, --preprocessed', 'output FSH produced by preprocessing steps')
-    .option('-s, --snapshot', 'generate snapshot in Structure Definition output', false)
     .option(
       '-r, --require-latest',
       'exit with error if this is not the latest version of SUSHI',
       false
     )
+    .option('-s, --snapshot', 'generate snapshot in Structure Definition output', false)
     .action(async function (projectPath, options) {
       await runBuild(projectPath, options);
     })
--- a/src/app.ts
+++ b/src/app.ts
@@ -58,7 +58,7 @@ async function app() {
 
   program
     .command('build')
-    .description('export a FSH project into a FHIR IG')
+    .description('build a SUSHI project')
     .argument('[path-to-fsh-project]')
     .option('-o, --out <out>', 'the path to the output folder')
     .option('-d, --debug', 'output extra debugging information')
--- a/src/app.ts
+++ b/src/app.ts
@@ -115,7 +115,7 @@ async function app() {
       console.log(
         'The --init option has been moved to a separate command. Instead, run the following command: sushi init'
       );
-      process.exit(0);
+      process.exit(1);
     } else if (process.argv.length === 2 || process.argv.slice(2).some(a => fs.existsSync(a))) {
       // If the old command structure was used, treat it as the build command
       // This includes support for things like `sushi` (no file path), `sushi [path-to-fsh-project]`,
--- a/src/app.ts
+++ b/src/app.ts
@@ -82,6 +82,14 @@ async function app() {
     });
 
   program
+    .command('init')
+    .description('initialize a SUSHI project')
+    .action(async function () {
+      await init();
+      process.exit(0);
+    });
+
+  program
     .command('update-dependencies')
     .description('update FHIR packages in project configuration')
     .argument('[path-to-fsh-project]')
@@ -96,14 +104,6 @@ async function app() {
       console.log('Additional information:');
       console.log('  [path-to-fsh-project]');
       console.log('    Default: "."');
-    });
-
-  program
-    .command('init')
-    .description('initialize a SUSHI project')
-    .action(async function () {
-      await init();
-      process.exit(0);
     });
 
   // Maintain backwards compatibility with prior SUSHI command structure
--- a/src/app.ts
+++ b/src/app.ts
@@ -43,7 +43,7 @@ import {
 } from './utils';
 
 const FSH_VERSION = '2.0.0';
-const SUPPORTED_COMMANDS = ['build', 'updateDependencies', 'init', '-h', '--help', 'help'];
+const SUPPORTED_COMMANDS = ['build', 'update-dependencies', 'init', '-h', '--help', 'help'];
 
 app().catch(e => {
   logger.error(`SUSHI encountered the following unexpected error: ${e.message}`);
@@ -82,8 +82,9 @@ async function app() {
     });
 
   program
-    .command('updateDependencies [path-to-fsh-project]')
+    .command('update-dependencies')
     .description('update FHIR packages in project configuration')
+    .argument('[path-to-fsh-project]')
     .action(async function (projectPath) {
       const input = ensureInputDir(projectPath);
       const config: Configuration = readConfig(input);
@@ -202,13 +203,6 @@ async function runBuild(input: string, program: OptionValues) {
 
   let tank: FSHTank;
   let config: Configuration;
-
-  // Update dependencies
-  if (program.updateDependencies) {
-    config = readConfig(originalInput);
-    await updateExternalDependencies(config);
-    process.exit(0);
-  }
 
   try {
     let rawFSH: RawFSH[];
--- a/src/app.ts
+++ b/src/app.ts
@@ -129,16 +129,12 @@ async function runUpdateDependencies(projectPath: string) {
 }
 
 async function runBuild(input: string, program: OptionValues, helpText: string) {
-  // NOTE: This is included to provide nicer handling for the previous CLI structure for building FSH projects
-  // Before doing anything else, we do our best to decide if the legacy SUSHI command structure was used
-  // (i.e. sushi . or sushi -d . or sushi -d or sushi)
-  const isLegacyBuildCommand =
-    !process.argv.includes('build') &&
-    (process.argv.slice(2).some(a => fs.existsSync(a)) ||
-      process.argv.slice(2).every(a => a.startsWith('-')));
-  if (!isLegacyBuildCommand) {
-    // This was probably an error (maybe a typo of a real command),
-    // not someone trying to use the old syntax, so exit
+  // NOTE: This is included to provide nicer handling for the previous CLI structure for building FSH projects.
+  // Check the first argument passed into sushi. If it is not "build", then this is a legacy build,
+  // in which case we should make sure that the first argument is a flag or a valid path.
+  const arg = process.argv[2];
+  if (arg != null && arg !== 'build' && !arg.startsWith('-') && !fs.existsSync(arg)) {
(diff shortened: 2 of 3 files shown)


src/app.ts
```diff
@@ -64,29 +68,50 @@ async function app() {
...
+    .command('init')
+    .description('initialize a SUSHI project')
+    .action(async function () {
+      await init();
+      process.exit(0);
+    });
+
+  // Maintain backwards compatability with prior SUSHI command structure by defaulting to build
+  if (process.argv.length >= 2 && !SUPPORTED_COMMANDS.includes(process.argv[2])) {
+    process.argv = [...process.argv.slice(0, 2), 'build', ...process.argv.slice(2)];
```
jafeltra, line 109: How thoroughly do we want support backwards compatibility? This won't support the old `sushi --init`.

jafeltra: Also, this block makes it so that if you make a typo in one of the commands (or try to use a made-up command), we try to run SUSHI. If we don't have this block, we get an error message that says `error: unknown command 'typo'`. I don't think there's much we can do about this, but it might be worth recognizing this quirk that comes up when trying to be backwards compatible.

On second thought, maybe this block could be smarter and check if the `process.argv.length == 2` or if `process.argv.length > 2 && fs.existsSync(process.argv[2])`. That might catch some of the typo cases, but wouldn't resolve the `sushi --init` backwards compatibility issue.

src/app.ts
```diff
@@ -43,18 +43,22 @@ import {
 } from './utils';
 
 const FSH_VERSION = '2.0.0';
+const SUPPORTED_COMMANDS = ['build', 'updateDependencies', 'init', '-h', '--help'];
```
jafeltra, line 46: This list needs to include `'help'` as well. This lets you do `sushi help build` to get the help text for a specific command.

src/app.ts
```diff
@@ -64,29 +68,50 @@ async function app() {
...
+    .action(async function () {
+      await init();
+      process.exit(0);
+    });
+
+  // Maintain backwards compatability with prior SUSHI command structure by defaulting to build
+  if (process.argv.length >= 2 && !SUPPORTED_COMMANDS.includes(process.argv[2])) {
+    process.argv = [...process.argv.slice(0, 2), 'build', ...process.argv.slice(2)];
   }
+  program.showHelpAfterError().parse(process.argv).opts();
```
jafeltra, line 111: `program.showHelpAfterError()` should be added higher up in this chain (it seems like) so that the help text shows up when an error occurs in the commands. For example, running `sushi build . --fake` won't show the help text when this is in its current place, but the help text does show up if it is moved up with the `name` and `version` options.

jafeltra (changes requested):
I found a few things that I think need to be updated. If the second reviewer agrees, I can plan to make the updates.

src/app.ts
jafeltra, line 109: I think this might simplify this line. I had to think about what was happening with the two spread operators on the `slice`s.

```suggestion
    process.argv = process.argv.splice(2, 0, 'build');
```

jafeltra: On third thought, we can probably make this even more robust by using the following conditional:

```
if (
  process.argv.length === 2 ||
  (process.argv.length > 2 &&
    !SUPPORTED_COMMANDS.includes(process.argv[2]) &&
    process.argv.slice(2).some(a => fs.existsSync(a)))
)
```

This better supports the fact that options could come before the path to the FSH project.

cmoesel: I like your third thought!

jafeltra: Fixed in [fe15249](https://github.com/FHIR/sushi/pull/1163/commits/fe15249a27e11d26faf9284e0d4a048f55960078).

src/app.ts
jafeltra: Updated in [fe15249](https://github.com/FHIR/sushi/pull/1163/commits/fe15249a27e11d26faf9284e0d4a048f55960078).

src/app.ts
jafeltra: Updated in [fe15249](https://github.com/FHIR/sushi/pull/1163/commits/fe15249a27e11d26faf9284e0d4a048f55960078).

src/app.ts
jafeltra: Yay! Added the third thought in [fe15249](https://github.com/FHIR/sushi/pull/1163/commits/fe15249a27e11d26faf9284e0d4a048f55960078).

src/app.ts
```diff
@@ -64,29 +69,64 @@ async function app() {
...
+      console.log('Additional information:');
+      console.log('  [path-to-fsh-project]');
+      console.log('    Default: "."');
+    });
 
-  if (program.init) {
-    await init();
-    process.exit(0);
+  program
+    .command('init')
```
cmoesel, line 101: When you run `sushi help`, it prints these out in the order we have them in the file.  So, right now, it shows:
```
Commands:
  build [options] [path-to-fsh-project]     export a FSH project into a FHIR IG
  updateDependencies [path-to-fsh-project]  update FHIR packages in project configuration
  init                                      initialize a SUSHI project
  help [command]                            display help for command
```

But (for whatever reason), I think I would prefer:
```
Commands:
  build [options] [path-to-fsh-project]      build a sushi project
  init                                      initialize a SUSHI project
  update-dependencies [path-to-fsh-project]  update FHIR packages in project configuration
  help [command]                            display help for command
```
Maybe because it is alphabetical (with `help` at the end)? Or maybe because `init` seems more useful than `update-dependencies`?  I dunno.  What do you think?

If you think you like `init` before `update-dependencies` too, then you'll need to move this block of code up.

src/app.ts
```diff
@@ -64,29 +69,64 @@ async function app() {
...
+  if (
+    process.argv.length === 2 ||
+    (process.argv.length > 2 && !SUPPORTED_COMMANDS.includes(process.argv[2]))
+  ) {
+    if (process.argv.some(a => a === '-i' || a === '--init')) {
+      // init was moved to a separate command, so log a message to indicate how to use it
+      console.log(
+        'The --init option has been moved to a separate command. Instead, run the following command: sushi init'
+      );
+      process.exit(0);
```
cmoesel, line 118: This should probably exit w/ an error code.
```suggestion
      process.exit(1);
```

src/app.ts
```diff
@@ -64,29 +69,64 @@ async function app() {
...
+    } else if (process.argv.length === 2 || process.argv.slice(2).some(a => fs.existsSync(a))) {
```
cmoesel, line 119: This, unfortunately, doesn't support running SUSHI w/ an argument but no path.  E.g., this no longer works:
```
sushi --debug
```

I think we can support this by adding another condition in the `else if` above, checking if all args after index 1 start with `-`.

cmoesel (changes requested):
Nice job, `/Julian?/`!  I've made a few small suggestions below for your consideration (@jafeltra).  Or I can fix them -- but I'd like your agreement first.

src/app.ts
```diff
@@ -43,18 +43,23 @@ import {
...
-
   const program = new Command()
     .name('sushi')
-    .usage('[path-to-fsh-project] [options]')
+    .version(getVersion(), '-v, --version', 'print SUSHI version')
+    .showHelpAfterError();
+
+  program
+    .command('build')
+    .description('export a FSH project into a FHIR IG')
```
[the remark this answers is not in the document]
jafeltra: I really like the symmetry (and the accuracy)!

src/app.ts
```diff
@@ -64,29 +69,64 @@ async function app() {
...
       console.log('');
       console.log('Additional information:');
       console.log('  [path-to-fsh-project]');
       console.log('    Default: "."');
       console.log('  -o, --out <out>');
       console.log('    Default: "fsh-generated"');
+    });
+
+  program
+    .command('updateDependencies [path-to-fsh-project]')
```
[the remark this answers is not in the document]
jafeltra: I like it! I felt like something felt off with it, but couldn't figure out what. But it's the camel case!

src/app.ts
jafeltra: I also like it, both for alphabetical order but also because I agree, `init` seems more useful. I'll move it up.

src/app.ts
jafeltra: Oh yep I forgot about that case.

src/app.ts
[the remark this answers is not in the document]
jafeltra: See: [b59d46f](https://github.com/FHIR/sushi/pull/1163/commits/b59d46f8f8f6f6c977e9545c45e6d2867ce08c92)

src/app.ts
jafeltra: See: [50da4cd](https://github.com/FHIR/sushi/pull/1163/commits/50da4cd43388038d724ac58ebcb4f622e3d533b2)

src/app.ts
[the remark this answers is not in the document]
jafeltra: I also found some old code we don't need for this option and removed that.

src/app.ts
jafeltra: So this is a fun one! I was not super loving how many conditional cases we were trying to handle, so I looked into it a little, talked to @mgramigna, and learned that there is an `isDefault` option on commands (see [Commands](https://github.com/tj/commander.js/#commands)). So I removed all the handling of processing the arguments and I removed the hardcoded list of options (yay!), and instead just used `{ isDefault: true }` on the `build` command. That should resolve the case you pointed out (e.g. `sushi --debug`) and it should more cleanly handle all the cases the if block was trying (and not really succeeding) to catch.

I also added in an option for `--init`, hid it from the `build` help text, and added in special case logic for logging the error message for that command. I think it works as we want it to, but I don't think this is a super necessary message because the IGP doesn't use that option, and this is a breaking change, so if it doesn't work or you don't like the extra bulk it adds, I'm happy to remove it.

jafeltra (changes requested):
This needs an equivalent README update. @cmoesel I can update that as well (assuming you agree).

src/app.ts
mgramigna: LGTM

jafeltra:
@cmoesel I updated the README in [37bd65a](https://github.com/FHIR/sushi/pull/1163/commits/37bd65a275929ee581feddf6d6b5d06ddadd66b2). I wanted to include something about `build`, but I'm not sure if this is too much help text to really be useful. I'm also going to put up a separate PR for the FSH School documentation shortly.

I think all your other comments should be addressed now as well.

jafeltra:
@cmoesel turns out I have another question - the alphabetical order/importance order of the commands got me thinking about the order of the options within the `build` command. Right now, when you use `sushi help build` or `sushi build --help`, the options are output in the order they're defined, which is

```
Options:
  -o, --out <out>       the path to the output folder
  -d, --debug           output extra debugging information
  -p, --preprocessed    output FSH produced by preprocessing steps
  -s, --snapshot        generate snapshot in Structure Definition output (default: false)
  -r, --require-latest  exit with error if this is not the latest version of SUSHI (default: false)
  -h, --help            display help for command
```

Should these be in alphabetical order? Or is `--out` so useful that it's purposefully at the top and `--require-latest` so rare that it's purposefully at the bottom? I have a feeling these are just in the order they were defined in.

Alphabetical order (with `help` at the end) would look like this:

```
Options:
  -d, --debug           output extra debugging information
  -o, --out <out>       the path to the output folder
  -p, --preprocessed    output FSH produced by preprocessing steps
  -r, --require-latest  exit with error if this is not the latest version of SUSHI (default: false)
  -s, --snapshot        generate snapshot in Structure Definition output (default: false)
  -h, --help            display help for command
```

cmoesel:
@jafeltra - I think I kind of like alphabeticalized -- because the other order seems kind of arbitrary anyway.  I don't think `--out` is really all that much more useful than the others.

src/app.ts
jafeltra: Ugh so I did realize that since this just automatically runs `build`, the check for if something is an actual directory is no longer included, so if you ran `sushi inti` (or some other typo), it will try to run `build` with a directory `inti`. I'm not sure if this is worth going back to the old approach with the multiple conditionals, but I feel like it's worth pointing out before decide if we definitely want to keep the `isDefault` approach.

cmoesel: I don't know if this adds more trouble than its worth, but could you make it so that one of the first things `build` does is check `argv[1]`?  If it's `build` then the user invoked it directly, so let it do its thing.  If it is not `build` and it does not start with `-`, then check if it points to an existing file.  If it doesn't, print SUSHI usage instructions?

jafeltra: I think that makes sense, and I've attempted it. I'm curious what you think about it. I had originally been excited about this, but I want to make sure my changes are actually solving more problems than the original approach.

The new commit with the handling in `build` is [45fd3ec](https://github.com/FHIR/sushi/pull/1163/commits/45fd3ec80f313b16b64900b8891681d529c48107). This commit plus [9e1ecc6](https://github.com/FHIR/sushi/pull/1163/commits/9e1ecc6d0068d3b6125f8829267c5997185eeed1) rework the `build` command to hopefully allow the old way of running SUSHI while still scoping it enough to provide useful information when people are trying to use the new commands but doing something wrong. Let me know what you think.

jafeltra:
A few more updates for this PR (that aren't specifically called out in conversations above):

- [6161823](https://github.com/FHIR/sushi/pull/1163/commits/616182348be11157fcfcdc9049bc37bdd62eb02d) adds better error handling for catching unexpected errors that get thrown in the commands. I don't think our previous error handling was being reached anymore because it was catching on `app()`, but `app` just sets up the CLI now.
- [cfd9c80](https://github.com/FHIR/sushi/pull/1163/commits/cfd9c8069c0ee0fe83915b53dded640a1ed10421) alphabetizes the `build` command options.

Also, a companion PR for FSH School documentation can be found [here](https://github.com/FSHSchool/site/pull/72).

src/app.ts
```diff
@@ -75,18 +80,69 @@ async function app() {
...
+  const isLegacyBuildCommand =
+    !process.argv.includes('build') &&
+    (process.argv.slice(2).some(a => fs.existsSync(a)) ||
+      process.argv.slice(2).every(a => a.startsWith('-')));
+  if (!isLegacyBuildCommand) {
+    // This was probably an error (maybe a typo of a real command),
+    // not someone trying to use the old syntax, so exit
+    console.log(helpText);
+    process.exit(1);
   }
```
cmoesel, line 144: Unfortunately, there are a few issues with this:
1. I think this would still fail for a command like `sushi -o /my/output/dir` when `/my/output/dir` doesn't exist (even though SUSHI can create it).
2. This would definitely fail if someone invoked SUSHI the recommended way (`sushi build .`) because `isLegacyBuildCommand` would be false (and rightly so), which would then cause us to exit with an error due to the condition on line 139.  Oops!
3. I think that the command _has_ to be the first argument to SUSHI. If I'm correct about that, then I think we can simplify this?

```suggestion
    // Check the first argument passed into sushi. If it is not "build", then this is a legacy build,
  // in which case we should make sure that the first argument is a flag or a valid path.
  // NOTE: The index of the argument of interest differs depending on how sushi is invoked, e.g.,
  // argv[1] for `sushi build /some/dir`, but argv[2] for `ts-node src/app.ts build /some/dir`.
  const arg = /(ts-)?node(\.[a-zA-Z]+)?/.test(process.argv[0]) ? process.argv[2] : process.argv[1];
  if (arg != null && arg !== 'build' && !arg.startsWith('-') && !fs.existsSync(arg)) {
    // It's not a flag or a path, so it's probably a typo of an existing command
    console.log(helpText);
    process.exit(1);
  }
```
(Although I guess you might argue it's not as simple now that we have to check for `node`/`ts-node`.  What do you think?)

cmoesel (changes requested):
Thanks for picking this up! I think there's an issue or two in the new logic to try to account for the legacy command.

.github/CODEOWNERS
```diff
@@ -1 +1 @@
-*   @jafeltra @cmoesel @julianxcarter @mint-thompson @guhanthuran
\ No newline at end of file
+*   @jafeltra @cmoesel @mint-thompson @guhanthuran
```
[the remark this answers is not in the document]
julianxcarter: 😭😭😭

src/app.ts
jafeltra: Some additional thoughts, but I think this is a good change!

1. Ah, I had forgotten we actually create it, even if it doesn't exist.
2. I thought `sushi build .` would work because the first part of `isLegacyBuildCommand` check if `'build` was not included. That still doesn't fix 1. though.
3. I think your simplified approach makes sense! But now I'm wondering if we can simplify it further. Is there ever a case where we do want to use `process.argv[1]`? When I tried to `npm pack` and use the packed version, the first argument is still a `node` argument, so I think it might be safe to just always have `const arg = process.argv[2]`.

cmoesel: > I thought `sushi build .` would work because the first part of `isLegacyBuildCommand` check if `build` was not included. That still doesn't fix 1. though.

Right, the first part of `isLegacyBuildCommand` is `!process.argv.includes('build')`.  So, if the args include `build`, then that evaluates to `false` -- and since it's `&&`ed w/ the rest, `isLegacyBuildCommand` becomes `false`.  Then line 139 is `if (!isLegacyBuildCommand)` -- so when `build` is included in the args, that becomes `if (!false)`, which is basically `if (true)`.  And what do we do `if (true)`? We `process.exit(1)`.

> When I tried to npm pack and use the packed version, the first argument is still a node argument, so I think it might be safe to just always have const arg = process.argv[2].

Oops.  You're probably right.  I likely made a bad assumption there!

cmoesel (approved):
This looks good! I like it! Thank you @julianxcarter and @jafeltra!

src/app.ts
```diff
@@ -44,29 +44,34 @@ import {
 
 const FSH_VERSION = '2.0.0';
 
-app().catch(e => {
+function logUnexpectedError(e: Error) {
   logger.error(`SUSHI encountered the following unexpected error: ${e.message}`);
   process.exit(1);
-});
+}
+app().catch(logUnexpectedError);
```
jafeltra, line 51: @jafeltra - super nitpick, can you put a blank line after the function?

jafeltra: No problem @jafeltra! See [19c7aa7](https://github.com/FHIR/sushi/pull/1163/commits/19c7aa78bf985da030dea19726aebd67edbbd399).

cmoesel (approved):
I don't know about the last newline that @jafeltra just added... It seems suspicious.  But... I guess I'll let it slide.  If anyone finds a bug though, it's her fault for adding that newline!

PR #1164, opened 2022-10-29
Update the IG parameters link

qligier:
The link https://confluence.hl7.org/display/FHIR/Implementation+Guide+Parameters now redirects to http://build.fhir.org/ig/FHIR/fhir-tools-ig/branches/master/CodeSystem-ig-parameters.html.

### DIFF
--- a/test/utils/fixtures/init-config/default-config.yaml
+++ b/test/utils/fixtures/init-config/default-config.yaml
@@ -56,7 +56,7 @@ publisher:
 # The parameters property represents IG.definition.parameter. Rather
 # than a list of code/value pairs (as in the ImplementationGuide
 # resource), the code is the YAML key. If a parameter allows repeating
-# values, the value in the YAML should be a sequence/array. 
+# values, the value in the YAML should be a sequence/array.
 # For parameters defined by core FHIR see:
 # http://build.fhir.org/codesystem-guide-parameter-code.html
 # For parameters defined by the FHIR Tools IG see:
--- a/test/utils/fixtures/init-config/default-config.yaml
+++ b/test/utils/fixtures/init-config/default-config.yaml
@@ -56,9 +56,11 @@ publisher:
 # The parameters property represents IG.definition.parameter. Rather
 # than a list of code/value pairs (as in the ImplementationGuide
 # resource), the code is the YAML key. If a parameter allows repeating
-# values, the value in the YAML should be a sequence/array. For a
-# partial list of allowed parameters see:
-# https://confluence.hl7.org/display/FHIR/Implementation+Guide+Parameters
+# values, the value in the YAML should be a sequence/array. 
+# For parameters defined by core FHIR see:
+# http://build.fhir.org/codesystem-guide-parameter-code.html
+# For parameters defined by the FHIR Tools IG see:
+# http://build.fhir.org/ig/FHIR/fhir-tools-ig/branches/master/CodeSystem-ig-parameters.html
 #
 # parameters:
 #   excludettl: true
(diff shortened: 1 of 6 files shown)


cmoesel (commented):
Thanks for this!  It's something I keep meaning to do and forgetting.

The truth is that while the page you linked represents _some_ of the supported parameters, it does not represent _all_ of the supported parameters (see [this comment from Grahame](https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/Unknown.20FHIRVersion.20code.20'5.2E0.2E0-ballot'/near/298697304)).

The R5 spec itself defines a core set of supported parameters here:
http://build.fhir.org/codesystem-guide-parameter-code.html

Would you mind updating this PR so it references both locations where allowed parameters are defined?

*Additional Details / Clarification (If You're Interested):*

The [R5 ImplementationGuide resource](http://build.fhir.org/implementationguide.html) binds [definition.parameter.code](http://build.fhir.org/implementationguide-definitions.html#ImplementationGuide.definition.parameter.code) to the [GuideParameterCode](http://build.fhir.org/valueset-guide-parameter-code.html) value set (comprised of the [guide-parameter-code](http://build.fhir.org/codesystem-guide-parameter-code.html) codes).  Since that binding is _extensible_, however, it allows for additional codes not in the value set.  So the [ig-parameters](http://build.fhir.org/ig/FHIR/fhir-tools-ig/branches/master/CodeSystem-ig-parameters.html) code system provides additional supported codes that aren't in the bound value set.

qligier:
Ok, the wiki content made me think it was a simple redirect, but it is seemingly much more complex.

cmoesel (commented):
Ha.  Yeah.  The reason for two separate lists is kind of complex -- but the only thing you really need to know is that there are two lists, and you need to look at both of them to get the _full_ list of supported parameters.

qligier:
Thank you for having updated the wiki, this PR is not needed anymore.

cmoesel:
I like the idea of changing the links (to remove one step).  I just think that we need to list both links. But if you don't want to do that, that's OK.  I can ask someone on my team to do it.

qligier:
Okay! I've no problem doing it; having a single link to an updated wiki page that explains the situation seemed better than updating 6 times the same comment, but if that situation is stable, I guess it's fine documenting it in the source code. I'll update the PR.

cmoesel:
Thanks! I think it's nice if people don't have to click through an intermediate link just to get to the other links.  And I think the situation is stable, so it should not change again any time soon!

cmoesel:
(Oops.  I clicked the wrong button and marked this "Ready for Review".  That was a mistake and I've reverted it back to "Draft". Sorry for the confusion!

qligier:
Updated as discussed.

test/utils/fixtures/init-config/default-config.yaml
```diff
@@ -56,9 +56,11 @@ publisher:
 # The parameters property represents IG.definition.parameter. Rather
 # than a list of code/value pairs (as in the ImplementationGuide
 # resource), the code is the YAML key. If a parameter allows repeating
-# values, the value in the YAML should be a sequence/array. For a
-# partial list of allowed parameters see:
-# https://confluence.hl7.org/display/FHIR/Implementation+Guide+Parameters
+# values, the value in the YAML should be a sequence/array.
```
cmoesel, line 59: The test for the default config file is failing because the fixture has an extra space after line 59. Removing the space should fix the test.
```suggestion
# values, the value in the YAML should be a sequence/array.
```

cmoesel (changes requested):
This is great!

Unfortunately, there is one failing test. I believe I found the issue and suggested the fix below.

test/utils/fixtures/init-config/default-config.yaml
qligier: Indeed 😳

jafeltra (approved):
I looked through the updates as well, and they look good to me!

PR #1167, opened 2022-11-07
Support fishing for profiles with a version

### LINKED ISSUE #1156: SUSHI cannot unfold elements that have a canonical URL of a profile that includes a version
When SUSHI tries to unfold elements, it tries to use a `profile` if one is specified in order to unfold the element. However, when the canonical URL for the profile includes a `|version` at the end of it, SUSHI cannot successfully fish for the definition. It should support this case. There are various places in SUSHI where we parse off the version before trying to fish for a definition, so we will likely need to do that here as well.

In addition to properly supporting URLs with versions, SUSHI should probably log a better error message when it cannot find the definition it i looking for and fallback to using the `type`, rather than the `profile`, if it cannot find the profile.

Zulip thread where this was reported: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/No.20element.20found.20at.20path

jafeltra:
Fixes #1156.

This PR updates how we unfold elements when we are trying to unfold based on a profile. If an element has a `type` with one `profile` defined, we try to fish for the profile when unfolding. However, if that profile's URL includes a `|version` at the end, we would not successfully fish for the profile. For example, the original bug was trying to find a profile like this:

```json
{
  "id": "Patient.identifier:reisepassnummer",
  ... some stuff that doesn't really matter here ...
  "type": [
    {
      "code": "Identifier",
      "profile": [
        "https://fhir.kbv.de/StructureDefinition/KBV_PR_Base_Identifier_Reisepassnummer|1.3.0"
      ]
    }
  ]
}
```

This PR supports ignoring that `version` (`1.3.0`) when fishing for the profile, which is similar to what we do in other places when canonicals have version.

This PR also supports falling back to the `type.code` if the profile cannot be found, even after removing the version.

The tests demonstrate these two cases as well.

### DIFF
--- a/src/fhirtypes/ElementDefinition.ts
+++ b/src/fhirtypes/ElementDefinition.ts
@@ -2278,13 +2278,25 @@ export class ElementDefinition {
       if (newElements.length === 0) {
         // If we have exactly one profile to use, use that, otherwise use the code
         const type = profileToUse ?? this.type[0].code;
-        const json = fisher.fishForFHIR(
-          type,
+        // There could possibly be a |version appended to the type, so don't include it while fishing
+        const typeWithoutVersion = type.split('|', 2)[0];
+        let json = fisher.fishForFHIR(
+          typeWithoutVersion,
           Type.Resource,
           Type.Type,
           Type.Profile,
           Type.Extension
         );
+        if (!json && profileToUse) {
+          // If we tried to fish based on a profile and didn't find anything, fall back to the type
+          json = fisher.fishForFHIR(
+            this.type[0].code,
+            Type.Resource,
+            Type.Type,
+            Type.Profile,
+            Type.Extension
+          );
+        }
         if (json) {
           const def = StructureDefinition.fromJSON(json);
           if (def.inProgress) {
--- a/test/fhirtypes/ElementDefinition.test.ts
+++ b/test/fhirtypes/ElementDefinition.test.ts
@@ -733,6 +733,32 @@ describe('ElementDefinition', () => {
       );
       expect(fooSliceExtensionTest.calculateDiff().type).toEqual(extensionSlice.type);
     });
+
+    it('should unfold an element with a profile that includes a version', () => {
+      const referenceRange = observation.elements.find(
+        e => e.id === 'Observation.referenceRange.low'
+      ) as ElementDefinition; // The element will exist based on the test fixtures
+
+      // Set type with a profile that has a version number
+      referenceRange.type[0].profile = [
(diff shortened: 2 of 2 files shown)


cmoesel (commented):
This looks good. It brings up a few questions, though:
* If we strip off the version and fish up a profile, maybe we should log a warning if the fished up profile has a different version than the one received.
* If we fail to find a profile and we fall back to a base type instead, maybe we should log a warning to indicate that happened as well.

You noted that you're following a similar approach as elsewhere in SUSHI -- so I'm guessing that maybe you aren't warning here because we don't warn in those other cases?

Anyway, what do you think?

jafeltra:
I like the idea of both of those warnings. I can add them!

> You noted that you're following a similar approach as elsewhere in SUSHI -- so I'm guessing that maybe you aren't warning here because we don't warn in those other cases?

It looks like there actually is one place where we do log a warning if the version we fish up is different ([here](https://github.com/FHIR/sushi/blob/master/src/fhirtypes/ElementDefinition.ts#L1207-L1216)). However, there are also some place (the ones I saw when I implemented this), where it doesn't look like we log a similar version mismatch warning even though maybe it would be applicable. Examples that I found are [here](https://github.com/FHIR/sushi/blob/master/src/fhirtypes/common.ts#L443-L446), [here](https://github.com/FHIR/sushi/blob/master/src/export/ValueSetExporter.ts#L57-L64), [here](https://github.com/FHIR/sushi/blob/master/src/fhirtypes/ElementDefinition.ts#L1103-L1106) and [here](https://github.com/FHIR/sushi/blob/master/src/fhirtypes/ElementDefinition.ts#L1823-L1827) although these are more concerned with lineage so may be slightly different, and [here](https://github.com/FHIR/sushi/blob/master/src/fhirtypes/StructureDefinition.ts#L860-L868) although I think this case might also be different.

So after more thoroughly looking through those examples, I think they might all be different enough cases from this current one that it is okay we are not logging a warning in those cases, but it does make sense to log a warning if the version mismatch (or the type fallback) occurs in this case.

jafeltra:
> * If we strip off the version and fish up a profile, maybe we should log a warning if the fished up profile has a different version than the one received.
> 
> * If we fail to find a profile and we fall back to a base type instead, maybe we should log a warning to indicate that happened as well.

[c055805](https://github.com/FHIR/sushi/pull/1167/commits/c0558051339f1dca96706e7419020443bb0a3449) adds both warnings.

guhanthuran:
Ran regression and ran the tests - everything looks green to me.

guhanthuran (approved):

PR #1175, opened 2022-11-22
Find CodeSystems from packages by id, name, and URL

### LINKED ISSUE #1172: SUSHI doesn't resolve code systems by name when processing codes
In US Core, there is a code system with the `name` [USCoreTagsCodeSystem](http://www.hl7.org/fhir/us/core/CodeSystem-us-core-tags.html) which contains the code `sdoh`.

As such, you should be able to use that name in a FSH code, like `USCoreTagsCodeSystem#sdoh`.  When you do, however, SUSHI complains:
> Resolved value "USCoreTagsCodeSystem" is not a valid URI.

First reported on Zulip [here](https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Aliases.2C.20terminology.2C.20and.20packages/near/310083886).

Simple FSH Online example [here](https://fshschool.org/FSHOnline/#/share/3tqBFHX).

jafeltra:
Fixes #1172 

This PR supports using CodeSystem `name` or `id`, as well as `url`, when assigning codes from CodeSystems that are either in the FHIR version or a dependency.

The tests are added in the first commit and they are fixed with the second commit. Using `name` or `id` worked when used with CodeSystems in on ValueSetComponentRules, so I took a similar approach as the one we used [there](https://github.com/FHIR/sushi/blob/master/src/export/ValueSetExporter.ts#L58-L62). Basically, fishing for metadata will correctly find the CodeSystem whether it is in the `fisher.tank` or `fisher.fhir` (which includes any FHIR resources and any dependency resources). If we find a CodeSystem, we want to use the `url` in the `system` on the rule.

### DIFF
--- a/src/export/Package.ts
+++ b/src/export/Package.ts
@@ -143,9 +143,8 @@ export class Package implements Fishable {
       return metadata;
     } else if (
       // If nothing is returned, perhaps the Package itself is being referenced
-      item === this.config.packageId ||
-      item === this.config.name ||
-      item === this.config.id
+      item != null &&
+      (item === this.config.packageId || item === this.config.name || item === this.config.id)
     ) {
       const metadata: Metadata = {
         id: this.config.packageId || this.config.id,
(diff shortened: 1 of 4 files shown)


cmoesel (approved):
I fixed a couple of small typos in the test names, but otherwise it looks good!

PR #1177, opened 2022-11-24
Introduce an ENV variable to force fancy characters use

qligier:
Fancy characters may be properly supported in a lot of terminal, including through the IG Publisher. This commits checks the presence of the 'FORCE_FANCY_CHARACTERS' environment variable and prevents replacing fancy characters if it's set.

Example in the IG Publisher:
fancy_chars

Not sure where it should be documented. The env variable FORCE_COLOR ([chalk](https://github.com/chalk/chalk/blob/main/source/vendor/supports-color/index.js#L32)) should also be documented, since both variables will probably be used together.


cmoesel:
Hi @qligier.  Thanks for the contribution!

We implemented the logic to disable fancy characters because the IG Publisher did not display them correctly. If that has been fixed (in the publisher or later versions of the JDK), then maybe we can just remove that logic entirely.

Do you know when it does and does not work in the publisher?

cmoesel (approved):
OK, yes, this works for me. After doing some investigation, we found that there are still cases in which we get undesired behavior on Windows -- so I think it makes sense to keep the default behavior as-is, but allow for the env var escape hatch!

Thanks, @qligier! Sorry it has taken us so long on this!

qligier:
Hi @cmoesel!
No problem at all! I am unable to reproduce that case, even with the windows command prompt, and the legacy mode enabled, the flag seems the best solution for now.

src/app.ts
```diff
@@ -315,13 +315,17 @@ function supportsFancyCharacters(): boolean {
   // There is no sure-fire way, but we know that most problems are when running in the IG Publisher,
   // so try to detect that situation (which is still actually pretty tricky and not guaranteed).
 
-  // 1. Many JVM will insert an environment variable indicating the main Java class being run.
+  // 1. If the user wants the fancy characters, allow them
+  if (Object.keys(process.env).some(k => k === 'FORCE_FANCY_CHARACTERS')) {
+    return true;
+  }
```
cmoesel, line 321: This returns true as long as the `FORCE_FANCY_CHARACTERS` environment variable exists. This means that even if someone sets the value to `false` or `0`, it would still force fancy characters. I think it would be better to explicitly look for the value to be `true` or `1`. Does that seem reasonable?

cmoesel (changes requested):
In reviewing this once more, I think there is one small opportunity for improvement.

src/app.ts
qligier: Indeed, I've added a check on `!!process.env.FORCE_FANCY_CHARACTERS`.

src/app.ts
```diff
@@ -315,13 +315,20 @@ function supportsFancyCharacters(): boolean {
   // There is no sure-fire way, but we know that most problems are when running in the IG Publisher,
   // so try to detect that situation (which is still actually pretty tricky and not guaranteed).
 
-  // 1. Many JVM will insert an environment variable indicating the main Java class being run.
+  // 1. If the user wants the fancy characters, allow them
+  if (
+    Object.keys(process.env).some(k => k === 'FORCE_FANCY_CHARACTERS') &&
+    !!process.env.FORCE_FANCY_CHARACTERS
```
cmoesel, line 321: I don't think this quite works as intended, as `!!'false'` is `true` (because a non-empty string is truthy in JS). Perhaps you should use `/(true|1)/i.test(process.env.FORCE_FANCY_CHARACTERS)` to do a case-insensitive test against true and 1.

cmoesel (changes requested):

src/app.ts
qligier: Oh yeah, of course, environment variables are only strings. Sorry :)

cmoesel (approved):
Looks good to me!

jafeltra (approved):
This change looks good to me as well! Thanks!

PR #1179, opened 2022-11-28
Include default type slicing in differential

### LINKED ISSUE #1160: SUSHI 3.0.0 does not put type slicing information in the differential
SUSHI 3.0.0 no longer does type-slice path-renaming.  E.g., if you do:
```
* valueQuantity MS
```
SUSHI 2.x creates an element in the differential w/ id `Observation.valueQuantity`.  SUSHI 3.0.0 Beta 1 creates an element in the differential with id `Observation.value[x]:valueQuantity`.

In the latter case (SUSHI 2.x approach), the use of `Observation.valueQuantity` _implied_ type slicing on `value[x]`.  When using the long form (SUSHI 3.x approach), however, I'm not sure that the type slicing is necessarily implied anymore. SUSHI should probably put the type slicing into the differential (assuming that it's not already present in the parent).

This came up in this Zulip thread: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Type.20Slices.20on.20Choices.20w.2F.20a.20Single.20Type/near/304048276

mint-thompson:
Fixes #1160 and completes task [CIMPL-1033](https://standardhealthrecord.atlassian.net/browse/CIMPL-1033).

If a choice element is sliced, include the slicing in the differential, even if the slicing is the default choice-type slicing and this is the only difference on the element.

### DIFF
--- a/src/fhirtypes/StructureDefinition.ts
+++ b/src/fhirtypes/StructureDefinition.ts
@@ -426,22 +426,7 @@ export class StructureDefinition {
       // we need to make sure the root element is included in the differential.
       if (e.hasDiff() || (this.derivation === 'specialization' && idx === 0)) {
         const diff = e.calculateDiff().toJSON();
-        const isTypeSlicingChoiceDiff =
-          diff.id.endsWith('[x]') &&
-          Object.keys(diff).length === 3 &&
-          diff.id &&
-          diff.path &&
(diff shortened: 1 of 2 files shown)


jafeltra (approved):
Makes sense to me!

cmoesel (approved):
Yes, looks good! Thank you!

PR #1182, opened 2022-12-02
Keep parent's fixedUri on Extension.url

### LINKED ISSUE #1178: Declaring Parent of Extension causes conflict in SD
In FSH, you can declare that an extension has a parent:
```
Extension: MyChildExtension
Parent: MyParentExtension
```
This will set the child extension's `baseDefinition` to the parent's canonical URL (indicating that it constrains the parent). It also fixes the child extensions `Extension.url` to the canonical URL of the child extension.  This is a problem because the parent has already fixed `Extension.url` to the parent's canonical URL -- so technically the child cannot fix it to a different value. This means that the child extension definition generated by SUSHI is actually invalid.

There are two possible approaches to fixing this:
1. Use the parent extension as the basis of the child extension, but set `baseDefinition` to the FHIR core `Extension`. This is essentially like cloning the parent and then modifying the clone for the child. This means that there is no _formal_ relationship between the parent and child, and all of the parents constraints need to be repeated in the child. The parent and child are _technically_ two completely different extensions, so the child _cannot_ be used in place of the parent (since it has a different fixed URL).
2. Formally set the child's `baseDefinition` to the parent's canonical URL and _do not_ fix `Extension.url` to the child extension's canonical URL. This is essentially the same as using `Profile:` and `Parent:` instead of `Extension:` and `Parent:`. In this case, the child is truly a constraint of the parent and uses the same fixed URL as the parent; and as such, can be used in place of the parent wherever the parent is expected.

The [FSH specification for Extension](https://hl7.org/fhir/uv/shorthand/reference.html#defining-extensions) is not explicit about which approach is correct, but I think the wording implies a constraint relationship, which would suggest that option 2 above is correct. The only downside of implementing 2 above is that while 2 already has an easy workaround (use `Profile:`), option 1 does not have any workaround -- so there would no longer be an easy way to accomplish option 1 if an author wanted that behavior.

mint-thompson:
Fixes #1178 and completes task [CIMPL-1042](https://standardhealthrecord.atlassian.net/browse/CIMPL-1042).

When an Extension's parent is another Extension (as opposed to the base Extension type), the Extension.url element will have a fixedUri set on the parent. To be valid FHIR, the child Extension should keep that same fixedUri.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-*   @jafeltra @cmoesel @julianxcarter @mint-thompson @guhanthuran
+*   @jafeltra @cmoesel @mint-thompson
--- a/src/export/StructureDefinitionExporter.ts
+++ b/src/export/StructureDefinitionExporter.ts
@@ -362,9 +362,11 @@ export class StructureDefinitionExporter implements Fishable {
       // Keep context, assuming context is still valid for child extensions.
       // Keep contextInvariant, assuming context is still valid for child extensions.
 
-      // Automatically set url.fixedUri on Extensions
+      // Automatically set url.fixedUri on Extensions unless it is already set
       const url = structDef.findElement('Extension.url');
-      url.fixedUri = structDef.url;
+      if (url.fixedUri == null) {
+        url.fixedUri = structDef.url;
+      }
       if (structDef.context == null) {
         // Set context to everything by default, but users can override w/ rules, e.g.
         // ^context[0].type = #element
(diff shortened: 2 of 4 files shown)


jafeltra (approved):
This all makes sense to me!

cmoesel (approved):
This looks good! I ran a regression to see how it would affect existing IGs.  Only three IGs were affected; all PACIO:
* fhir-pacio-congitive-status
* fhir-pacio-functional-status
* fhir-pacio-pfe

In all of these, the extensions w/ an IG-specific URL for `event-location` were affected, as their fixed URL changed to the URL of the extension they are based on: `http://hl7.org/fhir/StructureDefinition/event-location`.

PR #1186, opened 2022-12-08
Automatically load hl7.fhir.uv.tools dependency

### LINKED ISSUE #1166: SUSHI should always put hl7.fhir.uv.tools in scope
According to Grahame Grieve, the IG Tools package is considered a core part of building IGs now and should be loaded automatically (without needing to declare it as a dependency).

So... SUSHI should also automatically load the IG Tools since authors can use extensions defined by it.  It sounds like SUSHI should always load `current` -- but we may want to revisit how we do that to try to reduce unneeded re-downloads (when nothing has changed).  We may also want to be fairly flexible when it fails to download (for whatever reason), as not every IG actually needs it and we don't want it to become a blocker for people who have firewalls or otherwise cannot do dynamic downloads of dependencies.

mint-thompson:
Fixes #1166 and completes task [CIMPL-1036](https://standardhealthrecord.atlassian.net/browse/CIMPL-1036).

The current version of this package will always be loaded, even if it is not configured as a dependency. If it is configured as a dependency, use the configured version. When it fails to load, log a warning instead of an error if it was not a configured dependency.

My strategy for implementing this was to consider loading dependencies in two groups: automatic dependencies and configured dependencies. Hopefully, this makes it straightforward to add more automatic dependencies later. There are changes to a lot of the tests for `loadExternalDependencies`, since there's now an additional package that shows up every time. A lot of those tests are really testing `loadConfiguredDependencies`, though, so it may be worthwhile to adjust the tests accordingly. Please let me know if you think that would be useful.

### DIFF
--- a/src/utils/Processing.ts
+++ b/src/utils/Processing.ts
@@ -18,6 +18,7 @@ import {
 import { Package } from '../export';
 import { Configuration } from '../fshtypes';
 import { axiosGet } from './axiosUtils';
+import { ImplementationGuideDependsOn } from '../fhirtypes';
 
 const EXT_PKG_TO_FHIR_PKG_MAP: { [key: string]: string } = {
   'hl7.fhir.extensions.r2': 'hl7.fhir.r2.core#1.0.2',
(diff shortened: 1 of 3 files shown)


cmoesel (changes requested):
This looks good and works well! I tested it locally w/ an existing cache as well as a cleared cache. I also tweaked the package date in the cached package and saw that it was detected as being stale. That's good!

I've only asked for one small change (to have the warning indicate it is an automatic dependency). The changes in the test file are all optional.

jafeltra (approved):
These changes make sense to me and work as expected!

cmoesel (approved):
Hurray!

PR #1189, opened 2022-12-12
Support IG R5 properties in R4 IGs

### LINKED ISSUE #1171: SUSHI should include R5 IG properties as extensions in R4 IGs
When SUSHI updated to support the new FHIR R5 ImplementationGuide resource, we added support for the new R5 ImplementationGuide resources and include those properties only in R5 IGs. This was originally based on guidance [here](https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/New.20IG.20Publisher/near/298399557).

However, we should now support including any _new_ properties that are defined in R5's ImplementationGuide resource as extensions in R4 IGs, based on guidance [here](https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Dependency.20reason/near/309636002). The extension should follow the pattern defined in the documentation for extensions for converting between versions [here](http://hl7.org/fhir/2022Sep/versions.html#extensions).

For example, `IG.dependsOn.reason` was added in R5. Currently, SUSHI will not include `reason` in R4 IGs. However, it should instead include an extension with `url` of `http://hl7.org/fhir/5.0/StructureDefinition/extension-ImplementationGuide.dependsOn.reason` and `valueMarkdown` of the specified reason in `sushi-config.yaml`.

Support for adding extensions should only be added for properties that are _new_ to R5 ImplementationGuides. Any property that was moved or renamed does _not_ need to have an extension defined for it.

jafeltra:
Fixes #1171.

This PR takes any properties that were in the R5 ImplementationGuide resource and adds them as extensions to R4 ImplementationGuide resources that SUSHI creates. To do this, I followed the documentation for converting between FHIR versions [here](http://hl7.org/fhir/2022Sep/versions.html#extensions). The properties that were changed in R5 are left as is in the R4 IG, and only new properties are added as extensions. The properties that are new can be seen in the function that is updated in the SUSHI code, as well as seen in the ImplementationGuide section of diff documentation [here](http://hl7.org/fhir/2022Sep/diff.html).

Note: Although I believe the extensions are created correctly, the IG Publisher does not recognize the extension URLs and produces errors if you create an R4 IG with these extensions. So if you add the new properties in your `sushi-config.yaml`, run SUSHI, and run the resulting IG through the IG Publisher, you will get errors in the QA report. This was brought up on Zulip [here](https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/Converting.20resources.20between.20FHIR.20versions) and seems to be an issue with the validator and the IGP. Until that is resolved, we don't want to merge this PR, but it can be reviewed.


src/ig/IGExporter.ts
```diff
@@ -1427,6 +1427,76 @@ export class IGExporter {
           dependency.reason = configDependency.reason;
         }
       });
+    } else {
```
mint-thompson, line 1430: This PR introduces this `else` block, which means that the whole `updateForR5` function is now a single if-else at the top level. I feel like it might be simpler if instead, the contents of each block were split into their own functions, and the if-else moved into `export`, which is where `updateForR5` is called. Something like this:
```ts
// in export()
if (!isR4(this.config.fhirVersion)) {
  this.updateForR5(); // this would be the if-block contents
} else {
  this.addR5Extensions(); // this would be the else-block contents
}
```
Do you think this makes sense?

mint-thompson (changes requested):
This looks good! I think you have all the logic implemented correctly, so I just have a few requests about the code organization and tests.

test/ig/IGExporter.IG.test.ts
```diff
@@ -3119,4 +3119,263 @@ describe('IGExporter', () => {
...
+      const instanceDef3 = InstanceDefinition.fromJSON(
+        fs.readJSONSync(path.join(fixtures, 'examples', 'Patient-example.json'))
+      );
+      instanceDef3._instanceMeta.usage = 'Example';
+      pkg.instances.push(instanceDef3);
+
+      const exporter = new IGExporter(pkg, defs, fixtures);
+      // No need to regenerate the IG on every test -- generate it once and inspect what you
+      // need to in the tests
+      exporter.export(tempOut);
```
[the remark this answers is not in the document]
jafeltra: Ah good call. I'll update this to a `beforeAll`, although the content may change based on your other comment.

test/ig/IGExporter.IG.test.ts
```diff
@@ -3119,4 +3119,263 @@ describe('IGExporter', () => {
...
+      ];
+
+      const pkg = new Package(r4WithR5propsConfig);
+
+      // Example: Patient/patient-example
+      const instanceDef3 = InstanceDefinition.fromJSON(
+        fs.readJSONSync(path.join(fixtures, 'examples', 'Patient-example.json'))
+      );
+      instanceDef3._instanceMeta.usage = 'Example';
+      pkg.instances.push(instanceDef3);
```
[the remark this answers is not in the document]
jafeltra: Whoops I thought I caught these renames, but I guess it slipped through. Agreed and I'll update.

src/ig/IGExporter.ts
jafeltra: Ah I like that approach much better! I'll update.

test/ig/IGExporter.IG.test.ts
```diff
@@ -3119,4 +3119,263 @@ describe('IGExporter', () => {
       ]);
     });
   });
+
+  describe('#r5-properties-on-r4-igs', () => {
```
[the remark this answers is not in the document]
jafeltra: I agree that this is good to test, and I was considering adding a test for each property to ensure that there are not side effects in the output if the property isn't added, but as I thought about it, I think this might be covered by some existing tests. The `should generate an implementation guide for simple-ig` [test](https://github.com/FHIR/sushi/blob/master/test/ig/IGExporter.IG.test.ts#L171) covers the missing `copyrightLabel` and `versionAlgorithm[x]` top level properties, as well as the properties on `resources` and `dependencies`. In the `#customized-ig` describe block, the `should maintain page hierarchy for configured pages` [test](https://github.com/FHIR/sushi/blob/master/test/ig/IGExporter.IG.test.ts#L1117) (as well as a few others in that block) tests the missing properties on `pages`.

While these tests I think do test the functionality you described, it is certainly less direct than it would be if I just added a new test for each case. Before I add in the 5 new tests, I wanted to check if you thought they are beneficial given the other existing tests.

test/ig/IGExporter.IG.test.ts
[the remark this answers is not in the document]
jafeltra: Updated in [6b5d6f2](https://github.com/FHIR/sushi/pull/1189/commits/6b5d6f2d87ae6ad596c2bb3fd0fbd955afa74455) but it might change again if I add more tests.

test/ig/IGExporter.IG.test.ts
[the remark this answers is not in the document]
jafeltra: Fixed in [18c9741](https://github.com/FHIR/sushi/pull/1189/commits/18c9741f6359b663df48a75d993ee41122f4c65a).

src/ig/IGExporter.ts
jafeltra: Updated in [c3f6501](https://github.com/FHIR/sushi/pull/1189/commits/c3f6501f590435cf43fa93af910e3be923cb1fde).

test/ig/IGExporter.IG.test.ts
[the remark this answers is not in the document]
mint-thompson: Ah, I see what you're saying: since the extensions are only present on R4 IGs, the existing tests for R4 IGs cover this scenario. No new tests are needed just for this.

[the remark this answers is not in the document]
jafeltra: Yep! If we wanted to call out the fact that R4 IGs without R5 properties do not add anything extra, the new tests would do that more explicitly. But with the `toEqual` expectations in those tests, we should catch any unintended consequences if something were to break in the future.

jafeltra:
I think everything on this PR is set now, so it is ready for re-review. I did also just try using the latest publisher and I still get the same errors, so this will stay as "Draft" for now, but it is ready to be re-reviewed content-wise.

src/ig/IGExporter.ts
```diff
@@ -1338,96 +1342,169 @@ export class IGExporter {
...
-      if (this.config.versionAlgorithmString) {
-        this.ig.versionAlgorithmString = this.config.versionAlgorithmString;
-      } else if (this.config.versionAlgorithmCoding) {
-        this.ig.versionAlgorithmCoding = this.config.versionAlgorithmCoding;
+  // If an R4 IG uses any new properties from R5, they should be included as extensions.
+  // Converting between FHIR versions is documented here: http://hl7.org/fhir/2022Sep/versions.html#extensions
+  addR5ExtensionsToR4() {
+    // Any new properties added to the IG resource in R5 should be included on extensions in R4 IGs
+    this.ig.definition.resource.forEach(resource => {
+      // Assign IG.definition.resource.profile to an extensions if provided
```
cmoesel, line 1441: This one is a little tricky because R4 has a _mostly_ equivalent property: `exampleCanonical` -- but it's `0..1` instead of `0..*`.  I think that maybe we should try to be smart about leveraging the existing R4 properties when it makes sense and only falling back to extensions when we need to represent something we can't say using R4 properties.  Maybe something like this?
* If `profile?.length === 1`, then `exampleCanonical = profile[0]`
* Else if `profile?.length > 1` , then `exampleCanonical = profile[0]` and add profile extension
* Else if `isExample != null` and `exampleCanonical == null`, then `exampleBoolean = isExample`

I think that basically says that R5 properties take precedence but we will represent them as R4 properties when possible.  It's a little tricky because in R5 `isExample` and `profile` can exist at the same time, but in R4, `example[x]` can only be `exampleCanonical` or `exampleBoolean` (and never both at the same time).  But I think maybe the pseudo-code above is right?  What do you think? Is this even a good idea?

src/ig/IGExporter.ts
```diff
@@ -1338,96 +1342,169 @@ export class IGExporter {
...
-      this.ig.dependsOn?.forEach(dependency => {
-        const configDependency = this.config.dependencies?.find(
-          d => d.packageId === dependency.packageId
-        );
-        if (configDependency.reason) {
-          dependency.reason = configDependency.reason;
+    // Add new IG.definition.page.source[x] property to an extension if it is provided. No need to set a default like R5 needs to.
+    this.ig.definition.page.page.forEach(page => {
+      // this.ig.definition.page is the toc.html page we create
+      // All configured pages are at the next level, so start at that level
```
cmoesel, line 1459: This confused me at first, until I realized you were really commenting on why you iterate on `this.ig.definition.page.page` in the line _above_.  Should this comment maybe be above the line it is explaining?

src/ig/IGExporter.ts
```diff
@@ -1338,96 +1342,169 @@ export class IGExporter {
...
+      this.addPageSourceExtensionForR4(page, this.config.pages);
+    });
```
cmoesel, line 1461: Should we also do something with `definition.page.name`? I thought about if we should use it to set the R4 `nameUrl` property, but I think it's technically a little different. (In fact, I think maybe `sourceUrl` is closer to `nameUrl` than `name` it).  So... maybe if there is a `page.name` we should add an extension for it?

src/ig/IGExporter.ts
```diff
@@ -1479,6 +1556,53 @@ export class IGExporter {
...
+  ): void {
+    const extensionUrl =
+      'http://hl7.org/fhir/5.0/StructureDefinition/extension-ImplementationGuide.definition.page.source';
+    const configPage = configPages?.find(
+      p =>
+        p.nameUrl.substring(0, p.nameUrl.lastIndexOf('.')) ===
+        page.nameUrl.substring(0, page.nameUrl.lastIndexOf('.'))
+    );
+    if (configPage) {
+      if (configPage.sourceUrl) {
```
cmoesel, line 1573: I _think_ this is the same as `page.nameUrl` in R4 -- but I'm not 100% confident enough in that to say we should overwrite an existing `nameUrl` with a `sourceUrl`.  So... I think what you're doing is probably best.

cmoesel (changes requested):
Nice job. I think you found a pretty clean way to do this.

I'd like to consider slightly different behavior for the R5 `resource` properties that are similar to existing R4 properties. See my individual comment for details. I also think maybe you missed one R5 property?

src/ig/IGExporter.ts
```diff
@@ -1338,96 +1342,169 @@ export class IGExporter {
...
+        resource.isExample = true;
+        if (resource.exampleCanonical) {
+          resource.profile = [resource.exampleCanonical];
         }
+        delete resource.exampleBoolean;
+        delete resource.exampleCanonical;
+      } else if (resource.exampleBoolean === false) {
+        resource.isExample = false;
+        delete resource.exampleBoolean;
+      }
```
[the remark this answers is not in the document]
jafeltra: I'd recommend either changing the settings on the Files tab or using `-w` in the command line because wow I hadn't looked at the UI since making that change it is really is rough.

src/ig/IGExporter.ts
jafeltra: I think that's fair. I will confess that I think I copied that from the previous R5 functionality, so I'll update it in both places.

src/ig/IGExporter.ts
jafeltra: I think this is right. I think I missed some of the updates to existing R4 properties from the R5 properties potentially included in configuration, which is maybe also what happened with some of the `page` properties further down too.

I think the psuedo-code makes sense, with one question. If `profile?.length > 1`, should the extension that is added include all the profiles provided or just include `profile[1]` and up since `profile[0]` is already represented?

src/ig/IGExporter.ts
jafeltra: Yeah, I think we should do something to handle `definition.page.name`. I think this is one of the places I missed converting some of the exiting R4 properties into the correct format. Instead of an extension, I think I'm leaning toward setting `page.nameUrl` equal to the value of `page.name`. When going from R4 to R5, we set the `page.name` to whatever is the `page.nameUrl` [here](https://github.com/FHIR/sushi/blob/c3f6501f590435cf43fa93af910e3be923cb1fde/src/ig/IGExporter.ts#L1510-L1520). But if `sourceUrl` is closer to `nameUrl` than `name`, maybe the R5 implementation is wrong. What do you think?

src/ig/IGExporter.ts
jafeltra: Similar to above, I think going off the R4 to R5 implementation, we only set sourceUrl based on the direct `source[x]` property in configuration and don't infer anything from an R4 IG. So I think that's why I added it as an extension. But similar to above, maybe this was the wrong approach.

src/ig/IGExporter.ts
cmoesel: I think it should include all the profiles. I'd rather it be a little redundant but complete. Otherwise if a tool sees the profile extension and _only_ looks at that, then it's missing something.  So I guess we're basically saying "Here is the _limited_ R4 representation (limited by cardinality in R4) and here is the _complete_ R5 representation (no cardinality limitation).  But if `profile` is only length 1, then R4 is not limited and therefore R5 is not needed.

src/ig/IGExporter.ts
cmoesel: Let's talk and try to work it out together. I agree it is confusing.

cmoesel:
> Should this function for going from an R5 configuration property to an R4 IG also handle stripping out the system when setting `definition.parameter.code`? Maybe it should add `system` to an extension? Or maybe it should add the whole Coding to an extension and assign nothing to `definition.parameter.code` (this doesn't feel right but maybe it is)? What do you think?

Oh dang.  Ummm... I think the R5 `Coding.code` value should get copied to the R4 `code` -- and then I guess throw a an R5`definition.parameter.code` extension (representing the `Coding`) onto the `parameter` object?

src/ig/IGExporter.ts
jafeltra: Updated in [32aa694](https://github.com/FHIR/sushi/pull/1189/commits/32aa694e112d618781fdf90c598633cbb24334d2).

src/ig/IGExporter.ts
jafeltra: Comments moved in [f396edf](https://github.com/FHIR/sushi/pull/1189/commits/f396edfe2a6e15631a25977b29463882003c8384).

src/ig/IGExporter.ts
jafeltra: Similarly, this is taken care of with the approach from the comment above and fixed in [67deb44](https://github.com/FHIR/sushi/pull/1189/commits/67deb44fe0b08395742a935a5189fa6c7726e4b5).

jafeltra:
> > Should this function for going from an R5 configuration property to an R4 IG also handle stripping out the system when setting `definition.parameter.code`? Maybe it should add `system` to an extension? Or maybe it should add the whole Coding to an extension and assign nothing to `definition.parameter.code` (this doesn't feel right but maybe it is)? What do you think?
> 
> Oh dang. Ummm... I think the R5 `Coding.code` value should get copied to the R4 `code` -- and then I guess throw a an R5`definition.parameter.code` extension (representing the `Coding`) onto the `parameter` object?

Updated in [d24e2be](https://github.com/FHIR/sushi/pull/1189/commits/d24e2be3650805e5e6eedb1453cfd9a328618db3).

jafeltra:
@cmoesel - I believe all the changes we discussed on Friday are implemented and fully tested with the latest updates.

@mint-thompson and @cmoesel  - when you re-review, let me know if you think some of the logic for the `IG.definition.page` properties abd `IG.definition.resource` properties is clear. There were a bunch of cases I was trying to ensure we handled, but the resulting logic feels like it might be over-complicating things, so if you notice an obvious simplification, let me know.

Everything that has been brought up on this PR should be set now, and the IG publisher does not appear to be logging errors anymore, so I've moved this out of draft.

src/ig/IGExporter.ts
```diff
@@ -1338,96 +1342,202 @@ export class IGExporter {
...
+          configEntry.profile.length === 1 &&
+          configEntry.profile[0] !== resource.exampleCanonical
+        ) {
+          resource.extension = (resource.extension ?? []).concat({
+            url: 'http://hl7.org/fhir/5.0/StructureDefinition/extension-ImplementationGuide.definition.resource.profile',
+            valueCanonical: configEntry.profile
+          });
         }
-      });
+      }
```
cmoesel, line 1470: I think lines 1445 - 1470 above can be done more succinctly like so:
```ts
      if (resource.exampleCanonical == null && configEntry?.profile?.length) {
        resource.exampleCanonical = configEntry.profile[0];
        delete resource.exampleBoolean;
      }
      if (
        resource.exampleCanonical != null &&
        configEntry?.profile?.some(p => p !== resource.exampleCanonical)
      ) {
        resource.extension = (resource.extension ?? []).concat({
          url: 'http://hl7.org/fhir/5.0/StructureDefinition/extension-ImplementationGuide.definition.resource.profile',
          valueCanonical: configEntry.profile
        });
      }
```

I tried it locally and all the tests still pass.

cmoesel (changes requested):
I didn't have time to re-review everything, but I've suggested a simplification for the resource profile bit... I'll re-review the rest on Thursday.

src/ig/IGExporter.ts
```diff
@@ -1479,6 +1589,69 @@ export class IGExporter {
...
+    configPages: ImplementationGuideDefinitionPage[]
+  ): void {
+    const sourceExtensionUrl =
+      'http://hl7.org/fhir/5.0/StructureDefinition/extension-ImplementationGuide.definition.page.source';
+    const nameExtensionUrl =
+      'http://hl7.org/fhir/5.0/StructureDefinition/extension-ImplementationGuide.definition.page.name';
+    const configPage = configPages?.find(
+      p =>
+        (p.nameUrl || p.name).substring(0, (p.nameUrl || p.name).lastIndexOf('.')) ===
+        page.nameUrl.substring(0, page.nameUrl.lastIndexOf('.'))
```
cmoesel, line 1605: I think this might be easier to read if done using `replace` rather than `substring`.  What do you think?
```suggestion
        (p.nameUrl || p.name).replace(/\.[^.]+$/, '') === page.nameUrl.replace(/\.[^.]+$/, '')
```
(Also, I didn't test this, so if you like it you should probably test it locally just to be sure it works).

src/ig/IGExporter.ts
```diff
@@ -1479,6 +1589,69 @@ export class IGExporter {
...
+      if (configPage.sourceUrl) {
+        // If nameUrl and sourceUrl do not match, add the sourceUrl value to an extension
+        if (configPage.nameUrl !== configPage.sourceUrl) {
+          page.extension = (page.extension ?? []).concat({
+            url: sourceExtensionUrl,
+            valueUrl: configPage.sourceUrl
+          });
+        }
+
+        if (configPage.name) {
```
cmoesel, line 1617: Don't we only need to do this if `configPage.name && configPage.nameUrl !== configPage.name`?

src/ig/IGExporter.ts
```diff
@@ -1479,6 +1589,69 @@ export class IGExporter {
...
+          // If nameUrl and name do not match, add the name value to an extension
+          if (configPage.nameUrl !== configPage.name) {
+            page.extension = (page.extension ?? []).concat({
+              url: nameExtensionUrl,
+              valueUrl: configPage.name
+            });
+          }
+        }
+        // Once nameUrl is set, assign a configured source[x] to an extension
+        if (configPage.sourceString) {
```
cmoesel, line 1635: Don't we only need to do this if `configPage.sourceString and configPage.nameUrl !== configPage.sourceString`?

src/ig/IGExporter.ts
cmoesel: Honestly, I forget everything we decided that day in the lab, so I could be wrong here...

src/ig/IGExporter.ts
cmoesel: Honestly, I forget everything we decided that day in the lab, so I could be wrong here...

cmoesel (changes requested):
Overall this looks good. In addition to the suggestion from my previous partial review, I just have a few more questions and suggestions.

src/ig/IGExporter.ts
jafeltra: I do think we decided that, but I don't actually think that can ever happen in reality. `configPage` is constructed in `addConfiguredPage` in the IGExporter as we convert the configuration to the IG. However, that function only ever adds `nameUrl`.

I _thought_ that if a `name` is set in the configuration, it was added in the `updateForR5` function when we add the R5 properties. However, it doesn't look like that is actually the case. That should probably be added (my bad that it isn't in there now). Once we do, I think this function can stay the same.

src/ig/IGExporter.ts
cmoesel: My apologies. I was thinking of `sourceUrl`.  You're right. `sourceString` is different -- so leave this as-is.

src/ig/IGExporter.ts
cmoesel: OK.  I see that `addConfiguredPage` is only looking for `page.nameUrl`.  So really, it needs to work for `page.nameUrl` _or_ `page.name`, right?

I still think that if it's R4, but the person put a `name`, we only want to add the extension if the `name` is _different_ than the `nameUrl`.  And I do think that's possible if the author did something stupid like putting `nameUrl` _and_`name` in the config (with different values). But if they're the same, I don't think the `name` extension adds any value to the R4 representation.

jafeltra: I think I partially over-complicated this because that check for if `configPage.name` is defined is used elsewhere in this function. And assuming the `name` is properly imported from configuration, the logic should be in place here.

I updated this section in [809e207](https://github.com/FHIR/sushi/pull/1189/commits/809e207e1a5963fe1bbbb9acc9fe0af202002595) by adding in the logic you suggested and I also tried to simplify the if/else cases. Let me know what you think.

I do still think the import of the configuration and the R5 export needs to be updated to handle if `name` is included in the configuration, but I have a couple questions about it, so maybe we can discuss that on Monday before I push it here.

jafeltra: @cmoesel  as discussed earlier, I also pushed [382309b](https://github.com/FHIR/sushi/pull/1189/commits/382309b430d759663b037dd1b75dab5d0a6dc6a7), which properly imports `page.name` from a configuration file and sets the value in an R5 IG. This should not effect the handling for R4 since the logic updated in [809e207](https://github.com/FHIR/sushi/pull/1189/commits/809e207e1a5963fe1bbbb9acc9fe0af202002595) should correctly handle the cases for when `name` is present or not. This latest commit ([382309b](https://github.com/FHIR/sushi/pull/1189/commits/382309b430d759663b037dd1b75dab5d0a6dc6a7)) technically doesn't impact the functionality of this PR, but it was a bug in the R5 IG configuration export, so it is good that it's fixed now.

jafeltra:
@cmoesel  and @mint-thompson - this should be ready for re-review. I think I fixed all the previously mentioned comments and the IG publisher still does not log any errors when the extensions are used in the IG resource.

cmoesel (approved):
This all looks good to me! Thanks for going through all this. I know it was a bit tedious at times!

I ran a full regression on it and only one IG changed -- and it changed in a good way! It added the `dependsOn.reason` extensions!

BTW -- this branch needs to be updated w/ master before it can be merged.

Issue #1181, opened 2022-11-30
SUSHI should not discard path rules in the importer


cmoesel:
Currently, SUSHI discards path rules in the import phrase because we've thought of them only as a way to set up indenting -- which all gets normalized to non-indented paths during import.

*Use Case 1*

With the new manual slice ordering approach, authors may want to use a path rule to indicate order of slices. For example, imagine a `LabObservationProfile` with this slice:
```
* category contains lab 1..1
* category[lab] = $OBSCAT#laboratory
```

If I author an `InstanceOf: LabObservationProfile`, I might want the lab category to go first, followed by a more specific `blood-lab` category.  Since the lab category slice already says all it needs to, I don't need to explicitly assign it. SUSHI will put it at the end of the array by default, so if I want it first, I should be able to simply throw in a path rule like so:
```
* category[lab]
* category[+] = $LABCAT#blood-lab
```

Since SUSHI strips out the path rule during import, however, my intent is lost.  The only way I can force what I want is by _re-assigning_ the value of `category[lab]` (which is a waste and invites trouble).

See also: https://github.com/FHIR/sushi/pull/1121#pullrequestreview-1199603388

*Use Case 2*

Even outside of slices, I might want to use a path rule to indicate that an optional fixed value should be included in the output. For example, if I have a profile with this:
```
* foo = #bar // (assuming foo is 0..1)
```
Then, by default, SUSHI will not add it because it is optional.  But I should be able to reference it in an Instance to indicate it should be added:
```
* foo
```

This is a bit contrived, but it could happen (and I have seen optional fixed value elements before).

cmoesel:
See the last instance in this FSHSchool link for an example of Use Case 1: https://fshschool.org/FSHOnline/#/share/3FgsTD9

PR #1191, opened 2022-12-16
Import Path Rules

jafeltra:
This PR adds support for SUSHI to import PathRules and include those rules on the FSH entities that support them. Previously, PathRules were just used to add the full path context to the other rules on an entity, but they were just discarded after that and never made it onto a FSH entity. Now, the path rules still set the context for all the other rules, but they are also returned by the importer and added to the FSH entity classes (like `Instance` and `Mapping`). They are not used at all during export since all the path context is already added to each rule during import.

I used some of the code/ideas that Nick had originally implemented in #797 (before refactoring to eliminate most of these changes).

I ran a full regression and got no changes, and I tried to package this up and use it in GoFSH and was able to remove the ts-ignore there. ~However, I can't remember exactly the use case we had with the manual slicing mode, so I haven't tested to ensure this helps there.~ Edit: I found the use case, and I don't think it works yet.

[`ccc3af7`](https://github.com/FHIR/sushi/commit/ccc3af7be15479238454251b466a8aad8d6cf572) adds a bunch of type casting in the test files that I was updating. I'm not sure if this started when I updated my globally installed version of TypeScript, but I was getting a lot of "variable is possibly undefined" errors that were showing up in my editor. Since we know those variables are definitely defined in the test, I added a type cast to have the error go away. There are still some other places where we use variables we already have a test for `expect(variable).toBeDefined()` and a testing util function we pass `undefined` to that is expecting a `boolean`/`string`/etc, but I didn't update those cases yet and just stuck to the type casting. I kept the commit separate, so I can always revert the commit if it's too distracting/out of scope.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-*   @jafeltra @cmoesel @mint-thompson @guhanthuran
+*   @jafeltra @cmoesel @mint-thompson
--- a/src/export/InstanceExporter.ts
+++ b/src/export/InstanceExporter.ts
@@ -19,7 +19,7 @@ import { InstanceOfNotDefinedError } from '../errors/InstanceOfNotDefinedError';
 import { InstanceOfLogicalProfileError } from '../errors/InstanceOfLogicalProfileError';
 import { Package } from '.';
 import { cloneDeep, merge, padEnd, uniq, upperFirst } from 'lodash';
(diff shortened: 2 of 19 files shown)


jafeltra:
Converting to draft until support for both use cases in #1181 are supported.

jafeltra:
I'm just going to close this PR and will open a new PR when the functionality is ready.

PR #1193, opened 2023-01-04
Warn when ig.ini does not point to SUSHI-generated IG JSON

### LINKED ISSUE #1173: Warn when ig.ini does not point to the IG that SUSHI generated
It's usually a mistake when the ig.ini does not point to the IG resource that SUSHI generated; otherwise, why have SUSHI generate the IG resource at all?  So... have SUSHI generate a warning in this case.  Of course, if `FSHOnly` is true, then it's not necessary to check (since SUSHI is not expected to create an IG resource).

cmoesel:
Fixes #1173.

The SUSHI-generated ImplementationGuide resource is intended to be used as input to the IG Publisher. If the `ig.ini` file points to a _different_ resource, log a hard-to-miss warning -- because this is almost always a mistake.

To test, run this branch against a valid example IG. There should be no warnings. Then change the `ig.ini` file so the `ig` points to a different location and run this branch again. Now you should see a warning. Finally, change the `sushi-config.yaml` to contain `FSHOnly: true` and run this branch once more. Now there should no longer be a warning.


jafeltra (approved):
Makes sense to me! I reviewed and tested the cases you mentioned. They work as expected and I couldn't think of any other case.

mint-thompson:
Is it necessary for this to check that the path given in ig.ini is the actual path where the SUSHI-generated IG JSON will be created? My understanding is that it only checks the filename to make sure that the configured id is used in the given path. That could be somewhere else entirely on the filesystem, or an entirely bogus path.

cmoesel:
@mint-thompson - It is looking for more than filename; it is looking for a path that ends in `fsh-generated/resources/ImplementationResource-{id}.json`. I feel like this is enough to catch 99.9% of the mistakes since it is pretty specific.  Since this is basically just a courtesy warning to save users from future confusion if they got it wrong, I think it's OK if it doesn't cover super wacky scenarios (because I think those are typically scenarios where the user would be very aware of what they're doing).

At least that's my opinion on it. ;-)

mint-thompson (approved):
All clear, let's roll it out.

PR #1196, opened 2023-01-19
Use inserted rules when checking new concepts

### LINKED ISSUE #1190: Parameterized RuleSets not working with hierarchical code system
If you use RuleSets to add codes in a code system, SUSHI complains if you try to add nested codes.  E.g.:
```
RuleSet: AddCode(code, display, definition)
* {code} {display} {definition}

CodeSystem: MyCodeSystemWithRuleSets
* insert AddCode(#MyCode, "MyCode", "This is my code")
* insert AddCode(#MyCode #SubCode, "SubCode", "This is my sub-code")
```

The above should work, but ends up with an error like this:
> error CodeSystem MyCodeSystemWithRuleSets does not have the specified hierarchy to add code SubCode

Similarly, `#MyCode insert AddCode(#SubCode, "SubCode", "This is my sub-code")` also fails.

FSH School example for the above (and other broken variations): https://fshschool.org/FSHOnline/#/share/3htxOHF

More advanced example that led to discovering this issue: https://fshschool.org/FSHOnline/#/share/3VVkR8o

mint-thompson:
Fixes #1190 and completes task [CIMPL-1049](https://standardhealthrecord.atlassian.net/browse/CIMPL-1049).

When applying insert rules on a CodeSystem, provide the expanded rules when checking new concepts. This way, concepts added by RuleSets can provide a hierarchy. Additionally, this allows for duplicate concepts to be detected when they are added by RuleSets.

In addition to the new tests, you can try running the FSH samples from the issue description. The two example CodeSystems that use insert rules should work correctly with this update.

### DIFF
--- a/src/export/CodeSystemExporter.ts
+++ b/src/export/CodeSystemExporter.ts
@@ -1,5 +1,5 @@
 import { FSHTank } from '../import/FSHTank';
-import { CodeSystem, CodeSystemConcept, StructureDefinition } from '../fhirtypes';
+import { CodeSystem, CodeSystemConcept } from '../fhirtypes';
 import {
   setPropertyOnDefinitionInstance,
   applyInsertRules,
@@ -8,9 +8,10 @@ import {
 import { FshCodeSystem } from '../fshtypes';
 import { CaretValueRule, ConceptRule } from '../fshtypes/rules';
 import { logger } from '../utils/FSHLogger';
-import { MasterFisher, Type, resolveSoftIndexing } from '../utils';
+import { MasterFisher, resolveSoftIndexing } from '../utils';
 import { Package } from '.';
-import { CannotResolvePathError } from '../errors';
+import { CannotResolvePathError, CodeSystemDuplicateCodeError } from '../errors';
+import { isEqual } from 'lodash';
 
 export class CodeSystemExporter {
   constructor(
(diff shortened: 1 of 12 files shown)


jafeltra (approved):
These changes make sense to me! Very straightforward. I tested out the examples as well, and those work as expected.

cmoesel (changes requested):
As I reviewed this, it reminded me of our previous conversation today about where and when to expand RuleSets... and I wondered if we really _should_ be checking the codes when we apply insert rules, or if we should be doing it later? I think we do it here because the only other time we check the codes is during import (for each code rule) -- so since we already missed that check, we do it when we apply `insert` rules instead.

But then that made me realize that while this does fix the example in #1190, there is still a problem with a variation on that example:
```
RuleSet: AddCode(code, display, definition)
* {code} {display} {definition}

CodeSystem: MyCodeSystemWithRuleSets
* insert AddCode(#MyCode, "MyCode", "This is my code")
* #MyCode #SubCode "SubCode" "This is my sub-code"
```
In this example, `#MyCode #SubCode` will be processed and checked in the importer _before_ the `insert` rule is resolved -- and as a result, the code check fails.

So it seems to me like maybe we shouldn't check the codes in the importer and `insert` rule applicator.  Maybe we should check the codes during export instead (_after_ import and `insert` rule application). What do you think? Does that seem right or is there a better way?

mint-thompson:
@cmoesel I think you are correct that it no longer make sense to try to check codes during import. Before `insert` rules, it was a convenient way to catch the error as early as possible, but it is no longer reliable. Checking during export means that we can know that the `insert` rules have been handled already. I'll update this PR accordingly.

mint-thompson:
@cmoesel I've moved the code that checks concept uniqueness and hierarchy to `CodeSystemExporter`, and I've added a test case similar to your example. It's not exactly the same, since the parameters have been applied before the exporter handles anything, but I also did a test run of your FSH and received the expected result with no errors.

src/fshtypes/FshCodeSystem.ts
```diff
@@ -26,52 +23,6 @@ export class FshCodeSystem extends FshEntity {
...
-          return (
-            rule instanceof ConceptRule &&
-            rule.code === predecessor &&
-            isEqual(rule.hierarchy, newConcept.hierarchy.slice(0, i))
-          );
-        })
-      ) {
-        throw new CodeSystemIncorrectHierarchyError(this.id, newConcept.code);
-      }
-    });
```
jafeltra, line 70: Is there somewhere that needs to be doing this check still? Or is there a reason it isn't needed anymore?

jafeltra (commented):
I just had a few questions about some of the refactored code.

test/import/FSHImporter.CodeSystem.test.ts
```diff
@@ -362,48 +362,6 @@ describe('FSHImporter', () => {
         expect(loggerSpy.getAllMessages('error')).toHaveLength(0);
       });
 
-      it('should log an error when encountering a duplicate code', () => {
```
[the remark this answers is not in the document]
mint-thompson: Ah, you're right, I hadn't added any because I saw tests for duplicate codes, but upon further inspection, they also all involve `insert` rules. I'll add some tests for handling duplicates without involving `insert` rules.

src/fshtypes/FshCodeSystem.ts
mint-thompson: The equivalent logic is in `CodeSystemExporter.setConcepts`. The loop that iterates over `concept.hierarchy` traverses the existing tree of concepts, and will log an error if it can't find the complete path specified in the hierarchy. The error log is followed by a return, meaning that the concept is not added to the exported CodeSystem.

jafeltra: Ah okay I see. I got confused since that code wasn't specifically new and didn't have the `CodeSystemIncorrectHierarchyError` thrown, but I see they're accomplishing the same thing. In that case, do you think it's safe to get rid of the `CodeSystemIncorrectHierarchyError` class entirely? It doesn't look like it's used anywhere else.

src/fshtypes/FshCodeSystem.ts
```diff
@@ -26,51 +23,6 @@ export class FshCodeSystem extends FshEntity {
...
-      // and the existing ConceptRule with that code has a matching hierarchy,
-      // the user may simply be referencing the existing concept to establish context.
-      if (
-        newConcept.display == null &&
-        newConcept.definition == null &&
-        isEqual(existingConcept.hierarchy, newConcept.hierarchy)
-      ) {
-        return false;
-      } else {
-        throw new CodeSystemDuplicateCodeError(this.id, newConcept.code);
```
cmoesel, line 50: If we stop throwing `CodeSystemDuplicateCodeError` in the `CodeSystemExporter` then we won't be using it anywhere and we can delete that class.

src/fshtypes/FshCodeSystem.ts
```diff
@@ -26,51 +23,6 @@ export class FshCodeSystem extends FshEntity {
...
-      if (
-        !this.rules.some(rule => {
-          return (
-            rule instanceof ConceptRule &&
-            rule.code === predecessor &&
-            isEqual(rule.hierarchy, newConcept.hierarchy.slice(0, i))
-          );
-        })
-      ) {
-        throw new CodeSystemIncorrectHierarchyError(this.id, newConcept.code);
```
cmoesel, line 68: We're not using `CodeSystemIncorrectHierarchyError` anymore, so we can delete that class.

cmoesel (changes requested):
Overall, I think this is a nice improvement. It not only works better, but I think we deleted more code than we added!  That's always nice!

I noted a few things that should be looked at before we merge.

src/export/CodeSystemExporter.ts
```diff
@@ -41,54 +41,54 @@ export class CodeSystemExporter {
...
+          // and the existing concept has the same code and hierarchy,
+          // this concept may just be used to establish path context, which is fine.
+          if (
+            !(
+              concept.display == null &&
+              concept.definition == null &&
+              isEqual(concept.hierarchy, existingConcept.hierarchy)
+            )
+          ) {
+            // duplicates are prohibited: http://hl7.org/fhir/codesystem.html#invs
```
jafeltra, line 58: Nice, good call to bring over the spec reference.

jafeltra (commented):
This looks good to me! It looks like there are just a few conflicts that need to be resolved since the last merge.

cmoesel (approved):
This looks great - and it does not introduce any regressions.

It just needs those conflicts resolved and then I think we're good!

jafeltra (approved):

cmoesel (approved):
It's good!

PR #1198, opened 2023-01-23
Find entities by id, even when their id is set by a rule

### LINKED ISSUE #1183: SUSHI should look at assigned values when resolving References
If an instance sets its `id` via an Assignment Rule, that `id` is what should be referenced from inside a `Reference()` keyword, not the name given to the instance via the `Instance:` keyword. Currently, SUSHI does not properly resolve references that use the id that was assigned from an assignment rule. Neither references to typical example instances or inline instances resolve properly.

An example of both `inline` and `example` instances with this behavior can be seen here: https://fshschool.org/FSHOnline/#/share/3h1fzck.

This was originally reported in [this issue](https://github.com/FSHSchool/FSHOnline/issues/116) on FSH Online.

mint-thompson:
Fixes #1183 and completes task [CIMPL-1052](https://standardhealthrecord.atlassian.net/browse/CIMPL-1052).

When fishing for most FSH entities, one of the conditions for a match is that the provided string matches the entity's `id`. However, an entity's `id` may be set by one of its rules. To make this more reliable, this PR introduces two main changes:
- All insert rules are applied before any resources are exported.
- When getting the id from a FSH definition, check the rules on the definition for one that sets the `id`.
These two changes together help to make sure that references to resources defined in FSH that have a rule for setting the `id` will resolve correctly.

In addition to the test cases added as part of the PR, the [sample FSH](https://fshschool.org/FSHOnline/#/share/3h1fzck) in the linked issue will now resolve all references correctly.

This change applies only to FSH entities that can be fished up by `id`. This is why `Mapping` is not changed, even though it has an `id`.

### DIFF
--- a/src/export/InstanceExporter.ts
+++ b/src/export/InstanceExporter.ts
@@ -64,8 +64,6 @@ export class InstanceExporter implements Fishable {
     instanceDef: InstanceDefinition,
     instanceOfStructureDefinition: StructureDefinition
   ): InstanceDefinition {
-    // The fshInstanceDef.rules list may contain insert rules, which will be expanded to AssignmentRules
-    applyInsertRules(fshInstanceDef, this.tank);
     resolveSoftIndexing(fshInstanceDef.rules);
     let rules = fshInstanceDef.rules.map(r => cloneDeep(r)) as AssignmentRule[];
     // Normalize all rules to not use the optional [0] index
@@ -429,8 +427,15 @@ export class InstanceExporter implements Fishable {
     return this.fisher.fishForMetadata(item, Type.Instance);
   }
 
+  applyInsertRules(): void {
+    const instances = this.tank.getAllInstances();
+    for (const instance of instances) {
+      applyInsertRules(instance, this.tank);
+    }
+  }
+
   exportInstance(fshDefinition: Instance): InstanceDefinition {
-    if (this.pkg.instances.some(i => i._instanceMeta.name === fshDefinition.id)) {
+    if (this.pkg.instances.some(i => i._instanceMeta.name === fshDefinition.name)) {
       return;
     }
 
@@ -475,7 +480,7 @@ export class InstanceExporter implements Fishable {
 
     const instanceOfStructureDefinition = StructureDefinition.fromJSON(json);
     let instanceDef = new InstanceDefinition();
-    instanceDef._instanceMeta.name = fshDefinition.id; // This is name of the instance in the FSH
+    instanceDef._instanceMeta.name = fshDefinition.name; // This is name of the instance in the FSH
     if (fshDefinition.title == '') {
       logger.warn(`Instance ${fshDefinition.name} has a title field that should not be empty.`);
     }
--- a/src/fhirtypes/common.ts
+++ b/src/fhirtypes/common.ts
@@ -881,6 +881,23 @@ export function getUrlFromFshDefinition(
   return `${canonical}/${fhirType}/${fshDefinition.id}`;
 }
 
+export function getIdFromFshDefinition(
+  fshDefinition: Profile | Extension | Logical | Resource | FshValueSet | FshCodeSystem | Instance,
+  tank: FSHTank
+): string {
+  if (fshDefinition instanceof Instance) {
+    applyInsertRules(fshDefinition, tank);
+    const assignmentRules = fshDefinition.rules.filter(
+      rule => rule instanceof AssignmentRule && rule.path === 'id' && typeof rule.value === 'string'
+    ) as AssignmentRule[];
+    if (assignmentRules.length > 0) {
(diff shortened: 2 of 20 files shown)


cmoesel (commented):
Overall, this looks good and the solution seems like it may help us down the road too!

I made one suggestion and I am currently running a full regression to see if there is any unintended impact.

cmoesel:
Full regression showed two IGs with differences.  Both are good.

In exhibit A, we see some canonical URLs changed to more properly reflect the correct ID:
image

In exhibit B, we see some relative references changed to properly include the resourceType. Honestly, I'm not 100% sure why this was affected, but I'll take it:
image

src/export/InstanceExporter.ts
```diff
@@ -475,7 +480,7 @@ export class InstanceExporter implements Fishable {
 
     const instanceOfStructureDefinition = StructureDefinition.fromJSON(json);
     let instanceDef = new InstanceDefinition();
-    instanceDef._instanceMeta.name = fshDefinition.id; // This is name of the instance in the FSH
+    instanceDef._instanceMeta.name = fshDefinition.name; // This is name of the instance in the FSH
```
jafeltra, line 483: Lol even though I know what it was like this before, this diff is a little bit funny

src/fshtypes/FshCodeSystem.ts
```diff
@@ -26,6 +26,25 @@ export class FshCodeSystem extends FshEntity {
...
+        rule instanceof CaretValueRule &&
+        rule.path === '' &&
+        rule.caretPath === 'id' &&
+        typeof rule.value === 'string'
+    ) as CaretValueRule[];
+    if (caretValueRules.length > 0) {
+      const lastCaretValueRule = caretValueRules[caretValueRules.length - 1];
+      return lastCaretValueRule.value.toString();
+    }
+    return this._id;
```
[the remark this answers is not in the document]
jafeltra: I like this idea!

jafeltra (commented):
This looks really good. I didn't notice anything else beyond Chris's suggestion.

jafeltra (approved):
Thanks, Mint! I think this makes sense. I was going back and forth on if it would help to have a `findIdAssignmentRule` function next to the `findIdCaretRule` function with a note that says to change the two together, but they probably are distinct enough that I think it's okay as is. But if you or Chris like the idea, I'm certainly not against it.

I didn't notice anything else!

cmoesel (changes requested):
Julia suggested creating a `findIdAssignmentRule`. I think that's probably a good idea, because I just found at least one other place that looks for an `id` assignment rule (and it actually isn't as good as yours, @mint-thompson).

So I suggest creating `findIdAssignmentRule` and using it in the relevant location in this PR as well as the other place I found similar code: https://github.com/FHIR/sushi/blob/0238364ea51f97120d79121bc0f41deba210d752/src/fhirtypes/common.ts#L433-L435

Similarly, I also found a place where `findIdCaretRule` can be used: https://github.com/FHIR/sushi/blob/0238364ea51f97120d79121bc0f41deba210d752/src/fhirtypes/mixins.ts#L64-L67

_Also note the use of `_.findLast` above, which seems pretty nice._

src/fhirtypes/common.ts
```diff
@@ -429,14 +429,9 @@ export function replaceReferences<T extends AssignmentRule | CaretValueRule>(
     );
     // If we can't find a matching instance, just leave the reference as is
     if (instance && instanceMeta) {
-      // If the instance has a rule setting id, that overrides instance.id
-      const idRule = instance.rules.find(
-        r => r.path === 'id' && r instanceof AssignmentRule
-      ) as AssignmentRule;
-      const id = idRule?.value ?? instance.id;
```
cmoesel, line 436: Oh... right! We don't even need this anymore. How exciting!

src/fhirtypes/mixins.ts
```diff
@@ -61,10 +62,12 @@ export class HasId {
    * @param {FshStructure | FshCodeSystem | FshValueSet} fshDefinition - The entity who's id is being set
    */
   validateId(fshDefinition: FshStructure | FshCodeSystem | FshValueSet | Instance) {
-    const idRule = _.findLast(
-      fshDefinition.rules,
-      rule => rule instanceof CaretValueRule && rule.caretPath === 'id' && rule.path === ''
-    ) as CaretValueRule;
+    let idRule: AssignmentRule | CaretValueRule;
+    if (fshDefinition instanceof Instance) {
+      idRule = findIdAssignmentRule(fshDefinition.rules);
```
cmoesel, line 67: Good catch!

cmoesel (approved):
I like it! Well done!

jafeltra (approved):
Looks good to me!

PR #1200, opened 2023-01-26
Warn on array items that match required slices but don't use a slice name

### LINKED ISSUE #1180: Manual Slice Ordering should warn when authors explicitly add an entry that matches a slice
The new manual slice ordering in the `next` branch is pretty awesome, but it does open the door for some mistakes that are easily made. For example, imagine a `LabObservationProfile` with this slice:
```
* category contains lab 1..1
* category[lab] = $OBSCAT#laboratory
```
If I author an `InstanceOf: LabObservationProfile`, I might include this rule:
```
* category = $OBSCAT#laboratory "Laboratory"
```

Currently, SUSHI `next` w/ manual slice ordering on does not issue any warnings about this, but it exports the instance with _two_ laboratory category entries -- one that was put in by the assignment using implicit 0 index and one that was put in by the required slice:
```json
[
  {
    "coding": [{
      "code": "laboratory",
      "system": "http://terminology.hl7.org/CodeSystem/observation-category",
      "display": "Laboratory"
    }]
  },
  {
    "coding": [{
      "code": "laboratory",
      "system": "http://terminology.hl7.org/CodeSystem/observation-category"
    }]
  }
]
```

This is obviously not my intention, so it would be nice for SUSHI to detect that the entry I added manually actually fits the slice -- and suggest that I use the slice name in my assignment.

There are multiple ways to detect this, but I think the easiest would be to markup the array entries as they're added.  For example:
```json
[
  {
    "coding": [{
      "code": "laboratory",
      "system": "http://terminology.hl7.org/CodeSystem/observation-category",
      "display": "Laboratory"
    }]
  },
  {
    "_impliedSlice": {
      "sliceName": "lab",
      "discriminator": [{
        "type": "pattern",
        "path": "$this"
      }]
    },
    "coding": [{
      "code": "laboratory",
      "system": "http://terminology.hl7.org/CodeSystem/observation-category"
    }]
  }
]
```
The we could go through as a post-process and compare the implied entries to the other (non-implied) entries and if they're a match at the discrimator path (using the discriminator type approach), issue a warning.  The warning should include the path and slice name.  If necessary, we could potentially use the FHIRPath JS library for finding values at the discriminator path (since discriminator path is FHIRPath).

Of course, this sounds easy now, but there may be some gotchas.

Ideally, we should not release the beta with manual slice ordering until these warnings are implemented.

See also: https://github.com/FHIR/sushi/pull/1121#pullrequestreview-1199603388

cmoesel:
Fixes #1180 

This PR adds the capability to detect cases where an array item in an instance matches a required (implied) slice but was not assigned using a slice name. We want to detect and warn on this because it very often results in undesirable outputs like repeated and/or incomplete array items. This capability is most useful when manual slice ordering is turned on, because without manual slice ordering most of these cases just overwrite the implied slice instead of being prepended before the slice.

The current approach for providing the warning is to find all arrays that have been sliced, and split their items into two groups: **(1)** items with an associated slice name (added by required assigned values or slice path assignements), and **(2)** items without an associated slice name (added by index). We then iterate over the items in the 2nd group (no slice name) to see if they are "matches" for any items in the first group (with slice names). An unnamed item is a match if it is a superset of the named item (e.g., if it fits the _pattern_ of the first item). If we detect a match, we produce a warning w/ relevant information.

There are some unit tests, but I found it helpful to run this against real IGs.  Remember, however, that manual slice ordering needs to be on, so you need to either update each project's configuration or temporarily hardcode SUSHI to always use manualSliceOrdering. I found the latter approach to be more convenient because I could then use it in concert w/ the regression script.  So here's one approach:

First, edit line 889 of `src/import/importConfiguration.ts` so it forces `manualSliceOrdering`:
```ts
manualSliceOrdering: true
```

Then create the file `naughty-repos.txt` with the following list of repos that get this warning (or if you prefer, just choose a subset):
```
HL7/fhir-spl#main
HL7/genomics-reporting#master
hl7ch/ch-allergyintolerance#master
hl7-eu/x-ehealth#master
HL7/fhir-radiation-dose-summary-ig#main
HL7/fhir-COVID19Library-ig#master
HL7/fhir-pacio-cognitive-status#master
saulakravitz/fhir-mCODE-ig#master
ahdis/ch-crl#master
nightingaleproject/vital_records_fhir_messaging_ig#main
hl7ch/ch-lab-order#master
mcode/mcode-omop#master
hl7ch/ch-resp#master
hl7ch/ch-ems#master
HL7/cimi-vital-signs#master
HL7/vrdr#master
HL7/vulcan-rwd#master
davidhay25/cs-structuredpath#main
davidhay25/actnow#main
IHE/Test.Asbestos#master
```

Then run the regression on local and use the `slice-ordering-on` branch as the comparison (this avoids tons of diffs since `next` has slice ordering off):
```
ts-node regression/run-regression.ts regression/naughty-repos.txt github:slice-ordering-on local
```

Then you can look at the logs of the local runs to see the new warnings.

NOTE: This seems to be pretty effective at finding paths that should use slice names _when the slice is required (1..1)_, but it won't find as many paths that reference optional slices since those optional slices might not have corresponding items in the instance to compare to. Our next step should be to analyze the slicing discriminators and the corresponding fixed/pattern values at the discriminator paths -- and then match instance items against _those_. That approach, however, is more complex -- and still not complete since we can't easily handle all types of discriminators and paths.


src/export/InstanceExporter.ts
```diff
@@ -447,6 +455,74 @@ export class InstanceExporter implements Fishable {
...
+                logger.warn(
+                  `${instanceDef.id} has an array item that matches a required slice, but does not use the slice name ` +
+                    'in its path. This may result in unexpected items in your array. If the item is intended to be ' +
+                    "in the slice, modify your FSH to use the slice name in the item's path. See: " +
+                    'https://hl7.org/fhir/uv/shorthand/reference.html#sliced-array-paths\n' +
+                    `  Path:  ${slicingEl.path}\n` +
+                    `  Slice: ${matches.join(' or ')}\n` +
+                    `  Value: ${JSON.stringify(value)}`,
+                  fshDef.sourceInfo
+                );
```
mint-thompson, line 517: Thinking about these two messages together, I think it would be good if the first message also included the part about unexpected items. So maybe the first one could be something like:
> `${instanceDef.id}` has an array item that is exactly the same as a required slice, but does not use the slice name in its path. This may result in unexpected items in your array. Usually, you can remove this item from your FSH, since FSH instances inherit required assigned values from their profile. Alternately, you can modify your FSH to use the slice name in the item's path.

src/export/InstanceExporter.ts
```diff
@@ -650,3 +727,9 @@ export class InstanceExporter implements Fishable {
     return this.pkg;
   }
 }
+
+// This logic is extracted to its own function so we can easily test it
+// to ensure it exhibits the behavior we need.
+export function matchesPattern(obj: any, pattern: any): boolean {
+  return isMatch(obj, pattern);
+}
```
mint-thompson, line 735: I feel conflicted about having `matchesPattern` as a wrapper function for lodash's `isMatch` and the accompanying test cases. Since there's no additional logic in `matchesPattern`, these tests are against lodash's code, not ours. As you mentioned in the comment , you added the tests so that you're confident that `isMatch` is doing what you want it to do, and I also want to be confident of that. I recognize that the tests you have for `matchesPattern` could be rewritten to be tests that check for messages from `checkForNamelessSlices`, but that means rewriting a bunch of tests to be much longer. So, it's giving you more work to do, but it's also a test that more accurately targets SUSHI code, not an external library. Let me know what you think.

mint-thompson (changes requested):
This is really good! I think that the strategy you've used for implementing this makes sense. I have a few comments, and one (somewhat large, sorry) request: could you add some more test cases where `LabProfile` has multiple required slices defined on `category`? I found that I wasn't getting the result I expected when I tried out this code with multiple required slices. Specifically, when I had this FSH:
```
Profile: LabProfile
Parent: Observation
* category ^slicing.discriminator.type = #pattern
* category ^slicing.discriminator.path = "$this"
* category ^slicing.rules = #open
* category contains lab 1..1 and procedure 1..1
* category[lab] = http://terminology.hl7.org/CodeSystem/observation-category#laboratory
* category[procedure] = http://terminology.hl7.org/CodeSystem/observation-category#procedure

Instance: LabInstance
InstanceOf: LabProfile
* category = http://terminology.hl7.org/CodeSystem/observation-category#laboratory
* status = #final
* code = http://foo.com#a
* valueBoolean = true
```
I got the "matches a required slice" message, but I expected to get the "exactly the same as a required slice" message.

src/export/InstanceExporter.ts
```diff
@@ -447,6 +455,74 @@ export class InstanceExporter implements Fishable {
...
+            return { name: v._sliceName, value: cleanValue };
+          });
+        if (namedValues.length) {
+          for (const value of unnamedValues) {
+            // It's a match if the unnamed value is a compatible superset of the named value (i.e., slice)
+            const matches = namedValues
+              .filter(nv => matchesPattern(value, nv.value))
+              .map(nv => nv.name);
+            if (matches.length) {
+              if (namedValues.every(nm => isEqual(value, nm.value))) {
```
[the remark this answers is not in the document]
cmoesel: Good catch. My intent was that if an item exactly matches a slice but _also_ partially matches another slice, I want the second message (because I can't say for sure which slice is the right one). But clearly I messed that up because it fails the use case where there are multiple slices and only one match (that happens to be exact). I'll work on that.

src/export/InstanceExporter.ts
cmoesel: Yeah, you're probably right. I originally did this because I thought the `matchesPattern` function was going to be _more_ than just lodash `isMatch`. When I discovered that it is, in fact, just `isMatch`, I had already written all the tests and it was 11pm the night before the sprint closed. So I opted for not re-doing all the tests and putting this message in instead.  But... you're probably right.

src/utils/PathUtils.ts
```diff
@@ -265,3 +266,51 @@ export function resolveSoftIndexing(rules: Array<Rule | CaretValueRule>, strict
...
   });
 }
+
+/**
+ * Given an id or path, collect all of the values at that path in a flattened array. This works
+ * similar to FHIRPath path navigation, but has a few tweaks for our specific use case.
+ */
+export function collectValuesAtElementIdOrPath(idOrPath: string, object: any) {
+  const ignoredSliceRequirements: string[] = [];
+  let values = flatten([object]);
```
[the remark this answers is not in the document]
cmoesel: Yes. A prior version of this function featured recursive calls to `collectValuesAtElementIdOrPath` that could pass an array as the object. After I refactored out the recursion, I forgot to remove the unnecessary `flatten`.  Good find.

src/export/InstanceExporter.ts
```diff
@@ -447,6 +455,74 @@ export class InstanceExporter implements Fishable {
...
+    fshDef: Instance,
+    instanceDef: InstanceDefinition,
+    instanceOfStructureDefinition: StructureDefinition
+  ) {
+    const nonChoiceSlicedElements = instanceOfStructureDefinition.elements.filter(
+      e => e.slicing && !e.path.endsWith('[x]')
+    );
+    for (const slicingEl of nonChoiceSlicedElements) {
+      // Use the sliced element's path to get all potential element values from the instance
+      const values = collectValuesAtElementIdOrPath(slicingEl.id, instanceDef);
```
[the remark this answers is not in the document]
cmoesel: Addressed in [4fea2ef](https://github.com/FHIR/sushi/pull/1200/commits/4fea2efe5f1365762afddea29f0838f921ec72b0).

src/utils/PathUtils.ts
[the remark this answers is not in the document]
cmoesel: Addressed in [4fea2ef](https://github.com/FHIR/sushi/pull/1200/commits/4fea2efe5f1365762afddea29f0838f921ec72b0). I also updated the argument type to be InstanceDefinition.

src/export/InstanceExporter.ts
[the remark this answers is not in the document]
cmoesel: Addressed in [4fea2ef](https://github.com/FHIR/sushi/pull/1200/commits/4fea2efe5f1365762afddea29f0838f921ec72b0). Now it works according to my intent: (a) use the exactly message if _all_ matches are exact, (b) use general message when  some matches are not exact.

src/export/InstanceExporter.ts
cmoesel: Addressed in [4fea2ef](https://github.com/FHIR/sushi/pull/1200/commits/4fea2efe5f1365762afddea29f0838f921ec72b0).

src/export/InstanceExporter.ts
cmoesel: Addressed in [44901a5](https://github.com/FHIR/sushi/pull/1200/commits/44901a59d50bf9007351b99ce146eda82e4cb322).

cmoesel:
@mint-thompson - Thanks for the review with excellent feedback. Your feedback has been addressed, including adding that extra test you asked for (and a bunch more).

test/utils/PathUtils.test.ts
```diff
@@ -356,4 +359,256 @@ describe('PathUtils', () => {
...
+    });
+
+    it('should collect simple singular properties with resourceType prefix and resourceType', () => {
+      const results = collectValuesAtElementIdOrPath('Observation.status', object);
+      expect(results.values).toEqual(['final']);
+      expect(results.ignoredSliceRequirements).toEqual([]);
+    });
+
+    it('should collect simple singular properties with resourceType prefix and no resourceType', () => {
+      //@ts-ignore
```
mint-thompson, line 387: I don't think this usage of `@ts-ignore` or the following uses are necessary. I was able to remove them and run the tests without any problems.

mint-thompson (changes requested):
Thank you for adding all those new tests! Hopefully it wasn't too much extra work. I have one minor change to request.

test/utils/PathUtils.test.ts
cmoesel: I added these because VS Code was flagging them for me. It turns out that VS Code tends to behave better if you put a test-specific tsconfig straight into the `test` folder. Other TypeScript projects I work on have done this as well, so there is precedent. After doing this, I was able to remove the `//@ts-ignore` statements.  See 302970c.

mint-thompson (dismissed):
That change to the tsconfig is working fine for me. Thank you!

src/export/InstanceExporter.ts
```diff
@@ -447,6 +455,73 @@ export class InstanceExporter implements Fishable {
...
+                );
+              } else {
+                logger.warn(
+                  `${instanceDef.id} has an array item that matches a required slice, but does not use the slice name ` +
+                    'in its path. This may result in unexpected items in your array. If the item is intended to be ' +
+                    "in the slice, modify your FSH to use the slice name in the item's path. See: " +
+                    'https://hl7.org/fhir/uv/shorthand/reference.html#sliced-array-paths\n' +
+                    `  Path:  ${slicingEl.path}\n` +
+                    `  Slice: ${matchedNames.join(' or ')}\n` +
+                    `  Value: ${JSON.stringify(value)}`,
```
jafeltra, line 514: I'm wondering if the `Value` should be included in a `debug` level log. There was a project I noticed that had this warning on `Bundle.entry` and the `Value` log is pretty dense because it's a large JSON object. Maybe the usefulness most times outweighs this particular case coming up, just was curious what you thought.

jafeltra (commented):
This looks really good to me! I had one small comment about the warning, but either way I think this looks good and approve!

src/export/InstanceExporter.ts
cmoesel: That's a good point... I think the value is too useful to hide in a debug log, but perhaps we should truncate it after some number of characters... I'll think on that.

cmoesel: OK. Commit ed89951 truncates values > 300 characters. The number 300 is kind of arbitrary, but it seems like a good compromise to me.  Let me know what you think. Here's an example from `nightingaleproject/vital_records_fhir_messaging_ig`:
image

jafeltra: The 300 value looks reasonable to me, so I like it! I think this will give enough info to point them to the right place without cluttering the output. Thanks!

jafeltra (approved):

mint-thompson (approved):

PR #1203, opened 2023-02-01
FSH may contain numbers in scientific notation

### LINKED ISSUE #1132: SUSHI does not support assigning decimals with exponents
SUSHI should support assigning values that have exponents to `decimal` types in FHIR.

For example, the following rules should be supported:

```
* valueQuantity = 2.1E-3 'g' // type quantity and assigns the value of type decimal
* referenceRange.high.value = 9.1E+6 // type decimal
```

Anywhere a decimal values can be used, a decimal with exponents should also be supported. The quantity and decimal listed above may not be an exhaustive list of where this should work.

Originally reported on Zulip [here](https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/decimal.20with.20exponents).
Simple FSH Online example with the above rules [here](https://fshschool.org/FSHOnline/#/share/3pgNcau).
FHIR Observation example that shows decimal values in use [here](https://www.hl7.org/fhir/observation-decimal.json.html).

mint-thompson:
Fixes #1132 and completes task [CIMPL-1001](https://standardhealthrecord.atlassian.net/browse/CIMPL-1001).

A number can be expressed in scientific notation by ending it with "E" followed by the exponent. The "E" can be lower or upper case. The exponent may contain a sign. A negative exponent is indicated by using a minus sign. A positive exponent is indicated by using a plus sign or no sign.

The test case shows a few examples of numbers that are now valid. There are no changes to the application code, only to the lexer definition and the accompanying generated code.


test/import/FSHImporter.test.ts
```diff
@@ -466,6 +466,56 @@ Long statement:
     expect(profile.description).toBe(expectedDescriptionLines.join('\n'));
   });
 
+  it('should parse numbers using exponential notation', () => {
+    const input = leftAlign(`
+    Instance: MyObservation
+    InstanceOf: Observation
+    * component[0].valueQuantity = 2.3E11 'kg'
+    * component[1].valueQuantity = 155e-8 'm'
+    * component[2].valueQuantity = 6.453E+25 's'
```
cmoesel, line 475: Could you add some more components that test other types that could use exponents?
* `valueInteger` (integer)
* `valueSampledData.period` (decimal)
* `valueSampledData.dimensions` (positiveInt)

cmoesel (changes requested):
The grammar change looks good and it passed through full regression with _no changes_.

I noted that the notation doesn't carry through to the JSON in all cases, but I think that's OK. It seems like maybe it only converts to decimal when it can do so with no loss of precision.

The one thing I would like to see is one or two more tests -- specifically in places that accept a regular number (not a quantity).The grammar change looks good and it passed through full regression with _no changes_.

I noted that the notation doesn't carry through to the JSON in all cases, but I think that's OK. It seems like maybe it only converts to decimal when it can do so with no loss of precision.

The one thing I would like to see is a little more testing:
* Using exponents with a few more types (which I think can be done by just adding more `component`s to your existing test
* Using exponents in a fixed value for a profile (ideally a fixed `Quantity` and a fixed `integer`)
* Using exponents in a caret rule (perhaps `^minValueInteger` and/or `^maxValueUnsignedInt` from `ElementDefinition` on an element path)

mint-thompson:
@cmoesel Thank you for the feedback! Based on our conversation about handling large values, I've made a small update that should result in more consistent value representations. When an exponent is present, the number is inspected to see if it represents an integer. If it does, a `BigInt` is created. Otherwise, `parseFloat` is used to store the value. The test case should demonstrate the potential combinations.

jafeltra (approved):
The logic here is trickier than I was expecting, but it makes sense to me! I couldn't think of any cases that would be missed.

cmoesel (approved):
image

PR #1205, opened 2023-02-02
Autoload hl7.terminology.r4 package

### LINKED ISSUE #1169: SUSHI should always put hl7.terminology in scope
The [HL7 Terminology (THO) IG](https://terminology.hl7.org/) should also always be in scope so that authors can reference its value sets and code systems without declaring it as a dependency.

If an author puts any version of `hl7.terminology` in their dependencies, then just use that.  Otherwise, SUSHI needs to use the package API to determine the latest version of `hl7.terminology` and load that (_not_ current).

Since not all authors will need this, SUSHI should be forgiving when it's not able to load the package, marking it as a warning rather than an error.

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Canonical.20URL.20gets.20.22forgotten.22.20on.20a.20FHIR.20version.20change/near/308462002

cmoesel:
Load the HL7 terminology package implicitly unless the author explicitly puts it in the config.

Based on [this conversation](https://chat.fhir.org/#narrow/stream/179239-tooling/topic/New.20Implicit.20Package/near/325488084), tools can choose whatever variant of the package suits them (e.g., r4, r5, etc). In our case, we only use terminology dependencies to resolve id/name/url, so we don't have to match the IG version. So... we always load r4. (The other reason is because r5 is currently stale and in the wrong format anyway).

Since the different variations of the terminology package represent the same content, however, if we detect _any_ variant in the config, we won't load the implicit terminology package.

Fixes #1169.


package.json
```diff
@@ -6,7 +6,7 @@
     "build": "del-cli dist && tsc && copyfiles -u 3 \"src/utils/init-project/*\" dist/utils/init-project",
     "build:watch": "tsc -w",
     "build:grammar": "bash antlr/gradlew -p antlr generateGrammarSource",
-    "test": "jest --maxWorkers=4 --coverage",
+    "test": "jest --coverage",
```
cmoesel, line 9: Oh, whoops. I forgot I did this.

I did it because one of the VS Code Jest extensions doesn't work correctly when `--maxWorkers` is set. Please run the tests locally and let me know if removing this has unsatisfactory results (for example, if it makes test runs a lot longer).

mint-thompson (dismissed):
Looks good to me. Also, running the tests doesn't take my system any longer now than it did before, so the change to the `test` script seems safe.

src/utils/Processing.ts
```diff
@@ -220,16 +228,41 @@ export function loadAutomaticDependencies(
   defs: FHIRDefinitions
 ): Promise<void | FHIRDefinitions>[] {
   return AUTOMATIC_DEPENDENCIES.map(dep => {
-    if (configuredDependencies.some(cd => cd.packageId === dep.packageId)) {
+    const alreadyConfigured = configuredDependencies.some(cd => {
+      // hl7.some.package, hl7.some.package.r4, and hl7.somepackage.r5 all represent the same content,
+      // so they are essentially interchangable and we should allow for any of them in the config.
```
jafeltra, line 233: If the above question warrants a change, how do you feel about this monumental change while you're already in the file?
```suggestion
      // so they are essentially interchangeable and we should allow for any of them in the config.
```

src/utils/Processing.ts
```diff
@@ -220,16 +228,41 @@ export function loadAutomaticDependencies(
...
     } else {
-      return loadDependency(dep.packageId, dep.version, defs).catch(e => {
-        let message = `Failed to load automatically-provided ${dep.packageId}#${dep.version}: ${e.message}`;
-        if (/certificate/.test(e.message)) {
-          message += CERTIFICATE_MESSAGE;
-        }
-        logger.warn(message);
-      });
+      let p = Promise.resolve();
+      if (dep.version === 'latest') {
```
jafeltra, line 244: This doesn't effect this PR since this is going into `master`, but is this the first time we support `'latest'` versions? If so, is this something FPL should eventually handle?

package.json
jafeltra: The tests do take longer for me than they do on `master`, but I ran them a few times and I couldn't get them to be consistently shorter with `--maxWorkers=4` back. I think 3 out of 4 times using `--maxWorkers=4` was shorter, but that fourth time when it didn't make a difference makes me feel like it is probably fine to keep this as is. It could easily just be my computer.

jafeltra (commented):
I had a few pretty minor comments, but I don't think any of them are even required.

src/utils/Processing.ts
```diff
@@ -220,16 +228,41 @@ export function loadAutomaticDependencies(
   defs: FHIRDefinitions
 ): Promise<void | FHIRDefinitions>[] {
   return AUTOMATIC_DEPENDENCIES.map(dep => {
-    if (configuredDependencies.some(cd => cd.packageId === dep.packageId)) {
+    const alreadyConfigured = configuredDependencies.some(cd => {
+      // hl7.some.package, hl7.some.package.r4, and hl7.somepackage.r5 all represent the same content,
```
cmoesel, line 232:
```suggestion
      // hl7.some.package, hl7.some.package.r4, and hl7.some.package.r5 all represent the same content,
```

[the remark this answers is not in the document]
cmoesel: I used the github review feature to add my own suggestion and commit it.

src/utils/Processing.ts
cmoesel: It is. I don't think that `latest` as a version identifier is an official FHIR thing -- but it seemed like the right approach to me, given the requirements. As for supporting it in FPL... maybe? This logic is currently outside of FPL since it's part of determining the set of packages we'd pass into FPL.  But it might be nice to embed it in FPL as a feature.

package.json
cmoesel: OK. We'll leave it for now. We can always bring it back later if it ends up being a problem.

cmoesel:
@jafeltra and @mint-thompson - I updated the typos in the comments. I think that's all that needed to be updated. It dismissed Mint's review, but since it was just a change to comments, I'll assume they still approve. But I do at least need an official approval from @jafeltra.

src/utils/Processing.ts
jafeltra: That makes sense. I wonder if it could be a helper function that is available from FPL. Thoughts for later!

jafeltra (approved):
Looks good!

PR #1208, opened 2023-02-08
Configurable Log Level

### LINKED ISSUE #1194: Silent / errors only mode
It would be great for SUSHI to have a mode in which it will print warnings/errors only - quiet unless there's a problem. The practical reason for this is that in visual-space constrained environments (think enterprise remote sessions or simply small screens), you always have to scroll up after a SUSHI run to find the actual issues.

There some support from others for this in https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/print.20errors.20only

jafeltra:
Fixes #1194 

This PR makes it so the log level is configurable from the command line. The `-l --log-level` option can be used to specify either `debug`, `info`, `warn`, or `error`. These align with the same options we allow in GoFSH. If no option is specified, it defaults to `info` as usual.

I removed the `--debug` flag entirely because this will be going into `next`. There were some changes we tried to support backwards compatibility for in the CLI command structure, but I decided this was a straight forward change that wouldn't have too much impact and mostly introduces new functionality, so `--debug` can't be used.

There are no units tests for this, but I tested it locally by adding a `logger.debug`, `logger.warn`, and `logger.error` to `app.ts` and ensuring the right message levels were used. Alternatively, you could trigger each of those messages with a SUSHI project.


cmoesel (dismissed):
This seems to work well. I was worried that if you chose `-l error` that the summary might not show the warnings (because I think we count those via the logger) -- but my worries were unfounded. The warning count remains accurate!

Nice job!

src/app.ts
```diff
@@ -140,12 +143,15 @@ async function runBuild(input: string, program: OptionValues, helpText: string)
...
+  if (program.logLevel === 'debug' || program.logLevel === 'warn' || program.logLevel === 'error') {
+    logger.level = program.logLevel;
+  }
 
   logger.info(`Running ${getVersion()}`);
   logger.info('Arguments:');
-  if (program.debug) {
-    logger.info('  --debug');
+  if (program.logLevel) {
+    logger.info(`  --log-level ${program.logLevel}`);
```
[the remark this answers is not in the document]
jafeltra: Wait I actually didn't realize that... that's kind of weird but I don't know a nice way around it.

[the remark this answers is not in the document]
cmoesel: Yeah, I mean, if they don't want info messages then they don't want info messages. I think we're just doing what they asked.

jafeltra:
@cmoesel and @mint-thompson -- I added [99a3fd3](https://github.com/FHIR/sushi/pull/1208/commits/99a3fd3e315923877de628eb05d0c75330cd8b9f) to add specific choices to the log level option. Let me know what you think!

src/app.ts
```diff
@@ -140,12 +145,15 @@ async function runBuild(input: string, program: OptionValues, helpText: string)
     process.exit(1);
   }
 
-  if (program.debug) logger.level = 'debug';
+  // Set the log level. If no level is specified, logger defaults to info
+  if (program.logLevel === 'debug' || program.logLevel === 'warn' || program.logLevel === 'error') {
```
cmoesel, line 149: Now that commander halts on invalid log-levels, we don't need to check them all here. Instead we only need to change the level if it's _not_ the default.
```suggestion
  if (program.logLevel !== 'info') {
```

cmoesel (changes requested):
I like the addition of the choices!

I thought about if you should add `.default('info')`, but that would make `program.logLevel` always non-null, so it would always get logged in the `Arguments:` message, even when the person didn't pass it in as an argument. So I think the way it is probably makes sense.

Now that we have some value-checking built in, however, we can simplify one of the `if` statements. See my individual comment.

src/app.ts
jafeltra: Ah good call!

jafeltra:
>I thought about if you should add .default('info'), but that would make program.logLevel always non-null, so it would always get logged in the Arguments: message, even when the person didn't pass it in as an argument. So I think the way it is probably makes sense.

I actually went back and forth on this too and ultimately landed in the same place - if you don't include `--log-level`/`-l` then it is probably better to not be included in that Arguments list. But I probably could have been convinced. But apparently I won't be convinced by you, so it'll be up to @mint-thompson 😆

src/app.ts
jafeltra: [951fac3](https://github.com/FHIR/sushi/pull/1208/commits/951fac3ba456e505e7fbbc91294cb8a930bdac3c) updates the `if`. It also simplifies the description of the option to work better with the choices.

mint-thompson (approved):
Hooray for log levels!

cmoesel (approved):
happy-joy

PR #1212, opened 2023-02-15
Update uninherited extensions based on JIRA-28441

cmoesel:
JIRA-28441 contains a more complete list of the extensions that should not be inherited. Update our hard-coded list to reflect the latest list in Jira.

This is needed for work that Saul Kravitz is doing.


jafeltra (approved):
The new test makes sense and the list of extensions reflects that Jira issue, so I think this is good!

mint-thompson (approved):

PR #1213, opened 2023-02-16
Load dependencies in deterministic order

cmoesel:
This PR changes dependency loading logic so that automatic dependencies are loaded first (in the order specified in our code), then configured dependencies (in the order specified in the config), then FHIR core. Dependencies that are loaded later get higher priority in name and id resolution.

This makes builds more repeatable but also solves the issue that the automatic `hl7.terminology` dependency was replacing FHIR Core canonicals w/ `hl7.terminology` canonicals.

To test this, run a regression against master and include this repo:
```
saulakravitz/vrdr-fsh#master
```

The master branch of SUSHI will generate value set bindings for `http://terminology.hl7.org/ValueSet/location-physical-type` and `http://terminology.hl7.org/ValueSet/organization-type`, but this branch will generate bindings for `http://hl7.org/fhir/ValueSet/location-physical-type` and `http://hl7.org/fhir/ValueSet/organization-type`. The original R4 bindings are preferred.

### DIFF
--- a/src/utils/Processing.ts
+++ b/src/utils/Processing.ts
@@ -212,22 +212,20 @@ export async function loadExternalDependencies(
     );
   }
   dependencies.push({ packageId: fhirPackageId, version: fhirVersion });
-  const automaticPromises = loadAutomaticDependencies(dependencies, defs);
-  const configuredPromises = loadConfiguredDependencies(
-    dependencies,
-    fhirVersion,
-    config.filePath,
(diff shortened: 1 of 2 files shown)


jafeltra (commented):
... based on current events, I may not approve ...

jafeltra (approved):
My approval stands

mint-thompson (approved):

PR #1214, opened 2023-02-16
Fix logic for setting logger level

jafeltra:
Fixes the logic for setting the log level to avoid setting a log level of `undefined` if no `-log-level` option is specified.


cmoesel (approved):
Thank you for 10x-ing the log.
image

mint-thompson (approved):
hooray for teamwork!

PR #1216, opened 2023-02-17
Insert rule parameters can be enclosed in double brackets

### LINKED ISSUE #1154: Consider approaches to better support RuleSet parameters with commas and parentheses
Currently, the `,` and `)` characters need to be escaped when using them as input arguments to a RuleSet. This makes development tedious in some cases, especially when passing FHIRPath expressions in as parameters.

It can be problematic if you have a RuleSet that inserts another RuleSet and needs to pass through the parameter -- because the parameter will be passed into the second-level RuleSet _without_ the escapes!

See this thread for additional conversation/suggestions: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/sushi.3A.20Fhirpath.20escape.20rules/near/258562781

mint-thompson:
Fixes #1154 and completes task [CIMPL-1054](https://standardhealthrecord.atlassian.net/browse/CIMPL-1054).

When a parameter is enclosed in double brackets, commas and right parentheses do not need to be escaped. The entire parameter value must be enclosed for this feature to be used. Existing insert rules are not affected.

This feature comes in two main parts:
- handling insert rule parameters in `FSHImporter`
- performing the value substitution using the new `MiniFSHImporter`

`MiniFSHImporter` and the accompanying definitions in `MiniFSH.g4` and `MiniFSHLexer.g4` provide a parser with a very narrow purpose: performing value substitutions in the contents of a `ParamRuleSet`. Value substitutions are performed with an awareness of whether the substitution is within double-brackets for an `insert` rule, in which case, double-bracket escapes are performed. In actual use, I expect this to be needed pretty rarely, but it is needed for a robust implementation.

Most of the new tests are in `FSHImporter.Profile.test.ts`, showing off the handling of various double-bracketed parameter values. As noted in that file and in `FSHImporter.SDRules.test.ts`, recent changes to the FSH grammar mean that the `contents` of a `ParamRuleSet` will start with a newline character. The MiniFSH grammar expects these newlines to be present, and as such, they need to be present in the test objects.

I ran a full regression on 2023-03-08, and no repos showed any changes. So, this hopefully means that these changes are in fact fully backwards compatible.


cmoesel (changes requested):
Nice approach to a tricky feature. I think some of the stuff you did here will also be very helpful as we add more capabilities to RuleSets. Thanks!

I've asked a few questions that may (will probably) result in small changes.

As for the regression -- if it's not too difficult, we might consider submitting a PR against VRDR with the fixes, noting that we found this previously undetected issue when we were implementing new features for RuleSet.

src/import/MiniFSHImporter.ts
```diff
@@ -0,0 +1,122 @@
...
+
+class MiniFSHImporter extends MiniFSHVisitor {
+  private bracketParamUsage: RegExp;
+  plainParamUsage: RegExp;
+
+  constructor(private ruleSet: ParamRuleSet, private values: string[]) {
+    super();
+    const escapedParameters = this.ruleSet.parameters.map(escapeRegExp).join('|');
+    this.bracketParamUsage = new RegExp(
+      `\\[\\[{\\s*(${escapedParameters})\\s*}\\]\\]|{\\s*(${escapedParameters})\\s*}`,
```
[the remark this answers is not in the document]
mint-thompson: Yes, that's correct. Using double-brackets allows you to include leading or trailing space as part of the substituted value.

package.json
```diff
@@ -103,4 +103,4 @@
     "winston": "^3.3.3",
     "yaml": "^1.9.2"
   }
-}
+}
```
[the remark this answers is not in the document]
mint-thompson: I think that was done automatically by my formatter, so I'll revert it.

cmoesel (dismissed):
image

# Spoiler alert! Double Brackets win!

Nice job, @mint-thompson. I think this will be a very welcome feature for some of our users! Thanks for working through some tough issues!

jafeltra (approved):
This looks great to me! I couldn't think of any cases you haven't already covered. Yay!

src/fshtypes/rules/InsertRule.ts
```diff
@@ -17,7 +17,15 @@ export class InsertRule extends Rule {
   private fshifyParameters(): string {
     return this.params
       .map(originalParam => {
-        return originalParam.replace(/\\/g, '\\\\').replace(/,/g, '\\,').replace(/\)/g, '\\)');
+        if (/(^\s+)|(\s+$)|[,\)]/g.test(originalParam)) {
```
cmoesel, line 20: I don't think you need to escape `)` in a regex character class (e.g., inside `[` `]`). So this can be: `/(^\s+)|(\s+$)|[,)]/g`

src/fshtypes/rules/InsertRule.ts
```diff
@@ -17,7 +17,15 @@ export class InsertRule extends Rule {
...
+          return `[[${originalParam
+            .replace(/\\/g, '\\\\')
+            .replace(/(\]\]\s*),/g, '$1\\,')
+            .replace(/(\]\]\s*)\)/g, '$1\\)')}]]`;
```
cmoesel, line 24: I think you can swap the last two replaces with a single replace by making another capture group:
```ts
.replace(/(\]\]\s*)([,)])/g, '$1\\$2')
```

src/fshtypes/rules/InsertRule.ts
```diff
@@ -17,7 +17,15 @@ export class InsertRule extends Rule {
...
+        } else {
+          return originalParam.replace(/\\/g, '\\\\');
+        }
+        // return originalParam.replace(/\\/g, '\\\\').replace(/,/g, '\\,').replace(/\)/g, '\\)');
```
cmoesel, line 28: Is there a reason to keep this commented out code?

cmoesel (changes requested):
Just a few tiny things I think can be improved...

cmoesel (approved):
RE-APPROVED!

PR #1222, opened 2023-02-23
SUSHI init should default manualSliceOrdering to true

cmoesel:
For SUSHI 3.0.0, we want to encourage manual slice ordering, so set it to true for new projects created using `sushi init`.

To test, use `sushi init` and verify `manualSliceOrdering` is set correctly in the config.

Note that the config does not retain the empty comment line after the `manualSliceOrdering` line (even though it is in the file that init copies). This seems to be related to the yaml library we use and I couldn't get it to keep that blank comment line. I guess that's OK.


mint-thompson (approved):

jafeltra (approved):
That blank line bugged me, but it must just be something due to the fact that there are no other uncommented lines that come after it. Regardless, this looks good!

PR #1228, opened 2023-02-24
Support versions in instance meta.profile when setting exampleCanonical in IG

### LINKED ISSUE #1138: SUSHI does not properly output exampleCanonical for examples whose meta.profile contains a version
When SUSHI generates the IG JSON, it adds `exampleCanonical` for each instance that is an example of a profile in the IG.  This generally works, but if the example has a version appended in its `meta.profile` (e.g., `* meta.profile = "http://example.org/fhir/SomeProfile|1.0.0"`, then SUSHI fails to process it and uses `exampleBoolean` instead of `exampleCanonical`.

mint-thompson:
Fixes #1138 and completes task [CIMPL-1006](https://standardhealthrecord.atlassian.net/browse/CIMPL-1006).

When listing example instances in an IG, the instance's meta.profile is used if it is available. The URL in meta.profile may contain a version. If it does, check to make sure that the specified version matches the version of the profile. If so, set exampleCanonical to the unversioned URL. If the versions do not match, set exampleBoolean to true.


src/ig/IGExporter.ts
```diff
@@ -882,9 +882,13 @@ export class IGExporter {
       pkgResource instanceof InstanceDefinition &&
       pkgResource._instanceMeta.usage === 'Example'
     ) {
-      const exampleUrl = pkgResource.meta?.profile?.find(url => this.pkg.fish(url, Type.Profile));
+      const exampleUrl = pkgResource.meta?.profile?.find(url => {
+        const [baseUrl, version] = url.split('|', 2);
+        const availableProfile = this.pkg.fish(baseUrl, Type.Profile);
+        return version == null || version === availableProfile.version;
```
cmoesel, line 888: I think this should also be checking that `availableProfile` is defined -- otherwise this will return `true` in some cases we don't want it to.

E.g.
```
Profile: MyPatient
Parent: Patient
Description: "An example profile of the Patient resource."
* ^version = "0.1.0"
* name 1..* MS

Instance: PatientExampleWrongMeta
InstanceOf: MyPatient
* meta.profile[+] = "http://example.org/StructureDefinition/MyPatient|1.2.3"
* meta.profile[+] = "http://example.org/StructureDefinition/NotMyPatient"
* name
  * given[0] = "James"
  * family = "Pond"
```

This results in an IG resource w/ incorrect `exampleCanonical`:
```json
{
  "reference": {
    "reference": "Patient/PatientExampleWrongMeta"
  },
  "name": "PatientExampleWrongMeta",
  "description": "An example of a patient with a license to krill.",
  "exampleCanonical": "http://example.org/StructureDefinition/NotMyPatient"
}
```

cmoesel (changes requested):
I think this is _almost_ right. I found a test case that breaks the logic -- so I suggest you add the case to your unit test and fix the issue (which, if I understand it correctly, should be a very easy fix).

jafeltra (commented):
This looks good. I just noticed one other case that I think should be covered by this:

I think that we should support setting the `exampleCanonical` in the IG JSON for resources that are provided. There's a separate function for `addPredefinedResources` that I think should probably have the new logic as well. You can replicate this by including an Instance JSON in the `input/examples/` folder that has a `meta.profile` canonical URL with a version. The IG JSON just marks `exampleBoolean: true`, but I think it should include the `exampleCanonical` with the URL if the profile and specified version are available.

If you disagree and don't think we need to support this case, let me know.

cmoesel (approved):
Full regression passed. Looks good!

image

jafeltra (approved):
Looks good to me and fixes that additional case!

PR #1229, opened 2023-02-28
Update puns.ts

markkramerus:


src/utils/puns.ts
```diff
@@ -19,7 +19,24 @@ const CLEAN_RUN_PUNS = [
...
+  "You've earned a seal of approval.",
+  "You're swimming in success!",
+  "You're making waves now!",
+  "You've turned the tide.",
+  "You're catching on now!",
+  "You're in the swim of things now!",
+  "Catch of the day, right here!",
+  "That was a big fish to fry!",
+  "Nice job! Are you hooked yet?",
+  "You've found your sea legs!",
```
cmoesel, line 34: This is the same as line 23 above (except `!` instead of `.`).

src/utils/puns.ts
```diff
@@ -34,16 +51,26 @@ const WARNING_PUNS = [
...
+  'Let minnow if you need some help',
+  "You're kraken me up.",
+  "Don't be koi about asking for help.",
+  "You seem a bit rudderless.",
+  'Keep swimming, Dory.',
+  "You should gut it out.",
+  "Don't get stuck in the doldrums.",
+  "Are you in a no-wind situation?",
+  "Sailors, take warning!",
+  "You seem to be casting about.",
```
cmoesel, line 64: This is very similar to line 46.

src/utils/puns.ts
```diff
@@ -34,16 +51,26 @@ const WARNING_PUNS = [
...
+  "This looks a bit fishy."
```
cmoesel, line 65: This is very similar to line 44.

src/utils/puns.ts
```diff
@@ -56,9 +83,22 @@ const ERROR_PUNS = [
   'You are battering FSH',
   'Swim back to School of FSH',
   'You need to mullet over',
+  "You're krilling me!",
+  "You're fluking it up!",
```
cmoesel, line 87: Might this offend some of our more sensitive users?

src/utils/puns.ts
```diff
@@ -56,9 +83,22 @@ const ERROR_PUNS = [
   'You are battering FSH',
   'Swim back to School of FSH',
   'You need to mullet over',
```
cmoesel, line 85: I just noticed we also have `You should mullet over` in the warnings section. Maybe remove this one?

cmoesel (changes requested):
You, sir, are prolific.

I found a few duplicates...

src/utils/puns.ts
markkramerus: I thought it was kind of edgy... I'll cut it out.

src/utils/puns.ts
markkramerus: cut line 34

src/utils/puns.ts
markkramerus: Cut line 46

src/utils/puns.ts
markkramerus: Removed.

src/utils/puns.ts
```diff
@@ -56,9 +83,22 @@ const ERROR_PUNS = [
   'You are battering FSH',
```
mint-thompson, line 83: I've wanted to remove this one for awhile. Could we get rid of it?

mint-thompson (changes requested):

src/utils/puns.ts
markkramerus: No objection here. Let me do one more push.

cmoesel (commented):
This is failing CI because strings must be enclosed in single-quotes (not double-quotes). The only exception to this rule is if the string contains a single-quote, _then_ it can be wrapped in double-quotes.

The easiest way to fix this is just to run `npm run lint:fix` and the linter will fix it for you. Then add, commit, and push.

cmoesel (approved):
Thanks, @markkramerus. I expect our users will appreciate this more than any other feature we've released in the last six months.

mint-thompson (approved):

PR #1230, opened 2023-03-01
Check ElementDefinition invariants related to element name when adding elements to Logical or Resource

### LINKED ISSUE #1202: Support numeric element names in Logical Models
FHIR provides [Logical Modes](https://www.hl7.org/fhir/structuredefinition.html#logical) (scroll down, the link is [broken](https://jira.hl7.org/browse/FHIR-40368)) as a way to represent the structure of arbitrary JSON. FSH can be used to create a StructureDefinition representing a Logical Model.

Currently, SUSHI will not allow the following:

```
Logical: Testing
Title: "Testing"

* 1 0..1 string "Very important data"
``` 

This is necessary to model a JSON object like:

```json
{
  "1": "Some very important data"
}
```

It is possible to create a Logical Model using JSON with an element name of `"1"`:

```json
{
    "resourceType": "StructureDefinition",
    "id": "Testing",
    "url": "http://example.org/StructureDefinition/Testing",
    "version": "0.1.0",
    "name": "Testing",
    "title": "Testing",
    "status": "active",
    "fhirVersion": "4.0.1",
    "mapping": [
        {
            "identity": "rim",
            "name": "RIM Mapping",
            "uri": "http://hl7.org/v3"
        }
    ],
    "kind": "logical",
    "abstract": false,
    "type": "http://example.org/StructureDefinition/Testing",
    "baseDefinition": "http://hl7.org/fhir/StructureDefinition/Base",
    "derivation": "specialization",
    "differential": {
        "element": [
            {
                "id": "Testing",
                "path": "Testing",
                "short": "Testing"
            },
            {
                "id": "Testing.1",
                "path": "Testing.1",
                "short": "Very important data",
                "definition": "Very important data",
                "min": 0,
                "max": "1",
                "type": [
                    {
                        "code": "string"
                    }
                ]
            }
        ]
    }
}
```

This does result in a Publisher warning from [`eld-20`](http://hl7.org/fhir/elementdefinition-definitions.html), which states:

> Element names should be simple alphanumerics with a max of 64 characters, or code generation tools may be broken	
>
> `path.matches('^[A-Za-z][A-Za-z0-9]*(\\.[a-z][A-Za-z0-9]*(\\[x])?)*$')`

Can FSH take a similar approach at least for Logical Models where an element name that does not satisfy this regex raises a warning rather than an error?

(Note: the `eld-19` invariant on ElementDefinition defines what is allowable in an element name.)

mint-thompson:
Fixes #1202 and completes task [CIMPL-1064](https://standardhealthrecord.atlassian.net/browse/CIMPL-1064).

The specification for [ElementDefinition](http://hl7.org/fhir/elementdefinition-definitions.html#ElementDefinition) has two invariants related to an element's name.
- `eld-19` prohibits certain characters and enforces a maximum length of 64 (not including `[x]` or slice names)
- `eld-20` recommends a "simple alphanumeric" style, where the first character is in the set `A-Za-z` and the remaining characters are in the set `A-Za-z0-9`.

Note that while the description in `eld-20` mentions the length limit, it is the expression on `eld-19` that actually enforces it. Therefore, I considered an overly-long name to be a violation of `eld-19`.

When adding an element to a Logical or Resource, check the name. If `eld-19` is violated, log an error indicating if the violation is due to special characters, length, or both. If `eld-19` is fulfilled but `eld-20` is violated, log a warning recommending a simple alphanumeric. Regardless of any errors or warnings, the element will still be added. This is done to avoid additional errors from other rules referencing these elements.

The error and warning messages are taken from the specification. But, I feel like the warning message might not be specific or helpful enough, since "simple alphanumeric" is not, by my understanding, a formally defined term. I would like some suggestions on how to make this message more helpful.


src/export/StructureDefinitionExporter.ts
```diff
@@ -482,6 +482,20 @@ export class StructureDefinitionExporter implements Fishable {
...
+            nameErrors.push('cannot include some special characters');
+          }
+          if (lastIdPart.length > 64) {
+            nameErrors.push('must be at most 64 characters long');
+          }
+          if (nameErrors.length > 0) {
+            logger.error(`Element names ${nameErrors.join(' and ')}.`, rule.sourceInfo);
+          } else if (!/^[A-Za-z][A-Za-z0-9]*$/.test(lastIdPart)) {
+            logger.warn('Element names should be simple alphanumerics.', rule.sourceInfo);
+          }
```
cmoesel, line 498: I like the simple messages, but I think users would benefit from a bit more... I'd suggest you add something like `See eld-19 at https://hl7.org/fhir/elementdefinition.html#ElementDefinition-inv.` (and likewise for eld-20).

I also wonder if we should add the violating path to the message?

E.g.
> Invalid path: Foo.b@r. Element names cannot include some special characters. See eld-19 at https://hl7.org/fhir/elementdefinition.html#ElementDefinition-inv.

and:
> Unadvised path: Foo.bar_stool. Element names should be simple alphanumerics. See eld-20 at https://hl7.org/fhir/elementdefinition.html#ElementDefinition-inv.

(I don't love the word "unadvised" but having trouble coming up w/ a good word for not-recommended).

src/export/StructureDefinitionExporter.ts
```diff
@@ -482,6 +482,20 @@ export class StructureDefinitionExporter implements Fishable {
...
+          const nameErrors: string[] = [];
+          if (!/^[^\s\\.,:;\\'"\/|?!@#$%&*()\[\]{}]+(\[x\])?$/.test(lastIdPart)) {
+            nameErrors.push('cannot include some special characters');
+          }
+          if (lastIdPart.length > 64) {
+            nameErrors.push('must be at most 64 characters long');
+          }
+          if (nameErrors.length > 0) {
+            logger.error(`Element names ${nameErrors.join(' and ')}.`, rule.sourceInfo);
+          } else if (!/^[A-Za-z][A-Za-z0-9]*$/.test(lastIdPart)) {
```
cmoesel, line 496: Won't this warn on valid choice paths like `Foo.value[x]`? We shouldn't warn on those.

cmoesel (changes requested):
This is a good start -- and I like the approach. I just have a couple of recommendations (and one issue that I think needs to be addressed).

src/export/StructureDefinitionExporter.ts
```diff
@@ -482,6 +482,20 @@ export class StructureDefinitionExporter implements Fishable {
           // Note: newElement() method automatically adds the new element to its structDef.elements
           const newElement = structDef.newElement(rule.path);
           newElement.applyAddElementRule(rule, this);
+          const lastIdPart = newElement.id.slice(newElement.id.lastIndexOf('.') + 1);
+          // see http://hl7.org/fhir/elementdefinition-definitions.html#ElementDefinition eld-19 and eld-20
+          const nameErrors: string[] = [];
+          if (!/^[^\s\\.,:;\\'"\/|?!@#$%&*()\[\]{}]+(\[x\])?$/.test(lastIdPart)) {
```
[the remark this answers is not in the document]
mint-thompson: Ah, it's even more than that: I don't need any backslashes before the `.`, since that's already considered as a literal when used in a character class.

src/export/StructureDefinitionExporter.ts
mint-thompson: I think I'll go with "inadvisable," that's about the best I can think of.

src/export/StructureDefinitionExporter.ts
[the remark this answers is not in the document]
jafeltra: Oh even better! Good catch.

src/export/StructureDefinitionExporter.ts
jafeltra: I would have said `0/10 recommended path: Foo.bar_stool` but I like your choice better 😆

cmoesel (approved):
It's good!

image

PR #1233, opened 2023-03-08
Import and Parse Unicode Characters

### LINKED ISSUE #1209: SUSHI fails to parse strings w/ escaped unicode
If a string contains escaped unicode (e.g., `\u25BF`), the SUSHI parser chokes. In order to satisfy it, you must escape the `\` (e.g., `\\u25BF`) -- but then it's no longer a valid escape sequence in the JSON string and the unicode character is not rendered correctly. SUSHI needs to handle escape sequences better.

In addition, if you use `"""` multi-line strings, then you don't need to escape the `\` in `\u25BF` in your FSH -- but SUSHI auto-escapes it for you in the exported JSON.  Which is still wrong, because in the JSON string, it needs to be `\u25BF`, _not_ `\\u25BF`.

jafeltra:
Fixes #1209.

This PR adds support for importing unicode characters (e.g. `\u2603`) and then removing the escape `\` that is added, similar to how we handle `\r` or `\n`. This allows users to have a rule like this in their FSH:

```
* valueString = "This cool \u2603 can read music in \uD834\uDD1E!"
```

and it will be processed and included in the JSON with the unicode characters like this:

```json
{
  "id": "Observation.value[x]:valueString",
  "patternString": "This cool ☃ can read music in 𝄞!"
  ...
}
```

This is supported in both regular strings and multiline (`"""`) strings.


antlr/src/main/antlr/FSHLexer.g4
```diff
@@ -62,10 +62,10 @@ ARROW:              '->';
 // PATTERNS
 
                  //  "    CHARS    "
-STRING:             '"' (~[\\"] | '\\r' | '\\n' | '\\t' | '\\"' | '\\\\')* '"';
+STRING:             '"' (~[\\"] | '\\u' | '\\r' | '\\n' | '\\t' | '\\"' | '\\\\')* '"';
 
                  //  """ CHARS """
-MULTILINE_STRING:   '"""' .*? '"""';
+MULTILINE_STRING:   '"""' (. | '\\u')*? '"""';
```
cmoesel, line 68: Is this necessary? I reverted it back to `.*?`, rebuilt the grammar, and re-ran the tests -- and they all passed.

test/import/FSHImporter.test.ts
```diff
@@ -84,6 +84,33 @@ describe('FSHImporter', () => {
...
+    ].join('\n');
+    expect(profile.title).toEqual('Just going to \u270E in some unicode');
+    expect(profile.rules).toHaveLength(2);
+    assertAssignmentRule(
+      profile.rules[0],
+      'valueString',
+      'This cool \u2603 can read music in \uD834\uDD1E!'
+    );
+    assertAssignmentRule(profile.rules[1], 'category.text', expectedCategoryText);
+  });
```
cmoesel, line 112: I think you should add another test, but in this one use the real unicode characters.  E.g.:
```ts
    const input = leftAlign(`
    Profile: Escape
    Parent: Observation
    Title: "Just going to ✎ in some unicode"
    * valueString = "This cool ☃ can read music in 𝄞!"
    * category.text = """
      When there is a sign of ❄, often school is canceled.
      Kids can stay home and play with 🀱 or do anything!
    """
    `);
```

We want to make sure that works too. (I tested it manually and it seems to work.)

cmoesel (changes requested):
Thanks for figuring this out! I'm glad the solution ended up being pretty clean! I just have one question (which I think leads to a suggestion) and one suggestion for another test.

antlr/src/main/antlr/FSHLexer.g4
jafeltra: Hm I thought so, but I believe you, so I'll try it out again.

jafeltra: I thought I had hit a case with a parser error, but I haven't found that case again, so it must have been something different. Reverted this change in [f56425a](https://github.com/FHIR/sushi/pull/1233/commits/f56425a39cb4ca0491dd7bf630206391622956bd).

test/import/FSHImporter.test.ts
jafeltra: Added in [a490e74](https://github.com/FHIR/sushi/pull/1233/commits/a490e740b9eec3d9c12ea61cf51de7cc44a21950).

cmoesel (approved):
Excellent!

image

mint-thompson (approved):
hooray for unicode!

PR #1234, opened 2023-03-08
Automatically load hl7.fhir.uv.extensions for FHIR R5 (Draft Final and Final)

cmoesel:
Starting w/ FHIR 5.0.0-draft-final, FHIR no longer includes extensions in the core package. This PR will load the `hl7.fhir.uv.extensions` package automatically for IGs that target FHIR 5.0.0-draft-final or 5.0.0.  In the future, we'll update this to also support R4 once the R4 extensions package is ready. At that point, we'll also need to refine the loading strategy.


jafeltra (approved):
Tests look good and I ran it locally and only see the package being installed with R5.

mint-thompson (approved):

PR #1241, opened 2023-03-22
Support dependencies on specific CI build branches

cmoesel:
Add support for dependency versions that specify a current branch name. For example: `current$mybranchname`.

See: https://chat.fhir.org/#narrow/stream/179166-implementers/topic/Package.20cache.20-.20multiple.20dev.20versions/near/291131585

If you want to manually test this, add this dependency to a project:
```yaml
dependencies:
  hl7.fhir.be.mycarenet: current$issue-30
```

If you run SUSHI on the master branch, you'll get an error when it tries to download the dependency (assuming you don't have it cached locally).

If you run this PR branch, it will work. If you run SUSHI with `-d`, you'll see a debug message that the cached date matches the build date, so it will not download again.

### DIFF
--- a/src/fhirdefs/load.ts
+++ b/src/fhirdefs/load.ts
@@ -60,18 +60,28 @@ export async function loadDependency(
         `Downloading ${fullPackageName} since SUSHI cannot determine if the version in the local cache is the most recent build.`
       );
     }
-  } else if (version === 'current') {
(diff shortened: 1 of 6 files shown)


jafeltra (approved):
Makes sense to me! I tried out using the branch, as well as typical "current" and a version number, and everything worked as expected!

PR #1242, opened 2023-03-22
Allow assignment of Instance values in any entity using Assignment or Caret Value rules, even when Instance has a numeric or boolean id

### LINKED ISSUE #1184: SUSHI cannot support inline Instances that have numeric ids
Right now, it is not possible for SUSHI to assign inline Instances that have a `id` that is a number. When the Instance is assigned in an assignment rule, SUSHI logs an error that the type `number` does not match the `Resource` type that it is being assigned to.

This is likely an issue with how the importer in SUSHI detects if a value that is being assigned is referencing an Instance.

An example of this is here: https://fshschool.org/FSHOnline/#/share/3UqVw4K

mint-thompson:
Fixes #1184 and completes task [CIMPL-1051](https://standardhealthrecord.atlassian.net/browse/CIMPL-1051).

Rules that assign a value, with or without a caret path, may specify an Instance as the value to assign. This applies to Structure Definitions, Code Systems, Value Sets, and Instances.

An Instance may have an id that is parsed as a numeric or boolean value. In these cases, if assignment of the numeric value fails, try to find an Instance by fishing using that numeric or boolean value. If an Instance is found, try to assign that Instance. The "raw value" of the number or boolean from the rule is used when fishing. For example, if the number 150 was specified using exponent notation as "15E1", then an Instance with id "15E1" may be found. An Instance with id "150" would not be found.

If an Instance with the numeric or boolean id is found, but the assignment or validation of the Instance value fails, the error message to log depends on what kind of failure occurred. If a type mismatch occurred, log an error message with the original type of the value, not the Instance type. If a different type of error occurred when assigning the Instance, log a message for that error.

When printing the FSH representation of Assignment or Caret Value rules, use the raw numeric or boolean value.

The code that handles retrying the assignment in the various exporters is pretty similar in all cases, but there are some small differences that led to me only being able to create a reusable function for `CodeSystemExporter` and `ValueSetExporter`.

This ended up going a bit beyond the original reported issue, but it felt correct to make instance assignment work better in all the places where it is allowed. One other point that I noticed, but did not include, is that some keywords are still not permitted as entity names. Should this also be changed, and if so, should that be included in this PR, or as part of a separate task (such as [the one to allow a path named "exclude"](https://standardhealthrecord.atlassian.net/browse/CIMPL-1073))?


jafeltra (approved):
This all looks really good to me. This was a tricky problem, and I think the solution makes a lot of sense. I couldn't think of any other edge cases that you haven't already accounted for.

Also to answer this question in case we haven't already:

> This ended up going a bit beyond the original reported issue, but it felt correct to make instance assignment work better in all the places where it is allowed. One other point that I noticed, but did not include, is that some keywords are still not permitted as entity names. Should this also be changed, and if so, should that be included in this PR, or as part of a separate task (such as [the one to allow a path named "exclude"](https://standardhealthrecord.atlassian.net/browse/CIMPL-1073))?

I think keeping keyword names as a separate task is the way to go here. This was specific enough to numbers and hopefully sets up a framework that could be reused, so I don't think it is worth expanding the scope of this task/PR to include it. So I vote we create a task and prioritize it as usual.

src/export/InstanceExporter.ts
```diff
@@ -184,8 +185,53 @@ export class InstanceExporter implements Fishable {
...
+                  `Unable to assign value at ${rule.path}: choice elements on an instance must use a specific type`,
+                  rule.sourceInfo
+                );
+              } else {
+                ruleMap.set(rule.path, {
+                  pathParts: validatedRule.pathParts,
+                  assignedValue: validatedRule.assignedValue,
+                  sourceInfo: rule.sourceInfo
+                });
+              }
```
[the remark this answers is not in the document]
cmoesel: I was so curious about the looping idea that I tried it and (a) it was kind of messy, and (b) it broke tests. Oh well. But pulling out a function seems OK. I replaced lines 143 - 235 (the body of the `rules.forEach` block) with this and all went well:
```ts
      const inlineResourceTypes: string[] = [];
      // define function that will be re-used in attempting to assign a value or inline instance
      const doRuleValidation = (value: AssignmentValueType) => {
        const validatedRule = instanceOfStructureDefinition.validateValueAtPath(
          rule.path,
          value,
          this.fisher,
          inlineResourceTypes,
          rule.sourceInfo,
          manualSliceOrdering
        );
        // Record each valid rule in a map
        // Choice elements on an instance must use a specific type, so if the path still has an unchosen choice element,
        // the rule can't be used. See http://hl7.org/fhir/R4/formats.html#choice
        if (validatedRule.pathParts.some(part => part.base.endsWith('[x]'))) {
          logger.error(
            `Unable to assign value at ${rule.path}: choice elements on an instance must use a specific type`,
            rule.sourceInfo
          );
        } else {
          ruleMap.set(rule.path, {
            pathParts: validatedRule.pathParts,
            assignedValue: validatedRule.assignedValue,
            sourceInfo: rule.sourceInfo
          });
        }
      };

      try {
        const matchingInlineResourcePaths = inlineResourcePaths.filter(
          i => rule.path.startsWith(`${i.path}.`) && rule.path !== `${i.path}.resourceType`
        );
        // Generate an array of resourceTypes that matches the path, so if path is
        // a.b.c.d.e, and b is a Bundle and D is a Patient,
        // inlineResourceTypes = [undefined, "Bundle", undefined, "Patient", undefined]
        matchingInlineResourcePaths.forEach(match => {
          inlineResourceTypes[splitOnPathPeriods(match.path).length - 1] = match.instanceOf;
        });
        // if a rule has choice elements in the path that are constrained to a single type,
        // change the path to reflect the available type for each choice element.
        // if the path changes, warn the user that it is a better idea to use the more specific element name
        const updatedPath = instanceOfStructureDefinition.updatePathWithChoices(rule.path);
        if (rule.path !== updatedPath) {
          logger.warn(
            `When assigning values on instances, use the choice element's type. Rule path changed from '${rule.path}' to '${updatedPath}'.`,
            rule.sourceInfo
          );
          rule.path = updatedPath;
        }
        doRuleValidation(rule.value);
      } catch (originalErr) {
        // if an Instance has an id that looks like a number, bigint, or boolean,
        // we may have tried to validate with that value instead of an Instance.
        // try to fish up an Instance with the rule's raw value.
        // if we find one, try validating with that instead.
        if (
          originalErr instanceof MismatchedTypeError &&
          ['number', 'bigint', 'boolean'].includes(typeof rule.value)
        ) {
          const instanceToAssign = this.fishForFHIR(rule.rawValue);
          if (instanceToAssign == null) {
            logger.error(originalErr.message, rule.sourceInfo);
          } else {
            try {
              doRuleValidation(instanceToAssign);
            } catch (instanceErr) {
              if (instanceErr instanceof MismatchedTypeError) {
                logger.error(originalErr.message, rule.sourceInfo);
              } else {
                logger.error(instanceErr.message, rule.sourceInfo);
              }
            }
          }
        } else {
          logger.error(originalErr.message, rule.sourceInfo);
        }
      }
```

src/export/InstanceExporter.ts
```diff
@@ -184,8 +185,53 @@ export class InstanceExporter implements Fishable {
...
+      } catch (originalErr) {
+        // if an Instance has an id that looks like a number, bigint, or boolean,
+        // we may have tried to validate with that value instead of an Instance.
+        // try to fish up an Instance with the rule's raw value.
+        // if we find one, try validating with that instead.
+        if (
+          originalErr instanceof MismatchedTypeError &&
+          (typeof rule.value === 'number' ||
+            typeof rule.value === 'bigint' ||
+            typeof rule.value === 'boolean')
```
cmoesel, line 197: Wherever you have this (I think it's at least four places), consider this instead:
```ts
['number', 'bigint', 'boolean'].includes(typeof rule.value)
```
It's not really a big deal, but `prettier` let's it all fit on one line. But this is a suggestion only. Feel free to ignore.

src/export/StructureDefinitionExporter.ts
```diff
@@ -544,7 +548,41 @@ export class StructureDefinitionExporter implements Fishable {
...
+                    element.assignValue(instance, rule.exactly, this);
+                    rule.value = instance;
+                  } catch (instanceErr) {
+                    // if it's still the wrong type, the assignment will fail.
+                    // but, we want to log that error with the original type, not the instance type.
+                    if (instanceErr instanceof MismatchedTypeError) {
+                      throw originalErr;
+                    } else {
+                      throw instanceErr;
+                    }
```
cmoesel, line 579: Can we just always throw `originalErr` here? What type of exception could we get in this block (when we're _trying_ to assign an instance) for which we wouldn't want to just fall back to the original error anyway?

src/export/StructureDefinitionExporter.ts
```diff
@@ -637,30 +694,66 @@ export class StructureDefinitionExporter implements Fishable {
...
-            if (fishedValue instanceof InstanceDefinition && fishedValue.resourceType) {
-              try {
-                sd.setInstancePropertyByPath(rule.caretPath, fishedValue, this);
-              } catch (e) {
+          // if we can find the exported instance, it will have some useful metadata.
+          const instanceExporter = new InstanceExporter(this.tank, this.pkg, this.fisher);
+          const exportedInstance = instanceExporter.fishForFHIR(fishItem);
+          if (exportedInstance != null) {
+            fishedValue = exportedInstance;
+          }
```
cmoesel, line 718: If the `InstanceExporter` returns a result, then you re-assign `fishedValue` without having ever used its original value. If the preferred source is the `InstanceExporter`, then wouldn't it be better to _start_ with that on line 708? And then if it returns null, fall back to `this.fishForFHIR` and (assuming it returns something), use `InstanceDefinition.fromJSON(fishedValue)` (see line 737) to turn it into an `InstanceDefinition`.

I might be missing something, but I feel like if you start at line 708 with:
```ts
const instanceExporter = new InstanceExporter(this.tank, this.pkg, this.fisher);
let fishedValue = instanceExporter.fishForFHIR(fishItem);
if (fishedValue == null) {
  const result = this.fishForFHIR(fishItem);
  if (!(result instanceof InstanceDefinition) && result instanceof Object) {
    fishedValue = InstanceDefinition.fromJSON(fishedValue);
  }
}
```
I feel like that might actually allow you to get rid of a lot of repeated code here. (But again, I may be overlooking something).

src/export/StructureDefinitionExporter.ts
```diff
@@ -637,30 +694,66 @@ export class StructureDefinitionExporter implements Fishable {
...
+          const needsResourceType = /^contained(\[\d+\])?$/.test(rule.caretPath);
+          if (needsResourceType && fishedValue.resourceType == null) {
+            if (originalErr != null) {
+              logger.error(originalErr.message, rule.sourceInfo);
+            } else {
+              logger.error(`Could not find a resource named ${rule.value}`, rule.sourceInfo);
+            }
```
cmoesel, line 725: What's this stuff with `contained`? Is this addressing a previously undiscovered bug?

cmoesel (changes requested):
Cool. Cool. Cool. And Tricky. Tricky. Tricky.

Nice job untangling all that had to be untangles. I left a few comments / suggestions that _might_ help to reduce some lines of code and keep things a little DRYer. Of course, determining what is ok to repeat and what is not ok to repeat is subjective, so let me know what you think.

src/export/StructureDefinitionExporter.ts
cmoesel: Maybe not. I don't know. This is an honest question. If we're OK just always throwing `originalErr` when the inline instance assignment fails _for any reason_, then I think there are a bunch of `if`/`else` blocks in several files that can be simplified. So it's tempting... but only if we don't think there's much to lose.

mint-thompson: I was mostly considering `ValueAlreadyAssignedError` here, which would come up when the instance is the right type. I think in that case, it's more helpful to let the author know about the value conflict rather than an apparent type error which SUSHI has already managed to work through.

cmoesel: Ah, that makes sense. I agree.

src/export/CodeSystemExporter.ts
```diff
@@ -115,14 +116,40 @@ export class CodeSystemExporter {
...
+          const instance = instanceExporter.fishForFHIR(rule.value as string);
+          if (instance == null) {
+            logger.error(
+              `Cannot find definition for Instance: ${rule.value}. Skipping rule.`,
+              rule.sourceInfo
+            );
+            continue;
+          }
+          rule.value = instance;
+        }
```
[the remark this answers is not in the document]
cmoesel: Ah, right.  I missed the `continue`. Good point. There's limited value in reducing repetition if it introduces toomuch additional complexity.  I guess the function could return a boolean indicating if it succeeded or not. Or it could throw and you could catch -- but that seems a bit much, doesn't it? I think I'd prefer the boolean return over the thrown error. Anyway -- if there is a reasonable approach, great. If not, thanks for considering it!

src/export/StructureDefinitionExporter.ts
mint-thompson: So, this is here in order to keep the test in `FHIRExporter.test.ts:139` working the same. The previous implementation here always needed something in `fishedValue.resourceType`, but now that assigning a non-resource Instance is allowed, I added this check on the path.

Thinking about it more, though, we can probably do without this check. It will mean that a different error is logged if a non-Instance resource is found when assigning to `contained`: specifically, it won't be the right type, since it isn't a descendant of `Resource`. But that's actually a fine thing to tell the author, unless I've overlooked something.

jafeltra (approved):
This looks good to me! I like the refactors, they really simplify things.

I haven't re-run a full migration, but I can if that would be useful!

cmoesel (approved):
Good work! I re-ran regression and confirmed that there are only a few changes -- and @mint-thompson has very helpfully submitted PRs for all of them!

It's hard to find good images to represent instances. I just went for this one because the people in it look _so cool_.

image

PR #1244, opened 2023-03-23
Allow most FSH keywords to be used as paths.

mint-thompson:
Completes task [CIMPL-1073](https://standardhealthrecord.atlassian.net/browse/CIMPL-1073).

FSH keywords that are alphanumeric and don't cause any lexer mode changes are safe to use as paths. The main important path is "exclude," which appears on Group resources. But, other paths are allowed names for elements added to Logical or Resource definitions.

Eventually, we want to add support for a path named "insert," which may require things like quoting paths. But, for now, this is useful enough, since "exclude" appears on a core resource type.

I only added a special test for "obeys", since paths are optional in an obeys rule. With other rule types that use a keyword after a path, the path is required, so it shouldn't look ambiguous.

This change resulted in a slightly different error when a Value Set component uses both include and exclude. But, the error still refers to the correct line, so I feel like this is acceptable.

Full regression is currently in progress, I'll post in this thread when it finishes.


mint-thompson:
Full regression set is complete. Only one repo showed differences (HL7NZ/mdr#master), and rerunning on it had no difference. So, I don't think that's anything to do with this PR.

mint-thompson:
> @mint-thompson - It looks like your double-brackets PR caused conflicts on this one.

Thanks for the heads-up! The conflicts are resolved now.

cmoesel (commented):
This PR looks good and it ended up being a lot less messy than I feared it might be.

My only slight hesitation is how this might tie our hands in the future. Does supporting all (or most) keywords as paths mean that we can't allow a true keyword as the start of a rule in the future? For example, would something like this cause a problem since `from` could be a path?
```
* code
  * from MyValueSet
```

I think the answer for this example is "no, it's not a problem" -- because the second rule would still only match a parser rule for a binding. And indeed, we already have this situation, as this already works with this PR:
```
* code
  * obeys MyInvariant
```

That said, if you're really naughty, you could name your invariant `MS` -- in which case you end up with:
```
* code
  * obeys MS
```
Now it is amguous. Is it an `obeys` rule to specify an invariant on `code`? Or is it a `flag` rule to specify `must-support` on `code.obeys`? As it turns out, SUSHI thinks it is the latter:
> No element found at path code.obeys for FlagRule in MyObservationProfile, skipping rule

So... now the question is: should we be concerned? Do we want to support any and all keywords this way? Or only those that are in the core FHIR model? Would it be better to say that when there is a path name conflict with a keyword, you should wrap the path in quotes? Then this would be unambigous:
```
* code
  * "obeys" MS
```

I'm not sure what the right answer is. 99.9% of the time, I suspect that the implementation as it is in this PR will not be a problem. Do we worry about the 0.1%?

test/import/FSHImporter.Logical.test.ts
```diff
@@ -613,6 +626,185 @@ describe('FSHImporter', () => {
         expect(loggerSpy.getAllLogs('error')).toHaveLength(0);
       });
 
+      it('should parse add element rules where the element name is a FSH keyword', () => {
```
jafeltra, line 629: This whole test is so trippy but fun.

jafeltra (approved):
Wow yeah, that was surprisingly straightforward! Neat!

cmoesel (approved):
image

PR #1245, opened 2023-03-24
Populate exampleCanonical regardless of meta.profile configuration

### LINKED ISSUE #1243: Turning off meta.profile should not disassociate instance w/ profile in IG resource
SUSHI allows users to turn off the generation of `meta.profile` via `sushi-config.yaml`:
```yaml
instanceOptions:
  setMetaProfile: never
```

When this happens, the `exampleCanonical` should still be set for the instance in the IG JSON, but instead, only exampleBoolean is being set -- so it is no longer associated at all with the profile it is an instance of.

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/No.20profile.20declaration.3F/near/343571355

mint-thompson:
Fixes #1243 and completes task [CIMPL-1080](https://standardhealthrecord.atlassian.net/browse/CIMPL-1080).

The configuration option for controlling whether `meta.profile` on Instances automatically gets an entry based on what it is an `InstanceOf` means that you can get examples without anything in their `meta.profile`. But, if they're examples of Profiles from the current package, set `exampleCanonical` in the IG to the URL of that Profile.

The small regression set showed a change in [HL7/davinci-epdx](https://github.com/HL7/davinci-epdx). The [PdexPriorAuth](https://github.com/HL7/davinci-epdx/blob/master/input/fsh/ExamplePriorAuthEOB.fsh) Instance has its `meta.profile` manually set to a versioned URL, which doesn't match the actual version of the [PdexPriorAuthorization](https://github.com/HL7/davinci-epdx/blob/master/input/fsh/ProfilePdexPriorAuthorization.fsh) profile. But, since it's an Instance of PdexPriorAuthorization, we can safely set `exampleCanonical` to PdexPriorAuthorization's URL.

### DIFF
--- a/src/export/InstanceExporter.ts
+++ b/src/export/InstanceExporter.ts
@@ -703,6 +703,8 @@ export class InstanceExporter implements Fishable {
         }
       }
     }
+    // regardless of what we did with meta.profile, we may need the instanceOfStructureDefinition url later
+    instanceDef._instanceMeta.instanceOfUrl = instanceOfStructureDefinition.url;
     instanceDef.validateId(fshDefinition);
     this.validateRequiredElements(
       instanceDef,
--- a/src/fhirtypes/InstanceDefinition.ts
+++ b/src/fhirtypes/InstanceDefinition.ts
@@ -99,6 +99,7 @@ type InstanceMeta = {
   description?: string;
   usage?: InstanceUsage;
   sdType?: string;
+  instanceOfUrl?: string;
 };
 
 // eslint-disable-next-line @typescript-eslint/no-empty-interface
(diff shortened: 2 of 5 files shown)


cmoesel (approved):
Just what the doctor ordered!

image

jafeltra (approved):
This makes sense to me! I was initially questioning a couple things but I think I decided both of them are fine:
- Should we just let people assign a profile with a version we can't find? But the answer is - we decided _no_ a while ago!

Looks good to me!
- Should there be a way to have instances _not_ be an example of what they're an `InstanceOf`? But I couldn't think of any use case for that, so probably not worth it.

PR #1248, opened 2023-03-27
Change hl7.fhir.uv.extensions autoload to use 'latest' version

cmoesel:
With the release of FHIR R5, the `hl7.fhir.uv.extensions` package got its first official release (1.0.0). As per [this conversation](https://chat.fhir.org/#narrow/stream/179239-tooling/topic/New.20Implicit.20Package/near/327872896), SUSHI and IG Publisher should automatically load the _latest published_ version of `hl7.fhir.uv.extensions`. Prior to this PR, we autoloaded the `current` version instead.

Note that I needed to enhance the logic for detecting `latest` so that it will fall back to `packages2.fhir.org` if there is no manifest for the package on `packages.fhir.org` (which is the case for `hl7.fhir.uv.extensions`).

To test this manually, create an R5 IG. When you run the `master` branch, you'll see it pull down the `current` version of `hl7.fhir.uv.extensions`.  When you run this PR branch, you should see it pull down version `1.0.0` instead.

### DIFF
--- a/src/utils/Processing.ts
+++ b/src/utils/Processing.ts
@@ -57,8 +57,11 @@ export const AUTOMATIC_DEPENDENCIES: AutomaticDependency[] = [
     version: 'latest'
   },
   {
+    // Right now, auto-load the extensions package for R5 only. In the future, we'll do this for R4 as well, but as
+    // of 3/27/2023, the rules for resolution in R4 have not yet been determined.
+    // See: https://chat.fhir.org/#narrow/stream/179239-tooling/topic/New.20Implicit.20Package/near/344938535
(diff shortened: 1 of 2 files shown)


mint-thompson (approved):
All good on the backport, and all good over here as well. Thank you!

jafeltra (approved):
This makes sense to me. I wonder if we should pull out the check for the dist-tags into a function in FPL that SUSHI can use, but that is a thought for another day.

cmoesel:
Yeah, we may want to consider adding support for "latest" as a version in FPL. It seems it could have usefulness elsewhere too. But as you said, that is a thought for another day!

PR #1249, opened 2023-03-27
Improve import & export of ids/extensions on primitives (v2.x)

cmoesel:
In FHIR JSON, ids and extensions on primitives are represented by creating a shadow property prefixed with an underscore. For example, to set an extension on `StructureDefinition.baseDefinition`, you assign and object containing `extension` to `StructureDefinition._baseDefinition`.  For more information see [JSON representation of primitive elements](https://hl7.org/fhir/json.html#primitive).

While SUSHI partially supported setting ids and extensions on primitive elements using FSH assignment rules, it _dropped_ all properties starting with `_` when it imported an SD using `StructureDefinition.fromJSON(...)`. It also did not propertly consider `_` properties when calculating differentials.

This PR adds support for importing `_` properties in `StructureDefinition.fromJSON` and improves support for exporting `_` properties via `StructureDefinition.toJSON` and detecting differentials. Once these improvements were made, however, additional adjustments were required:
* When `ElementDefinition.type.profile` or `ElementDefinition.type.targetProfile` arrays are manipulated, the shadow `_profile` and `_targetProfile` arrays must also be adjusted to maintain the relationship between indices. In addition, when a profile or targetProfile is changed, the original extensions may no longer apply, so they are removed.
* When a SD is used as a parent and that SD has `_`-prefixed properties, we strip the `_`-prefixed properties because in many (or most) cases, they don't necessarily apply to the child SD.

_**NOTE: This is a SUSHI 2.x port of #1240, cherry-picked from its commit.**_

### DIFF
--- a/src/export/StructureDefinitionExporter.ts
+++ b/src/export/StructureDefinitionExporter.ts
@@ -394,6 +394,15 @@ export class StructureDefinitionExporter implements Fishable {
       delete structDef.context;
       delete structDef.contextInvariant;
     }
+
+    // Remove all top-level properties that start with _, as the top-level primitive
+    // extensions on the parent may not be valid for the child.
+    Object.keys(structDef).forEach(key => {
+      if (key.startsWith('_')) {
+        // @ts-ignore
+        delete structDef[key];
+      }
+    });
   }
 
   /**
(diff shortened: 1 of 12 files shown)


jafeltra (approved):
Wow a clean cherry-pick. What an exciting moment ✨

Issue #1204, opened 2023-02-02
Support specifying a version within a contains rule for extensions


jafeltra:
Currently, you cannot specify a version of an extension when using a contains rule to add an extension. It would be good to be able to specify the version directly in contains rule using either the extension name, id, or url as you would do in any other contains rule.

This would mean that FSH such as this:

```
* extension contains USCoreEthnicityExtension|5.0.1 named example-ext 1..1
```

should result in a profile with the following differential element:

```json
{
  "id": "Condition.extension:example-ext",
  "path": "Condition.extension",
  "sliceName": "example-ext",
  "min": 1,
  "max": "1",
  "type": [
    {
      "code": "Extension",
      "profile": [
        "http://hl7.org/fhir/us/core/StructureDefinition/us-core-ethnicity|5.0.1"
      ]
    }
  ]
}
```

FSH Online example that demonstrates current behavior: https://fshschool.org/FSHOnline/#/share/3HrN4hi
Originally reported on Zulip: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/How.20to.20add.20extension.20to.20a.20profile.20with.20canonical.20version

mint-thompson:
Fixed by #1251

PR #1251, opened 2023-03-30
Allow versioned URL in extension contains rule

mint-thompson:
Completes task [CIMPL-1065](https://standardhealthrecord.atlassian.net/browse/CIMPL-1065).

A version can be specified after a pipe. The version is not used when fishing. If the type and extension both have a version, log a warning if they don't match.

I ran the small regression repo set against this branch, and no changes appeared.

### DIFF
--- a/src/export/StructureDefinitionExporter.ts
+++ b/src/export/StructureDefinitionExporter.ts
@@ -689,13 +689,21 @@ export class StructureDefinitionExporter implements Fishable {
     }
     rule.items.forEach(item => {
       if (item.type) {
-        const extension = this.fishForFHIR(item.type, Type.Extension);
+        // there might be a |version appended to the type, so don't include it while fishing
+        const [typeWithoutVersion, version] = item.type.split('|', 2);
+        const extension = this.fishForFHIR(typeWithoutVersion, Type.Extension);
         if (extension == null) {
           logger.error(
             `Cannot create ${item.name} extension; unable to locate extension definition for: ${item.type}.`,
             rule.sourceInfo
           );
           return;
+        }
+        if (version != null && extension.version != null && version != extension.version) {
+          logger.warn(
+            `${item.type} is based on ${typeWithoutVersion} version ${version}, but SUSHI found version ${extension.version}`,
+            rule.sourceInfo
+          );
         }
         try {
           const slice = element.addSlice(item.name);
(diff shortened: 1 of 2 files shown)


src/export/StructureDefinitionExporter.ts
```diff
@@ -689,14 +689,22 @@ export class StructureDefinitionExporter implements Fishable {
...
         if (extension == null) {
           logger.error(
             `Cannot create ${item.name} extension; unable to locate extension definition for: ${item.type}.`,
             rule.sourceInfo
           );
           return;
         }
+        if (version != null && extension.version != null && version != extension.version) {
+          logger.warn(
+            `${item.type} is based on ${typeWithoutVersion} version ${version}, but SUSHI found version ${extension.version}`,
```
cmoesel, line 704: For some reason the phrase *based on* just doesn't sound right to me. Might it be more clear if we said:
> The ${typeWithoutVersion} extension was specified with version ${version}, but SUSHI found and used version ${extension.version}

Or:

> The extension ${item.type} was specified, but SUSHI found and used ${typeWithoutVersion}|${extension.version}

cmoesel (changes requested):
This looks like it should work (although I have not manually tested), but... I have one suggestion/request regarding the wording of the message...

src/export/StructureDefinitionExporter.ts
jafeltra: I'm not sure if this factors in at all, but it is worth noting that this is the same wording we have in the two other places we have a similar message in ElementDefinition [here](https://github.com/FHIR/sushi/blob/master/src/fhirtypes/ElementDefinition.ts#L1245-L1249) and [here](https://github.com/FHIR/sushi/blob/master/src/fhirtypes/ElementDefinition.ts#L2325-L2329).

jafeltra (commented):
When I tested this locally, the version was able to be parsed and I no longer get an error from it, but the version isn't included in the differential in the output. I think this task is addressing issue #1204, and I used the FSH linked there, and didn't see the version come up in the `type.profile` in the differential. Is that expected to work on this branch?

src/export/StructureDefinitionExporter.ts
cmoesel: Ah. I see. I should have guessed! The other ones are a little different because they're dealing w/ canonicals we pulled out of definitions (either `SD.baseDefinition` when traversing the hierarchy or `ED.type.profile` when unfolding); whereas this one is a canonical version explicitly specified by the author.  The "based on" wording feels a little weird to me in all of them -- but I don't think they necessarily should all get the same fixed up language.

I'd still suggest we update the message here, but leave it to Mint to decide if they want to update the other messages as well (as they're clearly outside the scope of this PR).

cmoesel (approved):
Hurray for versioned extensions!
image

jafeltra (approved):
Looks good to me!

PR #1252, opened 2023-03-31
Parse Path Rules on Instances

### LINKED ISSUE #1181: SUSHI should not discard path rules in the importer

jafeltra:
This PR adds support for keeping Path Rules on Instances when they are imported.

This allows Path Rules to be used to define the order of slices when using the `manualSliceOrdering` configuration property. An example of FSH you can use to test this case is [here](https://fshschool.org/FSHOnline/#/share/3U1vab6), although it is probably more helpful to test this locally. With `master`, when you have `manualSliceOrdering` set to `true`, the path rule for `category[lab]` has no effect and the `category[+]` rule for the Activity slice will overwrite the fixed value for the Lab slice. On this branch, with `manualSliceOrdering` set to `true`, the Lab slice comes first and the Activity slice is added after it. When `manualSliceOrdering` is set to `false`, the Lab slice still is overwritten by the Activity slice, which is expected.

This also allows Path Rules to be used to add any required fixed values on optional elements if the path is used in a Path Rule. An example of FSH you can use to test this case is [here](https://fshschool.org/FSHOnline/#/share/3KkEFh9), although it is probably more helpful to test this locally too. With `master`, the path rules for `dataAbsentReason` and `identifier` will have no effect. On this branch, they will include the fixed values for `dataAbsentReason` and for the required `identifier.use`.

I ran a full regression, and the only repos with changes are `HL7/carin-digital-insurance-card#master`, `standardhealth/fsh-icare#master`, `CEOsys/cpg-on-ebm-on-fhir#master`, which all just rearrange slices to be the order they are defined on the extension when a path rule is used, and `hl7-it/lab-report` repo, which has an extra `null`  included where a path rule with no required properties was used before a soft index. We decided this regression is okay. I also ran a full regression with `manualSliceOrdering` forced to be on, and the only regression was in the `hl7-it/lab-report` repo. 

Fixes #1181

Related FSH Spec PR here: https://github.com/HL7/fhir-shorthand/pull/152

### DIFF
--- a/src/export/InstanceExporter.ts
+++ b/src/export/InstanceExporter.ts
@@ -28,7 +28,7 @@ import { InstanceOfLogicalProfileError } from '../errors/InstanceOfLogicalProfil
 import { AbstractInstanceOfError } from '../errors/AbstractInstanceOfError';
 import { MismatchedTypeError } from '../errors/MismatchedTypeError';
 import { Package } from '.';
-import { cloneDeep, isEqual, isMatch, merge, padEnd, uniq, upperFirst } from 'lodash';
+import { cloneDeep, isEmpty, isEqual, isMatch, merge, padEnd, uniq, upperFirst } from 'lodash';
 import { AssignmentRule, AssignmentValueType, PathRule } from '../fshtypes/rules';
 import chalk from 'chalk';
 
@@ -432,7 +432,9 @@ export class InstanceExporter implements Fishable {
             arrayEl?._sliceName?.toString().startsWith(`${child.sliceName}/`)
         );
         instanceChild.forEach((arrayEl: any) => {
-          if (arrayEl != null) this.validateRequiredChildElements(arrayEl, child, fshDefinition);
+          if (arrayEl != null && !isEmpty(arrayEl)) {
+            this.validateRequiredChildElements(arrayEl, child, fshDefinition);
+          }
         });
       } else if (instanceChild != null) {
         this.validateRequiredChildElements(instanceChild, child, fshDefinition);


mint-thompson:
I tried a variation on the second FSH example you provided, and I have a question about the results. Since `Observation.identifier` is `0..*`, I tried adding another entry to the list:
```
Profile: OptionalObservation
Parent: Observation
// identifier is 0..*
* identifier.use 1..1
* identifier.use = #official
* dataAbsentReason = #unknown

Instance: MyOptionalObservation
InstanceOf: OptionalObservation
* dataAbsentReason
* identifier[0]
* identifier[1]
* code = #123
* status = #final
```
But, the resulting `identifier` array only had one element. I expected there to be two elements, since I had path rules for two elements. Is this expectation correct?

jafeltra:
Ahh this seems to be happening when `manualSliceOrdering` is `true`, so I think the case I changed in that logic is probably incorrect. I think you're probably right that two elements should be expected, so I'll take a look.

jafeltra:
@mint-thompson - [f285f46](https://github.com/FHIR/sushi/pull/1252/commits/f285f46041909e8ab67499f5328ac240069db866) updates the logic for removing arrays with all null values in the `cleanResource` step, which let me remove some of the other logic I had added in to try to avoid having them added in the first place, which caused the issue you found.

I re-ran regression, and the same 3 repos had differences still, which is good. I think this is ready for review again.

src/export/InstanceExporter.ts
mint-thompson, line 1: Now that `PathRule` is an allowed type for a rule, I feel a little weird about the type assertion on line 84:
```ts
let rules = fshInstanceDef.rules.map(r => cloneDeep(r)) as AssignmentRule[];
```
Everything ends up working fine, since the functions dealing with `AssignmentRule` can all handle situations where the rule's value is `null`. The type assertion is not as true as it used to be, since `PathRule` can now appear. But, because the validation function is always okay with a null value, it means that we get the extremely convenient behavior of getting all the implied required values out when sending a `PathRule` to `validateValueAtPath`. So I guess what I'm considering is: would it be worth the trouble of properly recognizing that sometimes a `PathRule` is being used here?

mint-thompson (changes requested):
Thank you for the updates! I found one scenario that I think we're handling in an acceptable way, but I wanted to point it out because it's a little weird, maybe? Consider this `Instance`:
```
Instance: RegularObservation
InstanceOf: Observation
* component[0]
* code = #123
* status = #final
```
The path rule `component[0]` means that it tries to create an entry. There are no values to assign from the Observation SD, so the entry is an empty object. Since `component.code` is `1..1`, SUSHI gives this error:
```
error Element Observation.component.code has minimum cardinality 1 but occurs 0 time(s).
```
This is probably fine, since that path rule is just sort of hanging out doing nothing, and that probably means the author will want to change something. But, it's also maybe a little weird, since there isn't an actual element there (or even a `component` array) in the exported instance.

test/export/InstanceExporter.test.ts
```diff
@@ -4424,6 +4425,132 @@ describe('InstanceExporter', () => {
...
+        const assignmentRule = new AssignmentRule('interpretation.text');
+        assignmentRule.value = 'Important interpretations here';
+        observation.rules.push(cardRule, assignmentRule);
+        doc.profiles.set(observation.name, observation);
+
+        const instance = new Instance('MyObsWithInterpretation');
+        instance.instanceOf = 'ObsWithInterpretation';
+        const pathRuleIndex0 = new PathRule('interpretation[0]');
+        const pathRuleIndex1 = new PathRule('interpretation[1]');
+        instance.rules.push(pathRuleIndex0, pathRuleIndex1);
```
cmoesel, line 4545: Can you add the FSH comments for these ones? I find that helpful to more quickly understand the test.

cmoesel (commented):
I'm currently running some regressions, but here are my comments so far. (BTW -- this isn't nearly as messy as I thought it might be!)

cmoesel:
In the regression without manual slice ordering, I noticed a few changes. Most were related to re-ordering now that we order instance arrays in the same order as the slices were declared. I wanted to check on one other change I saw:

In `hl7-it/lab-report`, the [CapabilityStatementConsumer](https://github.com/hl7-it/lab-report/blob/master/input/fsh/instances/CapabilityStatementConsumer.fsh) has a `CapabilityStatement` instance with:
```
* rest.resource.searchInclude 
* rest.resource.searchRevInclude 
// a bunch of comments here...
* rest.resource[+].type = #PractitionerRole
```
In the current version, those path rules are completely ignored, so the `PracititonerRole` is at `rest.resource[0]`. In this PR, the path rules introduce `rest.resource[0]`, which becomes `null`, and then the `PractitionerRole` is at `rest.resource[1]`. I think we might have already talked about this, and I think it's probably right (since `[+]` is supposed to increment the last referenced index), but I thought I'd mention it once more.

cmoesel (changes requested):
Hmm... related to my last comment... When `manualSliceOrdering` is `false`, then `rest.resource[0]` is `null`.  But when `manualSliceOrdering` is `true`, then `rest.resource[0]` (in the same instance) is `{}`.
image

I think `null` is acceptable, but I'm not sure that `{}` is. I think I'd prefer that be `null` as well. Is that a fair request?

src/export/InstanceExporter.ts
jafeltra: > How much trouble is it to have this code work on (AssignmentRule|PathRule)[]?

I was able to update the type assertion to be `(AssignmentRule | PathRule)[]`, which caused some issues, but I mostly put additional checks to check if a rule was an `instanceof AssignmentRule` before accessing properties that PathRules don't have. I don't think that's too much trouble to move to trying number 2 here, but I can push something up shortly so you can both let me know if you disagree.

jafeltra:
> In the current version, those path rules are completely ignored, so the PracititonerRole is at rest.resource[0]. In this PR, the path rules introduce rest.resource[0], which becomes null, and then the PractitionerRole is at rest.resource[1]. I think we might have already talked about this, and I think it's probably right (since [+] is supposed to increment the last referenced index), but I thought I'd mention it once more.

>When manualSliceOrdering is false, then rest.resource[0] is null. But when manualSliceOrdering is true, then rest.resource[0] (in the same instance) is {}.

Hm these changes are new since I ran the regression before making the PR ready for review again, so maybe they came in when I merged in the latest changes from `master`. I think that means they haven't been talked about yet. I'll take a look at them to see if any of the previous conversations we had might apply and see if I can resolve the issue when `manualSliceOrdering` is on, but it might be worth a conversation as well.

src/export/InstanceExporter.ts
jafeltra: I updated the type casting in [24aad43](https://github.com/FHIR/sushi/pull/1252/commits/24aad439ca0fff5ae4bb65e7b1d904a216af86a8). Let me know what you think.

test/export/InstanceExporter.test.ts
jafeltra: Done in [0e78678](https://github.com/FHIR/sushi/pull/1252/commits/0e78678cd1c2c888cae6437681de474c250a9831).

jafeltra:
@cmoesel - the issue from the regression with `manualSliceOrdering` on is fixed in [745146e](https://github.com/FHIR/sushi/pull/1252/commits/745146e629553ed1f7c5b3537fd717a36e102899).

@mint-thompson - the issue where logging an error for missing required elements on a path that eventually gets removed should be fixed in [ddac2c3](https://github.com/FHIR/sushi/pull/1252/commits/ddac2c3dc30d58aed49fec2cc86193209c527375). I tried to create a test for it, but I couldn't get it to fail without the fix even with the exact FSH that does produce the error. I think it had to do with what children were expanded already when that logic is reached.

src/export/InstanceExporter.ts
cmoesel: Yeah... it's kind of annoying, but I think it's probably right to do it.

mint-thompson (approved):
All of the issues I found have been resolved by these updates. Thank you! Being able to use path rules like this will be amazingly useful.

cmoesel (approved):
Excellent. Here are some more path rules for you:
image

PR #1254, opened 2023-04-14
Correctly resolve references to non-instance items

### LINKED ISSUE #1201: SUSHI does not create references to non-Instance definitions correctly
If an author specifies a reference targeting a definition that was created using the `Profile`, `ValueSet`, or `CodeSystem` declarations, the reference is incorrect. For example, given a profile with name `MySimpleProfile` and id `my-simple-profile`, the assignment `Reference(MySimpleProfile)` will result in a reference to `"MySimpleProfile"` instead of `"StructureDefinition/my-simple-profile"`.

I suspect that the code that resolves references is only looking at `Instance`s.

FSH Online: https://fshschool.org/FSHOnline/#/share/3Hheh66

cmoesel:
This PR fixes #1201 and [CIMPL-1063](https://standardhealthrecord.atlassian.net/browse/CIMPL-1063). References that refer to a non-Instance item by its name or id should be resolved so that the reference is exported using the correct relative URL (e.g., `StructureDefinition/my-profile`). Since it is possible to have clashing names/ids, resolution happens in a 3-tiered system based on priority: first local instances, then local definitions, and finally external instances and definitions.

In a full regression, the following two IGs had differences:
* hl7-be/medication-record
* HL7NZ/northernRegion

In both cases, SUSHI now detects that some references are pointing to a resource of the _wrong type_. As a result, it now logs an error and omits the offending assignments. I believe this is the correct behavior.

### DIFF
--- a/src/fhirtypes/common.ts
+++ b/src/fhirtypes/common.ts
@@ -766,8 +766,8 @@ export function getArrayIndex(pathPart: PathPart): number {
 }
 
 /**
- * Replaces references to instances by the correct path to that instance.
- * Replaces references to local code systems by the url for that code system.
+ * Replaces a reference to an item by name or id with the correct relative path to that item.
+ * Replaces a reference to a local code system by the url for that code system.
  * @param {AssignmentRule} rule - The rule to replace references on
  * @param {FSHTank} tank - The tank holding the instances and code systems
  * @param {Fishable} fisher - A fishable implementation for finding definitions and metadata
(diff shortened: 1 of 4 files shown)


src/fhirtypes/common.ts
```diff
@@ -781,21 +781,77 @@ export function replaceReferences<T extends AssignmentRule | CaretValueRule>(
...
+    if (type != null && id != null) {
       clone = cloneDeep(rule);
       const assignedReference = getRuleValue(clone) as FshReference;
-      assignedReference.reference = `${instanceMeta.sdType}/${instance.id}`;
-      assignedReference.sdType = instanceMeta.sdType;
+      // Set the sdType which will be used for type-checking during assignment
+      assignedReference.sdType = type;
+      // Only replace references that are specified using name or id. Absolute URLs or
+      // relative URLs w/ a type should be left as-is to allow the user more control.
+      if (!value.reference.includes('/')) {
```
jafeltra, line 852: This is pretty nifty. I was trying to think of other times someone might have a `/` inside the Reference, but I couldn't think of one that wasn't already using the `${type}/${id}` format or a URL. So this feels kinda sneaky/sneakily simple to me, but I guess it works!

jafeltra (approved):
This makes sense to me! I tested the example and looked through your examples, and it all makes sense to me. I didn't run a full regression, since it sounded like you already had, but I can if you'd like another set of eyes.

jafeltra:
Also, 

> In both cases, SUSHI now detects that some references are pointing to a resource of the wrong type. As a result, it now logs an error and omits the offending assignments. I believe this is the correct behavior.

Agreed on this. The changes and logs look correct to me.

mint-thompson (approved):
hooray for references!

PR #1255, opened 2023-04-18
Allow versioned URL in extension contains rule (#1251)

mint-thompson:
Completes task [CIMPL-1086](https://standardhealthrecord.atlassian.net/browse/CIMPL-1086).

* Allow versioned URL in extension contains rule

A version can be specified after a pipe. The version is not used when fishing. If the type and extension both have a version, log a warning if they don't match.

* Include version in extension type profile when specified

Modify warning message for version mismatch on extension type profile to be less verbose.

(cherry picked from commit 892ccf8a2a965710e8156bf9e7bbbd3464f07b37)

### DIFF
--- a/src/export/StructureDefinitionExporter.ts
+++ b/src/export/StructureDefinitionExporter.ts
@@ -665,7 +665,9 @@ export class StructureDefinitionExporter implements Fishable {
     }
     rule.items.forEach(item => {
       if (item.type) {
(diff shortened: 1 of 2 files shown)


cmoesel (approved):
Looks good! I tested it manually as well and it worked as expected. Thanks, @mint-thompson!

image

PR #1257, opened 2023-04-19
Replace references to contained canonicals w/ fragment reference

### LINKED ISSUE #1185: SUSHI should detect when ValueSets are used as contained resources and create the proper references
SUSHI should detect when a value set is used as a contained resource and use the correct reference format (with a `#`) when that contained value set is referenced. Right now, if the value set is added as a contained resource and the value set is used in a reference (e.g. `Reference(MyValueSet)`), the reference does not reference the inline resource and creates a regular reference that points to the standalone ValueSet resource (e.g. `"reference": "EgValueSet"`). Instead, it should reference the contained resource using the `#` reference format (e.g. `"reference": "#MyValueSet"`).

A simple FSH Online example demonstrates this: https://fshschool.org/FSHOnline/#/share/3UM8Hxg

This was originally reported on Zulip here: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Reference.20to.20contained.20value.20set

cmoesel:
Fixes #1185 and implements [CIMPL-1050](https://standardhealthrecord.atlassian.net/browse/CIMPL-1050).

If a reference that is specified w/ a canonical refers to a resource in the contained array, replace the reference w/ the fragment id. This works whether the reference is specified using name, id, or URL. Most of this was actually fixed by #1254, but some additional work was needed to make it auto-substitute the fragment when the reference used a canonical URL.

This also adds logic to ensure we only do this replacement on references, whereas the previous logic did it on _any_ value that matched a contained resource type/id.

### DIFF
--- a/src/fhirtypes/common.ts
+++ b/src/fhirtypes/common.ts
@@ -1034,9 +1034,12 @@ export function cleanResource(
   // Update references to any contained resources to be #id instead of resourceType/id
   resourceDef.contained?.forEach((containedResource: any) => {
     const referenceString = `${containedResource.resourceType}/${containedResource.id}`;
+    const referenceUrl = containedResource.url;
     replaceField(
       resourceDef,
-      (o, p) => o[p] === referenceString,
+      (o, p) =>
+        /(^|\.)reference$/.test(p) &&
+        (o[p] === referenceString || (referenceUrl && o[p] === referenceUrl)),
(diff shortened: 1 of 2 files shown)


jafeltra (approved):
Makes sense to me!

mint-thompson (approved):
hooray for references to contained resources!

PR #1265, opened 2023-04-26
Warn on unrecognized configuration properties and on configured pages with unusual names

### LINKED ISSUE #1225: Consider reporting a warning when top-level configuration properties are indented
When a top-level `sushi-config` property is incorrectly indented in the YAML file, SUSHI will not process it correctly. This is because it is technically still valid YAML, but the results are likely not what the user intended. For example, something like this is not processed how it it was likely intended:

```yaml
id: fhir.example
canonical: http://example.org
name: ExampleIG
status: draft
version: 0.1.0
fhirVersion: 4.3.0
copyrightYear: 2021+
releaseLabel: ci-build
pages:
  index.md:
    title: IG Home

  menu:
    Home: index.html

```

`menu` is considered a `page`.

Consider adding a warning in SUSHI if an item that is expected at the top level of the config file is present but not at the top level.

Some things to consider:
- This may be an intentional choice. The YAML is valid so it could be exactly what the user intended. Should we be warning on a case that could be intentional and there will be no way to resolve the warning?
- Is this even a case that is feasible to detect with the YAML library we are using? 
- This is only an issue when a property is incorrectly indented underneath a list because it is grouped with that list. In the above example, without the `pages` property, SUSHI will already report an error because the YAML is not valid (`"All collection items must start at the same column."`).

mint-thompson:
Fixes #1225 and completes task [CIMPL-1084](https://standardhealthrecord.atlassian.net/browse/CIMPL-1084).

An unrecognized configuration property is likely to represent something like a misspelled property, so emit a warning to bring it to the author's attention. Similarly, a mis-indented property may end up in the list of pages, so emit a warning when an entry in the list of pages has a URL that doesn't have .md or .xml at the end.

The giant array of property names is, admittedly, a little bit awkward. But in my research, I couldn't find a clean way to generate the array.

Try this change out locally, and you may find some old, leftover configuration properties that are hanging around in the FSH projects you use for testing. I was surprised to discover I still had `ignoreWarnings` in my configuration.


cmoesel (commented):
I think this is a great start, but it might be even more helpful if it handled more than top-level properties. Several of the properties have nested properties and it's certainly possible to get those wrong too. In fact, I think a fairly easy mistake is to indent a property too much -- which would potentially make it nested under another property and go undetected.

I've made this a comment rather than requesting changes, because I'd like your assessment on how much it would take to go this next step. There are some gotchas since some of our sub-properties allow arbitrary names -- but then require specific subproperties under them.

A couple of quick ideas off the top of my head:
* I think we have code somewhere that flattens objects so nested keys end up like `foo.bar.baz`. We could use that w/ regexes to determine allowable keypaths.
* We could create a JSON schema and then use a JSON schema validator against it. I forget if JSON schema allows for arbitrary key names in places, but if so, this might be a good bet and it also better documents what is allowed.

src/import/importConfiguration.ts
```diff
@@ -893,6 +896,28 @@ function parseInstanceOptions(
...
+  if (yaml.pages != null) {
+    const unusualPages = Object.keys(yaml.pages).filter(nameUrl => !/\.(md|xml)$/.test(nameUrl));
+    if (unusualPages.length > 0) {
+      const pageWord = unusualPages.length === 1 ? 'Page' : 'Pages';
+      logger.warn(
+        `${pageWord} not ending in .md or .xml found in configuration: ${unusualPages.join(', ')}`
+      );
+    }
+  }
+}
```
jafeltra, line 919: One configuration file I had that I tried this on had a pages object like so:

```yaml
pages:
  index.html:
    title: Home
    background.xml: 
      title: Background Info
```
I can't remember exact if this will cause issues down the line (and I haven't run the IG Publisher on this yet), but do you know if html files are allowed? Or if `index.html` is a special case?

jafeltra (commented):
image
I uncommented a whole block I had in an old config file and just blindly ran it, and this was a fun message to see!

Besides the one question I asked inline, I think this looks good. I agree with Chris's comment that this could be cool if we handled deeper validation, but I'll refer to you and Chris on if that is reasonable.

mint-thompson:
@cmoesel I think using a JSON schema for detecting potential mistakes would be a good strategy. There are enough subproperties to check that I don't want to do all the validation by hand, and the [`patternProperties` keyword](https://json-schema.org/understanding-json-schema/reference/object.html#pattern-properties) would let us enforce rules about extra properties. So from there, it's just a matter of taking the errors reported by the validator and turning them into useful warning messages.
@jafeltra I think you're right about `.html` files being allowed here. I'm about to test it out to be sure, but my guess is based on the fact that a configured page can specify `#html` for the `generation` property.

src/import/YAMLschema.ts
```diff
@@ -0,0 +1,449 @@
...
+        name: { type: 'string' },
+        description: { type: 'string' },
+        exampleBoolean: { type: 'boolean' },
+        exampleCanonical: { type: 'string' },
+        isExample: { type: 'boolean' },
+        profile: {
+          oneOf: [{ type: 'string' }, { type: 'array', items: { type: 'string' } }]
+        },
+        groupingId: { type: 'string' }
+      },
```
cmoesel, line 383: We also support `extension` on each resource, as this is how you get the special binary examples of logical models to work.

src/import/YAMLschema.ts
```diff
@@ -0,0 +1,449 @@
...
+              $ref: '#/definitions/configPage'
+            }
+          },
+          additionalProperties: {
+            oneOf: [
+              // { $ref: '#/definitions/configPage' },
+              { type: 'string' },
+              { type: 'array', items: { $ref: '#/definitions/extension' } }
+            ]
+          }
```
cmoesel, line 414: What are these `additionalProperties` intended to cover?

cmoesel (changes requested):
This is cool. I like how once the schema is defined, the validation is a lot simpler.

I've suggested a few changes for your consideration.

src/import/YAMLschema.ts
mint-thompson: I put that in to align with `[key: string]: YAMLConfigurationPage | string | Extension[];` in the `YAMLConfigurationPage` type. But, looking at it more closely, the purpose it serves in the type definition is to allow for nested pages, which is covered by the `patternProperties` in the schema, so these `additionalProperties` can be removed.

mint-thompson: Also, this points out that the message when a page doesn't have the extension we like, the current implementation just says "unrecognized property". In this case, I think it's worth it for me to try to make the message more specific.

jafeltra (commented):
I think this makes a lot of sense and is really going to be very helpful! I just had a couple questions on the JSON schema. Let me know if you want to talk about any of them.

src/import/YAMLschema.json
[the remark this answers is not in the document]
mint-thompson: Hmm... I hadn't thought about that as much. The original focus had been on finding properties appearing in unexpected places due to incorrect spelling or indentation. Those types of issues are found setting rules on what can't appear by using `additionalProperties`. For properties that are required by SUSHI, I feel like the existing logic in the various helper functions in `importConfiguration.ts` are handling cases where values are required for processing, and not just required by a type definition. In the case of `publisher` specifically, the `name` is expected to exist on the object when using this part of the configuration in `parseContact`. But since it's used to populate `ContactDetail.name`, which is optional, nothing bad actually happens if it's undefined? Which maybe means it shouldn't be actually required in the type definition? This is a little tricky to navigate. I think that if it is actually required, the check should not be in the schema, but in the parsing functions, just so that the usage of the schema keeps its original narrow purpose.

cmoesel (approved):
This looks good to me! I confirmed that you addressed Julia and my comments -- and I also confirmed that the schema JSON is packaged up in the module tgz and works when globally installed.

@jafeltra said that as long as you addressed her feedback, she was good with us assuming her approval, so I think we can merge this whenever you want, @mint-thompson!

PR #1269, opened 2023-05-02
Add support for Requirements: keyword to invariant

cmoesel:
This PR pretty much does what it says. Now you can use the `Requirements:` keyword to indicate a value for [ElementDefinition.constraint.requirements](https://hl7.org/fhir/R5/elementdefinition-definitions.html#ElementDefinition.constraint.requirements).


cmoesel:
Closing this PR as we have opted to support the `requirements` element through a rule instead of a keyword.

PR #1275, opened 2023-05-11
Fish With Version

jafeltra:
This PR supports fishing for definitions with a version. This PR adds support for fishing with a specified version in the `FSHTank` and in the `Package`. Support for fishing with a specified version in `FHIRDefinitions` is provided by FPL. There is an FPL PR with this support [here](https://github.com/standardhealth/fhir-package-loader/pull/11).

In addition to supporting fishing with a version, there are a number of places that SUSHI tries to fish for the version specified but is satisfied if any version is found, but logs a warning. To support this, three util functions are included: `fishForFHIRBestVersion`, `fishForMetadataBestVersion`, and `fishInTankBestVersion`. These functions will call either `fishForFHIR`, `fishForMetadata`, or `fish` (in a Tank) first with the item and specified version. If nothing is found, they'll then try again without a version, and if a definition/metadata is found with a differing version, a warning is logged. I replaced the special case handling SUSHI was doing to slice off the version and check for discrepancies between versions with these functions where possible.

To test this, you can use the FPL PR (until a release is ready and available directly in this PR) and link your local FPL to SUSHI. Then you can run the tests and test FSH projects locally.

I ran a full regression and found that two repos had regressions, [HL7/vrdr](https://github.com/HL7/vrdr) and [nightingaleproject/vital_records_fhir_messaging_ig](https://github.com/nightingaleproject/vital_records_fhir_messaging_ig), both with the same issue. Both repos specify a line similar to:

```
* valueDateTime.extension[partialDateTime].extension[time].valueTime.extension[dataabsent].valueCode = http://hl7.org/fhir/ValueSet/data-absent-reason|4.0.1#unknown
```
This line is setting a URL with a version as a system in a FSH code. Before, SUSHI was just fishing for this, didn't find a definition so didn't realize it was a ValueSet, checked that it was a valid URI, and successfully used the code system. In this branch, fishing with the version will work correctly and SUSHI will find a ValueSet definition for this definition, which is a problem, and it will throw a MismatchBindingTypeError and not set the value. I think this regression is actually the correct behavior for SUSHI now.

~_Note: this PR is marked as draft until a new version of FPL is released with support for fishing with versions so that the FHIR Definitions fishing works correctly._~ FPL v.0.3.0 is used now, so this is no longer draft!


src/export/StructureDefinitionExporter.ts
```diff
@@ -780,24 +796,20 @@ export class StructureDefinitionExporter implements Fishable {
...
-            rule.sourceInfo
-          );
-        }
         try {
           let profileUrl = extension.url;
+          const [, ...versionPart] = item.type.split('|');
+          const version = versionPart.join('|') || null;
           if (version) {
             profileUrl += `|${version}`;
           }
```
cmoesel, line 815: I think this is another case where a variant of my regex replace trick might work.
```
const profileUrl = item.type.replace(/^[^|]+/, extension.url);
```

src/export/StructureDefinitionExporter.ts
```diff
@@ -596,7 +604,15 @@ export class StructureDefinitionExporter implements Fishable {
...
-            const vsURI = this.fishForMetadata(rule.valueSet, Type.ValueSet)?.url ?? rule.valueSet;
+            const vsMetadata = this.fishForMetadata(rule.valueSet, Type.ValueSet);
+            const [, ...versionParts] = rule.valueSet.split('|');
+            const version = versionParts.join('|') || null;
+            const vsURIWithVersion = vsMetadata?.url
+              ? `${vsMetadata?.url}${
+                  version && !vsMetadata.url.endsWith(version) ? `|${version}` : ''
+                }`
+              : null;
+            const vsURI = vsURIWithVersion ?? rule.valueSet;
```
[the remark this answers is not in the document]
cmoesel: Also -- this is one of the places we didn't preserve the version before and now we do. I think that's a good idea. Thanks!

src/export/ValueSetExporter.ts
```diff
@@ -57,7 +57,7 @@ export class ValueSetExporter {
         if (component.from.system) {
           const systemParts = component.from.system.split('|');
           const foundSystem = (
-            this.fisher.fishForMetadata(systemParts[0], Type.CodeSystem)?.url ??
+            this.fisher.fishForMetadata(component.from.system, Type.CodeSystem)?.url ??
```
cmoesel, line 60: I see that this change makes it more consistent w/ the approach we take for value sets below -- so that's good.

I also see that if there is a version, we drop the version when there's a match but keep it when there isn't (value sets below does the same thing). I think this is one of those places where we want to keep the version when there is a match. I think we had decided, however, that these kinds of changes aren't necessarily in scope for this PR and can be deferred to another PR if we want.

src/fhirtypes/common.ts
```diff
@@ -855,14 +855,16 @@ export function replaceReferences<T extends AssignmentRule | CaretValueRule>(
...
+    const codeSystemMeta = fishForMetadataBestVersion(fisher, value.system, Type.CodeSystem);
     if (codeSystemMeta) {
       clone = cloneDeep(rule);
       const assignedCode = getRuleValue(clone) as FshCode;
-      assignedCode.system = `${codeSystemMeta.url}${version ? `|${version}` : ''}`;
+      const [, ...versionParts] = value.system?.split('|') ?? [];
+      const version = versionParts.join('|');
+      assignedCode.system = `${codeSystemMeta.url}${
+        !version || codeSystemMeta.url.endsWith(version) ? '' : `|${version}`
+      }`;
```
cmoesel, line 867: Another opportunity for the regex replace (I think):
```ts
assignedCode.system = value.system.replace(/^[^|]+/, codeSystemMeta.url);
```

src/fhirtypes/common.ts
```diff
@@ -1255,6 +1294,42 @@ export function getUrlFromFshDefinition(
...
+    | Instance
+    | Invariant
+    | RuleSet
+    | Mapping,
+  version: string
+): string {
+  // TODO - this early exit can be removed once rules are supported on invariants
+  if (fshDefinition instanceof Invariant) {
+    return version;
+  }
```
cmoesel, line 1323: An `Invariant` will never have a version in it. Nor will a `Mapping`.  I guess a `RuleSet` might, but I'm not sure that's relevant here since you don't know where the RuleSet is applied?

src/import/FSHTank.ts
```diff
@@ -295,50 +312,64 @@ export class FSHTank implements Fishable {
...
+              (version == null || version === getVersionFromFshDefinition(i, this.config.version))
+          );
           break;
         case Type.Invariant:
-          result = this.getAllInvariants().find(i => i.name === item);
+          result = this.getAllInvariants().find(
+            i =>
+              i.name === base &&
+              (version == null || version === getVersionFromFshDefinition(i, this.config.version))
+          );
```
cmoesel, line 366: Invariants don't carry versions, so I don't think this is needed.

src/utils/FishingUtils.ts
```diff
@@ -0,0 +1,106 @@
...
+// If found, return it. If not found, try to remove any |version info and fish for any available version.
+export function fishForFHIRBestVersion(
+  fisher: Fishable,
+  item: string,
+  ...types: Type[]
+): any | undefined {
+  let result = fisher.fishForFHIR(item, ...types);
+
+  // If the item was not found, try slicing off a version and searching for any version
+  if (result == null && item != null && item.includes('|')) {
```
cmoesel, line 28:
```suggestion
  if (result == null && item?.includes('|')) {
```

src/utils/FishingUtils.ts
```diff
@@ -0,0 +1,106 @@
...
+    const [base, ...versionParts] = item.split('|');
+    const version = versionParts.join('|') || null;
+    result = fisher.fishForFHIR(base, ...types);
+    if (version != null && result?.version != null && version != result?.version) {
```
cmoesel, line 32:
```suggestion
    if (version != null && result?.version != null && version != result.version) {
```

src/utils/FishingUtils.ts
```diff
@@ -0,0 +1,106 @@
...
+  ...types: Type[]
+): Metadata | undefined {
+  if (fisher == null) {
+    return;
+  }
+
+  let metadata = fisher.fishForMetadata(item, ...types);
+
+  // If the item was not found, try slicing off a version and searching for any version
+  if (metadata == null && item != null && item.includes('|')) {
```
cmoesel, line 56:
```suggestion
  if (metadata == null && item?.includes('|')) {
```

src/utils/FishingUtils.ts
```diff
@@ -0,0 +1,106 @@
...
+    const [base, ...versionParts] = item.split('|');
+    const version = versionParts.join('|') || null;
+    metadata = fisher.fishForMetadata(base, ...types);
+    if (version != null && metadata?.version != null && version != metadata?.version) {
```
cmoesel, line 60:
```suggestion
    if (version != null && metadata?.version != null && version != metadata.version) {
```

src/utils/FishingUtils.ts
```diff
@@ -0,0 +1,106 @@
...
+  let result = fisher.fishForFHIR(item, ...types);
+
+  // If the item was not found, try slicing off a version and searching for any version
+  if (result == null && item != null && item.includes('|')) {
+    const [base, ...versionParts] = item.split('|');
+    const version = versionParts.join('|') || null;
+    result = fisher.fishForFHIR(base, ...types);
+    if (version != null && result?.version != null && version != result?.version) {
+      logger.warn(
+        `The ${base} definition was specified with version ${version}, but SUSHI found version ${result.version}`
```
cmoesel, line 34: I feel like the language is not quite right here because I initially interpreted "The ${base} definition was specified with version ${version}" to mean it was _defined_ with version ${version}. But what we really mean is it was _requested_ with that version. So maybe:

> The ${base} definition was requested with version ${version}, but SUSHI found version ${result.version}

or even just putting the original value with the `|` in it, e.g.:

> ${item} was requested, but SUSHI found ${base}|${result.version}

src/utils/FishingUtils.ts
```diff
@@ -0,0 +1,106 @@
...
+  let metadata = fisher.fishForMetadata(item, ...types);
+
+  // If the item was not found, try slicing off a version and searching for any version
+  if (metadata == null && item != null && item.includes('|')) {
+    const [base, ...versionParts] = item.split('|');
+    const version = versionParts.join('|') || null;
+    metadata = fisher.fishForMetadata(base, ...types);
+    if (version != null && metadata?.version != null && version != metadata?.version) {
+      logger.warn(
+        `The ${base} definition was specified with version ${version}, but SUSHI found version ${metadata?.version}`
```
cmoesel, line 62: Whatever language you decide on for above should be reflected here too.

src/utils/FishingUtils.ts
```diff
@@ -0,0 +1,106 @@
...
+  | FshCodeSystem
+  | Instance
+  | Invariant
+  | RuleSet
+  | Mapping
+  | undefined {
+  let result = tank.fish(item, ...types);
+
+  // If the item was not found, try slicing off a version and searching for any version
+  if (result == null && item != null && item.includes('|')) {
```
cmoesel, line 91:
```suggestion
  if (result == null && item?.includes('|')) {
```

src/utils/FishingUtils.ts
```diff
@@ -0,0 +1,106 @@
...
+    const [base, ...versionParts] = item.split('|');
+    const version = versionParts.join('|') || null;
+    result = tank.fish(base, ...types);
+    if (result != null) {
+      const resultVersion = getVersionFromFshDefinition(result, tank.config.version);
+      if (version != null && resultVersion != null && version != resultVersion) {
+        logger.warn(
+          `The ${base} FSH definition was specified with version ${version}, but SUSHI found version ${resultVersion}`
```
cmoesel, line 99: And again, make sure the final language gets reflected here.

test/export/InstanceExporter.test.ts
```diff
@@ -5886,6 +5886,30 @@ describe('InstanceExporter', () => {
...
+      brightInstance.instanceOf = 'Observation';
+      const assignedCodeRule = new AssignmentRule('code');
+      assignedCodeRule.value = new FshCode('bright', 'Visible|1.0.0');
+      brightInstance.rules.push(assignedCodeRule);
+      doc.instances.set(brightInstance.name, brightInstance);
+
+      const visibleSystem = new FshCodeSystem('Visible');
+      const visibleSystemVersion = new CaretValueRule('');
+      visibleSystemVersion.caretPath = 'url';
+      visibleSystemVersion.value = 'http://hl7.org/fhir/us/minimal/CodeSystem/Visible|1.0.0';
```
cmoesel, line 5900: I don't think a `CodeSystem` definition would declare its `url` with a version. The `url` would be the version-less URL and the `version` element would contain the version. E.g.
```ts
const visibleSystemUrl = new CaretValueRule('');
visibleSystemUrl.caretPath = 'url';
visibleSystemUrl.value = 'http://hl7.org/fhir/us/minimal/CodeSystem/Visible';
const visibleSystemVersion = new CaretValueRule('');
visibleSystemVersion.caretPath = 'version';
visibleSystemVersion.value = '1.0.0';
```

test/import/FSHTank.test.ts
```diff
@@ -151,6 +225,15 @@ describe('FSHTank', () => {
...
     doc3.invariants.get('Invariant1').severity = new FshCode('error');
+    // TODO - add back in once rules on invariants are supported
+    // const invariant2 = new Invariant('Invariant2');
+    // invariant2.description = 'second invariant';
+    // invariant2.severity = new FshCode('error');
+    // const invariant2Version = new CaretValueRule('');
+    // invariant2Version.caretPath = 'version';
+    // invariant2Version.value = 'v-2.0.0';
+    // invariant2.rules.push(invariant2Version);
+    // doc3.invariants.set(invariant2.name, invariant2);
```
cmoesel, line 236: I think you can delete this since invariants don't have a version.

test/import/FSHTank.test.ts
```diff
@@ -527,6 +772,19 @@ describe('FSHTank', () => {
...
+
+    // TODO - add back in once rules on invariants are supported
+    it.skip('should find an invariant when fishing with a version that is set by a caret rule', () => {
+      expect(tank.fish('Invariant2|v-2.0.0', Type.Invariant).name).toBe('Invariant2');
+    });
+
+    it('should not find an invariant when fishing with a version that does not match', () => {
+      expect(tank.fish('Invariant1|2.0.0', Type.Invariant)).toBeUndefined();
+    });
+
```
cmoesel, line 787: I think these can be removed. Also, Invariants don't support caret rules anyway. :-)

cmoesel (changes requested):
Wow. This turned out to be pretty intense! But... we finally have a more consistent approach to handling versioned canonicals! Thank you!

I found a few things that I think can be improved and/or fixed and/or removed. But... this is looking very good!

src/export/Package.ts
```diff
@@ -30,85 +30,123 @@ export class Package implements Fishable {
...
           if (!def) {
             def = this.instances.find(
               i =>
                 i._instanceMeta.usage === 'Definition' &&
                 i.resourceType === 'StructureDefinition' &&
                 i.derivation === 'constraint' &&
                 i.type !== 'Extension' &&
-                (i.id === item || i._instanceMeta.name === item || i.url === item)
+                (i.id === base || i._instanceMeta.name === base || i.url === base) &&
+                (version == null || (i.version && i.version === version))
```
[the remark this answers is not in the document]
jafeltra: Well... I thought there was a reason, but now I can't think of what that reason was.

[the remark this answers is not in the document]
cmoesel: I think `(version == null || (i.version && i.version === version))` can be shortened to `(version == null || i.version === version)` because:
* If `version` is null or undefined, it is true without need to evaluate the 2nd part;
* If `version` is not null/undefined, then it won't be equal to `i.version` if `i.version` is null anyway.

src/export/ValueSetExporter.ts
jafeltra: Now that I have your nifty regex replace in my back pocket, I don't mind doing it here. But I can also hold off for a separate PR if that's preferred.

jafeltra: Actually, wait, now I'm not sure if I agree we should keep the version. The version the author specified gets added to the `composeElement.version` regardless of if we find a match on line 63. And because the `foundSystem` variable does a final `split('|')` and then only `foundSystem[0]` is used later, the version is always spliced off and never added to `composeElement.system`. So do we want to include the `|version` on the `system` property, as well as on the `version` property?

src/export/Package.ts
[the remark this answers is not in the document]
jafeltra: Yeah agreed. Maybe I had something different at some point and changed things around and didn't realize the redundancy in the end.

src/fhirtypes/common.ts
jafeltra: I wonder if it would be convenient or annoying to wrap the regex replace in a helper function...

test/export/InstanceExporter.test.ts
jafeltra: I thought I had added this case because of a regression, but I'm not sure which repo without rerunning the regression. I believe this was also tied to [this](https://github.com/FHIR/sushi/pull/1275#discussion_r1196899479) change, and updating to use the regex replace actually causes this test to fail. I guess it is worth keeping the test with the version caret value rule, so I'll update this!

src/utils/FishingUtils.ts
jafeltra: I like the simple second version!

src/export/Package.ts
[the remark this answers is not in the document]
jafeltra: Addressed in [9668cdc](https://github.com/FHIR/sushi/pull/1275/commits/9668cdc2e8174378a573a33170c4740424a1b2cd)

src/export/StructureDefinitionExporter.ts
jafeltra: Part of [70e31fc](https://github.com/FHIR/sushi/pull/1275/commits/70e31fc1b4bdb5eaab6119f21d08e44e038c24db)

src/fhirtypes/common.ts
jafeltra: Part of [70e31fc](https://github.com/FHIR/sushi/pull/1275/commits/70e31fc1b4bdb5eaab6119f21d08e44e038c24db).

And I'm leaning that it would be annoying to wrap in a helper function... so didn't do that.

src/import/FSHTank.ts
jafeltra: Part of [0d0852d](https://github.com/FHIR/sushi/pull/1275/commits/0d0852d19cbf5a24e3e63fdf9a7f359ac288bcce)

src/utils/FishingUtils.ts
jafeltra: Part of [18b8ba8](https://github.com/FHIR/sushi/pull/1275/commits/18b8ba89e0c6aa3a9ec12bfc0c288154ec6be7f5)

src/utils/FishingUtils.ts
jafeltra: Part of [18b8ba8](https://github.com/FHIR/sushi/pull/1275/commits/18b8ba89e0c6aa3a9ec12bfc0c288154ec6be7f5)

src/utils/FishingUtils.ts
jafeltra: Part of [18b8ba8](https://github.com/FHIR/sushi/pull/1275/commits/18b8ba89e0c6aa3a9ec12bfc0c288154ec6be7f5)

src/utils/FishingUtils.ts
jafeltra: Part of [18b8ba8](https://github.com/FHIR/sushi/pull/1275/commits/18b8ba89e0c6aa3a9ec12bfc0c288154ec6be7f5)

src/utils/FishingUtils.ts
jafeltra: Simple second version updated in Part of [18b8ba8](https://github.com/FHIR/sushi/pull/1275/commits/18b8ba89e0c6aa3a9ec12bfc0c288154ec6be7f5). Let me know if you prefer something else!

src/utils/FishingUtils.ts
jafeltra: Part of [18b8ba8](https://github.com/FHIR/sushi/pull/1275/commits/18b8ba89e0c6aa3a9ec12bfc0c288154ec6be7f5)

src/utils/FishingUtils.ts
jafeltra: Part of [18b8ba8](https://github.com/FHIR/sushi/pull/1275/commits/18b8ba89e0c6aa3a9ec12bfc0c288154ec6be7f5)

src/utils/FishingUtils.ts
jafeltra: Part of [18b8ba8](https://github.com/FHIR/sushi/pull/1275/commits/18b8ba89e0c6aa3a9ec12bfc0c288154ec6be7f5) (although you must know that by now ;) )

test/export/InstanceExporter.test.ts
jafeltra: Included in [70e31fc](https://github.com/FHIR/sushi/pull/1275/commits/70e31fc1b4bdb5eaab6119f21d08e44e038c24db)

test/import/FSHTank.test.ts
jafeltra: Part of [0d0852d](https://github.com/FHIR/sushi/pull/1275/commits/0d0852d19cbf5a24e3e63fdf9a7f359ac288bcce)

test/import/FSHTank.test.ts
jafeltra: Part of [0d0852d](https://github.com/FHIR/sushi/pull/1275/commits/0d0852d19cbf5a24e3e63fdf9a7f359ac288bcce)

src/export/ValueSetExporter.ts
cmoesel: Oh, dang. You're right! For whatever reason, `ValueSet.compose.include` splits the version off into its own element. And the `system` element is `uri` _not_ `canonical`. So, yep, we definitely do _not_ want to attach the `|version`.  Good call!

test/export/InstanceExporter.test.ts
cmoesel: Technically, `CodeSystem.url` is a `uri`, so you're not supposed to append a version to it. It even says:
> URL should not contain | or # - these characters make processing canonical references problematic

So if someone is making a CodeSystem with a URL like that in a real IG, we shouldn't support it anyway.

jafeltra: Oh! Well maybe I just invented that test case. It seems so specific so I thought I must have pulled it from somewhere, but maybe not. Well, it is no longer supported now!

cmoesel (approved):
This looks very good! I ran a regression and got the same two regressions as you did. Upon studying them, however, I realized that they're assigning the value to a `code` -- so while the system in the FSH is incorrect, it isn't actually _used_ because primitive `code` does not include the system. This means the assignment can still safely happen.

That being the case, I thought it would be more appropriate to detect that situation and _warn_ on it rather than _error_ -- and proceed with the assignment rather than skipping it. So that is what I've done in 3a9f3e9.

Now everything looks good to me and has my approval!

src/fhirtypes/common.ts
```diff
@@ -855,14 +855,12 @@ export function replaceReferences<T extends AssignmentRule | CaretValueRule>(
       }
     }
   } else if (value instanceof FshCode) {
-    const [system, ...versionParts] = value.system?.split('|') ?? [];
-    const version = versionParts.join('|');
-    const codeSystem = tank.fish(system, Type.CodeSystem);
-    const codeSystemMeta = fisher.fishForMetadata(system, Type.CodeSystem);
+    const codeSystem = fishInTankBestVersion(tank, value.system, Type.CodeSystem);
+    const codeSystemMeta = fishForMetadataBestVersion(fisher, value.system, Type.CodeSystem);
```
mint-thompson, line 859: This is not really a problem, but it's possible for each of these functions to separately log a warning for a version mismatch if the CodeSystem is retrieved from the package when fishing for metadata. Here's an example that can make this happen:
```
CodeSystem: BreakfastCS
* ^version = "1.0.0"
* #toast "Toast"
* #donut "Donut"

Instance: MyBreakfast
InstanceOf: Observation
* status = #final
* code = BreakfastCS|2.0.0#toast
```

mint-thompson (approved):
This looks great, thank you! There's one slightly unusual thing that I noticed, but I don't think it necessarily needs to be changed, so I'm giving my approval.

src/fhirtypes/common.ts
cmoesel: Ah, yeah. We technically need to do both fishes -- so the only way I can think of to avoid the double warning is to do the 2nd one based on whatever was actually returned by the first (rather than the original request). E.g.,:

```ts
    const codeSystemMeta = fishForMetadataBestVersion(fisher, value.system, Type.CodeSystem);
    if (codeSystemMeta) {
      clone = cloneDeep(rule);
      const assignedCode = getRuleValue(clone) as FshCode;
      assignedCode.system = value.system.replace(/^[^|]+/, codeSystemMeta.url);

      // Find the code system using the returned metadata to avoid duplicate warnings if version mismatches
      const matchedCanonical = codeSystemMeta.url
        ? `${codeSystemMeta.url}${codeSystemMeta.version ? `|${codeSystemMeta.version}` : ''}`
        : value.system;
      const codeSystem = fishInTankBestVersion(tank, matchedCanonical, Type.CodeSystem);
      if (codeSystem && (codeSystem instanceof FshCodeSystem || codeSystem instanceof Instance)) {
        // if a local system was used, check to make sure the code is actually in that system
        listUndefinedLocalCodes(codeSystem, [assignedCode.code], tank, rule);
      }
    }
```

@mint-thompson - I have this change locally and it _does_ worth. Do you think I should push it up?

mint-thompson: I think so, yes. It's not too complicated, and it makes the output more consistent. Thank you!

cmoesel: OK. Fixed in 858c75d. I also noticed that the warning messages did not log the source info -- so we'd say there was a mismatch, but it would be up to the author to find it in their FSH files. This didn't seem optimal, so I also updated it to pass along the source info and log that. It's in the same commit.

cmoesel (commented):
OK. I've addressed the double-warning and source info logging. @mint-thompson - can you give it one more review?

mint-thompson (approved):

PR #1282, opened 2023-05-31
Support Context keyword when defining Extensions

### LINKED ISSUE #1259: Add ContextElement:, ContextFHIRPath:, and ContextExtension: keywords to Extension
The FHIR chiefs recommend that extensions should always declare their context -- and broad contexts (such `Element`) are rarely (if ever) correct. As such, FSH should make it easier to declare contexts -- and it seems that adding new keywords is the most appropriate approach to this.

Since [StructureDefinition.context](https://hl7.org/fhir/R5/structuredefinition-definitions.html#StructureDefinition.context) supports three types (`element`, `fhirpath`, and `extension`), we will introduce a keyword for each type:
```
ContextElement: Patient.name, Practitioner.name
ContextFHIRPath: (Observation | Condition).code
ContextExtension: RaceExtension
```

@markkramerus and I discussed various approaches to supporting the different types, but this was the one we finally agreed upon.

A few notes on how it works:
* Each keyword can have multiple values separated by commas
* Since FHIRPath can have commas, authors can escape them (`\,`) _or_ use `[[` `]]` (like insert rules)
* If no contexts are provided, default to the broadest context: `ContextElement: Resource, Element`

mint-thompson:
Fixes #1259 and completes task [CIMPL-1090](https://standardhealthrecord.atlassian.net/browse/CIMPL-1090).

This adds the new `Context` keyword that can be used when defining an Extension. The value of the context can be specified in a few different ways:
- A quoted string
- A FSH path that starts with the name or id of some StructureDefinition
- A URL for a StructureDefinition, optionally followed by `#` and a FSH path

Multiple contexts can be specified in a comma-separated list.

The resulting `type` of the context depends on what is specified:
- For a quoted string, the `fhirpath` type is always used.
- For an unquoted value that specifies an Extension definition or an element that is part of a complex extension, the `extension` type is used.
- For any other element, the `element` type is used.

This allows for some flexibility for how authors specify the context, while consistently resolving contexts to the more specific `extension` type whenever possible. The included test cases demonstrate this flexibility.

For more information about extension contexts, the [official documentation](http://hl7.org/fhir/defining-extensions.html#context) provides some useful details.


test/export/StructureDefinition.ExtensionExporter.test.ts
```diff
@@ -216,4 +221,415 @@ describe('ExtensionExporter', () => {
...
+        }
+      ];
+      doc.extensions.set(extension.name, extension);
+      const exported = exporter.exportStructDef(extension);
+      expect(exported.context).toBeUndefined();
+      expect(loggerSpy.getLastMessage('error')).toMatch(
+        'Could not find resource MysteryExtension to use as extension context.'
+      );
+      expect(loggerSpy.getLastMessage('error')).toMatch(/File: Context\.fsh.*Line: 8\D*/s);
+    });
```
jafeltra, line 486: The last test in the custom resource block reminded me of possibly another test case that might be useful in this section. Is it worthwhile also testing that an invalid FSH path on a resource? I think the error will be triggered by the same logic, so it might be a duplicate test, so curious what you think.

jafeltra (commented):
This looks really good! I had a few comments inline, as well as this:

One other thing related to this feature: while reviewing this, I realized that we need to add `Context` to the keywords that get highlighted in the VS Code extension. Ideally, I think we could release a new version of the extension around the same time, or even before, we release this feature in SUSHI.

src/export/StructureDefinitionExporter.ts
```diff
@@ -383,26 +384,16 @@ export class StructureDefinitionExporter implements Fishable {
...
-        // ^context[0].expression = "Patient"
-        // TODO: Consider introducing metadata keywords for this
-        structDef.context = [
-          {
-            type: 'element',
-            expression: 'Element'
-          }
-        ];
+        this.setContext(structDef, fshDefinition);
       }
```
cmoesel, line 397: I don't think we talked about what to do with inherited contexts... The way it is implemented now, if a parent defined a context, then even if the author specified a `Context:`, it would be dropped (since `this.structDef context === null` would be `false`). I don't think we want that.

I think that if there is an inherited context and the author specifies a `Context:`, it should _overwrite_ the existing context. The other option would be to append to the existing context, but I don't think this is right because a profiled extension might be more specific than its parent and therefore apply to fewer (or more specific) contexts as well.

src/export/StructureDefinitionExporter.ts
```diff
@@ -420,6 +411,124 @@ export class StructureDefinitionExporter implements Fishable {
...
+          // this could be the path to an element on some StructureDefinition
+          // it could be specified either as a url, a path, or a url with a path
+          // url: http://example.org/Some/Url
+          // path: resource-name-or-id.some.path
+          // url with path: http://example.org/Some/Url#some.path
+          // we split on # to make contextItem and contextPath, so if contextPath is not empty, we have a url with a path
+          // otherwise, check if contextItem is a url or not. if it is not a url, we have a fsh path that starts with a name or id
+          const splitContext = extContext.value.split('#');
+          let contextItem = splitContext[0];
+          let contextPath = splitContext.slice(1).join('#');
```
cmoesel, line 437: This assumes that if there are multiple `#` in the value, the _first_ one is the separator and the remaining ones are in the path.  But... is `#` a valid symbol in a path? I don't _think_ it is (although I'm not 100% sure). But we know `#` is a valid symbol in a URI (which allows fragments). So I'd suggest that we actually should operate on the assumption that the _last_ `#` is the separator.

cmoesel: That said, there is a bit of a wrinkle here if a URL has a `#` in the URL itself and there is no path. We'd (incorrectly) assume the last bit is a path. But... maybe you could get around this by putting a `#` at the end of the URL (e.g., _blank_ path)?

src/export/StructureDefinitionExporter.ts
```diff
@@ -420,6 +411,124 @@ export class StructureDefinitionExporter implements Fishable {
...
+                  contextExpression = resourceSD.url;
+                  contextType = 'extension';
+                } else if (
+                  resourceSD.derivation === 'specialization' &&
+                  resourceSD.url.startsWith('http://hl7.org/fhir/StructureDefinition/')
+                ) {
+                  contextExpression = targetElement.id;
+                } else {
+                  contextExpression = `${resourceSD.url}#${targetElement.id}`;
+                }
```
cmoesel, line 474: This is all _very nice_.

src/export/StructureDefinitionExporter.ts
```diff
@@ -420,6 +411,124 @@ export class StructureDefinitionExporter implements Fishable {
...
+      ];
+    }
+  }
+
+  private setContextForComplexExtension(
+    structDef: StructureDefinition,
+    targetElement: ElementDefinition,
+    extContext: ExtensionContext
+  ) {
+    const extensionHierarchy = [
```
cmoesel, line 509: I think I'd appreciate a comment here indicating the target format you're going for and maybe even including a direct link to Lloyds message where he says this is what to do.

test/export/StructureDefinition.ExtensionExporter.test.ts
```diff
@@ -216,4 +221,415 @@ describe('ExtensionExporter', () => {
...
+            value: 'my-obs',
+            isQuoted: false
+          }
+        ];
+        doc.extensions.set(extension.name, extension);
+        const exported = exporter.exportStructDef(extension);
+        expect(exported.context).toEqual([
+          {
+            type: 'element',
+            expression: 'http://hl7.org/fhir/us/minimal/StructureDefinition/my-obs#Observation'
```
cmoesel, line 509: Oh... interesting. I was going to ask why, but now I see. The spec says:
> If there is no #, the Element id is one defined in the base specification.

Good call.

test/export/StructureDefinition.ExtensionExporter.test.ts
cmoesel, line 1: This is a really nice set of tests, but I think we also want to test this for specifying profiles (and their elements) as context. We want to be sure it exports it with the profile URL (as opposed to profile name or id).

cmoesel (changes requested):
This is a really nice implementation of a tricky feature. Very well done. I think you did a great job leveraging some of the functions we already had, resulting in a very clean implementation.

I've left a few comments for your consideration. Thank you!

src/export/StructureDefinitionExporter.ts
mint-thompson: That is an interesting case. I think your workaround would work fine, in that case. Also, if the author uses an alias, that should also work.

test/export/StructureDefinition.ExtensionExporter.test.ts
mint-thompson: The tests in the `#withCustomResource` block all use a profile, and show that the URL is used.

src/fshtypes/Extension.ts
```diff
@@ -1,24 +1,56 @@
...
+  metadataToFSH(): string {
+    const sdMetadata = super.metadataToFSH();
+    const contextLines: string[] = [];
+    this.contexts.forEach(extContext => {
+      if (extContext.isQuoted) {
+        contextLines.push(`Context: "${fshifyString(extContext.value)}"`);
+      } else {
+        contextLines.push(`Context: ${extContext.value}`);
+      }
+    });
```
cmoesel, line 32: As far as I can tell, this is dead code now and can be deleted.

cmoesel (changes requested):
Everything is looking really good! I just found one small bit of dead code that can probably be deleted.

test/export/StructureDefinition.ExtensionExporter.test.ts
jafeltra: Do you think adding that type of test would be redundant? I'm still just curious to hear what you think.

jafeltra (commented):
I only wanted to check in one outstanding question, other than what Chris mentioned.

test/import/FSHImporter.Extension.test.ts
```diff
@@ -78,17 +94,89 @@ describe('FSHImporter', () => {
...
       });
 
+      it('should parse an extension with multiple contexts', () => {
+        const input = `
+        Extension: SomeExtension
+        Parent: ParentExtension
+        Id: some-extension
+        Context: "some.fhirpath()", Observation.component, http://example.org/MyPatient#identifier,
+                 "another.fhirpath(var, 0)"
+        `;
```
cmoesel, line 104: We need to add an assertion that confirms that no errors are emitted. As of now, this test _does_ produce errors because the parser doesn't know what to do w/ the spaces between the contexts.

cmoesel (changes requested):
I just did some manual testing on this, and unfortunately, it looks like it's not quite ready yet.

Given an extension with multiple contexts like this:
```
Extension: Foo
Context: Patient, Practitioner, "some.fhirpath"
* value[x] only boolean
```

We get the following parse errors:
```
error token recognition error at: ' '
  File: /Users/cmoesel/data/fsh/ScratchPad/input/fsh/scratch.fsh
  Line: 12
error token recognition error at: ' '
  File: /Users/cmoesel/data/fsh/ScratchPad/input/fsh/scratch.fsh
  Line: 12
```

It also emits errors if we try to separate them on separate lines (one error for every space character).

So... we'll hold this feature back for SUSHI 3.1.0.

cmoesel (approved):
YES!

image

jafeltra (approved):

PR #1284, opened 2023-06-02
Assignment on primitive satisfies cardinality of value element

### LINKED ISSUE #1281: SUSHI does not validate cardinality of primitive .value element correctly
If a profile constrains a primitive's `.value` to have min `1`, SUSHI doesn't recognize instances of the primitive as fulfilling that constraint. It emits an error like:

> error Element Provenance.target.reference.value has minimum cardinality 1 but occurs 0 time(s).

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Reference.2Evalue.20SUSHI.20error/near/361407164

FSH Online: https://fshschool.org/FSHOnline/#/share/3OFFtRc

mint-thompson:
Fixes #1281 and completes task [CIMPL-1109](https://standardhealthrecord.atlassian.net/browse/CIMPL-1109).

All elements with primitive types have a child element named "value". When this child element is required, an assignment to the primitive satisfies this cardinality requirement.

Because the primitive's "value" element will only be checked if the parent primitive element is not null, it is already known that a value has been assigned to the parent primitive when the "value" element's cardinality is being checked.


cmoesel (changes requested):
This solves the false negative quite elegantly.

Unfortunately, it introduces a false positive. If a primitive's `.value` is `1..1`, then it _must_ have a value; setting just an extension on the primitive should not satisfy it. Prior to this PR, SUSHI would emit an error, but now it lets it pass with no error.

E.g., the following should cause an error if `target.reference.value` is `1..1`, but it no longer does in this PR:
```
* target.reference.extension[http://hl7.org/fhir/StructureDefinition/data-absent-reason].valueCode = #unsupported
```

[Full example on FSH Online](https://fshschool.org/FSHOnline/#/share/42iZqAI)

mint-thompson:
There are some existing problems with extensions on primitive arrays that are not fixed in this PR. Two test cases are added that give examples of these problems, but they are currently skipped. I've added [CIMPL-1116](https://standardhealthrecord.atlassian.net/browse/CIMPL-1116) for this to be addressed in the future.

jafeltra (commented):
I think the code makes sense, but when I went to run a regression, the repo `HL7/davinci-epdx#master` has differences that I don't quite understand. Also, there are errors being logged saying `Cannot read properties of undefined (reading 'length')`, which shouldn't be happening. I haven't looked into why that's happening yet and if the differences and that unexpected error are happening, but I can if that is helpful.

cmoesel:
@jafeltra - I'm in the middle of running a regression on this one -- so I'll report back.

cmoesel (approved):
Code looks good and regression passes w/ no changes. Thank you!

PR #1290, opened 2023-06-09
Handle spaces after bracketed parameters

### LINKED ISSUE #1285: Insert rules with [[parameter]] not parsed correctly when spread across lines
Using normal parameters with escaping and no `[[` `]]`, SUSHI handles line breaks just fine. Given:
```
RuleSet: AddVariableToTestScript(name, expression)
* variable[+].name = "{name}"
* variable[=].expression = "{expression}"

Instance: OldWayWorks
InstanceOf: TestScript
Title: "My Test Script"
Description: "A small test with a few FHIRPath expressions"
* name = "Foo"
* status = #final
* insert AddVariableToTestScript(
    firstObservation,
    component.all(valueSampledData.exists(\)\)
  )
* insert AddVariableToTestScript(
    testResponse,
    resource.repeat(item\).answer.value.extension.value.aggregate($this+$total\,0\)
  )
```
You get:
```json
"variable": [
  {
    "name": "firstObservation",
    "expression": "component.all(valueSampledData.exists())"
  },
  {
    "name": "testResponse",
    "expression": "resource.repeat(item).answer.value.extension.value.aggregate($this+$total,0)"
  }
]
```

But if you use `[[` `]]`, it seems to not recognize the breaks between parameters. Given:
```
Instance: NewWayBroken
InstanceOf: TestScript
Title: "My Test Script"
Description: "A small test with a few FHIRPath expressions"
* name = "Foo"
* status = #final
* insert AddVariableToTestScript(
    firstObservation,
    [[component.all(valueSampledData.exists())]]
  )
* insert AddVariableToTestScript(
    testResponse,
    [[resource.repeat(item).answer.value.extension.value.aggregate($this+$total,0)]]
  )
```
You get:
```json
"variable": [
  {
    "name": "firstObservation",
    "expression": "component.all(valueSampledData.exists())]]\n  )\n* insert AddVariableToTestScript(\n    testResponse,\n    [[resource.repeat(item).answer.value.extension.value.aggregate($this+$total,0)]]\n"
  }
]
```

This is occurring in SUSHI 3 Beta 3.

mint-thompson:
Fixes #1285 and completes task [CIMPL-1117](https://standardhealthrecord.atlassian.net/browse/CIMPL-1117).

It has always been permitted by FSH spec to put any amount of spaces or newlines between the closing brackets of a bracketed parameter and the following comma or right-parentheses character. Update the grammar to avoid matching extra characters after the end of the bracketed parameter. Update the importer to correctly trim comma, right-parentheses, space, and bracket characters from bracketed parameter tokens.

To explain in a little more detail about what was going wrong here, there are two parts to this PR: the lexer change and the importer change.

In the lexer, the rules for matching bracketed parameters were defined a little too broadly. They allowed `]]` to be followed by any amount of space, and then anything that wasn't `,)`. But, what I think was happening was this: "any amount" could be zero, and then the space character would match the `~[,)]` rule! The lexer would then continue grabbing as many characters as it could, only stopping on the last bracketed parameter. By adding space characters to the disallowed set, this scenario is avoided.

Once that was fixed, the code that handled bracketed parameters needed to be fixed. The previous implementation did things a little out of order, since it always would try to remove the last three characters from a bracketed parameter. This works as long as there's no space between the closing brackets and the `)` or `,` character, but fails if there are spaces. So, the process is to instead trim off the `)` or `,` that will always be the last character of the token, then trim spaces, then trim off two characters from the start and finish to get rid of the brackets.

With both of these changes, the issue is resolved. You can try this out with the FSH in the linked issue, and as a bonus, try putting all of the parameters in double brackets. You can get some kind of wild results before this fix.


cmoesel:
@mint-thompson - Awesome! But also, oops -- I just merged the Context keyword PR which ended up causing a conflict in the generated ANTLR files. Sorry!

jafeltra (approved):
These changes make a lot of sense, thanks for the helpful breakdown of the changes! The example in the issue is resolved if there are new lines or spaces around the parameter. Looks good to me!

cmoesel (approved):
image

PR #1291, opened 2023-06-13
Slicing rules should not be on slices

### LINKED ISSUE #1283: Don't allow (or require) slicings in slices
Based on [this thread](https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/Reslicing.20extensions.20causes.20validator.20errors/near/360153471), you cannot introduce new `slicing`s in existing slices. Or expressed as a FHIRPath invariant expression:
```
sliceName.exists() implies slicing.empty()
```

This is a departure from how we thought reslicing worked -- and as a result, SUSHI complains if you try to introduce re-slices into a slice that does not have a `slicing`.  SUSHI needs to be updated to look for a slicing on the base element (not the slice that is being re-sliced). In addition, if SUSHI finds a `slicing` inside a slice, it should log a warning saying that the slicing info should be moved to the base slicing. If possible, we should consider having SUSHI move the discriminators for the author (but still have the warning).

mint-thompson:
Fixes #1283 and completes task [CIMPL-1108](https://standardhealthrecord.atlassian.net/browse/CIMPL-1108).

You can add reslices of a slice without having to define slicing rules on the slice. If you want more discriminators, you set those on the base element. Show a warning if the user adds slicing rules to an element that has a slice name.

I decided to somewhat expand the `ValidationError` class and `validate` functions for this implementation, since that let me manage the new condition being checked alongside the existing checks.

The tests on `sliceIt` are added for completeness, but probably don't come up during actual program execution. This is because `sliceIt` is only called in specific circumstances where SUSHI wants to define the slicing rules. If the author defines the slicing using caret rules, this function is not used.


jafeltra (commented):
I think these changes make sense! I added two small suggestions, but neither are really required if you don't like them.

src/export/StructureDefinitionExporter.ts
```diff
@@ -1184,7 +1184,11 @@ export class StructureDefinitionExporter implements Fishable {
     structDef.inProgress = false;
 
     structDef.validate().forEach(err => {
-      logger.error(err.message, fshDefinition.sourceInfo);
+      if (err.severity === 'warn') {
+        logger.warn(err.message, fshDefinition.sourceInfo);
+      } else {
+        logger.error(err.message, fshDefinition.sourceInfo);
+      }
```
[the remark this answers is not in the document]
cmoesel: Note that this only safe if we also default `ValidationError.severity` to `'error'` in its constructor (since it is marked as optional).

src/fhirtypes/ElementDefinition.ts
```diff
@@ -319,10 +319,20 @@ export class ElementDefinition {
   }
 
   validate(): ValidationError[] {
+    const errors: ValidationError[] = [];
     if (this.slicing) {
-      return this.validateSlicing(this.slicing);
+      if (this.sliceName) {
+        errors.push(
+          new ValidationError(
+            'An element with a slice name does not need to define its own slicing. Instead, define additional slicing discriminators on the base element.',
```
cmoesel, line 327: I think it needs to be just a tad stronger:

```suggestion
            'An element with a slice name should not define its own slicing. Instead, define additional discriminators to the original slicing on the base element.',
```

test/fhirtypes/ElementDefinition.slicing.test.ts
```diff
@@ -363,9 +363,7 @@ describe('ElementDefinition', () => {
       component.sliceIt('pattern', 'code', false, 'open');
       const systolicBP = component.addSlice('SystolicBP');
       const diastolicBP = component.addSlice('DiastolicBP');
-      systolicBP.sliceIt('pattern', 'interpretation', false, 'open');
```
cmoesel, line 366: To keep the use case somewhat realistic, maybe instead of removing this (and the other one below it), we should modify it to call `sliceIt` on `component` (which just appends the discriminator to the existing slicing). The test works either way, but in the real world, the re-slice only works if there is another discriminator.

cmoesel (changes requested):
Thanks for this. It looks pretty good! I've left a few comments for your consideration.

jafeltra (approved):
Thanks for making those changes! This makes sense to me!

src/fhirtypes/ElementDefinition.ts
```diff
@@ -319,10 +319,20 @@ export class ElementDefinition {
   }
 
   validate(): ValidationError[] {
+    const errors: ValidationError[] = [];
     if (this.slicing) {
-      return this.validateSlicing(this.slicing);
+      if (this.sliceName) {
+        errors.push(
+          new ValidationError(
+            'An element with a slice name should not define its own slicing. Instead, define additional discriminators to the original slicing on the base element.',
```
cmoesel, line 327: You took my suggestion, but it turns out my suggestion was a tiny bit awkward. Sorry. I think this reads better:
```suggestion
            'An element with a slice name should not define its own slicing. Instead, append additional discriminators to the original slicing on the base element.',
```

cmoesel (approved):
Just what we needed!

jafeltra (approved):

PR #1293, opened 2023-06-15
Support Characteristics keyword when defining Logical models

### LINKED ISSUE #1279: Add Characteristics keyword for Logicals
Logical models should declare their characteristics so that it's clear how they can and cannot be used. This is done via the [Structure Type Characteristics (Constraint Control)](https://hl7.org/fhir/extensions/StructureDefinition-structuredefinition-type-characteristics.html) extension and the [Type Characteristic Codes](https://hl7.org/fhir/extensions/ValueSet-type-characteristics-code.html) value set.

We've recently learned that logicals can only be targeted as a reference if they declare the `can-be-target` characteristic ([zulip](https://chat.fhir.org/#narrow/stream/179177-conformance/topic/Reference.20to.20Logical.20Model/near/358606109)). Unfortunately, that code was missed in the value set, so in the interim, the `https://hl7.org/fhir/tools/StructureDefinition/logical-target` extension should be used instead...

To make this easier for users, we propose a new `Characteristics:` keyword for logicals. This keyword accepts the characteristic codes (separated by commas). E.g.:
```
Logical: Treadmill
Characteristics: #can-be-target
// ...
```

Until the `#can-be-target` code is available, SUSHI can automatically substitute the temporary extension.

SUSHI should also detect when a logical is used in a `Reference` and produce an _error_ if the logical has not specified the `#can-be-target` characteristic.

mint-thompson:
Fixes #1279 and completes task [CIMPL-1104](https://standardhealthrecord.atlassian.net/browse/CIMPL-1104).

Logical models can define Characteristics that indicate how they should be used. Since there can be multiple characteristics, a list of codes can be provided. Only allowed codes will be set on the logical model. Any disallowed codes result in error messages. When exporting the logical model, the codes are checked against the [type-characteristics-code](http://hl7.org/fhir/type-characteristics-code) CodeSystem. If the code is not present in that CodeSystem, a warning is logged. Since that CodeSystem is in the FHIR core extensions package, it should usually be available, but if it isn't, log a warning to let the author know that we aren't checking any of their characteristic codes.

There's some special handling for the `can-be-target` code, since that is supposed to be supported, but isn't in the CodeSystem, so it has to be set with a different extension for now.


src/import/FSHImporter.ts
```diff
@@ -996,6 +1008,37 @@ export class FSHImporter extends FSHVisitor {
...
 
+  visitCharacteristics(ctx: pc.CharacteristicsContext): LogicalCharacteristic[] {
+    const characteristics: LogicalCharacteristic[] = [];
+    ctx.CODE_ITEM().forEach(codeCtx => {
+      const characteristicCode = codeCtx.getText().slice(0, -1).trim().slice(1);
+      if (isLogicalCharacteristic(characteristicCode)) {
+        characteristics.push(characteristicCode);
+      } else {
+        logger.error(
+          'Invalid Characteristic. Supported characteristic codes are "#has-target", "#has-range", "#is-continuous", "#has-length", "#has-size", "#can-bind", "#has-units", "#do-translations", and "#can-be-target".',
```
jafeltra, line 1019: This is also just a question/opinion thing, but I wonder if it is worth including the codes in a defined array that could be `.join()`'ed and used in these two error messages, and maybe in the `isLogicalCharacteristic` function as well, in case the codes ever change.

jafeltra (commented):
This looks really good. I just had a few questions, I think they're all just around code style, so feel free to push back on any of them if you don't like them. I also did have one other question below:

- The issue linked also mentions the following:
  > SUSHI should also detect when a logical is used in a Reference and produce an error if the logical has not specified the #can-be-target characteristic.

  Is that something that we plan to implement as part of this task since this task is resolving that issue? If it is not planned to be part of this task, can we hold off on closing that issue until we do implement that check?

antlr/src/main/antlr/FSH.g4
```diff
@@ -10,7 +10,7 @@ alias:              KW_ALIAS SEQUENCE EQUAL (SEQUENCE | CODE);
 
 profile:            KW_PROFILE name sdMetadata+ sdRule*;
 extension:          KW_EXTENSION name (sdMetadata | context)* sdRule*;
-logical:            KW_LOGICAL name sdMetadata* lrRule*;
+logical:            KW_LOGICAL name (sdMetadata | characteristics)* lrRule*;
```
cmoesel, line 13: I wondered if we should also support this on `resource` -- but after looking at some R5 resources, it looks like FHIR core does not use it on resources, so it probably isn't necessary. Maybe there are just assumptions that can be made about resources in general (like they always can be a target). Anyway... I think we're good!

src/export/StructureDefinitionExporter.ts
```diff
@@ -541,6 +547,58 @@ export class StructureDefinitionExporter implements Fishable {
...
+    // http://hl7.org/fhir/ValueSet/type-characteristics-code
+    // the value set is composed of all codes in the type-characteristics-code code system.
+    // http://hl7.org/fhir/type-characteristics-code
+    // however, the can-be-target code may be missing. if so, use the logical-target extension:
+    // http://hl7.org/fhir/tools/StructureDefinition/logical-target
+    // see this thread for a discussion of this:
+    // https://chat.fhir.org/#narrow/stream/179177-conformance/topic/Reference.20to.20Logical.20Model/near/358606109
+    if (this.typeCharacteristicCodes == null) {
+      logger.warn(
+        `Type characteristics code system not found. ${fshDefinition.name} may include invalid characteristics.`,
```
cmoesel, line 563: I think people might worry about this message as it is worded now (e.g., "what did it see that made it think these codes may be invalid?"). Perhaps it would be better to say something like this?
```suggestion
        `Type characteristics code system not found. Skipping validation of characteristics for ${fshDefinition.name}.`,
```

src/export/StructureDefinitionExporter.ts
```diff
@@ -1258,6 +1316,10 @@ export class StructureDefinitionExporter implements Fishable {
    */
   export(): Package {
     this.checkInvariants();
+    this.typeCharacteristicCodes = this.fishForFHIR(
+      'http://hl7.org/fhir/type-characteristics-code',
+      Type.CodeSystem
+    ) as CodeSystem;
```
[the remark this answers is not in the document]
cmoesel: I think `export()` only gets called once, and then it loops through all the SDs on line 1324. So it's probably OK, but I still think it's worth considering initializing it in the constructor since it's a private member of the class.

src/fhirtypes/common.ts
```diff
@@ -1456,3 +1457,18 @@ export function orderedCloneDeep(input: any, keys?: string[]): any {
...
+  if (cs.concept == null) {
+    return false;
+  }
+  return cs.concept.some(concept => {
+    if (concept.code === target) {
+      return true;
+    }
+    if (concept.concept != null) {
+      return includesConcept(target, concept);
+    }
```
cmoesel, line 1471: Oooooh... Smart.

src/fshtypes/Logical.ts
```diff
@@ -22,3 +24,31 @@ export class Logical extends FshStructure {
...
+      'is-continuous',
+      'has-length',
+      'has-size',
+      'can-bind',
+      'has-units',
+      'do-translations',
+      'can-be-target'
+    ].indexOf(s) >= 0
+  );
+}
```
cmoesel, line 54: Hard-coding these codes here is a little brittle because as soon as the next version (or pre-release) of FHIR comes out, the list could change. We look up the code system in the exporter so that we can avoid hard-coding the valid codes there; but I think we lose the value of that if we're going to validate against a hard-coded list in the importer anyway. Do we need to validate in the importer _and_ exporter?

src/import/FSHImporter.ts
cmoesel: If we have to hard-code them, that's probably a good approach to localize the values to one place, but... I do think we should consider validating only in the exporter -- and we could dynamically draw the list of valid codes from the code system in that case and still log an error telling the user allowed values. I do like how helpful this message is.

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -561,7 +561,7 @@ describe('StructureDefinitionExporter R4', () => {
     it('should throw ParentDeclaredAsNameError when the extension declares itself as the parent', () => {
       const extension = new Extension('Foo');
       extension.parent = 'Foo';
-      doc.logicals.set(extension.name, extension);
+      doc.extensions.set(extension.name, extension);
```
cmoesel, line 564: Whoopsy!

cmoesel (changes requested):
This is a lovely feature. I've made a few suggestions. In particular, if we can avoid hard-coding valid characteristics, I think that would be good. If we can't avoid it, then we might as well use the hard-coded ones in the exporter too (since at that point there's not much value in dynamically deriving the valid characteristics).

src/export/StructureDefinitionExporter.ts
```diff
@@ -541,6 +547,58 @@ export class StructureDefinitionExporter implements Fishable {
...
+        `Type characteristics code system not found. ${fshDefinition.name} may include invalid characteristics.`,
+        fshDefinition.sourceInfo
+      );
+    }
+    const characteristicExtensions: StructureDefinition['extension'] = [];
+    fshDefinition.characteristics.forEach(characteristic => {
+      if (this.typeCharacteristicCodes != null) {
+        if (includesConcept(characteristic, this.typeCharacteristicCodes)) {
+          characteristicExtensions.push({
+            url: 'http://hl7.org/fhir/StructureDefinition/structuredefinition-type-characteristics',
```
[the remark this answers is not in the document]
mint-thompson: it's unlikely to change, but the same string is reused enough times that it makes sense to put it in a variable.

src/import/FSHImporter.ts
mint-thompson: I think you're right, both that we should avoid hard-coding the list, and that it's fine to only validate on the export side. In that case, I'll update the message on the export side to give a listing of the supported codes (with a little extra woozeling to make sure that `can-be-target` appears in the message, even if it's not in the code system.)

src/fhirtypes/common.ts
```diff
@@ -854,6 +858,25 @@ export function replaceReferences<T extends AssignmentRule | CaretValueRule>(
       if (!value.reference.includes('/')) {
         assignedReference.reference = `${type}/${id}`;
       }
+      const typeFhir = fisher.fishForFHIR(type);
+      // if typeFhir is a logical, it needs to have can-be-target characteristic.
+      // this characteristic may appear as one of two extensions.
+      if (typeFhir?.kind === 'logical') {
```
jafeltra, line 864: I'm not completely sure, but when we differentiate between profiles and logicals in FPL, it looks like we check that it is a `resourceType` of `'StructureDefinition'`, with a `kind` of `'logical'` and a derivation of `'specialization'`. I'm not sure if all that is needed here, but maybe?

src/import/FSHTank.ts
```diff
@@ -412,6 +418,82 @@ export class FSHTank implements Fishable {
...
+                    );
+                  })
+                  .pop() as CaretValueRule;
+                if (valueRule?.value === true) {
+                  meta.canBeTarget = true;
+                  break;
+                }
+              }
+            }
+          }
```
jafeltra, line 496: Wow! This logic was tricky. Nice!

jafeltra (commented):
This looks really good. I just had one more question about how we decide if a definition is a logical model.

src/fhirtypes/common.ts
mint-thompson: That's a good question. I was implementing this under the assumption that a profile of a logical model would also need to have the `can-be-target` characteristic, which it would typically inherit from the parent definition. If this assumption is correct, I can update the comment to state this more clearly.

jafeltra: Ohhh that is a good point. I actually don't know the answer to that, but I guess I could test it out in the IG Publisher and see what happens.

cmoesel (changes requested):
The checks really add some complexity, don't they?

I have a few questions and suggestions. I'm also curious -- has anyone ran a regression on this? We touch enough stuff that I feel like we want to ensure we're not unintentionally breaking anything.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1011,6 +1011,30 @@ export class ElementDefinition {
       }
       typeMatches.get(typeMatch.code).push(typeMatch);
     }
+    // check Reference and CodeableReference type matches for logical targetProfiles.
+    // logical models should have an extension set to indicate that they can be used as reference targets.
```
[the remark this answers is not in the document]
mint-thompson: My implementation might be a little too subtle: the metadata attribute `canBeTarget` is `undefined` for non-logical types. So, when filtering the list of types, the attribute won't be equal to `false`, and those types won't be included in the list.

src/fhirtypes/common.ts
```diff
@@ -854,6 +858,25 @@ export function replaceReferences<T extends AssignmentRule | CaretValueRule>(
...
+          );
+        });
+        if (!canBeTarget) {
+          logger.warn(
+            `Referenced type ${typeFhir?.name ?? typeFhir?.id ?? type} for logical instance ${
+              value.reference
+            } does not specify that it can be the target of a reference.`,
+            rule.sourceInfo
+          );
+        }
```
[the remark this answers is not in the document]
mint-thompson: This is a good idea, thank you! It makes sense for these functions to be as consistent as possible.

mint-thompson:
> The checks really add some complexity, don't they?
> 
> I have a few questions and suggestions. I'm also curious -- has anyone ran a regression on this? We touch enough stuff that I feel like we want to ensure we're not unintentionally breaking anything.

The checks are definitely... a little intense. But, I think your suggestions will help make the implementation a lot cleaner. Also, I ran the small regression set, which showed no changes. I'm running the full regression set now.

mint-thompson:
@cmoesel Thank you again for your suggestions. I've made fishing for metadata more consistent and added some logic to hopefully reduce the amount of fishing. Also, I ran the full regression set, and no changes were reported. That's about what I expected, given that this PR only adds new warnings, not errors.

jafeltra (approved):
This was tricky, but I think it all makes sense to me!

PR #1294, opened 2023-06-15
Log an error when applying type and binding constraints on CodeableReference subelements

### LINKED ISSUE #1264: Error on applying type and binding constraints on CodeableReference sub-elements
[The spec says](https://hl7.org/fhir/R5/references.html#CodeableReference):
> The targetProfile and binding constraints on the CodeableReference apply to the CodeableReference.reference or the CodeableReference.concept respectively as appropriate and they SHALL NOT be specified directly on the .reference or .concept elements. 

It would be nice of SUSHI to error when people try to do these invalid things. The error should suggest changing the binding to the parent.

jafeltra:
Fixes #1264 

When constraining the `reference` portion of the CodeableReference or applying a binding to the `concept` portion of the CodeableReference, the constraints should be applied to the top level CodeableReference element. If an author tries to add the to the `reference` or `concept` portion of the CodeableReference element directly, log an error.

As it is implemented now, SUSHI will still apply the constraint or binding to the element as the author specified, but this is not correct according the the spec. If we think SUSHI should not even apply the element at all, I can adjust the implementation to throw an error, which will not apply the constraint or binding.

### DIFF
--- a/src/fhirtypes/ElementDefinition.ts
+++ b/src/fhirtypes/ElementDefinition.ts
@@ -998,6 +998,22 @@ export class ElementDefinition {
       );
     }
 
+    // Check if a CodeableReference is attempting to constraint directly to the reference element
+    if (
+      this.type.length === 1 &&
+      this.type[0].code === 'Reference' &&
+      this.path.endsWith('.reference') &&
+      this.parent()?.type?.[0]?.code === 'CodeableReference'
+    ) {
+      const errorMessage =
+        'Constraining references on CodeableReference elements should not be applied directly on the .reference element. ' +
+        `The constraint should be applied to the parent element: ${this.path
+          .split('.')
+          .slice(1, -1) // slice of the resource name and the final .concept portion in order to match the recommended FSH path
+          .join('.')}`;
+      logger.error(errorMessage);
+    }
+
     // Setup a map to store how each existing element type maps to the input types
     const typeMatches: Map<string, ElementTypeMatchInfo[]> = new Map();
     targetTypes.forEach(t => typeMatches.set(t.code, []));
@@ -1591,6 +1607,21 @@ export class ElementDefinition {
     );
     if (isEmpty(validTypes)) {
       throw new CodedTypeNotFoundError(this.type ? this.type.map(t => t.code) : []);
+    }
+    // Check if a CodeableReference is attempting to bind directly to the concept element
+    if (
+      this.type.length === 1 &&
+      this.type[0].code === 'CodeableConcept' &&
+      this.path.endsWith('.concept') &&
+      this.parent()?.type?.[0]?.code === 'CodeableReference'
+    ) {
+      const errorMessage =
+        'Binding constraints on CodeableReference elements should not be applied directly on the .concept element. ' +
+        `The constraint should be applied to the parent element: ${this.path
+          .split('.')
+          .slice(1, -1) // slice of the resource name and the final .concept portion in order to match the recommended FSH path
+          .join('.')}`;
+      logger.error(errorMessage);
     }
     const strengths = ['example', 'preferred', 'extensible', 'required'];
     // Check if this is a valid strength (if the binding.strength already exists)
--- a/src/fhirtypes/ElementDefinition.ts
+++ b/src/fhirtypes/ElementDefinition.ts
@@ -1006,7 +1006,7 @@ export class ElementDefinition {
       this.parent()?.type?.[0]?.code === 'CodeableReference'
     ) {
       logger.error(
-        'Constraining references on CodeableReference elements should not be applied directly on the .reference element',
+        "Constraining references on a CodeableReference element's underlying .reference path is not allowed. Instead, constrain the references directly on the CodeableReference element.",
         rule.sourceInfo
       );
     }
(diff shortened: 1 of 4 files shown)


mint-thompson (commented):
I feel like SUSHI shouldn't apply a constraint that we know is wrong. So the options would be either "move the constraint to the correct element" or "throw out the constraint." Between these two, I think I would prefer the latter. The error message directly tells the author what they need to change. If SUSHI fixes it for them, then there's not as much reason for the author to actually fix their FSH.

jafeltra:
Sorry, I'm just getting back to answering these questions!

1. Yes. It violates `sdf-24` and `sdf-25` (in [this](http://hl7.org/fhir/structuredefinition.html#invs) list). The publisher adds an error for each constraint violation so the qa.html file will have two errors like:
  ```
error | Rule  sdf-24: 'For CodeableReference elements, target profiles must be listed  on the CodeableReference, not the CodeableReference.reference' Failed
  ```
  and
  ```
error | Rule  sdf-25: 'For CodeableReference elements, bindings must be listed on the  CodeableReference, not the CodeableReference.concept' Failed
  ```
2.  I ran a full regression last night, and there is only one repo, `HL7/fhir-spl`, that hits this case, and it actually logs an error for binding and an error for constraining references.

jafeltra:
After talking to @cmoesel, I think it makes sense to continue to add the constraints and log an error. We would be applying the change that the author specified, and the violation is a little less strong than the other places where we throw an error and don't process a rule. In addition, since SUSHI will log an error, the author should be aware of the issue, and if they use the IG Publisher, the publisher will stop once they encounter the error, at which point they can go back and edit their FSH to correct the issue.

@mint-thompson - if you disagree or have additional thoughts we should take into consideration, please let me know!

src/fhirtypes/ElementDefinition.ts
```diff
@@ -972,6 +972,22 @@ export class ElementDefinition {
...
+      this.path.endsWith('.reference') &&
+      this.parent()?.type?.[0]?.code === 'CodeableReference'
+    ) {
+      const errorMessage =
+        'Constraining references on CodeableReference elements should not be applied directly on the .reference element. ' +
+        `The constraint should be applied to the parent element: ${this.path
+          .split('.')
+          .slice(1, -1) // slice of the resource name and the final .concept portion in order to match the recommended FSH path
+          .join('.')}`;
+      logger.error(errorMessage);
```
mint-thompson, line 988: I think it would be useful to include `rule.sourceInfo` in the logged message. Also, if the rule was applied with a path that includes a slice, this message won't include it, since `this.path` doesn't include slices. So, maybe it should use `this.id` instead?

mint-thompson (changes requested):

src/fhirtypes/ElementDefinition.ts
jafeltra: Great point on both. I wonder if maybe trying to include the path is too much here. Could there be places where the `this.id` value doesn't actually match what the FSH path should be? Like for slices, won't the syntax not match? I think maybe this message is trying to be too helpful and is becoming unhelpful. I'm leaning towards just taking out the second sentence and leaving the first. This would apply to both the error messages.

jafeltra: I sort of just went for it and removed the path sentence because I talked myself into thinking this was adding unnecessary complication, especially now that the rule info is listed and the line number will be right there. So [e2d9bdb](https://github.com/FHIR/sushi/pull/1294/commits/e2d9bdbc5b443fd198efd3a280049be84b6bdcc3) removes that second sentence where the path could have been wrong.

That same commit also adds the rule source info to the message. I did have to pass in `rule.sourceInfo` to `bindToVs` in order to support that. I made it an optional parameter in the function because there are some places where we bind without it coming directly from a rule.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1592,6 +1610,18 @@ export class ElementDefinition {
...
     }
+    // Check if a CodeableReference is attempting to bind directly to the concept element
+    if (
+      this.type.length === 1 &&
+      this.type[0].code === 'CodeableConcept' &&
+      this.path.endsWith('.concept') &&
+      this.parent()?.type?.[0]?.code === 'CodeableReference'
+    ) {
+      logger.error(
+        'Binding constraints on CodeableReference elements should not be applied directly on the .concept element.',
```
cmoesel, line 1621: Similar to the other message, we need to also tell them what they _should_ do. If you liked my other suggestion, this one kind of mirrors it:
```suggestion
        'Applying value set bindings to a CodeableReference element's underlying .concept path is not allowed. Instead, apply the binding directly to the CodeableReference element.',
```

cmoesel (changes requested):
This looks good! It's a nice clean approach. I just have a few suggestions regarding the wording of the error messages.

jafeltra:
@cmoesel - those are both good ideas. I should have replaced the sentence I took out with something like this, so thanks! I added them in a separate commit (rather than taking the suggestions) so I could fix the tests as well. [d6904ca](https://github.com/FHIR/sushi/pull/1294/commits/d6904cafd437cd91752c33e39379d407f4393f41) has everything.

mint-thompson (approved):

cmoesel (approved):
Splendid! Please merge at will!

PR #1295, opened 2023-06-27
Include version in Only Rules

### LINKED ISSUE #1270: SUSHI drops version in versioned "only" rules
Given this FSH:
```
Profile: MyPatient
Parent: Patient
* ^version = "1.0"

Profile: MyComposition
Parent: Composition
* section.entry only Reference(MyPatient|1.0)
```

SUSHI drops the version when it constrains `section.entry`:
```json
      {
        "id": "Composition.section.entry",
        "path": "Composition.section.entry",
        "type": [
          {
            "code": "Reference",
            "targetProfile": [
              "http://example.org/StructureDefinition/MyPatient"
            ]
          }
        ]
      }
```

Since `targetProfile` is a `canonical`, it should retain the `|1.0`.

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Adding.20version.20to.20Reference

jafeltra:
This PR ensures that if a version is specified in a `Reference` or `CodeableReference`, the version is included in the targetProfile information as well. Previous updates already ensure that the version is used when fishing and the fallback if the version is not found works as intended. This just updates the url to include the version so it is not lost when building the targetProfile list. This was already in place for `Canonical`s specifically, so I included both `Reference` and `CodeableReference` in the same place.

Fixes #1270.

### DIFF
--- a/src/fhirtypes/ElementDefinition.ts
+++ b/src/fhirtypes/ElementDefinition.ts
@@ -1006,7 +1006,10 @@ export class ElementDefinition {
     for (const type of types) {
       const typeMatch = this.findTypeMatch(type, targetTypes, fisher);
       // if the type is Canonical, it may have a version. preserve it in the match's metadata.
-      if (type.isCanonical && type.type.indexOf('|') > -1) {
+      if (
+        (type.isCanonical || type.isReference || type.isCodeableReference) &&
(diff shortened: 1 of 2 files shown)


cmoesel (approved):
I love PRs like this.

PR #1302, opened 2023-07-11
Fix problems when setting values on child elements in primitive arrays

mint-thompson:
Completes task [CIMPL-1116](https://standardhealthrecord.atlassian.net/browse/CIMPL-1116).

Child elements of primitives are represented in JSON by using the name of the primitive element prefixed with an underscore. When the primitive element is also an array, create both the regular and underscore-prefixed arrays at the same time. When adding an element to one array, add a corresponding element to the other. Both arrays track slice names for sliced arrays. While this will sometimes result in the creation of arrays that contain no meaningful elements, these arrays are already cleaned up during export.

This PR changes the way soft index resolution works with `manualSliceOrdering` enabled to make it more accurately fulfill the requirement of "no access to elements with slice names without using the slice name." As a result, while the index values returned from `convertSoftIndicesStrict` have changed, the resulting instances are the same. A similar change is made in `createUsefulSlices`, which is used when `manualSliceOrdering` is enabled.

The useful outcomes of this PR are:
- The correct number of array elements are created when assigning values to a primitive array and its children.
- Cardinality requirements that apply to a primitive array and its children are correctly enforced. This has been observed in cases where the primitive array's `value` child is required on a defined slice, but not required in general.

As there are a lot of potential paths through the new code depending on the order in which primitive elements and their children are assigned, I tried to write tests as well as I could to achieve code coverage. Additionally, there were some places in my initial implementation where code coverage was logically impossible, and the unreachable code was removed. In particular, there are places in [`createUsefulSlices`](https://github.com/FHIR/sushi/compare/cimpl-1116-primitive-array?expand=1#diff-c3ea71b50f004cf55770cf8d3e9c5483a08b05e68093720ea2fd106da774ded6R196) and [`setPropertyOnInstance`](https://github.com/FHIR/sushi/compare/cimpl-1116-primitive-array?expand=1#diff-c3ea71b50f004cf55770cf8d3e9c5483a08b05e68093720ea2fd106da774ded6R659) where I think it is logically impossible to reach the code when `sliceName` is not null. Therefore, these code blocks do not attempt to use `sliceName`. I would appreciate a careful review of these sections in particular.

No changes appeared in the partial regression set. I'm going to be running a full regression tonight, and I will post the results in this thread.

### DIFF
--- a/src/fhirtypes/common.ts
+++ b/src/fhirtypes/common.ts
@@ -157,18 +157,6 @@ export function createUsefulSlices(
             }
           } else {
             // This is an array entry that does not have a named slice (so a typical numeric index)
-            // but "no slice name" is its own group.
-            const sliceIndices: number[] = [];
-            current[pathPart.base]?.forEach((el: any, i: number) => {
-              if (el?._sliceName == null) {
-                sliceIndices.push(i);
-              }
-            });
-            if (ruleIndex >= sliceIndices.length) {
-              effectiveIndex = ruleIndex - sliceIndices.length + current[pathPart.base].length;
-            } else {
-              effectiveIndex = sliceIndices[ruleIndex];
-            }
             knownSlices.set(
               currentPath,
               Math.max(effectiveIndex + 1, knownSlices.get(currentPath) ?? 0)
@@ -440,8 +428,23 @@ export function setImpliedPropertiesOnInstance(
       }
       // check the children for instance of this element
       const children = currentElement.def.children(true);
+      // special handling: if our current element has no slice name, we need guarantee the defined minimum
+      // this is the only place where we do this, in order to accomodate cases where some named slices already exist
+      let existingSliceCount = 0;
+      if (finalMin < currentElement.def.min && currentElement.def.sliceName == null) {
+        // our final min was lowered by slices, so add that to our indices.
+        const slicePaths = currentElement.def
+          .getSlices()
+          .map(el => `${tracePath}[${el.sliceName}]`);
+        slicePaths.forEach(slicePath => {
+          if (knownSlices.has(slicePath)) {
+            existingSliceCount += knownSlices.get(slicePath);
+          }
+        });
+      }
       for (let idx = 0; idx < finalMin; idx++) {
-        const newHistory = traceParts.slice(-1)[0] + (idx > 0 ? `[${idx}]` : '');
+        const effectiveIdx = idx + existingSliceCount;
+        const newHistory = traceParts.slice(-1)[0] + (effectiveIdx > 0 ? `[${effectiveIdx}]` : '');
         elementsToCheck.push(
           ...children.map(
             child =>
@@ -540,7 +543,7 @@ export function setImpliedPropertiesOnInstance(
   }
   sortedRulePaths.forEach(path => {
     const { pathParts } = instanceOfStructureDefinition.validateValueAtPath(path, null, fisher);
-    setPropertyOnInstance(instanceDef, pathParts, sdRuleMap.get(path), fisher, manualSliceOrdering);
+    setPropertyOnInstance(instanceDef, pathParts, sdRuleMap.get(path), fisher);
   });
 }
 
@@ -582,8 +585,7 @@ export function setPropertyOnInstance(
   instance: StructureDefinition | ElementDefinition | InstanceDefinition | ValueSet | CodeSystem,
   pathParts: PathPart[],
   assignedValue: any,
-  fisher: Fishable,
-  manualSliceOrdering = false
+  fisher: Fishable
 ): void {
   if (assignedValue != null) {
     // If we can assign the value on the StructureDefinition StructureDefinition, then we can set the
@@ -629,19 +631,6 @@ export function setPropertyOnInstance(
               el?._sliceName === sliceName ||
               (isExtension(pathPart.base) && el?.url && el?.url === sliceExtensionUrl)
             ) {
-              sliceIndices.push(i);
-            }
-          });
-          // Convert the index in terms of the slice to the corresponding index in the overall array
-          if (index >= sliceIndices.length) {
-            index = index - sliceIndices.length + current[key].length;
-          } else {
-            index = sliceIndices[index];
-          }
-        } else if (manualSliceOrdering) {
-          const sliceIndices: number[] = [];
-          current[key]?.forEach((el: any, i: number) => {
-            if (el?._sliceName == null) {
               sliceIndices.push(i);
             }
           });
--- a/src/fhirtypes/common.ts
+++ b/src/fhirtypes/common.ts
@@ -610,14 +610,22 @@ export function setPropertyOnInstance(
       let index = getArrayIndex(pathPart);
       let sliceName: string;
       if (index != null) {
-        // If the array doesn't exist, create it
-        if (current[key] == null) {
+        if (pathPart.primitive) {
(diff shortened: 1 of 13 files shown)


src/fhirtypes/common.ts
```diff
@@ -171,14 +194,34 @@ export function createUsefulSlices(
...
                 current[key].push({ _sliceName: sliceName });
+                if (underscored) {
+                  current[pathPart.base][current[key].length - 1] = { _sliceName: sliceName };
+                } else if (pathPart.primitive) {
+                  current[`_${pathPart.base}`][current[key].length - 1] = { _sliceName: sliceName };
+                }
               } else if (j === effectiveIndex) {
                 current[key].push({});
+                if (underscored) {
+                  current[pathPart.base][current[key].length - 1] = null;
```
cmoesel, line 214: In several of the situations above, when it is `underscored`, we set the `pathPart.base` value to `{}`, but here we set it to `null`. Is this difference intentional?

src/fhirtypes/common.ts
```diff
@@ -171,14 +194,34 @@ export function createUsefulSlices(
...
+              }
             } else if (j >= current[key].length) {
               if (sliceName) {
                 // _sliceName is used to later differentiate which slice an element represents
                 current[key].push({ _sliceName: sliceName });
+                if (underscored) {
+                  current[pathPart.base][current[key].length - 1] = { _sliceName: sliceName };
+                } else if (pathPart.primitive) {
+                  current[`_${pathPart.base}`][current[key].length - 1] = { _sliceName: sliceName };
+                }
```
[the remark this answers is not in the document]
cmoesel: And this might apply to similar patterns below as well, so I'll stop calling it out.

cmoesel (commented):
I have not fully reviewed all the tests, and I'm not sure when I'll be able to, so I thought I'd at least submit some of my comments/questions for the code...

mint-thompson:
In the full regression set, I've discovered some problems with handling caret rules on a `type.profile.extension` and `type.targetProfile.extension`. I don't think this will require a complete overhaul, but some of the suggestions @cmoesel made may help with those scenarios.

mint-thompson:
I've made some updates that resolve problems that occurred when assigning extensions on `type.targetProfile`. [The PACarePlan profile](https://github.com/HL7/physical-activity/blob/master/input/fsh/Planning.fsh#L71) does this, so that's how I caught it. The array fill on creation was something in my original implementation, but I removed it when I couldn't think of a test case to make it important. I'll have to keep extensions on types in mind for the future. These updates also make use of @cmoesel's suggestions for simplifying some of the branching logic.

test/utils/PathUtils.test.ts
```diff
@@ -170,9 +170,9 @@ describe('PathUtils', () => {
       resolveSoftIndexing(rules, true);
       expect(rules.map(r => r.path)).toEqual([
         'component[Bread]',
-        'component[1]',
-        'component[1]',
-        'component[2]'
+        'component[0]',
+        'component[0]',
+        'component[1]'
```
cmoesel, line 175: From your PR description:
> This PR changes the way soft index resolution works with manualSliceOrdering enabled to make it more accurately fulfill the requirement of "no access to elements with slice names without using the slice name." As a result, while the index values returned from convertSoftIndicesStrict have changed, the resulting instances are the same.

Hmmm.... As long as the author is using `[+]`, I think it makes sense to them for it to come after `[Bread]`. But if the author uses literal `[0]`, I think they would expect it to be the _first_ element in the array. I know I would.

If `[0]` _isn't_ always the first element in the array, then I think we might be violating the [Array Paths Using Numerical Indices](https://build.fhir.org/ig/HL7/fhir-shorthand/reference.html#array-paths-using-numerical-indices) part of the spec.

So, IMO, I'm not sure this is a good change. Like I said, when thinking about it in turns of soft-indexing, it's fine. But I think that what you did to make this work also affects numeric indexing (right?) - and I'm not sure that's OK.

cmoesel (changes requested):
It's looking good, but this might be the gift that keeps on giving, as I've left at least one sort of significant thing for you to consider... perhaps it is worth discussing as a team.

test/utils/PathUtils.test.ts
mint-thompson: Unfortunately, having looked at the spec a bit more closely, I think you're right. I think I got caught up in the idea that manual slice ordering means that an array element without a slice name can't be referenced without the slice name, when that's technically not in the spec. All we can try to guarantee is that with manual slice ordering turned on, if you use soft indexing and named slices, none of your soft indices without slice names will end up referring to an element with a slice name. But, if an author makes a bunch of slices, and then a bunch of rules with hard numeric indices, we can't make that guarantee. This matches up with the descriptions in the [SUSHI docs](https://fshschool.org/docs/sushi/tips/#manual-slice-ordering).

Somehow, I got the idea in my head of a much stricter restriction: that even hard numeric indices wouldn't be able to access named slices when manual slice ordering is enabled. Then I went and implemented that. So, I'll revert that, and adjust the implementation accordingly. Thank you for reminding me of what we've already written in the FSH spec and SUSHI docs!

mint-thompson: Just re-read some of the SUSHI docs, and this appears in the section about manual slice ordering:

```
Instance: ExampleByName
InstanceOf: ExampleBPObservation
// some required elements omitted for brevity
* component[systolicBP].valueQuantity = 108 'mm[Hg]'
* component[diastolicBP].valueQuantity = 45 'mm[Hg]'

Instance: ExampleByNumber
InstanceOf: ExampleBPObservation
// some required elements omitted for brevity
* component[0].valueQuantity = 108 'mm[Hg]'
* component[1].valueQuantity = 45 'mm[Hg]'
```

> When manual slice ordering is enabled, rules that set values on the systolicBP or diastolicBP slices must use the slice name. With this option enabled, the ExampleByName instance would produce the same entries in component, but the ExampleByNumber instance would contain four entries in the component list: two entries that are not part of a named slice, followed by the systolicBP slice, and finally the diastolicBP slice.

This appears to directly contradict the preceding part of that section of the SUSHI docs. So, this part of the SUSHI docs may need to be revised.

cmoesel: I don't think it contradicts it. Remember, if you don't explicitly refer to a required slice (with `manualSliceOrdering` on), then the required slices will go at the _end_ of the array. So `component[0]` and `component[1]` are not overwriting anything since they go first -- and then the required slices are applied _after_ that.

cmoesel: It's only with `manualSliceOrdering` OFF that the required slices would get written first, and _then_ `component[0]` and `component[1]` would overwrite them.

mint-thompson: @cmoesel you're right: there's no contradiction here. I've clearly been down in the slice mines for too long.

cmoesel (approved):
Is it complicated? Yup. Is it working? Yup. Is it working _well_? For sure!

I just tried this with the original examples from CIMPL-1116, and wow -- they were messed up before; but now they are perfect. This is a good change and I'm glad to get those pretty nasty bugs out of the way. Thanks, @mint-thompson!

jafeltra (approved):
I think all this makes sense. I followed along with the tests, and I think the changes here make sense and don't add too much extra complexity. Nice job wrangling some very tricky code.

PR #1304, opened 2023-07-13
Update fhir-package-loader dependency

mint-thompson:
Completes task [CIMPL-1126](https://standardhealthrecord.atlassian.net/browse/CIMPL-1126).

fhir-package-loader now supports "latest" as a version string, so there is no more need for custom logic to handle it.


cmoesel (approved):
I love deleting code!

jafeltra (approved):
Neat!

PR #1307, opened 2023-07-17
Log the name of the file that is not processed by SUSHI

### LINKED ISSUE #1306: Log the path and name of files that are not processed
When SUSHI finds a file ini the `input` directory that it does not process, it prints a log message:

> Sushi: info  Found 1 file in an input/* resource folder that was neither XML nor JSON. This file was not processed as a resource by SUSHI.

This is not super helpful. For me, it is often an OSX file system file `.DS_Store`. But sometimes it is a file with a mis-spelled file suffix, or a file that's supposed to be in a different directory. From the message, I cannot tell.

It would be nice if SUSHI would just print the name (and the path) of the file too.

mrinnetmaki:
Add a log message for each of the files not processed by SUSHI.

I get that the current implementation is faster, and also results in fewer log messages when there are a lot of files in the `input` directory that are not processed by SUSHI. But there should not be that many such files, right? So let's just print the name of each of the files, and then the total count too. It's much more helpful.

Fixes #1306.

### DIFF
--- a/src/fhirdefs/load.ts
+++ b/src/fhirdefs/load.ts
@@ -68,9 +68,7 @@ export function loadCustomResources(
             resourceJSON = converter.xmlToObj(xml);
           } else {
             invalidFileCount++;
-            logger.info(
-              `File not processed by SUSHI: ${file}`
-            );
+            logger.debug(`File not processed by SUSHI: ${file}`);
             continue;
           }
         } catch (e) {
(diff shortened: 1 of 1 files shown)


cmoesel:
Thanks for submitting a PR, @mrinnetmaki. It looks like it failed our file formatting check. The easiest way to fix this is to run: `npm run prettier:fix`.

You can check that all the files follow our formatting rules (and that all tests pass) by running: `npm run check`.

As for the functionality, I worry a little issuing lots of extra `info` messages for people who might have many non-xml/non-json files in their input (for whatever reason). How would you feel about changing this in one of the following ways:
* Change the logger to log the extra files using `debug` logging -- and in the message that tells you there were non-json/non-xml files, add this at the end: `To see the unprocessed files in the logs, run SUSHI with the "--log-level debug" flag.`
* OR... Instead of logging each file as a separate info message, collect the file paths in an array and then append them to the end of the message that says there were non-xml/non-json files.

How would you feel about either of those changes?

mrinnetmaki:
Thanks for the feedback, @cmoesel! Using debug level log message is OK.

cmoesel (approved):
Thank you, @mrinnetmaki. This looks good to me!

jafeltra (approved):
This looks like it will be helpful. I like the debug message approach. Thanks!

PR #1309, opened 2023-07-20
Update antlr4 dependency

mint-thompson:
Completes task [CIMPL-1140](https://standardhealthrecord.atlassian.net/browse/CIMPL-1140).

Update import paths for antlr4 classes. Remove tsconfig-paths, as it is no longer needed for remapping. Remove @types/antlr4, since types are now included in the main antlr4 package.

I tried this out with GoFSH, npx, and FSH Online. There were no problems with GoFSH, and it appeared to work fine with npx when using a packed-up version of SUSHI. With FSH Online, the situation is not yet fully resolved. The application itself runs correctly, and the problems that were appearing when using the `Context` keyword are resolved. But, the test scripts still have some import problems. This is probably something that will have to get resolved by some adjustments to FSH Online's jest configuration. I talked about this with @jafeltra, and we agreed that it would be fine to make that something to investigate separately from this task.


cmoesel (approved):
I love this PR. All those nasty hoops we had to jump through for ANTLR: GONE!

I ran a full regression and there were no changes in output.

Thanks @mint-thompson!

jafeltra (approved):
I tested out running SUSHI with `ts-node`, `node`, and a packed version. And they all worked out well! Hopefully the FSH Online fix will be simpler now 🤞 Thanks!

PR #1310, opened 2023-07-21
Support patch wildcard when including package versions in configuration

### LINKED ISSUE #1219: SUSHI should support x patch wildcard in package versions
The Confluence doc on versions says that the following is allowed as a dependency version: `3.0.x`, meaning the latest patch of `3.0`. SUSHI should support this.

See: https://confluence.hl7.org/pages/viewpage.action?pageId=35718629#NPMPackageSpecification-Versionreferences

jafeltra:
This PR supports using a patch wildcard in a package version that is listed in the `sushi-config.yaml` `dependencies` list. In order to support this functionality, the latest version of FPL is used, which automatically determines the latest satisfying version of the package.

In order to test this, add a dependency with a version that uses a patch wildcard, such as:

```
dependencies:
  hl7.fhir.us.mcode: 2.0.x
```

Fixes #1219

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -15,7 +15,7 @@
         "chalk": "^3.0.0",
         "commander": "^8.2.0",
         "fhir": "^4.9.0",
-        "fhir-package-loader": "^0.4.0",
+        "fhir-package-loader": "^0.5.0",
         "fs-extra": "^8.1.0",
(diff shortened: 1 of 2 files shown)


mint-thompson (approved):
hooray for patch versions!

cmoesel (approved):
Splendid!

PR #1311, opened 2023-07-24
Support looser type checking when constraining profile types

### LINKED ISSUE #1289: Consider issuing a warning (instead of an error) when constraining a type that is not a formal subtype
SUSHI currently only allows authors to constrain a type to a formal subtype of the resource or profile type. However, it is possible that a type could be constrained to another type that does conform to the type but is not a _formal_ subtype. SUSHI should consider loosening this error to a warning. There is a chance that someone used the wrong type and did make a mistake. However, it is possible the author knows the type they are trying to constrain to does actually conform to the base type.

SUSHI may be able to check that the base resource type of the original type and the constrained type match. If they don't match, then SUSHI knows the constraint is not valid. If they do match, SUSHI can't truly tell if the constrained type is allowed and should consider logging a warning saying so.

Zulip thread where this came up: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Question.20about.20inheriting.20profiles
A simple FSH Online example that demonstrates the error SUSHI currently logs: https://fshschool.org/FSHOnline/#/share/3J4LxPR

cmoesel:
When constraining an element's type, the previous behavior required the new type to be a formal child of the old type. If the old type is a profile, however, the new type shouldn't have to be a formal subtype; it need only conform to the old type. When this occurs, log a warning but let it proceed.

You can test this using [Julia's FSH Online example](https://fshschool.org/FSHOnline/#/share/3J4LxPR). On master branch, it will cause an error and not allow the constraint. On this branch it will log a warning but allow the constraint.

But... Oh no! I just tested this w/ IG Publisher and the IG Publisher gives this error:
> Unable to generate snapshot for http://example.org/StructureDefinition/SimpleExampleObservation in /Users/cmoesel/data/fsh/LooserTypeConstraints/fsh-generated/resources/StructureDefinition-SimpleExampleObservation because StructureDefinition http://example.org/StructureDefinition/SimpleExampleObservation at Observation.performer: invalid constrained type Reference from Reference in http://example.org/StructureDefinition/SimpleExampleObservation

It seems that maybe the IG Publisher requires formal inheritance even though that might not be strictly necessary. Putting this into DRAFT mode until we resolve this with the IG Publisher team.

Fixes #1289 and CIMPL-1122.


cmoesel:
Closing this PR, as the IG Publisher does not support looser type checking -- so even if SUSHI supports it, authors will just get errors further down the line.

cmoesel:
Re-opening this PR as draft since Grahame has specified an approach to indicating when a profile conforms to another even if it is not in its direct hierarchy. See this thread: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Question.20about.20inheriting.20profiles/near/389046996

This PR will need to be updated to consider the `structuredefinition-imposeProfile` extension.

cmoesel:
This PR is now obsolete and replaced by #1366.

PR #1313, opened 2023-07-25
Improve Handling of ElementDefinition.binding without valueSet

### LINKED ISSUE #1312: Unexpected error when applying implied extension with a binding, but no value set
Originally reported at https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Issue.20with.20Specimen.2Ecollection.2Edevice.20R5.20extension

This error occurs when applying the implied extension `http://hl7.org/fhir/5.0/StructureDefinition/extension-Specimen.collection.device` for [FHIR 5.0.0 `Specimen.collection.device`](http://hl7.org/fhir/R5/specimen-definitions.html#Specimen.collection.device), which has a binding, but no value set. The following FSH will produce the error:
```
Alias: $specimen-collection-device-r5 = http://hl7.org/fhir/5.0/StructureDefinition/extension-Specimen.collection.device
Profile: CustomSpecimen
Parent: Specimen
* collection.extension contains $specimen-collection-device-r5 named SpecimenCollectionDeviceR5 0..1
* collection.extension[SpecimenCollectionDeviceR5]
```

The error occurs in `applyToValueXElement` in `fhirdefs/impliedExtensions.ts`, which assumes that a binding will include some value set information.

cmoesel:
This PR improves handling of element definition bindings that don't have a specified value set. Due to assumptions in the code, if the `ElementDefinition` class's `bindVS` was called with a null `valueSet`, it would throw an error. Similarly, when processing implied extensions, it was assumed that either `valueSet`, `valueSetUri`, or `valueSetReference` would exist. Now we defensively check against null value sets and allow bindings to set just a strength (since FHIR has bindings like this).

You can test using the example in #1312 or by checking out just the first commit in this PR (e07cc4a) that includes just the tests without the fix. After verifying the bug, you can verify the solution by running the tests again with the solution commit (53bfe45).

Fixes #1312 and CIMPL-1146


jafeltra (approved):
This makes sense to me! I also tested out the example in the issue on each commit, and they both worked as expected.

PR #1314, opened 2023-07-26
v3.2.0

cmoesel:
This PR upgrades FPL to v0.6.0 and bumps its own version number to 3.2.0. I chose to consider it a minor release since support for wildcard versions is a feature, not a fix.


jafeltra (approved):
Seems like a good justification to me! ✅

PR #1315, opened 2023-07-26
Allow caret and insert rules to be applied to concepts within a ValueSet

mint-thompson:
Completes task [CIMPL-1130](https://standardhealthrecord.atlassian.net/browse/CIMPL-1130).

Some of the elements within `ValueSet.compose.include.concept` and `ValueSet.compose.exclude.concept` can't be reached using the normal concept component rules on a ValueSet. To access those elements, you can now use a caret rule with a path that is a contained concept. Similarly, you can insert a RuleSet at such a path. These rules are similar to the code caret rules and code insert rules used by CodeSystem, but since the `concept` elements are a flat list, the path has to be exactly once concept long.

A concept component rule with exactly one concept in it establishes a context for an indented rule to appear after it. So, these two rules:
```
* ZOO#hippo "Hippopotamus"
* ZOO#hippo insert DesignationRules("hipopótamo", #es)
```
are equivalent to these two rules:
```
* ZOO#hippo "Hippopotamus"
  * insert DesignationRules("hipopótamo", #es)
```

The grammar changes to support these rules resulted in some different parser errors in cases where commas are used to list concepts. Additionally, these rules represent a sort of weird case where a rule may set a context, but not itself be indented. Rather than change the set of rules that are allowed to be used in context operations, I added a flag to the context-operation functions so that errors relating to improper indents only get logged once.


src/import/FSHImporter.ts
```diff
@@ -2426,7 +2480,8 @@ export class FSHImporter extends FSHVisitor {
       };
       if (
         !(pc.containsPathContext(ctx) || pc.containsCodePathContext(ctx)) &&
-        location.startColumn - DEFAULT_START_COLUMN > 0
+        location.startColumn - DEFAULT_START_COLUMN > 0 &&
+        !suppressError
```
cmoesel, line 2484: This is a little nit-picky, but `!suppressError` is the easiest operation of this boolean logic -- and there's no need to check anything else if we're suppressing errors -- so I'd suggest this should be _first_ in the chain of boolean logic so it can short-circuit when suppressing errors.

src/import/FSHImporter.ts
```diff
@@ -1760,22 +1787,32 @@ export class FSHImporter extends FSHVisitor {
...
     const fullCodePath = this.getArrayPathWithContext(localCodePath, ctx);
     insertRule.pathArray = fullCodePath;
     return this.applyRuleSetParams(ctx, insertRule);
   }
 
-  visitInsertRule(ctx: pc.InsertRuleContext): InsertRule {
-    const insertRule = new InsertRule(this.getPathWithContext(this.visitPath(ctx.path()), ctx))
+  visitInsertRule(ctx: pc.InsertRuleContext, withPathArray = false): InsertRule {
+    const localPath = this.visitPath(ctx.path());
+    const fullPathArray = this.getArrayPathWithContext(localPath === '' ? [] : [localPath], ctx);
```
cmoesel, line 1809: Is this array stuff only needed because of the undocumented syntax that allows introducing multiple codes from the same system in a single VS compose rule? If so, should we consider removing support for that syntax since it's not documented in the spec? (We'd want to check to see if people are using it before we removed it -- and check with Mark too).

test/import/FSHImporter.ValueSet.test.ts
```diff
@@ -1246,6 +1285,32 @@ describe('FSHImporter', () => {
         expect(vs.rules).toHaveLength(1);
         assertInsertRule(vs.rules[0] as Rule, '', 'MyRuleSet');
       });
+
+      it('should parse an insert rule at a concept', () => {
+        const input = leftAlign(`
+        ValueSet: ZooVS
+        * ZOO#hippo insert DesignationRules
```
cmoesel, line 1292: In practice, this would be invalid, right? Because you need to introduce the code first before you can use it as a context for insert?

test/import/FSHImporter.context.test.ts
```diff
@@ -1027,5 +1027,84 @@ describe('FSHImporter', () => {
...
+      ]);
+      expect(loggerSpy.getAllMessages('error')).toHaveLength(0);
+    });
+
+    it('should parse a value set that uses an indented CaretValueRule after a ValueSetConceptComponentRule with the from keyword', () => {
+      const input = leftAlign(`
+      ValueSet: ZooVS
+      * #hippo "Hippopotamus" from system ZOO
+        * ^designation.value = "hipopótamo"
+        * ^designation.language = "es"
```
cmoesel, line 1059: In addition to the variations you've already done, I think we also need to test variations with `include` and `exclude`:
```
* include ZOO#hippo "Hippopotamus"
  * ^designation.value = "hipopótamo"
```
and
```
* exclude ZOO#hippo "Hippopotamus"
  * ^designation.value = "hipopótamo"
```
and also:
* Cases with multiple codes to ensure context gets properly set and reset each time
* Cases with root-level caret rules after concepts and concept caret rules to ensure the root-level caret doesn't get the concept context.
* Trying to indent a caret rule after a _filter_ rule.

cmoesel (changes requested):
This isn't as messy as you lead me to believe. That's good!

I do have a few questions and suggestions I'd like you to consider.

Thanks!

src/import/FSHImporter.ts
```diff
@@ -1240,17 +1261,18 @@ export class FSHImporter extends FSHVisitor {
...
+    return this.prependPathContext(pathArray, parentCtx, isPathRule, isInstanceRule, suppressError);
   }
 
   visitPath(ctx: pc.PathContext): string {
     return ctx?.getText() || '';
   }
 
   visitCaretPath(ctx: pc.CaretPathContext): string {
-    return ctx.CARET_SEQUENCE().getText();
+    return ctx.CARET_SEQUENCE()?.getText();
```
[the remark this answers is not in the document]
mint-thompson: This scenario occurred when commas were used to list concepts, since the parser would think it had encountered a `codeCaretValueRule`. But, since we're removing the lists of concepts, it should be safe for me to change this back.

test/import/FSHImporter.ValueSet.test.ts
mint-thompson: Yes, this would produce an error during export, since it wouldn't be able to find the `ZOO#hippo` concept anywhere in the compose lists. I can add the concept component in so that this is a more valid scenario.

src/import/FSHImporter.ts
mint-thompson: I'm storing information in `pathArray` because the path might be an actual path or a list of concepts, and that determination gets made elsewhere. So even with removing the syntax for multiple concepts in a single rule, tracking both is useful.

Alternatively, since ValueSets have the special condition of disallowing paths on their normal CaretValueRules, it might be possible to do this without needing to bother with `pathArray` here if I can be careful about always removing the path when it isn't a concept. Then, when exporting the ValueSet, any CaretValueRule with a path is assumed to be referring to something within a compose element.

src/import/FSHImporter.ts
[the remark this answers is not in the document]
cmoesel: I noticed that you did not revert this back. It's probably OK either way, but just calling it out in case you think it is better to revert it.

cmoesel (dismissed):
Fantastic. This works well and has a great set of tests. I tried it out with all the examples in the balloted spec and they went great!

cmoesel (dismissed):
Still good!

test/import/FSHImporter.ValueSet.test.ts
```diff
@@ -1244,7 +1209,37 @@ describe('FSHImporter', () => {
...
+        ValueSet: ZooVS
+        * ZOO#hippo ZOO#big-hippo insert DesignationRules
+        `);
+        const result = importSingleText(input, 'Insert.fsh');
+        const vs = result.valueSets.get('ZooVS');
+        expect(vs.rules).toHaveLength(0);
+        expect(loggerSpy.getAllMessages('error')).toHaveLength(1);
+        expect(loggerSpy.getLastMessage('error')).toMatch(
+          /Only one concept may be listed before an insert rule on a ValueSet\..*File: Insert\.fsh.*Line: 3\D*/s
+        );
```
jafeltra, line 1242: I have a question about this case. We log an error if there are multiple concepts listed before the insert rule, but I don't think we log any error if the caret value rule is included after multiple concepts. For example, something like this doesn't log an errors and adds the designation to the hippo concept. Is that expected?

```
ValueSet: ZooVS
* ZOO#hippo "Hippopotamus"
* ZOO#rhino
* ZOO#hippo ZOO#rhino ^designation.value = "hipopótamo"
```

jafeltra (commented):
I just had one question that could easily just be due to something I missed. Other than that, it looks great and I didn't notice anything else.

test/import/FSHImporter.ValueSet.test.ts
cmoesel: Oh dang. Yeah, multiple concepts in a VS should trigger an error no matter what. Good find.

mint-thompson:
@jafeltra thank you for pointing out that case with the caret rule! I've added a test for that and resolved the issue.

cmoesel (approved):
How many times can I approve the same PR? Hopefully this is the last one! (But if not, that is OK).

jafeltra (approved):
Thanks for making that update! Looks good to me!

PR #1316, opened 2023-07-31
Warn and ignore version when replacing reference

### LINKED ISSUE #1298: Warn when users assign references to resources w/ `|version`
Reference assignments should not use `|` to indicate a version (see [zulip thread](https://chat.fhir.org/#narrow/stream/179177-conformance/topic/syntax.20of.20Reference.2Ereference.20with.20version/near/373027152)). It's valid for constraining reference types, but not assigning references to instances.

mint-thompson:
Fixes #1298 and completes task [CIMPL-1142](https://standardhealthrecord.atlassian.net/browse/CIMPL-1142).

When assigning a reference, versions should not be included. Ignore the version, warn the user to not do that, and try to fish up the referenced item.

A side effect of this is that it is no longer possible to fish up a locally defined entity by name if its name has a `|` character in it. However, it can still be fished up by id, and also, putting `|` in a name is not recommended by the FHIR spec. Please don't do it.

### DIFF
--- a/src/fhirtypes/common.ts
+++ b/src/fhirtypes/common.ts
@@ -940,6 +940,16 @@ export function replaceReferences<T extends AssignmentRule | CaretValueRule>(
           rule.sourceInfo
         );
       }
+    } else {
+      // if we still haven't found anything, there's one more possibility:
+      // the reference includes a version, which it doesn't need.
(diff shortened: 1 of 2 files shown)


jafeltra (approved):
This makes sense to me, and the test demonstrates the new intended behavior.

PR #1317, opened 2023-08-02
System version doesn't need to match when assigning a code

### LINKED ISSUE #1305: SUSHI should not log a warning when specifying the version of a CodeSystem
In SUSHI 3.0.0 and later, a warning is logged when trying to look up a CodeSystem with a version specified when assigning codes. When SUSHI fishes for the CodeSystem resource, it treats the version specified as if it is the version of the resource, and not the version of the system itself. The version is assigned in the output correctly. However, SUSHI does not need to fish for the CodeSystem with the version, which would avoid the incorrect warnings being logged when it cannot find a resource with the matching version.

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Strange.20behaviour.20concerning.20CodeSystem.20versions.20and.20aliases/near/375208251

FSH from the discussion: https://fshschool.org/FSHOnline/#/share/3NT8LKd (note: warning is only logged in SUSHI 3.0.0 and later).

mint-thompson:
Fixes #1305 and completes task [CIMPL-1147](https://standardhealthrecord.atlassian.net/browse/CIMPL-1147).

A code assigned with a system may include a version as part of the system. This is generally intended to refer to the system's published version in the world, and is not guaranteed to match the version of the CodeSystem FHIR resource. Fish for any version when resolving the system, and do not emit a warning even if the versions don't match.


src/fhirtypes/common.ts
```diff
@@ -952,9 +952,12 @@ export function replaceReferences<T extends AssignmentRule | CaretValueRule>(
       }
     }
   } else if (value instanceof FshCode) {
+    // the version on a CodeSystem resource is not the same as the system's actual version out in the world.
+    // so, they don't need to match.
+    const baseSystem = value.system?.split('|')[0];
     const codeSystemMeta = fishForMetadataBestVersion(
```
cmoesel, line 958: Since we're removing the version, I think this can just be `fishForMetadata` now, right @jafeltra?

jafeltra: Yep! For simplicity, `fishForMetadata` should be sufficient. Looking at the implementation of `fishForMetadata`, if there is no `|` in the the value passed in, it actually doesn't make any difference and will only try to call `fishForMetadata`. But based on how we've used these functions, we've only been using `fishForBestMetadata` in the cases where we want the version match or the next best thing.

jafeltra:
I didn't notice anything else, besides the small change @cmoesel suggested.

cmoesel (approved):
Looks good to me!

jafeltra (approved):
Looks good to me!

PR #1319, opened 2023-08-03
Add more error puns

vadi2:
Very important community contribution.


cmoesel:
Very important, indeed! I like these, @vadi2!

It looks like there's still an issue with code style. I think that when a string contains literal `'`, prettier prefers using `"` to surround the string so you don't have to escape the `'`. You should be able to fix it by running `npm run prettier:fix`.

vadi2:
I have succeeded.

PR #1323, opened 2023-08-15
Antlr4 performance workaround

mint-thompson:
Completes task [CIMPL-1153](https://standardhealthrecord.atlassian.net/browse/CIMPL-1153).

Modifying the babel configuration and rebuilding antlr4 restores performance to an acceptable speed. This workaround can be removed when the configuration change is available in the official antlr4 package.

To test, change to the PR's branch, then run `npm pack` to create a `.tgz` file for the package. Then, run `npm install -g <package-name>.tgz` to globally install the package. This will add the workaround to your globally installed version of SUSHI. Then, go to any SUSHI project and build it. You should notice that the FSH text import is restored to an acceptable speed and is much faster than in v3.3.0. A [test repo](https://github.com/amosref/GitTest) is available that clearly demonstrates the performance difference.

### DIFF
--- a/dev/fixAntlr4.js
+++ b/dev/fixAntlr4.js
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
 /**
  * The purpose of this script is to modify the configuration.
  * of the antlr4 dependency. This is necessary as a workaround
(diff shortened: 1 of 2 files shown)


jafeltra (approved):
I confirmed that this has much better performance than 3.2.0 and 3.3.0, and it is comparable to 3.1.0. Thanks for making this fix! Hopefully antlr gets an update out soon, but this seems like a safe workaround for us in the meantime.

PR #1325, opened 2023-08-21
Better antlr4 workaround

mint-thompson:
The first workaround #1323 for fixing up `antlr4` did not behave correctly under a variety of circumstances, such as when `antlr4` was not contained within SUSHI's `node_modules` directory, or on linux/mac systems. This workaround is more consistent, and should work regardless of OS or other installed packages.


jafeltra (approved):
The previous workaround did work on my Mac but didn't work well when things in `node_modules` weren't where they were expected. This workaround still works on my Mac and seems to be working better in FSH Online. Yay!

PR #1327, opened 2023-08-21
Add build command to antlr4 workaround script

mint-thompson:
The updated script in #1325 should have included the `npm run build` step that was removed from `package.json`.


jafeltra (approved):
Looks good and the workaround works again!

PR #1329, opened 2023-08-24
Log a warning when conformance and terminology instances default to Usage: #example

### LINKED ISSUE #1303: Warn when instances look like examples but omit "Usage: #example"?
It seems somewhat common for developers to create example resources without adding

    Usage: #example 

There may be some heuristics that could help here, if the title/description says "example" but the usage does not.

jafeltra:
Fixes #1303

This PR adds a warning when instances of conformance or terminology resources will use the default `#example` usage. Because we default the usage within the importer, I added the warning within the importer so we can distinguish between a usage that was set in FSH and a usage set by the default value.

Any feedback on the wording of the warning is appreciated.

### DIFF
--- a/src/fhirtypes/common.ts
+++ b/src/fhirtypes/common.ts
@@ -40,6 +40,28 @@ import { buildSliceTree, calculateSliceTreeCounts } from './sliceTree';
 import { InstanceExporter } from '../export';
 import { MismatchedTypeError } from '../errors';
 
+// List of Conformance and Terminology resources from http://hl7.org/fhir/R4/resourcelist.html
+// and http://hl7.org/fhir/5.0.0-snapshot1/resourcelist.html
+export const CONFORMANCE_AND_TERMINOLOGY_RESOURCES = new Set([
+  'CapabilityStatement',
+  'CapabilityStatement2', // pre-release R5
+  'StructureDefinition',
+  'ImplementationGuide',
+  'SearchParameter',
+  'MessageDefinition',
+  'OperationDefinition',
+  'CompartmentDefinition',
+  'StructureMap',
+  'GraphDefinition',
+  'ExampleScenario',
+  'CodeSystem',
+  'ValueSet',
+  'ConceptMap',
+  'ConceptMap2', // pre-release R5
+  'NamingSystem',
+  'TerminologyCapabilities'
+]);
+
 // characteristics are set using the structuredefinition-type-characteristics extension
 export const TYPE_CHARACTERISTICS_EXTENSION =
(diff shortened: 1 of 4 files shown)


mint-thompson (commented):
The code changes look good. I have some thoughts about the warning message. The first sentence feels like it goes on a little long. I think it would be okay to break it into two sentences. Maybe something like one of these:

- `No usage was specified on ${instance.name}, so the default #example usage will be applied. ${instance.instanceOf} Instances are typically definitions. Specify a usage to remove this warning.`
- `No usage was specified on ${instance.name}. The default #example usage will be applied, but ${instance.instanceOf} Instances are typically definitions. Specify a usage to remove this warning.`

jafeltra:
Thanks for the suggestions @mint-thompson! I'm leaning toward the second one, so I can add that in!

jafeltra:
Updated in [53be172](https://github.com/FHIR/sushi/pull/1329/commits/53be1729ad2e61407bd484b5b53be40e68ed0549)

src/fhirtypes/common.ts
```diff
@@ -40,6 +40,28 @@ import { buildSliceTree, calculateSliceTreeCounts } from './sliceTree';
 import { InstanceExporter } from '../export';
 import { MismatchedTypeError } from '../errors';
 
+// List of Conformance and Terminology resources from http://hl7.org/fhir/R4/resourcelist.html
+// and http://hl7.org/fhir/5.0.0-snapshot1/resourcelist.html
```
cmoesel, line 44: We should change this to reference the official R5 resource list at https://hl7.org/fhir/R5/resourcelist.html

cmoesel (changes requested):
I like the message. There is a limitation in that this won't work if the `InstanceOf` points to a _profile_ of one of the conformance or terminology resources -- but that should be pretty rare, and this is really just intended as a helpful hint -- so I'm not too concerned.

I did note one thing I think we should update in the code you moved. Other than that I think it looks good!

src/fhirtypes/common.ts
jafeltra: Good catch! Updated in [f4b8e27](https://github.com/FHIR/sushi/pull/1329/commits/f4b8e27f3ab1dcee1fbb315661be1ab052b7eb00).

jafeltra:
> There is a limitation in that this won't work if the `InstanceOf` points to a _profile_ of one of the conformance or terminology resources -- but that should be pretty rare, and this is really just intended as a helpful hint -- so I'm not too concerned.

Hmm I hadn't thought about that case. If we think it is worth covering this case, I'm happy to change up the implementation.

cmoesel (approved):
This looks good to me! Thanks!

PR #1331, opened 2023-08-24
Debug log stacks from unexpected errors

### LINKED ISSUE #1330: SUSHI errors when an instance adds an extension on meta.profile without explicitly setting the profile
Given FSH like the following:
```
Instance: PatientExample
InstanceOf: MyPatient
* meta.profile.extension[MyExtension].valueString = "Howdy"
* name
  * given[0] = "James"
  * family = "Pond"
```
SUSHI will emit the following error: `Cannot read properties of undefined (reading 'startsWith')`

This is because the InstanceExporter is trying to check if the `meta.profile` value starts with a particular string -- but in this case, although `meta.profile` is not null, it does not yet have an assigned value. In fact, the code where this is happening is the code that is intended to _set_ the `meta.profile` value based on the `InstanceOf`. Oops.

cmoesel:
This PR debug logs error stacks whenever we log an arbitrary error message (which usually corresponds to unexpected errors). This will allow us to more easily find the root of unexpected problems, especially things like trying to access properties on null/undefined.

This also fixes a log statement that passed in `e.sourceInfo` instead of `instance.sourceInfo` -- which resulted in receiving no source information when an unexpected error arose while exporting an instance.

You can test this using the example project I attached to #1330. This PR does _not_ fix #1330, but I implemented this PR while investigating #1330. Using that example project and SUSHI 3.3.2, you'll get an error with no source information and no stack trace (even when using debug logging). Using this PR's branch and the default logging level, you'll get the source information. Using this PR's branch and the `-l debug` flag, you'll additionally get the stack trace.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -38,6 +38,9 @@ const FSH_VERSION = '3.0.0-ballot';
 
 function logUnexpectedError(e: Error) {
   logger.error(`SUSHI encountered the following unexpected error: ${e.message}`);
+  if (e.stack) {
+    logger.debug(e.stack);
+  }
   process.exit(1);
 }
 
@@ -243,6 +246,9 @@ async function runBuild(input: string, program: OptionValues, helpText: string) 
     // If no errors have been logged yet, log this exception so the user knows why we're exiting
     if (stats.numError === 0) {
       logger.error(`An unexpected error occurred: ${e.message ?? e}`);
+      if (e.stack) {
+        logger.debug(e.stack);
+      }
     }
     process.exit(1);
   }
--- a/src/app.ts
+++ b/src/app.ts
@@ -71,6 +71,7 @@ async function app() {
     )
     .option('-s, --snapshot', 'generate snapshot in Structure Definition output', false)
     .action(async function (projectPath, options) {
+      setLogLevel(options);
       await runBuild(projectPath, options, program.helpInformation()).catch(logUnexpectedError);
     })
     .on('--help', () => {
@@ -100,7 +101,14 @@ async function app() {
   program
     .command('init')
     .description('initialize a SUSHI project')
-    .action(async function () {
+    .addOption(
+      new Option(
+        '-l, --log-level <level>',
+        'specify the level of log messages (default: "info")'
+      ).choices(['error', 'warn', 'info', 'debug'])
+    )
+    .action(async function (options) {
+      setLogLevel(options);
       await init().catch(logUnexpectedError);
       process.exit(0);
     });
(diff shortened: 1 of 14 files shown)


src/import/FSHImporter.ts
mint-thompson, line 1: In `parseGeneratedRuleSet`, we log the collected errors and warnings from the secret logger. But, the stack traces aren't kept as part of that collection. Do you think it would be useful to keep those stack traces around so they can be logged? This is somewhat complicated by the fact that we output aggregated messages for all the errors and warnings, which makes it less obvious which message goes with which stack trace. It might be okay to just output them all in order, so something like:
```
error: Errors parsing insert rule etc.
- error 1
- error 2
- error 3
debug: error 1 stack trace
debug: error 2 stack trace
debug: error 3 stack trace
```

src/app.ts
```diff
@@ -38,6 +38,9 @@ const FSH_VERSION = '3.0.0-ballot';
 
 function logUnexpectedError(e: Error) {
   logger.error(`SUSHI encountered the following unexpected error: ${e.message}`);
+  if (e.stack) {
+    logger.debug(e.stack);
+  }
```
mint-thompson, line 43: Only the `build` command can set the log level, which means that the stack trace is only available for that command. It might be useful to be able to get the stack trace for other commands by adding the log level option for those commands.

mint-thompson (commented):
Hooray for stack traces! This should make troubleshooting a lot faster. I have some questions about getting even more stack traces.

src/import/FSHImporter.ts
cmoesel: I thought about this when I was going through the code looking for times we log arbitrary error messages. At the time, I deemed that adding traces here added more complexity than was justified by the potential gain -- and that having a long list of stack traces could be pretty annoying (since each trace is pretty long). I figured that if this ever happened to be the spot where we needed a stack trace, we could add it at that time.

But... that was a very subjective decision. So I am willing to be overridden if you feel differently.

src/app.ts
cmoesel: Ah... I hadn't thought of that. It probably makes sense to add it to the other commands too. I'm heavy on my other project today, but I can maybe look at it next week.

src/import/FSHImporter.ts
mint-thompson: I think your decision here is fine. If we get anything logged in `parseGeneratedRuleSet`, that's probably enough information to work with.

src/app.ts
cmoesel: 3737ce4 adds support for --log-level to the other commands (`init` and `update-dependencies`). To test it, I added `throw new Error('Boo!')` in the code those commands invoke and confirmed that without `-l debug` there is no stack trace, but with it, there is.

jafeltra (approved):
Looks good to me! I didn't notice anything else that would be needed, and I like setting the log level is available to the other commands now.

PR #1332, opened 2023-08-29
Better regression report

cmoesel:
A number of improvements to the regression report, including:
- Link to repositories
- Call out changes in error count
- Capture warning count and call out changes
- Capture time per run and call out significant slowdowns
- Sort to put diffs at top of list
- Include skipped repos at bottom of list
- Add styling

In addition:
- Update `repos-all-.txt`
- Use Node 18.17.1


jafeltra (approved):
I think this looks really good! The new links are really helpful, and the new metrics + sorting of rows will be very helpful.

PR #1339, opened 2023-09-07
Use title and name values in IG resource list

### LINKED ISSUE #1296: ImplementationGuide.definition.resource.name not populated correctly for SUSHI generated ActorDefinition, Requirements, TestPlan and TestScript resource types
The FHIR IG Publisher and HL7 IG Template have been updated to include a new "Testing" tab page on all canonical resource types when TestPlan and/or TestScript resources are linked or reference those canonical resources. When reviewing the rendered "Testing" tab page, the listings of TestPlan and TestScript resource links are displaying the TestPlan.name and TestScript.name values instead of the expected TestPlan.title and TestScript.title values. However, this is only happening when the TestPlan and TestScript resource instances are generated by SUSHI from FSH definitions.

I have reviewed the SUSHI source code and found where this occurs - the **IGExporter.ts** **_addResourcePackage(referenceKey, pkgResource, configResource)_** method that processes these SUSHI generated resources is casting the ActorDefinition, Requirements, TestPlan and TestScript resource types as an **InstanceDefinition** class where the title attribute is not expected to exist. Since the ActorDefinition, Requirements, TestPlan and TestScript resource types implement the CanonicalResource interface, they do have the title attribute defined which needs to be used to populate the corresponding **ImplementationGuide.definition.resource.name** value.

_FYI - The **ImplementationGuide.definition.resource.name** value is used for the link name(s) in the IG pages._

The current logic is:

```
  private addPackageResource(
    referenceKey: string,
    pkgResource: StructureDefinition | ValueSet | CodeSystem | InstanceDefinition,
    configResource?: ConfigurationResource
  ): void {
    const newResource: ImplementationGuideDefinitionResource = {
      reference: { reference: referenceKey }
    };
    if (pkgResource instanceof InstanceDefinition) {
      newResource.name =
        configResource?.name ?? pkgResource._instanceMeta.title ?? pkgResource._instanceMeta.name;
      ...
```

A simple update to this logic to test for the existence of the title attribute will correct this issue as follows:

```
  private addPackageResource(
    referenceKey: string,
    pkgResource: StructureDefinition | ValueSet | CodeSystem | InstanceDefinition,
    configResource?: ConfigurationResource
  ): void {
    const newResource: ImplementationGuideDefinitionResource = {
      reference: { reference: referenceKey }
    };
    if (pkgResource instanceof InstanceDefinition) {
      const title = typeof pkgResource.title === 'string' ? pkgResource.title : null;
      if (title) {
        newResource.name = title;
      } else {
        newResource.name =
          configResource?.name ?? pkgResource._instanceMeta.title ?? pkgResource._instanceMeta.name;
      }
```

mint-thompson:
Fixes #1296 and completes task [CIMPL-1138](https://standardhealthrecord.atlassian.net/browse/CIMPL-1138).

Some resource types have title and name elements that are strings or similar datatypes. If present, these can be used to set the name in the IG resource list.

For package resources, only the title element value is used for this purpose. A title set with a keyword will have higher precedence than a title set by assigning to an element.

For predefined resources, the name element can be used for this purpose as well. The title element has a higher precedence than the name element. Note that the name element will only be used for predefined resources, not package resources. This change applies to both example and non-example predefined resources.

As expected, this change affects many IGs using FSH. It can change the order of resources in the IG if they are being sorted by name, which also changes the order of generated link references. The link references themselves are unchanged. It will also change the titles shown on pages in the IG, since many artifacts will now have a name based on something other than their `id`.


src/ig/IGExporter.ts
```diff
@@ -836,10 +836,16 @@ export class IGExporter {
       reference: { reference: referenceKey }
     };
     if (pkgResource instanceof InstanceDefinition) {
+      const title = typeof pkgResource.title === 'string' ? pkgResource.title : null;
```
cmoesel, line 839: I am seeing this pattern in enough places that I'd suggest creating a simple function like:
```ts
function stringOrNull(value: string): string | null {
  return typeof value === 'string' ? value : null;
}
```
Then you don't have to define intermediate variables like `title`, `name`, or `description` and instead just use that function where you need to reference them. E.g., instead of `title ??` on line 843 below, you'd have `stringOrNull(pkgResource.title) ??`.

cmoesel (changes requested):
I haven't had a chance to fully test this, but I have one initial (but small) suggestion.

cmoesel (approved):
This looks great! I ran a regression over all repos updated in the last year and a lot of things changed, but I think the changes are definitely for the better!

jafeltra (approved):
The changes make sense to me!

PR #1342, opened 2023-09-13
Update Automatic Dependencies

cmoesel:
Update the autodependency mechanism to add the extensions package to R4 IGs. It will have lower precedence than R4 core, which is the behavior I observed in the IG Publisher. In addition, use the appropriate `.r4` and `.r5` terminology and extension packages based on the IG's FHIR version.


jafeltra (approved):
These changes make sense to me! I tested it out locally and the correct packages were loaded based on the FHIR version I was using.

mint-thompson (approved):

PR #1344, opened 2023-09-14
Improve performance with Code Systems and Value Sets

jafeltra:
This PR improves the performance when exporting CodeSystems and ValueSets by setting all implied properties from `CaretValueRules` one time during the export.

Instead of setting implied properties one rule at a time each time that `setPropertyOnDefinitionInstance` is called, collect all known slices for all rules once and set all implied properties once. This helps speed up performance because `setImpliedPropertiesOnInstance` is expensive and caused the processing of Code Systems and Value Sets with a lot of `CaretValueRules` to be very slow.

This approach is only implemented for those two exporters right now because to make the same change for StructureDefinition exports required more changes to the code that made things more complex, and there were no repos that I saw that stood out as one that would benefit significantly from that change.

The performance degradation was originally reported on Zulip [here](https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/.E2.9C.94.20Sushi.203.2E0.2E0.2E.20.D0.A2imeout.20300.20seconds/near/363994155) in the [HL7EE/ig-ee-base#master](https://github.com/HL7EE/ig-ee-base/tree/master/) repo. In addition to that repo, when I ran a regression on the old `repos-all` file, there were three other repos that had significant time improvements when using this branch: [HL7/fhir-us-ph-library#master](https://github.com/HL7/fhir-us-ph-library/tree/master/), [HL7/fhir-directory-attestation#master](https://github.com/HL7/fhir-directory-attestation/tree/master/), and [ahdis/chmed#master](https://github.com/ahdis/chmed/tree/master/). There should not be any actual changes to output on this branch.

### DIFF
--- a/src/export/CodeSystemExporter.ts
+++ b/src/export/CodeSystemExporter.ts
@@ -1,15 +1,17 @@
 import { FSHTank } from '../import/FSHTank';
-import { CodeSystem, CodeSystemConcept } from '../fhirtypes';
+import { CodeSystem, CodeSystemConcept, PathPart } from '../fhirtypes';
 import {
   setPropertyOnDefinitionInstance,
   applyInsertRules,
   cleanResource,
-  assignInstanceFromRawValue
+  determineKnownSlices,
+  setImpliedPropertiesOnInstance,
+  validateInstanceFromRawValue
 } from '../fhirtypes/common';
 import { FshCodeSystem } from '../fshtypes';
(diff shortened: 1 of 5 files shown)


mint-thompson (approved):
hooray for performance improvements!

cmoesel (commented):
Interesting. I can see why setting implied properties on every assignment could slow things down! I'm just holding off approval until I run a full regression using our new regression script. I'm interested to see how the times change as well (which the new script will show us).

PR #1345, opened 2023-09-15
convert 'logicals' also

mrcerf:


mint-thompson (approved):
Looks good to me. Thank you!

PR #1346, opened 2023-09-15
Update type signature for loadAutomaticDependencies

mint-thompson:
Completes task [CIMPL-1143](https://standardhealthrecord.atlassian.net/browse/CIMPL-1143).

The base FHIRDefinitions from fhir-package-loader can be used in this function. This makes it easier to reuse loadAutomaticDependencies when importing SUSHI as a dependency. This is relevant to GoFSH, but also to anyone else using importing SUSHI.


jafeltra (approved):
Ah, lovely!

cmoesel (approved):
Makes sense to me!

PR #1348, opened 2023-09-18
Handle context StructureDefinition without derivation

### LINKED ISSUE #1333: Context unnecessarily uses [url]#[elementid] form for resources in the base specification
An extension defined with:
```
Context: Resource
```
results in:
```json
"context": [
  {
    "expression": "http://hl7.org/fhir/StructureDefinition/Resource#Resource",
    "type": "element"
  }
]
```
[FSH Online](https://fshschool.org/FSHOnline/#/share/3suM7Ro)

The [FHIR spec](https://hl7.org/fhir/R5/defining-extensions.html#context) says:
> Element Ids are unique within the base specification, and within a structure definition. The full path for the element is [url]#[elementid]. If there is no #, the Element id is one defined in the base specification

Since `Resource` is in the base specification, SUSHI should output the expression as `Resource`, not `http://hl7.org/fhir/StructureDefinition/Resource#Resource`.

mint-thompson:
Fixes #1333 and completes task [CIMPL-1160](https://standardhealthrecord.atlassian.net/browse/CIMPL-1160).

Some core FHIR StructureDefinition resources do not have a derivation attribute. These resources should still use the shortened form when used as an Extension context.

Curiously, the Resource definition in R4 and R4B does not have a derivation, but the derivation is present in R5. There are other definitions in each FHIR version that have no derivation. So, this PR takes the approach that any core FHIR StructureDefinition that doesn't specifically have a derivation of `constraint` can use the short form.

### DIFF
--- a/src/export/StructureDefinitionExporter.ts
+++ b/src/export/StructureDefinitionExporter.ts
@@ -497,7 +497,7 @@ export class StructureDefinitionExporter implements Fishable {
                   contextExpression = resourceSD.url;
                   contextType = 'extension';
                 } else if (
-                  resourceSD.derivation === 'specialization' &&
+                  resourceSD.derivation !== 'constraint' &&
                   resourceSD.url.startsWith('http://hl7.org/fhir/StructureDefinition/')
                 ) {
                   contextExpression = targetElement.id;
--- a/test/export/StructureDefinition.ExtensionExporter.test.ts
+++ b/test/export/StructureDefinition.ExtensionExporter.test.ts
@@ -367,6 +367,78 @@ describe('ExtensionExporter', () => {
       ]);
     });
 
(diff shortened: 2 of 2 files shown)


jafeltra (approved):
Makes sense and lines up with what we talked about earlier. I tested it out with the example in the issue and it works as expected.

cmoesel (dismissed):
Good catch. I did a little more research into this, and it turns out that you do not need to specify `derivation` if you do not have a `baseDefinition` (see `sdf-27` in [StructureDefinition.derivation](https://hl7.org/fhir/R5/structuredefinition-definitions.html#StructureDefinition.derivation).

cmoesel (approved):
I _still_ like it!

PR #1350, opened 2023-09-19
Support Allowed R5 resources in R4 IGs

cmoesel:
Allow authors to create instances of R5 `ActorDefinition`, `Requirements`, `SubscriptionTopic`, and `TestPlan` in R4 IGs (and R5 `ActorDefinition`, `Requirements`, and `TestPlan` in R4B IGs). Also allow these R5 resources to be referred to via references and canonicals (e.g., from an [obligation extension](https://build.fhir.org/ig/FHIR/fhir-tools-ig/branches/master/StructureDefinition-obligation.html)). This was discussed in [this Zulip thread](https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/using.20R5.20resources.20in.20FSH/near/377870473) and the specific allowed R5 resources are listed in [this Java class](https://github.com/HL7/fhir-ig-publisher/blob/master/org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/SpecialTypeHandler.java).

This capability is primarily supported by adding these special resources to FHIRDefinitions when it is initialized -- so they are always fishable. We also add a `_timeTraveler` flag to these resources so we can appropriately reject them where they're not allowed (namely as parents of profiles/logicals). Versions of FHIR that have these resources already simply overwrite the ones we added (and thus, do not have the `_timeTraveler` flag).

This PR also introduces a new `FHIRVersionsUtility` that can be used for determining information about FHIR versions based on their version number. Some existing code dealing with FHIR versions was refactored to leverage this new approach.

You can manually test this by creating an R4 or R4B IG and creating instances of one of the special R5 classes. You can also try using the [obligation extension](https://build.fhir.org/ig/FHIR/fhir-tools-ig/branches/master/StructureDefinition-obligation.html) to reference one of these R5 classes (likely `ActorDefinition`).

_NOTE: My initial implementation supported this capability by automatically loading all of R5 as a "supplemental FHIR package" -- and then extending the `MasterFisher` to support fishing by supplemental versions. This approach, however, was more complicated and more costly (in terms of memory). I believe my final implementation is better, but I've left the 1st implementation as a [separate commit](https://github.com/FHIR/sushi/commit/fe89fdea9d15fad12362ba6662ba2209cf92fa4a) if you'd like to review it._


src/utils/Processing.ts
```diff
@@ -56,22 +54,22 @@ export const AUTOMATIC_DEPENDENCIES: AutomaticDependency[] = [
...
     packageId: 'hl7.fhir.uv.extensions.r4',
     version: 'latest',
-    fhirVersion: R4_OR_4B_REGEX
+    fhirVersions: ['R4', 'R4B']
   },
   {
     packageId: 'hl7.fhir.uv.extensions.r5',
     version: 'latest',
-    fhirVersion: R5_OR_CURRENT_REGEX
+    fhirVersions: ['R5']
```
mint-thompson, line 72: While this wasn't the primary purpose of this PR, I really like that the automatic dependencies are now associated with specific version constants rather than just a regex.

src/utils/Processing.ts
```diff
@@ -317,45 +315,33 @@ export async function loadExternalDependencies(
...
-    fhirPackageId = 'hl7.fhir.r5.core';
-  } else {
-    fhirPackageId = 'hl7.fhir.r5.core';
-    prerelease = true;
-  }
-  if (prerelease) {
+  const fhirVersionInfo = config.fhirVersion
+    .map(v => getFHIRVersionInfo(v))
+    .find(v => v.isSupported);
+  if (fhirVersionInfo.isPreRelease) {
```
mint-thompson, line 321: Ahhhhh yeah. That's the good stuff right there. Thank you for `getFhirVersionInfo`.

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -7434,6 +7434,57 @@ describe('StructureDefinitionExporter R4', () => {
...
+      const name = sd.findElement('Patient.name');
+      expect(name.extension).toEqual([
+        {
+          url: 'http://hl7.org/fhir/tools/StructureDefinition/obligation',
+          extension: [
+            { url: 'actor', valueCanonical: 'http://hl7.org/fhir/us/minimal/ActorDefinition/AD1' },
+            { url: 'code', valueCode: 'SHALL' }
+          ]
+        }
+      ]);
```
jafeltra, line 7485: Should this have a real quick check for empty error and warning logs?

mint-thompson (approved):
This hits a lot of files, but I think I understand it. Basically, it's okay to have an Instance of a time-traveler, but not a Profile. Just for fun, I also included a time-travel Instance as a `contained` resource in a Profile, and that worked fine. That's what I expected, since it's consistent with allowing Instances to exist on their own. Thank you!

jafeltra (commented):
This looks really excellent. I looked through all the code changes and tests, and tried out some simple FSH. I really only had one minor suggestion and a question for myself.

src/utils/MasterFisher.ts
```diff
@@ -18,7 +18,11 @@ import { Instance } from '../fshtypes';
  * only in the tank).
  */
 export class MasterFisher implements Fishable {
-  constructor(public tank?: FSHTank, public fhir?: FHIRDefinitions, public pkg?: Package) {}
+  public defaultFHIRVersion?: string;
+  constructor(public tank?: FSHTank, public fhir?: FHIRDefinitions, public pkg?: Package) {
+    this.defaultFHIRVersion =
+      fhir?.fishForFHIR('StructureDefinition')?.fhirVersion ?? tank?.config.fhirVersion?.[0];
```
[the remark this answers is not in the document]
cmoesel: Technically speaking, the tank configuration allows multiple FHIR versions to be specified (as does the `ImplementationGuide` resource). I don't think we really support that use case well, but if someone were to put multiple FHIR versions in the config, I want to use whichever one is given priority in the FHIRDefinitions (i.e., the version that will be most often fished up).

test/export/StructureDefinitionExporter.test.ts
cmoesel: Sure!

cmoesel: Done in 491e587.

src/utils/MasterFisher.ts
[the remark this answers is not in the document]
jafeltra: Ahh that makes sense! Thanks for the explanation.

jafeltra (approved):
💯

PR #1351, opened 2023-09-20
Allow names that resemble a datetime

mint-thompson:
Completes task [CIMPL-1173](https://standardhealthrecord.atlassian.net/browse/CIMPL-1173).

The definition of the DATETIME lexer token is broad enough to encompass values that are reasonable to use as entity names. Add this token to the grammar rule for name. This also allows these names to be used in rules that refer to other entities, such as obeys rules.


mint-thompson:
Closing in favor of #1352

PR #1352, opened 2023-09-20
Improve parsing of date-like names

cmoesel:
This PR allows FSH entities to have names that resemble dates or datetimes, such as `2000`, `2000-10`, `2000-10-30`, etc. Prior to this PR, SUSHI would fail to parse these names. 

To support this, the following changes were made to the parser grammar g4 file:
- Update "name" to allow DATETIME and TIME lexer tokens
- Update "name" to allow more keyword lexer tokens
- Update "alias" to allow "name" rather than "SEQUENCE"

Note that `name` and `path` now have identical definitions in the grammar. I decided to leave them both in place rather than consolidating to one since they are semantically distinct and could diverge in the future. But... let me know if you disagree.

I've run a full regression and confirmed that this does not result in any changes to existing repos.

This issue was reported on Zulip here: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Certain.20pattern.20not.20allowed.20in.20names.2Finvariants.3F/near/391919370


mint-thompson (approved):

jafeltra (approved):
DateTimes for everyone!

PR #1353, opened 2023-09-20
Improve handling of child elements on meta.profile

### LINKED ISSUE #1330: SUSHI errors when an instance adds an extension on meta.profile without explicitly setting the profile

mint-thompson:
Fixes #1330 and completes task [CIMPL-1159](https://standardhealthrecord.atlassian.net/browse/CIMPL-1159).

An Instance may have extensions or other child elements defined on meta.profile. When these elements exist, they should not interfere with checking for the InstanceOf within the profile array. Additionally, if values are set on meta.profile and on child elements, keep those values at aligned indices when adding InstanceOf to meta.profile.

It sounds sort of complicated when written out, but if you take a look at the test cases I added, I think it makes sense. Please let me know if you think that the behavior I've implemented will cause problems. I ran a 1-year 100-count regression and nothing changed, and I'll go for a bigger set tonight.


mint-thompson:
Regression update: ran the full one-year regression, no changes.

test/export/InstanceExporter.test.ts
```diff
@@ -766,6 +766,91 @@ describe('InstanceExporter', () => {
...
+          extension: [
+            {
+              url: 'http://hl7.org/fhir/us/minimal/StructureDefinition/SimpleExt',
+              valueString: 'simple string'
+            }
+          ]
+        }
+      ]);
+    });
+
```
cmoesel, line 853: Can we add two more tests?
1. Set a different canonical (and extension) directly on `meta.profile[1]` (leaving `meta.profile[0]` empty for SUSHI to put in the default.
2. Set a different canonical at `meta.profile[0]` (w/ the extension) and the instanceOf canonical at `meta.profile[1]` (also w/ the extension). We want to be sure that the depuplication code is working as expected (spoiler alert: I'm pretty sure it's not, see [InstanceExporter.ts#839](https://github.com/FHIR/sushi/blob/e10d628e2534bd07df1b8894c847d1eaec0900be/src/export/InstanceExporter.ts#L839)).

cmoesel (changes requested):
I've made one suggestion (which you can take or leave) and a request for some more tests, one of which I think will expose a flaw in keeping `meta.profile` and `meta._profile` in sync.

src/export/InstanceExporter.ts
```diff
@@ -836,7 +852,31 @@ export class InstanceExporter implements Fishable {
...
+            // arrays are concatenated, primitives prefer the existing value if both values are non-null
+            // this makes use of the fact that typeof null === 'object'
+            if (isArray(a) && isArray(b)) {
+              return a.concat(b);
+            } else if (typeof a !== 'object' || typeof b !== 'object') {
+              return a ?? b;
+            }
+          });
+        });
+        [instanceDef.meta.profile, instanceDef.meta._profile] = unzip(uniqueProfiles);
```
cmoesel, line 875: Wow. What a block of code. Things like this remind me that I should try to keep myself more familiar w/ lodash.

cmoesel (approved):
This is definitely an improvement on the old behavior. Thank you!

Do you think we should run another regression, just in case?

PR #1355, opened 2023-09-21
Don't return instance rules when looking for id

mint-thompson:
Completes task [CIMPL-1169](https://standardhealthrecord.atlassian.net/browse/CIMPL-1169).

When searching for a rule that sets an id, check the rule's isInstance flag. Only return the rule if the flag is not set.

To explain how this was going infinite: an Instance had a rule that looked like this:
```
* id = procedure-death-certification-ut-example1
```
Because the value is unquoted, the rule's `isInstance` flag is set. So, when applying this rule during export, the exporter attempts to fish up an Instance with a matching name or id. When checking the same instance, `findIdAssignmentRule` would find this rule, meaning that it now wants to assign this Instance. But, the Instance must be exported before it can be assigned. So we start exporting the Instance, attempt to apply this rule, and thus we are stuck until the call stack fills up and the computer drops it all on the ground.

This FSH still results in an error, but a much more helpful one: the fisher reports that it can't find an Instance called `procedure-death-certification-ut-example1` to assign for this rule.

### DIFF
--- a/src/fhirtypes/common.ts
+++ b/src/fhirtypes/common.ts
@@ -880,7 +880,7 @@ export function replaceReferences<T extends AssignmentRule | CaretValueRule>(
   fisher: Fishable
 ): T {
   let clone: T;
-  const value = getRuleValue(rule);
+  const value = rule.value;
   if (value instanceof FshReference) {
     let type: string, id: string, instanceMeta: Metadata;
     // Prefer resolving to instances, so look them up first
@@ -945,7 +945,7 @@ export function replaceReferences<T extends AssignmentRule | CaretValueRule>(
     }
     if (type != null && id != null) {
       clone = cloneDeep(rule);
-      const assignedReference = getRuleValue(clone) as FshReference;
+      const assignedReference = clone.value as FshReference;
       // Set the sdType which will be used for type-checking during assignment
       assignedReference.sdType = type;
       // Only replace references that are specified using name or id. Absolute URLs or
@@ -982,7 +982,7 @@ export function replaceReferences<T extends AssignmentRule | CaretValueRule>(
     const codeSystemMeta = fisher.fishForMetadata(baseSystem, Type.CodeSystem);
     if (codeSystemMeta) {
       clone = cloneDeep(rule);
-      const assignedCode = getRuleValue(clone) as FshCode;
+      const assignedCode = clone.value as FshCode;
       assignedCode.system = value.system.replace(/^[^|]+/, codeSystemMeta.url);
 
       // Find the code system using the returned metadata to avoid duplicate warnings if version mismatches
@@ -1002,19 +1002,6 @@ export function replaceReferences<T extends AssignmentRule | CaretValueRule>(
     }
   }
   return clone ?? rule;
-}
-
-/**
- * Function to get a value from a rule that has a value (AssignedValue or CaretValue)
- * @param rule - The rule to get a value from
- * @returns - The value on the rule
- */
-function getRuleValue(rule: AssignmentRule | CaretValueRule): AssignmentValueType {
-  if (rule instanceof AssignmentRule) {
-    return rule.value;
-  } else if (rule instanceof CaretValueRule) {
-    return rule.value;
-  }
 }
 
 export function listUndefinedLocalCodes(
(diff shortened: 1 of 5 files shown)


test/fshtypes/Instance.test.ts
```diff
@@ -15,6 +15,26 @@ describe('Instance', () => {
...
+
+    it('should not return an id set by an instance assignment rule', () => {
+      const p = new Instance('MyInstance');
+      const idRule = new AssignmentRule('id');
+      idRule.value = 'different-id';
+      idRule.isInstance = true;
+      p.rules.push(idRule);
+      // the id is still the default
+      expect(p.id).toBe('MyInstance');
+    });
```
cmoesel, line 35: Cool. Can we assert no errors in the first test and the expected error in the second test?

test/fshtypes/Profile.test.ts
```diff
@@ -13,6 +19,27 @@ describe('Profile', () => {
...
+
+    it('should not return an id set by an instance caret rule', () => {
+      const p = new Profile('MyObservation');
+      const idRule = new CaretValueRule('');
+      idRule.caretPath = 'id';
+      idRule.value = 'different-id';
+      idRule.isInstance = true;
+      p.rules.push(idRule);
+      expect(p.id).toBe('MyObservation');
+    });
```
cmoesel, line 40: Here too. Can we assert no errors in the first test and the expected error in the second test?

test/fshtypes/Instance.test.ts
mint-thompson: There aren't any expected errors in either test, since nothing is being exported. While it's unusual to assign an Instance here, it's allowed. Although, I just tested out some FSH that does this, and the output is not quite right:
```
Instance: MyObs
InstanceOf: Observation
Usage: #example
* status = #final
* code = #123
* id = SomeId

Instance: SomeId
InstanceOf: http://hl7.org/fhir/StructureDefinition/id
Usage: #inline
* value = "my-obs-id"
```

The Instance gets assigned using its object structure, the resulting file for `MyObs` is named `Observation-[object Object].json`, and SUSHI emits an error: `The string "[object Object]" does not represent a valid FHIR id.` We might need to investigate assignment of inline primitive instances.

mint-thompson:
@cmoesel I'm glad you asked about `url`, since that got me more curious about other related scenarios. As it turns out, _any_ situation where an Instance refers to itself leads to a stack overflow. For example, both of these Instances cause stack overflows:
```
Instance: MyQuestionnaire
InstanceOf: Questionnaire
* url = MyQuestionnaire
* status = #draft

Instance: MyObs
InstanceOf: Observation
Usage: #example
* status = #final
* code = #123
* valueString = MyObs
```

So, it looks like we may need to do a better job handling this in the more general case of AssignmentRules on Instances.

But to answer your original question: when checking for a `url`, we use `getValueFromFshRules`, which doesn't check if the rule has `isInstance` set, but also just returns the value as a string when a rule with the correct path is found. This implementation is safe as long as `isInstance` is not set, but either way, it does not cause a stack overflow. Also, it's not until processing the actual CaretValueRule that we get into trouble. For example, using this FSH:
```
Profile: ObsProfile
Parent: Observation
* ^url = http://hl7.org/some/example/StructureDefinition/ObsProfile
```
The call to `getUrlFromFshDefinition` at `StructureDefinitionExporter.ts#L309` returns the string `http://hl7.org/some/example/StructureDefinition/ObsProfile` without any trouble. But when processing the CaretValueRule, we get down to `StructureDefinitionExporter.ts#L1000`, with this line:
```ts
sd.setInstancePropertyByPath(rule.caretPath, fishedValue, this);
```
We get an error, since `fishedValue` is an InstanceDefinition that represents ObsProfile, which is not the right type here:
```
Cannot assign StructureDefinition value: ObsProfile. Value does not match element type: uri
```
This error is much better than filling up the stack, of course.

test/fshtypes/Instance.test.ts
cmoesel: Yeah, primitive instances in general are a pretty weird case. I'm not very surprised that we don't handle them well in some (or all?) cases.

src/fhirtypes/common.ts
```diff
@@ -1004,19 +1004,6 @@ export function replaceReferences<T extends AssignmentRule | CaretValueRule>(
...
- * @returns - The value on the rule
- */
-function getRuleValue(rule: AssignmentRule | CaretValueRule): AssignmentValueType {
-  if (rule instanceof AssignmentRule) {
-    return rule.value;
-  } else if (rule instanceof CaretValueRule) {
-    return rule.value;
-  }
-}
-
```
cmoesel, line 1019: Well that's not a very useful function, is it? Good riddance!

cmoesel (approved):
This looks good. Thanks for improving the code that looks for the url as well.

jafeltra (approved):
Makes sense to me!

PR #1362, opened 2023-09-28
Remove antlr4 workaround

mint-thompson:
Completes task [CIMPL-1175](https://standardhealthrecord.atlassian.net/browse/CIMPL-1175).

The 4.13.1 release of antlr4 changes the .babelrc file to have the desired configuration. It is no longer necessary to manually change it during installation.


cmoesel (approved):
Oh good. I like it when we can remove workarounds!

I tested this by npm packing it and globally installing the tgz. It seems to still work just fine!

cmoesel (changes requested):
@jafeltra is right. When I run `npm install`, my `npm-shrinkwrap.json` is modified to remove the `"hasInstallScript": true,` line. That should be reflected in this PR. I am using Node 18 w/ NPM 9.

jafeltra (approved):
I don't get any changes from `npm install` now! Yay!

cmoesel (approved):
Two thumbs up.

PR #1363, opened 2023-09-28
Allow binding to types with the #can-bind characteristic

### LINKED ISSUE #1347: SUSHI should allow binding to Logical Models that have the #can-bind characteristic
Logical Models support specifing type characteristics through the `Characteristics` keyword. One of the values supported by this extension is `can-bind`, which states that the type can be bound to a terminology. SUSHI should allow Logical Models that have this characteristic specified to be used in bindings. If the characteristic is not specified, but the type is used in a binding, a warning should be logged (this should be similar to the behavior in SUSHI when a Logical Model is used as a reference without the `can-be-target` characteristic).

jafeltra:
Fixes #1347.

This PR makes it so that if a Logical Model includes `#can-bind` included in the `Characteristics` or if it has the `type-characteristics` extension with the `can-bind` value, then that type can have bindings applied to it.

If an element tries to add a binding and is not one of the bindable types and one of the allowed types is a Logical Model, SUSHI will now log a warning. If an element tries to add a binding and it is not a bindable type and none of the types are a Logical Model, SUSHI's original behavior is the same and an error is thrown.

I did update the message in that error to reflect that the type can have a binding if it is a logical model with #can-bind because it seemed to make the error message the most correct, but there technically is not a case when that error message will be logged and the fix will be to add `#can-bind` because that case would case the warning to be logged instead. If you disagree with this approach, let me know.

The approach I took to implement this was to add another metadata field `canBind`, similar to `canBeTarget`. After talking to @cmoesel, we might want to consider refactoring this to have one `characteristics` metadata field, which can be an object that has each characteristic's value. We decided that this refactor could be a separate task, but I wanted to note it here for context.

### DIFF
--- a/src/errors/CodedTypeNotFoundError.ts
+++ b/src/errors/CodedTypeNotFoundError.ts
@@ -8,7 +8,7 @@ export class CodedTypeNotFoundError extends Error implements Annotated {
     super(
       `Cannot bind value set to ${foundTypes.join(
         ','
-      )}; must be coded (code, Coding, CodeableConcept, Quantity, CodeableReference), the data types (string, uri), or a Logical Model with the #can-bind Characteristic.`
+      )}; must be coded (code, Coding, CodeableConcept, Quantity, CodeableReference), the data types (string, uri), or a logical model with the #can-bind characteristic.`
     );
   }
 }
--- a/src/errors/CodedTypeNotFoundError.ts
+++ b/src/errors/CodedTypeNotFoundError.ts
@@ -8,7 +8,7 @@ export class CodedTypeNotFoundError extends Error implements Annotated {
     super(
       `Cannot bind value set to ${foundTypes.join(
         ','
-      )}; must be coded (code, Coding, CodeableConcept, Quantity, CodeableReference), or the data types (string, uri).`
+      )}; must be coded (code, Coding, CodeableConcept, Quantity, CodeableReference), the data types (string, uri), or a Logical Model with the #can-bind Characteristic.`
     );
   }
 }
--- a/src/fhirtypes/ElementDefinition.ts
+++ b/src/fhirtypes/ElementDefinition.ts
@@ -1636,10 +1636,10 @@ export class ElementDefinition {
       'string',
       'uri'
     );
-    const isBindable = this.type?.some(
-      type => fisher?.fishForMetadata(type.code, Type.Logical)?.canBind
-    );
-    if (isEmpty(validTypes) && !isBindable) {
+    const isBindable =
+      validTypes.length ||
+      this.type?.some(type => fisher?.fishForMetadata(type.code, Type.Logical)?.canBind);
+    if (!isBindable) {
       if (this.type?.some(type => fisher?.fishForMetadata(type.code, Type.Logical) != null)) {
         // Only warn if it was from a logical model that doesn't have #can-bind characteristic
         logger.warn(
--- a/src/fhirtypes/ElementDefinition.ts
+++ b/src/fhirtypes/ElementDefinition.ts
@@ -1623,7 +1623,8 @@ export class ElementDefinition {
   bindToVS(
     vsURI: string,
     strength: ElementDefinitionBindingStrength,
-    ruleSourceInfo?: SourceInfo
+    ruleSourceInfo?: SourceInfo,
+    fisher?: Fishable
   ): void {
     // Check if this is a valid type to be bound against
     const validTypes = this.findTypesByCode(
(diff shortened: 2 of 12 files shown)


src/errors/CodedTypeNotFoundError.ts
```diff
@@ -8,7 +8,7 @@ export class CodedTypeNotFoundError extends Error implements Annotated {
     super(
       `Cannot bind value set to ${foundTypes.join(
         ','
-      )}; must be coded (code, Coding, CodeableConcept, Quantity, CodeableReference), or the data types (string, uri).`
+      )}; must be coded (code, Coding, CodeableConcept, Quantity, CodeableReference), the data types (string, uri), or a Logical Model with the #can-bind Characteristic.`
```
cmoesel, line 11: I think you're right to add this to the message (for completeness sake) -- but I suggest tweaking the casing just a bit:
```suggestion
      )}; must be coded (code, Coding, CodeableConcept, Quantity, CodeableReference), the data types (string, uri), or a logical model with the #can-bind characteristic.`
```

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1635,8 +1636,19 @@ export class ElementDefinition {
       'string',
       'uri'
     );
-    if (isEmpty(validTypes)) {
-      throw new CodedTypeNotFoundError(this.type ? this.type.map(t => t.code) : []);
+    const isBindable = this.type?.some(
+      type => fisher?.fishForMetadata(type.code, Type.Logical)?.canBind
+    );
+    if (isEmpty(validTypes) && !isBindable) {
```
cmoesel, line 1642: I'd suggest that `isBindable` should look at the `validTypes` before doing the potentially expensive fishing expedition. In 99.95% of the cases, the requirement for a bindable type will already be fulfilled before we even look at logical models. It also makes `isBindable` complete (i.e., it represents all possibility of the element being bindable).
```suggestion
    const isBindable = validTypes.length || this.type?.some(
      type => fisher?.fishForMetadata(type.code, Type.Logical)?.canBind
    );
    if (!isBindable) {
```

src/fhirtypes/ElementDefinition.ts
```diff
@@ -1635,8 +1636,19 @@ export class ElementDefinition {
...
+      if (this.type?.some(type => fisher?.fishForMetadata(type.code, Type.Logical) != null)) {
+        // Only warn if it was from a logical model that doesn't have #can-bind characteristic
+        logger.warn(
+          'Binding on a Logical Model type without the #can-bind Characteristic is discouraged',
```
cmoesel, line 1646: I think it's more than discouraged -- it's technically disallowed. Also, it looks like we usually do not capitalize _Logical Model_ in error messages and warnings. So maybe something more like:
```suggestion
          `Bindings can only be applied to logical model types with the #can-bind characteristic. Update the target logical model to declare the #can-bind characteristic or remove the binding from ${this.id}.`,
```

src/import/FSHTank.ts
```diff
@@ -419,82 +419,8 @@ export class FSHTank implements Fishable {
...
-                  .pop() as CaretValueRule;
-                if (valueRule?.value === true) {
-                  meta.canBeTarget = true;
-                  break;
-                }
-              }
-            }
-          }
+          meta.canBeTarget = hasLogicalCharacteristic(result, 'can-be-target');
+          meta.canBind = hasLogicalCharacteristic(result, 'can-bind');
```
cmoesel, line 423: I refactored all this code in the branch I'm currently working on. This will be a fun rebase for me! (But seriously, don't worry about it).

cmoesel (changes requested):
Thanks, @jafeltra. This is a nice feature for our users who are working with logical models. I've suggested a few small changes to consider.

As you noted, I think we will want to do some refactoring to simplify this -- but I think that will be a lot easier after we get my upcoming PR in that already refactors some of th code in these areas. So let's get this PR in, then mine, then we can look at refactoring!

jafeltra:
Thanks for the feedback @cmoesel. I've taken all your suggestions, but I didn't hit the "commit suggestion" buttons so I could update prettier and the tests for them as well.

cmoesel:
> I've taken all your suggestions, but I didn't hit the "commit suggestion" buttons so I could update prettier and the tests for them as well.

No problem. I actually prefer it that way. I just use the suggestions feature of GitHub because it makes it very easy to see exactly what I'm suggesting.

I'll re-review when I have a moment.

cmoesel (approved):
Alright, I like it!

mint-thompson (approved):
I like the strategy you used to implement this, where you generalized the idea of searching for a specific characteristic. Thank you!

PR #1365, opened 2023-10-03
Stop collecting ParamRuleSet contents when encountering a Logical or Resource

mint-thompson:
Completes task [CIMPL-1185](https://standardhealthrecord.atlassian.net/browse/CIMPL-1185).

Add the keywords for Logical and Resource definitions to the list of keywords that are not allowed in ParamRuleSet contents. [This example](https://fshschool.org/FSHOnline/#/share/3F6jDAg) shows the bug in action: the contents don't stop until a Profile appears.


jafeltra (approved):
Great work here 🕵️

PR #1373, opened 2023-10-12
Unfold choice type slices when no unfolded elements match

### LINKED ISSUE #1337: SUSHI fails to resolve IPS MedicationStatement.effectivePeriod path in an instance
A SUSHI project that depends on `hl7.fhir.uv.ips: 1.1.0` cannot successfully create an instance of the IPS MedicationStatement. Given the following:
```
Instance: InstanceOfIPSMedicationStatement
InstanceOf: MedicationStatement-uv-ips
Usage: #example
* status = #active
* effectivePeriod.start = "2023"
* subject = Reference(Patient/Bob)
* medicationReference = Reference(Medication/9ac3356c-4ea4-4814-84c3-235484f2ef13) "enbrel"
```
SUSHI reports:
> error The element or path you referenced does not exist: effectivePeriod.start
>   File: /Users/cmoesel/data/fsh/DependsOnIPS/input/fsh/patient.fsh
>   Line: 5
> error Element MedicationStatement.effective[x] has minimum cardinality 1 but occurs 0 time(s).
>   File: /Users/cmoesel/data/fsh/DependsOnIPS/input/fsh/patient.fsh
>   Line: 1 - 7

Zulip: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Unexpected.20error.20instantiating.20an.20IPS.20profile/near/389171755

jafeltra:
Fixes #1337 

This PR fixes a bug where if the choice type was referenced by slice name (i.e. `valueString`) but there were other non-slice elements that could be unfolded, SUSHI would fail to ever unfold and find the choice type. This PR adds another place where choice type elements are unfolded so that all elements can be considered when finding a provided path.

I ran a regression on all repos from the last year and there were not changes and no significant differences in processing time. That said, if anyone doesn't think this is the right approach to resolving this issue, let me know!

### DIFF
--- a/src/fhirtypes/StructureDefinition.ts
+++ b/src/fhirtypes/StructureDefinition.ts
@@ -216,6 +216,7 @@ export class StructureDefinition {
     const parsedPath = parseFSHPath(path);
 
     let fhirPathString = this.pathType;
+    let previousPathPart = '';
     let matchingElements = this.elements;
     let newMatchingElements: ElementDefinition[] = [];
     // Iterate over the path, filtering out elements that do not match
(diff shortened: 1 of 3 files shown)


mint-thompson (approved):
Looks good, thank you!

cmoesel (approved):
The code looks good. I'm glad it didn't end up being too tricky to implement!

I tried it on my example project that originally produced the error. No more error and the output is as expected. Great!

PR #1377, opened 2023-10-25
Create index files for FSH-defined entities

### LINKED ISSUE #1361: Generate file mapping FHIR exports to FSH sources
There's some interest from the community in providing a file that maps exported FHIR definitions back to the FSH file (and line range) where they are defined. In particular, it could be used to support FSH-related features in Simplifier.

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Tracing.20FHIR.20resources.20back.20to.20FSH.20file/near/392518075

mint-thompson:
Fixes #1361 and completes task [CIMPL-1188](https://standardhealthrecord.atlassian.net/browse/CIMPL-1188)

There is one entry in the index for each FSH definition that results in an exported JSON file. Inline instances do not appear in their own file, so they do not appear in the index. The index is created in two formats. There is a txt file that is formatted for human readability, and there is a JSON file that is useful for machine processing.


src/utils/Processing.ts
```diff
@@ -564,6 +567,42 @@ export function writeFHIRResources(
...
+        endLine: fshInfo.location.endLine,
+        outputFile: fileName
+      });
+    }
+  }
+  // write txt with nice formatting
+  index.unshift(['Name', 'Type', 'File', 'Lines', 'Output']);
+  fs.outputFileSync(path.join(outDir, 'fsh-generated', 'fsh-index.txt'), table(index));
+  // write json for machine usage
+  fs.outputJsonSync(path.join(outDir, 'fsh-generated', 'fsh-index.json'), otherIndex);
```
cmoesel, line 605: I know this is for machines, but I think we should consider formatting the JSON so that humans can more easily parse it too. It takes a little more space, but I think it's worth it.

cmoesel (changes requested):
I really like this capability -- and the text table looks great! I've made just a few minor suggestions for your consideration.

src/utils/Processing.ts
```diff
@@ -564,6 +567,42 @@ export function writeFHIRResources(
...
+      otherIndex.push({
+        fshFile: relativeInput,
+        fshName: fshInfo.fshName,
+        fshType: fshInfo.fshType,
+        startLine: fshInfo.location.startLine,
+        endLine: fshInfo.location.endLine,
+        outputFile: fileName
+      });
+    }
+  }
```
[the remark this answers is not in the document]
jafeltra: I definitely like the idea of sorting. I'm torn on the sort order because it sounded like when this issue first came up that someone was looking to find the FSH file based on the JSON file, which would make your third suggestion most useful. But looking at the table and the contents, I'd prefer either of the other two.

All that to say, I like the idea, and I'm torn so I don't have a strong preference either.

src/utils/Processing.ts
```diff
@@ -564,6 +567,42 @@ export function writeFHIRResources(
...
+}
+
+export function writeFSHIndex(
+  outDir: string,
+  outPackage: Package,
+  inputDir: string,
+  skippedResources: string[] = []
+) {
+  const index: string[][] = [];
+  const otherIndex: any[] = [];
```
jafeltra, line 580: Is it possible to name these a little more descriptively? Maybe `textIndex` and `jsonIndex`? I'm not sure but I had to figure out for a second what the difference was.

jafeltra, line 580: The `any[]` type is pretty well defined in this case, but is this `any` because there is no exported type that fits exactly here?

jafeltra (commented):
This seems super useful! I just had a few small comments/questions.

src/utils/Processing.ts
```diff
@@ -564,6 +576,44 @@ export function writeFHIRResources(
...
+        relativeInput,
+        `${fshInfo.location.startLine} - ${fshInfo.location.endLine}`
+      ]);
+      jsonIndex.push({
+        fshFile: relativeInput,
+        fshName: fshInfo.fshName,
+        fshType: fshInfo.fshType,
+        startLine: fshInfo.location.startLine,
+        endLine: fshInfo.location.endLine,
+        outputFile: fileName
```
cmoesel, line 609: I can't decide if it's silly to ask for this or not, but... what do you think about ordering these properties in the same order as for the text file? (E.g., outputFile, fshName, fshType, fshFile, startLine, endLine).

cmoesel (commented):
One more question/request that may be silly...

jafeltra (approved):
Huge fan of this.

PR #1378, opened 2023-10-31
Handle recursive extension contexts

### LINKED ISSUE #1375: Recursion Bug for Extensions w/ Context Set to Self
Complex extensions may be recursive (e.g., contain a nested extension referring to themself). In this case, the extension should add itself to its `Context:`. If an author attempts to do this, however, SUSHI fails due to infinite recursion.

Example (on [FSH Online](https://fshschool.org/FSHOnline/#/share/401ycyI)):
```
Extension: Node
Context: Device, Node
Title: "Node"
Description: "A node in a file system"
* extension contains name 1..1 and isFolder 1..1 and Node named children 0..*
* extension[name].value[x] only string
* extension[isFolder].value[x] only boolean
```

cmoesel:
This PR updates the SD exporter to process extension contexts last so that all the necessary components of the SD are in place and available in the Package. This allows proper lookup of sub-extensions and avoids infinitely recursive fishing.

I ran a full regression and one project came back with a change. It turns out they were only setting `^context.expression` and not `^context.type`. This worked because SUSHI was adding a default context and `^context.expression` was just overwriting the expression part and leaving the type part in place. With this PR change in place, the `^context.type` remains blank in the JSON (since they didn't set it). I think this is the proper behavior because a default context should not have been applied in the first place since they were setting their own context. I've filed a PR to fix this on that one project [here](https://github.com/HL7/davinci-dtr/pull/102).

Fixes #1375

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -92,15 +92,87 @@
       }
     },
     "node_modules/@babel/code-frame": {
-      "version": "7.22.5",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.22.5.tgz",
-      "integrity": "sha512-Xmwn266vad+6DAqEB2A6V/CcZVp62BbwVmcOJc2RPuwih1kw02TjQvWVWlcKGbBPd+8/0V5DEkOcizRGYsspYQ==",
+      "version": "7.22.13",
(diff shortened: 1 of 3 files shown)


mint-thompson (approved):
Looks perfect to me. Hooray for recursive extension contexts!

jafeltra (approved):
This looks good! And the bonus PR looks great too!

PR #1379, opened 2023-11-01
Cimpl 1177 implied codeableconcept

### LINKED ISSUE #1341: SUSHI behavior for fixed CodeableConcept may have unexpected results
SUSHI applies pattern values before processing rules (I think). As a result, in some cases (like CodeableConcept), it results in unexpected output if an instance rule overwrites a previously applied pattern value.

For example, given this profile and instance:
```
Profile: ConditionProfile
Parent: Condition
* code = http://foo.org|foov0.0.1#bar

Instance: MyCondition
InstanceOf: ConditionProfile
* code.coding[0] = http://notfoo.org#notbar
* code.coding[1] = http://foo.org|foov0.0.1#bar
* subject = Reference(Patient/John)
```
the instance ends up with the following code:
```json
{
  "coding": [
    {
      "code": "notbar",
      "system": "http://notfoo.org",
      "version": "foov0.0.1"
    },
    {
      "code": "bar",
      "system": "http://foo.org",
      "version": "foov0.0.1"
    }
  ]
}
```
Note that the `http://notfoo.org#notbar` code ends up with the `foov0.0.1` system from the pattern.

Ideally, SUSHI should be smarter about how it does this so that it recognizes that `coding[0]` is assigned a different value that does not match the pattern -- and as a result, puts the pattern at `coding[1]` instead (ideally recognizing that although `coding[1]` also is also assigned a value, that value matches the pattern).

FSHOnline: https://fshschool.org/FSHOnline/#/share/3rfNkMb

mint-thompson:
Fixes #1341 and completes task [CIMPL-1177](https://standardhealthrecord.atlassian.net/browse/CIMPL-1177).

When a profile element has a `patternCodeableConcept`, make sure that the `coding` it contains doesn't get overwritten in instances of that profile by assignment rules on that element's `coding` sub-element. At the same time, make sure that if there is an assignment on the `coding` that matches the implied value, don't create an extra element for it.

A full regression for this branch has changes on some repos. All of the changes are one of these:
- reordering elements of a `coding` array, which is equally correct
- preserving the coding from a `patternCodeableConcept`, which is more correct
- assigning a `coding` from an instance rule at the same index as the implied value when it matches the pattern, which is more correct

None of these repos require changes in order to handle this PR.

All of that said, while this implementation is correct, it feels a little sloppy. Specifically, adding in a new function for a different way of assigning a complex value is a little awkward, but I needed to do something very slightly different than what I'd get from `Object.assign`. Additionally, the current implementation only really handles `CodeableConcept`, but maybe it should handle any type that contains an array? And that could be basically anything, given that a profile could have a pattern value on `contained`, or the pattern could include extensions. So, maybe this needs to cook a little longer. Please let me know what you think.

### DIFF
--- a/src/export/InstanceExporter.ts
+++ b/src/export/InstanceExporter.ts
@@ -22,7 +22,8 @@ import {
   isModifierExtension,
   createUsefulSlices,
   determineKnownSlices,
-  setImpliedPropertiesOnInstance
+  setImpliedPropertiesOnInstance,
+  replaceField
 } from '../fhirtypes/common';
 import { InstanceOfNotDefinedError } from '../errors/InstanceOfNotDefinedError';
 import { AbstractInstanceOfError } from '../errors/AbstractInstanceOfError';
@@ -36,7 +37,6 @@ import {
   isEmpty,
   isEqual,
   isMatch,
-  merge,
   mergeWith,
   padEnd,
   uniq,
@@ -344,8 +344,7 @@ export class InstanceExporter implements Fishable {
         }
       });
     });
-    instanceDef = merge(instanceDef, ruleInstance);
-    return instanceDef;
+    return ruleInstance; // do we even need to clone anymore
   }
 
   /**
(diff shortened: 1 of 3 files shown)


mint-thompson:
I discovered some additional cases that were not handled by the current implementation. Trying to handle those breaks the cases that are currently handled. At this point, I feel like this has become too tangled and bizarre to be useful. A new strategy for Instances is needed, and that is beyond the scope of this PR.

PR #1381, opened 2023-11-07
Value Set concept components have one code per line

mint-thompson:
Completes task [CIMPL-1201](https://standardhealthrecord.atlassian.net/browse/CIMPL-1201).

A ValueSetConceptComponentRule may have more than one code. When producing FSH for the rule, each code needs to be written as a separate line of FSH. This is true regardless of whether there are any ValueSets associated with the rule.


cmoesel (commented):
This is kind of weird. The [FSH spec](https://hl7.org/fhir/uv/shorthand/2023Sep/reference.html#defining-value-sets) doesn't include this syntax because it doesn't make a lot of sense. Given that the result is the intersection of the code and the value set(s), then you have these possibilities for `abc#123 from valueset X and valueset Y`:
* `abc#123` is in both of those value sets - so the result is `abc#123`
* `abc#123` is not in one (or both) of those value sets - so the result is empty.

But, in theory, you should know the content of those value sets. So you know it will or won't be included. So either include the code or don't. The VS just confuses things.

All that said, I guess FHIR does allow that to happen (even if it's senseless) -- so I guess maybe we need to support it in case GoFSH runs into it, but maybe we leave it as an undocumented syntax?

mint-thompson:
@cmoesel I completely agree that it is weird. This is a case where FSH allows it because FHIR allows it, and SUSHI allows it because FSH allows it. SUSHI requires a system when listing specific concepts, which is aligned with ValueSet invariant `vsd-2`, and at that point, the concept is uniquely identified. So, leaving this kind of thing out of examples in the documentation seems completely reasonable to me: this doesn't feel like it is ever useful or necessary in order to build a perfectly good ValueSet.

jafeltra:
This PR was confusing to me when I first looked at it, and maybe this is why. I guess one question I have is, shouldn't GoFSH, and therefore the toFSH, produce FSH that we encourage people to use? But are we possibly losing information if we leave out the "from valueset" parts?

mint-thompson:
We definitely want to preserve information. But, leaving out the value sets from the include/exclude concept rule doesn't really cause any loss of information, by my understanding of ValueSet definitions. It would mean that the round trip comparison will show a difference here, though. I'll take some more time to think about this.

cmoesel:
Since the overall result is defined as the intersection between the concepts and the value sets, there is a difference _if_ the value set does not have the concept. Again, it would be silly if anyone did that -- but removing the value set is not necessarily harmless.

jafeltra (approved):

PR #1382, opened 2023-11-07
CaretValueRule codes are quoted if they contain spaces

mint-thompson:
Completes task [CIMPL-1200](https://standardhealthrecord.atlassian.net/browse/CIMPL-1200).

Codes in FSH are quoted if they contain any space characters. When producing FSH for a CaretValueRule, if it is a code caret rule, quote codes that contain space characters.


cmoesel (approved):
Yep. This checks out.

src/fshtypes/rules/CaretValueRule.ts
```diff
@@ -44,11 +44,22 @@ export class CaretValueRule extends Rule {
...
-        : '';
+      if (this.pathArray.length) {
+        printablePath =
+          this.pathArray
+            .map(code => {
+              if (/\s/.test(code)) {
+                return `#"${fshifyString(code)}"`;
+              } else {
+                return `#${code}`;
+              }
```
jafeltra, line 57: Maybe this is a silly question, but why does the case with spaces need to fishify the code but the case without spaces doesn't?

jafeltra: You must fishify the string because it is a string!

PR #1385, opened 2023-11-14
Produce correct FSH for concept CaretValueRules on ValueSets

mint-thompson:
Completes task [CIMPL-1199](https://standardhealthrecord.atlassian.net/browse/CIMPL-1199).

When a CaretValueRule on a ValueSet has a concept as its path, the system is also included. However, this is not the case when the rule is on a CodeSystem. CaretValueRule should produce FSH that is correct in both cases.

My strategy here was to try to use `pathArray` a little more consistently between CodeSystems and ValueSets. In both cases, the `#` will be included in this array's elements. ValueSets include the system `system#code` and CodeSystems don't `#code`. This means needing to trim off the leading `#` in some situations when exporting CodeSystems.

### DIFF
--- a/src/export/CodeSystemExporter.ts
+++ b/src/export/CodeSystemExporter.ts
@@ -209,9 +209,9 @@ export class CodeSystemExporter {
     const conceptIndices: number[] = [];
     let conceptList = codeSystem.concept ?? [];
     for (const codeStep of codePath) {
-      const stepIndex = conceptList.findIndex(concept => concept.code === codeStep);
+      const stepIndex = conceptList.findIndex(concept => `#${concept.code}` === codeStep);
       if (stepIndex === -1) {
-        throw new CannotResolvePathError(codePath.map(code => `#${code}`).join(' '));
+        throw new CannotResolvePathError(codePath.join(' '));
       }
       conceptIndices.push(stepIndex);
       conceptList = conceptList[stepIndex].concept ?? [];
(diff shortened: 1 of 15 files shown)


jafeltra (dismissed):
Looks good to me and works as expected with the example from the Jira issue.

cmoesel:
This seems good, but since it hit a variety of areas in the codebase, I'm running a regression. @mint-thompson - If you already ran one, let me know. Also, it looks like there is a conflict that needs to be resolved.

cmoesel (changes requested):
The full regression came back with some differences that I think will need to be addressed. It looks like some code caret rules got dropped and some child codes got dropped.

If you want to start w/ a regression that highlights the repos that had these differences, you can run this command:
```
npm run regression -- run -r HL7EE/ig-ee-base#master -r HL7/cqf-recommendations#master -r HL7/davinci-crd#master -r HL7/physical-activity#master -r HL7/sdc#master -r JohanGHole/alldevs-demo-ig#master -r JohanGHole/dhis2-hiv-ig#master -r medcomdk/dk-medcom-terminology#master -r medigree/translatable#master -r steveleshan/Measles1232#main -r WorldHealthOrganization/smart-immunizations-measles#main
```

mint-thompson:
The latest update fixes the problems in the regression, as well as correcting some related issues that were popping up involving codes that start with a double-quote `"` character.

jafeltra (approved):
I think the changes here make sense. I also tested that the smaller set of regression repos all have no changes now. If you think it is worthwhile (if no one else has done it yet), I can also run a full regression to be sure nothing new came up.

cmoesel:
I ran a full regression Monday afternoon and got back some changes. I think they might be related to this branch not being up to date w/ master though -- so I'm going to synch the branch and run regression on just the projects that had differences in my last run.

cmoesel (approved):
OK. The regression on those repos came out clean after merging in master -- so it all looks good to me!

PR #1387, opened 2023-11-15
Improve IG exampleCanonical matching on versioned meta.profile for profiles that don't specify a version

### LINKED ISSUE #1386: SUSHI does not set exampleCanonical for examples when meta.profile is versioned, but profile doesn't declare version
If a predefined profile does not declare its version, but an example refers to it in `meta.profile` with a versioned canonical, SUSHI does not properly detect the match. As a result, the example gets documented in the IG JSON w/ `exampleBoolean` instead of `exampleCanonical`.

cmoesel:
If an example uses a versioned canonical in its `meta.profile`, SUSHI will attempt to verify that the versioned profile is in scope for the IG before exporting it as an `exampleCanonical` in the IG JSON. In cases where the target profile did not specify its version, SUSHI failed to find a match. Now SUSHI will fallback to matching on the IG's version (specified in the config) if the profile does not specify a version.

I've separated this into two commits, the first of which adds (failing) tests for this, and the second of which adds the fix. You can also test this against [US Core](https://github.com/HL7/US-Core/).

Fixes #1386

### DIFF
--- a/src/ig/IGExporter.ts
+++ b/src/ig/IGExporter.ts
@@ -870,7 +870,8 @@ export class IGExporter {
           const availableProfileOrLogical = this.pkg.fish(baseUrl, Type.Profile, Type.Logical);
           return (
             availableProfileOrLogical != null &&
-            (version == null || version === availableProfileOrLogical.version)
+            (version == null ||
+              version === (availableProfileOrLogical.version ?? this.config.version))
           );
         }
       );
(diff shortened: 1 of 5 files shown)


jafeltra (approved):
Looks good and I see the appropriate changes in US Core.

mint-thompson (approved):
Looks ship-shape to me!

PR #1390, opened 2023-11-20
Add Characteristics to Logical FSH representation

mint-thompson:
Completes task [CIMPL-1206](https://standardhealthrecord.atlassian.net/browse/CIMPL-1206).

Adds an implementation of `metadataToFSH` to `Logical` so that characteristics will be included in the FSH representation. If no characteristics are present, the keyword is not added.


cmoesel (approved):
I dig it.

PR #1397, opened 2023-11-29
Use rules for most Invariant keywords

mint-thompson:
Helps with [CIMPL-1164](https://standardhealthrecord.atlassian.net/browse/CIMPL-1164).

Description still gets to be a keyword, but all the others use assignment rules. This is the style we want to encourage.


cmoesel (approved):
I like it!

jafeltra (approved):
💯

PR #1400, opened 2023-12-08
Use https to download scripts during sushi init

### LINKED ISSUE #1393: Pulling IG Publisher Scripts from wrong address
sushi/src/utils/Processing.ts currently pulls from http://raw.githubusercontent.com/HL7/ig-publisher-scripts/main/*
This causes an issue behind a proxy because the proxy is only used for https requests.

In my opinion the simplest solution would be to have SUSHI pull from https://raw.githubusercontent.com/HL7/ig-publisher-scripts/main/* instead. 

Is there any reason that this should be http?

cmoesel:
SUSHI init should use https instead of http to download scripts. The message it logs to the console _says_ https, but the actual network request was over http. The PR fixes that so it uses https to download the IG Publisher scripts.

To test, run `sushi init` using this PR and ensure it successfully downloads the scripts.

Fixes #1393

### DIFF
--- a/src/utils/Processing.ts
+++ b/src/utils/Processing.ts
@@ -747,7 +747,7 @@ export async function init(): Promise<void> {
     '_updatePublisher.bat',
     '_updatePublisher.sh'
   ]) {
-    const url = `http://raw.githubusercontent.com/HL7/ig-publisher-scripts/main/${script}`;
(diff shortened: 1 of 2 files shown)


jafeltra (approved):
I ran `sushi init` and the scripts downloaded and worked when I used them.

mint-thompson (approved):

PR #1405, opened 2023-12-14
Canonical to non-Instance contained resources create fragment references

### LINKED ISSUE #1367: Canonical() should result in #fragment when referring to a contained resource
If a resource contains another resource, then within the container resource, `Canonical(ContainedResource)` should result in a fragment like `#contained-resource-id` instead of the full Canonical URL of the resource.

Example: https://fshschool.org/FSHOnline/#/share/3tnpgrc

Zulip: https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/Canonical.20reference.20to.20Contained.20resource/near/391011891

jafeltra:
Fixes #1367, addresses [CIMPL-1189](https://standardhealthrecord.atlassian.net/browse/CIMPL-1189)

This PR makes it so that `Canonical()` references that refer to a contained resource (not created via an inline instance) result in a fragment (`#id`) rather than the full canonical URL. I've had some back and forth on the implementation on this, so a few areas to consider when reviewing this:

- completeness: This approach only addresses the issue within Instances. With the current version of SUSHI, using `Canonical` with an inline instance creates a fragment as expected. This adds support for using either assignment rules or non-Instance FSH entity to add a contained resource to an Instance. However, this PR doesn't change anything for other entity types, so StructureDefinitions still are not able to contain a ValueSet entity directly or with assignment rules. See some examples here: https://fshschool.org/FSHOnline/#/share/47Q1KTc.
- order of rules matters: This approach means that the contained rule needs to come before the reference using `Canonical`. I talked to Chris about this, and we seemed comfortable with that assumption at the time, but anyone is free to change their mind.
- maintainability: One reason I went with this approach to make a PR for is that for the most part, the changes are isolated to a function that can check for a relevant contained resource. Right now, that is only applied within the InstanceExporter. This could fairly easily be added to CodeSystemExporter and ValueSetExporter. It will take more work to support the same functionality in StructureDefinitionExporter.
- performance: Is looping through the validated contained resource rules for every rule that uses a Canonical keyword reasonable? When running a full regression, no repos indicated significant changes in execution time. I tried to limit when we loop through rules (only when the Assignment Rule uses the Canonical keyword) and the number of rules we loop through (only rules on `contained`), but it does add a new loop in our already slow Instance Exporter, so it is worth thinking about.

I ran a full regression and there were only four repos that had changes, which all were expected changes, `DigitalSQR/smart-immunizations-mini#main`, `ritikarawlani/smart-outbreak-measles#main`, `steveleshan/Measles1232#main`, and `WorldHealthOrganization/smart-ot#main`.

For bonus points / inquiring minds, the other approach for this task that I've implemented is in `[fragments-in-canonicals-v2](https://github.com/FHIR/sushi/compare/fragments-in-canonicals-v2)`. The quick summary of that PR is that it covers both Instances and StructureDefinitions with contained resources that reference FSH entities, but it does _not_ support contained resources added via direct assignment rules at the various contained paths on Instances or StructureDefinitions. It also touches more code and changes some fairly basic functions, like `assignString` and `assignFHIRValue`, which makes the code a little harder to maintain and felt a little less clean than this approach. The outcome of the full regression is the same on this branch.


src/fhirtypes/common.ts
```diff
@@ -1601,3 +1601,40 @@ export function findImposeProfiles(sd: any): string[] | undefined {
...
+  );
+  const hasMatchingContainedResourceName = containedResources.some(
+    r => r.pathParts.slice(-1)[0].base === 'name' && r.assignedValue === value
+  );
+  const hasMatchingContainedResourceUrl = containedResources.some(
+    r => r.pathParts.slice(-1)[0].base === 'url' && r.assignedValue === value
+  );
+  if (matchingContainedResource != null) {
+    return matchingContainedResource.assignedValue.id;
+  } else if (
```
mint-thompson, line 1632: Since the return value is known right away if `matchingContainedResource` is not null, this check could happen right after trying to find `matchingContainedResource`.

src/fhirtypes/common.ts
```diff
@@ -1601,3 +1601,40 @@ export function findImposeProfiles(sd: any): string[] | undefined {
...
+  );
+  const hasMatchingContainedResourceId = containedResources.some(
+    r => r.pathParts.slice(-1)[0].base === 'id' && r.assignedValue === value
+  );
+  const hasMatchingContainedResourceName = containedResources.some(
+    r => r.pathParts.slice(-1)[0].base === 'name' && r.assignedValue === value
+  );
+  const hasMatchingContainedResourceUrl = containedResources.some(
+    r => r.pathParts.slice(-1)[0].base === 'url' && r.assignedValue === value
+  );
```
mint-thompson, line 1629: This implementation can get the incorrect result if you have more than one contained resource that has its elements assigned directly in the containing resource. Here's an example:
```
Instance: MyObservation
InstanceOf: Observation
Usage: #example
* status = #draft
* code = #123
* contained[0]
  * resourceType = "Observation"
  * id = "obs-one"
  * status = #draft
  * code = #111
* contained[1]
  * resourceType = "Observation"
  * id = "obs-two"
  * status = #draft
  * code = #222
* valueString = Canonical(obs-two)
```
The value of `valueString` will be `#obs-one`, since the rule on `contained[0].id` is the one that is found for `matchingContainedResourceId`.

mint-thompson (changes requested):
This one is definitely tricky. I took a look at this implementation and at `fragments-in-canonicals-v2`, and I think that I prefer the general approach here, rather than messing around with `assignString` and `assignFHIRValue` on `ElementDefinition`. But, the v2 strategy does have the nice feature of not relying on rule ordering. I think a combination of the two may be useful.

While exporting an Instance or StructureDefinition, if you keep track of all of your rules that assigned a `FshCanonical`, it may be possible to double-check those rules at the end and update the assigned value if the referenced entity is found in `contained`. That way, you don't need to modify the functions on `ElementDefinition` that actually assign the value. Let me know if you want to work together on this.

src/fhirtypes/common.ts
jafeltra: Good catch! I'll fix that, thanks!

src/export/InstanceExporter.ts
```diff
@@ -187,6 +201,10 @@ export class InstanceExporter implements Fishable {
             assignedValue: validatedRule.assignedValue,
             sourceInfo: rule.sourceInfo
           });
+          // Check if the rule we just validated was at a valid contained path to keep track for resolving Canonical references
+          if (/^contained(\[\d\])*/.test(rule.path)) {
```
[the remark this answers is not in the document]
jafeltra: Fixed in [e4837bd](https://github.com/FHIR/sushi/pull/1405/commits/e4837bd810dd83692f6be497b890f27510ceb6e6).

src/fhirtypes/common.ts
jafeltra: Fixed in [e4837bd](https://github.com/FHIR/sushi/pull/1405/commits/e4837bd810dd83692f6be497b890f27510ceb6e6)

src/fhirtypes/common.ts
jafeltra: Fixed in [e4837bd](https://github.com/FHIR/sushi/pull/1405/commits/e4837bd810dd83692f6be497b890f27510ceb6e6).

jafeltra:
Thanks for the comments and suggestion @mint-thompson. I've addressed the 3 individual comments. As for the overall suggestion of how to eliminate the rule ordering constraint, I have a few additional considerations to discuss before implementing.

We haven't yet deferred any rules in the `InstanceExporter`. I can't remember if that was for a particular reason or not, but I don't think there is a super easy way to pull out the logic we have and re-apply it to deferred rules. I don't necessarily think it is impossible, but it does require more refactoring.

For the `StructureDefinitionExporter`, I think this requires slightly more work than just deferring the rules for the assignment rules with canonical values because right now the `StructureDefinitionExporter` doesn't even allow users to assign a ValueSet entity to the contained array. So we need to defer both the contained rules so that all the other exporters can finish exporting, then apply the contained array rules, and then reapply the deferred assignment rules with canonicals. I think this would defer all the assignment rules that use the canonical keyword because we can't know if the canonical keyword is referencing a contained entity until we add the contained rules.

Since that feels a little less straightforward than I was thinking when first reading your comment, I wanted to discuss it a little more before going ahead and implementing something that requires significant changes. Since we know this implementation isn't perfect, maybe it is worth keeping it simpler until we can get a complete solution. Then again, maybe deferring all these rules does make for a complete solution and it is worth doing now. I'm open to thoughts.

src/export/InstanceExporter.ts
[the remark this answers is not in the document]
jafeltra: I added a test for double digit indices and slice names in [a615ead](https://github.com/FHIR/sushi/pull/1405/commits/a615ead98505d448463e0e597b2ba8938c359dc1), but through that I realized that the `*` portion of the original regex actually let us account for those cases as well. So technically they worked all along (which is probably why the regression didn't catch any issues with those cases), but the regex was confusing and misleading. So along with the test in that commit, I also updated the regex again from what it was in the original fix in [e4837bd](https://github.com/FHIR/sushi/pull/1405/commits/e4837bd810dd83692f6be497b890f27510ceb6e6) to be simpler. Now it just checks if the path was either `contained` or it was `contained[` or `contained.`, which both indicate that we're at the contained element. That excludes if there was some other element called `containedThing`, which we do not want to match on here. Let me know what you think.

jafeltra:
[a615ead](https://github.com/FHIR/sushi/pull/1405/commits/a615ead98505d448463e0e597b2ba8938c359dc1) - I pushed up a few small fixes, along with the new test. The fixes include:
- a simpler regex for contained rules (explanation above)
- a better null check when cleaning up resources with null values in `contained`. Current SUSHI throws an error `Cannot read properties of null (reading 'resourceType')`. I noticed this when I took a shortcut in the test and just plopped another value at `contained[10]`, but I think that is allowed so we shouldn't throw a JS error.
- some cleaned up comments in the tests

jafeltra:
I also talked with @mint-thompson about deferring rules so rule order is not important and getting this behavior to work in the other exporters as well. We agreed that containing resources to the `contains` array before referencing them with `Canonical` is a reasonable restriction, so we aren't going to worry about complicating the code to defer rules in Instance Exporter just to eliminate that restriction. We also decided that handling this case in the other exporters is more work than we want to include in this task, so I'll make separate tasks for the different exporters and we can prioritize them as needed (which may be when a user encounters an issue with those resource types).

mint-thompson (dismissed):
The new changes and tests look good. Thank you!

src/fhirtypes/common.ts
```diff
@@ -1602,3 +1604,50 @@ export function findImposeProfiles(sd: any): string[] | undefined {
...
+) {
+  const matchingContainedResource = containedResources.find(
+    r =>
+      r.assignedValue?.url === value ||
+      r.assignedValue?.name === value ||
+      r.assignedValue?.id === value
+  );
+  if (matchingContainedResource != null) {
+    return matchingContainedResource.assignedValue.id;
+  }
```
cmoesel, line 1623: If the contained resource rules can be any rule starting w/ `contained`, then isn't it possible that the assigned value might be at a nested element within a contained resource? E.g.
```
* contained[0].resourceType = "Bundle"
* contained[0].entry[0].resource = MyPlanDefinitionInstance
```
Now if a canonical reference referred to `MyPlanDefinitionInstance`'s url, name, or id, I think this code would create the fragment even though `MyPlanDefinitionInstance` isn't a directly contained resource. But I don't think we want that.

I'd actually suggest that in the code where we build up the list of contained resources (in `InstanceExporter`), we use a more picky regex that only catches paths that end with the `contained` element or end with `url`/`name`/`id` directly on the contained element.

So... collect these ones:
* `contained = Foo`
* `contained[0] = Foo`
* `contained[sliceName] = Foo`
* `contained.url = "http://example.org/foo"`
* `contained[0].name = "Foo"`
* `contained[sliceName].id = "foo"`
* _etc._

But do _NOT_ collect these ones:
* `contained.status = #final`
* `contained[0].entry[0].resource = MyPlanDefinitionInstance`
* `contained[mySlice].entry.resource.id = "foo"`
* _etc._

This will prevent us from matching on resources nested in a contained resource _and_ it should make the looping faster since you're only looping through the truly relevant rules.

cmoesel (changes requested):
This turned out to be pretty clean. I like it!

That said, I think I may have found an edge case that needs addressing (and a bonus opportunity for a small performance improvement).

src/fhirtypes/common.ts
jafeltra: [8977105](https://github.com/FHIR/sushi/pull/1405/commits/8977105549e9cdda2a0f8a6981a304ba33873e38) updates the regex and adds a test case to avoid this edge case.

src/export/InstanceExporter.ts
```diff
@@ -202,8 +202,8 @@ export class InstanceExporter implements Fishable {
             sourceInfo: rule.sourceInfo
           });
           // Check if the rule we just validated was at a valid contained path to keep track for resolving Canonical references
-          // It doesn't matter the index/slice name/child path, so just check the first part of the path is 'contained'
-          if (/^contained(?:$|[[.])/.test(rule.path)) {
+          // Only check if the rule was a directly contained resources (aka 'contained' or 'contained.name/url/id', optionally with slice names or indices)
+          if (/^contained(\[([^\]]+)\](\[\d+\])?)?((\.url|\.name|\.id))?$/.test(rule.path)) {
```
mint-thompson, line 206: This regex looks correct, but it may be possible to simplify it a little bit. I think that `^contained(\[[^\]]+\])?(\[\d+\])?(\.url|\.name|\.id)?$` should be equivalent? Also, the capture group for the slice name could be stricter based on what characters are actually allowed for slice names, but since we've already validated the rule, I feel like we probably don't need that here.

mint-thompson (approved):
The updated regex looks correct! I've got a minor suggestion about it, but I don't think it's necessary.

src/export/InstanceExporter.ts
cmoesel: I like @mint-thompson's simplification, but if `contained` is sliced and then resliced, you could have _multiple_ slice components, so that first `?` really ought to be `*`:
```
/^contained(\[[^\]]+\])*(\[\d+\])?(\.url|\.name|\.id)?$/
```

Since we're not really _validating_ here, however, I think we could be even _looser_ since a numeric index matches the named slice expression anyway:
```
/^contained(\[[^\]]+\])*(\.url|\.name|\.id)?$/
```
Right? `contained`, optionally followed by any number of bracketed tokens and/or optionally followed by url, name, or id subproperty.

jafeltra: Yep! It checks out to me and the tests pass. [29f5bf0](https://github.com/FHIR/sushi/pull/1405/commits/29f5bf0515f6ee24f5e14390039345ddaff58750) has the update. I'm also running another regression to just make sure nothing pops up there.

src/export/InstanceExporter.ts
```diff
@@ -159,10 +160,23 @@ export class InstanceExporter implements Fishable {
...
+        if (value instanceof FshCanonical) {
+          const entityName = value.entityName;
+          const matchingContainedReferenceId = getMatchingContainedReferenceId(
+            entityName,
+            containedRules
+          );
+          if (matchingContainedReferenceId) {
+            value = `#${matchingContainedReferenceId}`;
+          }
+        }
```
cmoesel, line 179: I'm just thinking about the bug report we got yesterday regarding matching references to contained resources. Do you think implementing this for references is as simple as taking the same approach when `value instanceof FshReference`?

cmoesel (approved):
Thanks, Julia! This all looks good to me. I had one question, but it's regarding the possibility of follow-on work. It should have no impact on this PR.

src/export/InstanceExporter.ts
jafeltra: I know we talked about this in person but just wanted to also reply here so we have it. I do think we can reuse this implementation with references to cover the case where we don't create correct references to contained resources created with individual assignment rules. However, I think this approach can only be easily applied to Instances that use `Reference(some-contained-resource)`. It would be a fairly straightforward approach to fill that gap and the case where this came up was in an instance, so it's a good option to implement.

This approach is less straightforward to use in other entities, as mentioned at the top of this PR.

We'd also want to consider how the cleanup in `cleanResource` relates to this case. We could consider if the place in `cleanResource` that tries to resolve the references can be updated to use this approach or if it isn't needed if we can make the changes somewhere else. All thoughts we can consider in a follow-on task.

PR #1409, opened 2024-01-03
Use actual version in IG JSON when dependency version is "latest"

### LINKED ISSUE #1398: IG Publisher doesn't support "latest" as a version
If an IG uses "latest" as a dependency version in sushi-config.yaml, SUSHI should replace "latest" with the actual version number when it exports the IG JSON. If it exports the IG JSON w/ "latest" as the version number, the IG Publisher will not be able to find it.

See: https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/.E2.9C.94.20Getting.20a.20bug.20in.20_genonce/near/406081927

mint-thompson:
Fixes #1398 and completes task [CIMPL-1213](https://standardhealthrecord.atlassian.net/browse/CIMPL-1213).

The IG publisher does not support `latest` as a version for packages in the IG's `dependsOn` array. When adding package information to this array, if the version is `latest`, look for the dependency's IG in the available FHIR definitions. If the dependency's IG is found, use the version it specifies.

### DIFF
--- a/test/ig/IGExporter.IG.test.ts
+++ b/test/ig/IGExporter.IG.test.ts
@@ -443,6 +443,30 @@ describe('IGExporter', () => {
       ]);
     });
 
+    it('should use the resolved version of a package when a dependency version is "latest"', () => {
+      config.dependencies = [{ packageId: 'hl7.fhir.us.core', version: 'latest' }];
+      exporter.export(tempOut);
+      const igPath = path.join(
+        tempOut,
+        'fsh-generated',
+        'resources',
+        'ImplementationGuide-sushi-test.json'
+      );
+      expect(fs.existsSync(igPath)).toBeTruthy();
+      const content = fs.readJSONSync(igPath);
+      const dependencies: ImplementationGuideDependsOn[] = content.dependsOn;
+      expect(loggerSpy.getAllLogs('error')).toHaveLength(0);
+      // Ensure US Core is exported with the resolved version
+      expect(dependencies).toEqual([
+        {
+          id: 'hl7_fhir_us_core',
+          uri: 'http://hl7.org/fhir/us/core/ImplementationGuide/hl7.fhir.us.core',
+          packageId: 'hl7.fhir.us.core',
+          version: '3.1.0'
+        }
+      ]);
+    });
+
     it('should issue an error when a dependency version is not provided', () => {
       config.dependencies = [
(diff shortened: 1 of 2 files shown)


test/ig/IGExporter.IG.test.ts
```diff
@@ -443,6 +443,30 @@ describe('IGExporter', () => {
...
+      );
+      expect(fs.existsSync(igPath)).toBeTruthy();
+      const content = fs.readJSONSync(igPath);
+      const dependencies: ImplementationGuideDependsOn[] = content.dependsOn;
+      expect(loggerSpy.getAllLogs('error')).toHaveLength(0);
+      // Ensure US Core is exported with the resolved version
+      expect(dependencies).toEqual([
+        {
+          id: 'hl7_fhir_us_core',
+          uri: 'http://hl7.org/fhir/us/core/ImplementationGuide/hl7.fhir.us.core',
```
jafeltra, line 463: I think this is an actively good change, but I just want to make sure since it wasn't explicitly mentioned in the issue or the PR. Because we now have a real version set, not just `"latest"`, the `uri` field is also (correctly, I think), taken from the IG, rather than the default we build [here](https://github.com/FHIR/sushi/pull/1409/files#diff-20b28aa731fc636ee110f2e0e3b7f52fe153d4a10cc1f78f8c9299463cda71f2R290).

So in this test, we get the US Core URI above, rather than the constructed `http://fhir.org/packages/hl7.fhir.us.core/ImplementationGuide/hl7.fhir.us.core`that released SUSHI would assign.

jafeltra (commented):
I think this looks really good. There was just one other side effect change that I wanted to confirm we are good with before approving.

test/ig/IGExporter.IG.test.ts
mint-thompson: Thank you for pointing this out! This is a side effect of the fact that changing the version means that the step that tries to get the URL from the IG now correctly finds the IG, since version is part of that search. I'm pretty sure there are no downsides to this.

jafeltra (approved):
Looks good!

cmoesel (approved):
Looks good - and I like the simplicity of the solution.

Technically this _could_ break if someone listed two versions of the same IG and one of them was "latest" -- but that would be a _terrible_ idea and lots of other things could break too, so I'm not concerned.

PR #1412, opened 2024-01-17
Fix fishing type order to match default order

mint-thompson:
Follow-on to #1395 that fixes the type order in two places where types were added, but the order was not fixed before the PR was merged. Oops. Oops!!!!

I ran a full regression, and no changes occurred.


cmoesel (approved):
Hurray for consistency!

jafeltra (approved):
Checks out!

PR #1417, opened 2024-01-23
Update assorted dependencies

mint-thompson:
Completes task [INT-1235](https://standardhealthrecord.atlassian.net/browse/INT-1235).

These updates include any dependency that can be updated without requiring significant code changes.

Some dependencies listed by `npm outdated` are not updated as part of this PR.
- `chalk`: major version 5 causes problems for jest. Updated to latest 4.x release.
- `html-minifier-terser`: major version 6 changes the functions we use to become async, which would require changing more or less the entirety of SUSHI's export functions to async.
- `junk`: major version 4 is an esmodule.
- `title-case`: major version 4 is an esmodule.
- `yaml`: changes to `Document.toString()` behavior makes the comments in the config file produced by `sushi init` move around a bunch. Potentially something we are okay with, but not what I would call a minor enough change to include in this PR.


cmoesel (changes requested):
Fun with dependencies!

I noticed that `npm audit` still reports two vulnerabilities related to the axios that come with FPL. We probably ought to update FPL to resolve that. (But that doesn't have to be part of this PR).

I appreciated the notes you put into the PR description. I think it would be helpful to document that in the project itself so that next time we update dependencies, we remember these things. Could you add a DEPENDENCY-NOTES.md file with that same information in it?

I am also running a 6-month regression just to make sure none of these changes had adverse effects not caught by the tests.

src/errors/BindingStrengthError.ts
```diff
@@ -2,7 +2,10 @@ import { Annotated } from './Annotated';
 
 export class BindingStrengthError extends Error implements Annotated {
   specReferences = ['http://hl7.org/fhir/R4/terminologies.html#strength'];
-  constructor(public foundStrength: string, public requestedStrength: string) {
```
jafeltra, line 5: Prettier is supposed to be the one with the opinions, but I just think the original format was prettier than this. But oh well.

src/fhirtypes/common.ts
```diff
@@ -482,7 +482,7 @@ export function setImpliedPropertiesOnInstance(
                   currentElement.def.min > idx
                     ? currentElement.requirementRoot
                     : [...currentElement.history, newHistory].join('.')
-              } as ElementTrace)
```
jafeltra, line 485: Oh this one is prettier. Nice, Prettier.

test/fhirdefs/FHIRDefinitions.test.ts
```diff
@@ -710,8 +710,8 @@ describe('FHIRDefinitions', () => {
             r === 'Base'
               ? undefined
               : r === 'CodeableReference'
-              ? 'http://hl7.org/fhir/StructureDefinition/DataType'
-              : 'http://hl7.org/fhir/StructureDefinition/Element',
```
jafeltra, line 714: Oh, also prettier. Cool.

jafeltra (commented):
The changes themselves make sense to me. Besides the `DEPENDENCY-NOTES.md` file and pending Chris's regression outcome, this looks good to me.

And based on my tally and my opinion, I guess Prettier had 2 nice changes and 1 meh change, so Prettier wins.

src/errors/BindingStrengthError.ts
cmoesel: It seems like it only made this change for constructors with `public`/`private` modifiers, so I suspect the intent of prettier here is to make these arguments stand out more since they're also a class member declaration.

jafeltra: Ah, perhaps.

jafeltra (approved):
Looks good!

PR #1418, opened 2024-01-24
Support logical model examples without resourceType or id

### LINKED ISSUE #1406: SUSHI chokes on JSON logical model instances that don't have resourceType/id
Technically, logical model instances are not required to have a `resourceType` or `id`, but if you include a JSON instance of a logical model in any of the `input/*` resource folders, SUSHI will try to load it and will emit an error when it can't find `resourceType`/`id`.

@lmckenzi has suggested an approach to detect which of these files can be safely ignored by looking at the resource listing in `sushi-config.yaml`. We should investigate further and implement as suggested and/or only _warn_ when `resourceType` can't be found (instead of _error_) since warnings can be suppressed.

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/How.20do.20I.20get.20SUSHI.20to.20ignore.20a.20binary.20JSON.20logical.20instance.3F/near/407798801

jafeltra:
Fixes #1406 

This PR supports including logical model examples that do not have a `resourceType` and/or an `id`. If the correct extension is added to the resource in the `sushi-config.yaml` file, then SUSHI will detect that and not log any warnings. This was based off the suggested approach from Zulip [here](https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/How.20do.20I.20get.20SUSHI.20to.20ignore.20a.20binary.20JSON.20logical.20instance.3F/near/407798801). I confirmed that this approach works with the IG Publisher, and it will properly process the resource.

One thing to note is that we document including logical models [here](https://fshschool.org/docs/sushi/tips/#instances-of-logical-models) and indicate that the extension `http://hl7.org/fhir/StructureDefinition/implementationguide-resource-format` should be used. However, that Zulip thread indicated the extension `http://hl7.org/fhir/tools/StructureDefinition/implementationguide-resource-format` should be used. I tested this with the publisher, and both extension URLs work the same within the publisher, so I think it is valid for SUSHI to support both as well. If we hear back that one URL is preferred over the other, we can make sure that the Tips and Tricks page in FSH School reflects the best practice.

I also updated the error that was logged when a `resourceType` or `id` was missing to be a warning. If an example without an `id` or `resourceType` is included but configured correctly with the extension, this will not be logged. If the resource isn't configured correctly, the IG Publisher will crash, so we _could_ keep it as an error, so I'm open to feedback here. My reasoning was that it seemed like the error should say something about configuring the resource type to resolve the error since it could be valid that `resourceType` or `id` was missing. Instead of including IGPublisher specific configuration in a SUSHI error message, I chose to just soften it to a warning. But again, I'm open to switching it back.

There's an example in the tests that shows this feature, but I've included a sample project here: [LogicalExamplesWithoutResourceTypeId.zip](https://github.com/FHIR/sushi/files/14030580/LogicalExamplesWithoutResourceTypeId.zip)

There is also a FSH School PR [here](https://github.com/FSHSchool/site/pull/86) that updates the Tips and Tricks to include how to configure an example without a `resourceType` and `id`.


src/ig/IGExporter.ts
```diff
@@ -962,8 +966,39 @@ export class IGExporter {
...
+                resource.extension?.some(e => igResourceFormatExtensionUrls.includes(e.url))
+            );
+
+            // This implementation is based on discussion on Zulip here:
+            // https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/How.20do.20I.20get.20SUSHI.20to.20ignore.20a.20binary.20JSON.20logical.20instance.3F/near/407861211
+            const hasBinaryResourceReference = (this.config.resources ?? []).some(
+              resource =>
+                resource.reference?.reference === `Binary/${path.parse(file).name}` &&
+                resource.extension?.some(e => igResourceFormatExtensionUrls.includes(e.url))
+            );
```
cmoesel, line 988: I think we should consider optimizing this a bit. Right now, for every predefined resource, it will loop through the sushi-config resource list twice.

We could combine the `hasBinaryExampleReference` and `hasBinaryResourceReference` into one loop that checks for the id-based Binary reference plus exampleCanonical, OR the file-based Binary reference. That would get us down to one loop per predefined resource.

OR... we could create a `configuredBinaryReferences` variable _before_ the top-level for loop (line 955-956) initialized using `(this.config.resources ?? []).filter(...)` to get the list of resource references starting with `Binary/` and having one of those extensions on it. Then just check against the `configuredBinaryReferences` list of resources here.

cmoesel (changes requested):
I think a case can be made for having that error be an error or warning (and in fact, you made both cases). I think perhaps warning is fine for our needs -- and also is more flexible if other use cases turn up in the future that allow for resources without id or resourceType.

I've made one suggestion for optimizing/simplifying some of the new code. Feel free to push back if you think it is unnecessary.

src/ig/IGExporter.ts
jafeltra: I like both of these simplifications, but I think I'm leaning towards the second one. I'll try it out and pick one!

jafeltra: Went with the second one in [047b02b](https://github.com/FHIR/sushi/pull/1418/commits/047b02bd6bdd1df628029f3ab40a8939272b22b4).

src/ig/IGExporter.ts
```diff
@@ -949,6 +953,11 @@ export class IGExporter {
       if (pathResourceDirectories) predefinedResourcePaths.push(...pathResourceDirectories);
     }
     const deeplyNestedFiles: string[] = [];
+    const configuredBinaryResources = (this.config.resources ?? []).filter(
+      resource =>
+        resource.reference?.reference.startsWith('Binary/') &&
```
cmoesel, line 958: We probably need to be defensive about both references here to avoid crashing if the nested reference is missing.
```suggestion
        resource.reference?.reference?.startsWith('Binary/') &&
```

jafeltra: Whoops! and oops I forgot to accept the the suggestion so the fix is in [e16e997](https://github.com/FHIR/sushi/pull/1418/commits/e16e997417025854c6a4b99d7aeaa9fc5acbc314).

cmoesel (approved):
This looks good and works well. I tested it using the provide project (thanks!) and confirmed that SUSHI build the IG JSON correctly and then the IG Publisher successfully processes it as well. I also confirmed that a warning is logged if there is no corresponding resource entry in the sushi-config.yaml. Thanks, Julia!

mint-thompson (approved):
Implementation and tests all look good. Thank you!

PR #1419, opened 2024-01-25
Regression script: handle errors thrown during repo compare

cmoesel:
Prior to this PR, when running a regression, if the code that compares repos threw an error, it crashed the regression script. This fix catches the error, marks it as such, and generates the diff files with a message indicating that there was an error comparing the repos.

I don't if this particular error is an idiosyncracy on my system or reproducible on other systems. You can try to reproduce it and see if my PR fixes it.

1. Checkout the master branch and run `npm install`
2. Run `npm run regression -- run -r FHIR/sql-on-fhir-v2#master`
3. It may crash with something like: _Unexpected error:  [Error: ELOOP: too many symbolic links encountered, stat '/path/to/sushi/master/regression/output/FHIR-sql-on-fhir-v2-master/sushi-gh-master/test_report/public/implementations.json']_
4. Now checkout this PR branch
5. Run `npm run regression -- run -r FHIR/sql-on-fhir-v2#master`
6. It should report the status of the compare as ERROR in the CLI
7. Click the diff links in the regression report and confirm they contain the error


regression/run.ts
```diff
@@ -401,87 +401,95 @@ async function generateDiff(
...
-    const v1File = path.join(repoSUSHIDir1, file);
-    const v2File = path.join(repoSUSHIDir2, file);
-    const [v1Contents, v2Contents] = await Promise.all([readFile(v1File), readFile(v2File)]);
-    let v1Json: any = null;
-    let v2Json: any = null;
-    try {
-      v1Json = JSON.parse(v1Contents);
-      v2Json = JSON.parse(v2Contents);
-      if (isEqual(v1Json, v2Json)) {
+  try {
```
cmoesel, line 404: Wow. GitHub really messes this diff up if you're showing whitespace changes. The only change is the surrounding try/catch. You can see this more easily if you tell it to hide whitespace in the diff: 
CleanShot 2024-01-24 at 19 50 50@2x

cmoesel: Holy gigantic image, Batman!

jafeltra (approved):
I also get that error (so it isn't just on your machine) and this branch handles it quite nicely.  Looks good.

mint-thompson (approved):
Tried out the repo you suggested on my system (Win10), and it fails at the `downloadAndExtractRepo` try-catch block. But, the code change looks reasonable, and it works for y'all, so I'm fine with it.

PR #1423, opened 2024-02-02
Fall back to package.json when resolving "latest" version in IG

cmoesel:
When a dependency uses the "latest" version in sushi-config.yaml, SUSHI must populate the generated IG JSON with the real version number (not "latest"). PR #1409 fixes this for dependency packages that contain an ImplementationGuide resource, but it does not address the case were a dependency package does not have an ImplementationGuide resource. In this case, we should fall back to the package's package.json file.

This was raised on Zulip here: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/automatically.20get.20latest.20dependencies/near/419233182

### DIFF
--- a/src/fhirdefs/FHIRDefinitions.ts
+++ b/src/fhirdefs/FHIRDefinitions.ts
@@ -25,6 +25,11 @@ export class FHIRDefinitions extends BaseFHIRDefinitions implements Fishable {
     }
   }
 
+  // Expose the package.json files to support extracting the version when "latest" is used
+  allPackageJsons(): any[] {
+    return Array.from(this.packageJsons?.values() ?? []);
+  }
+
   // This getter is only used in tests to verify what supplemental packages are loaded
   get supplementalFHIRPackages(): string[] {
     return flatten(Array.from(this.supplementalFHIRDefinitions.keys()));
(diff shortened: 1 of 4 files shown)


src/fhirdefs/FHIRDefinitions.ts
```diff
@@ -25,6 +25,11 @@ export class FHIRDefinitions extends BaseFHIRDefinitions implements Fishable {
     }
   }
 
+  // Expose the package.json files to support extracting the version when "latest" is used
+  allPackageJsons(): any[] {
+    return Array.from(this.packageJsons?.values() ?? []);
+  }
+
```
jafeltra, line 32: I think it makes sense to include this in FPL, since that's where all the other `all*` functions are, like `allProfiles`, and the FPL base FHIRDefinitions defines the `packageJsons` list, so it isn't a SUSHI-specific concept. However, since we're trying to get this out in the next release, would you rather wait to do that?

jafeltra (commented):
I just had one question that I was curious about before approving this.

src/fhirdefs/FHIRDefinitions.ts
jafeltra: We talked about this offline, and we'll move this into FPL eventually, but we're okay with keeping it here until we're able to move into FPL. [CIMPL-1238](https://standardhealthrecord.atlassian.net/browse/CIMPL-1238) tracks that task.

jafeltra (approved):

PR #1427, opened 2024-02-14
Allow path-resource paths with a wildcard

### LINKED ISSUE #1415: SUSHI should allow wildcard directory paths in the path-resource parameter
It appears that the IG publisher allows using a directory wildcard `/*` in the `path-resource` parameter. However, SUSHI does not.

In SUSHI, you can list out the path to each subfolder to include, but trying to include multiple directories with a `/*` at the end does not work. For example, you need to do something like this:

```yaml
parameters:
  path-resource:
    - input/resources/group1
    - input/resources/group2
```

rather than something like

```yaml
parameters:
  path-resource:
    - input/resources/*
```

A small project that replicates the issue: [NestedFolders.zip](https://github.com/FHIR/sushi/files/13993751/NestedFolders.zip)

Zulip: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/sub-folder.20support

jafeltra:
Fixes #1415 

This PR allows a user to specify a `*` wildcard in the `path-resource` parameter file path they specify. When a single `<path>/*` is used, SUSHI will allow any level of nested folders and load any file within those folders. I tested this with the IG Publisher, and specifying `<path>/*` allows you to add any resources at any nested level within that path. The IG Publisher does not support specifying `<path>` or `<path>/` or `<path>/**`, so this PR doesn't support those cases.

To test this, you can use the various options in this sample project: [NestedFoldersExample.zip](https://github.com/FHIR/sushi/files/14285686/NestedFoldersExample.zip)

I did change up my implementation a little bit because I was originally using something that was added in Node 18, but I figured since I found a way to accomplish the same thing pretty easily with Node 16 features, that was better than knowingly breaking SUSHI support for Node 16 (even though we don't explicitly support it).


mint-thompson (changes requested):
This implementation still emits a warning when there are subdirectories that match the wildcard path. To try it out, I made another directory, `path-resource-double-nest/john/morenested`, and created another Patient resource at `path-resource-double-nest/john/morenested/Patient-Morenested.json`. The more nested Patient was present in the "deeply nested resources" warning message when I ran the test on line 3011 that has a configured path of `input/resources/path-resource-double-nest/*`.

jafeltra:
Darn.. I guess my workaround to work with Node 16 was actually not correct. I'll either revert it back to the Node 18 features or find a better workaround in Node 16.

mint-thompson (approved):
Everything is working for me, thank you!

src/ig/IGExporter.ts
```diff
@@ -971,6 +971,19 @@ export class IGExporter {
       pathResourceDirectories = pathResources
         .map(directoryPath => path.join(this.inputPath, directoryPath))
         .filter(directoryPath => existsSync(directoryPath));
+      pathResources
+        .filter(directoryPath => directoryPath.endsWith(`${path.sep}*`))
```
cmoesel, line 975: I'm wondering about how this works cross-platform (specifically on Windows). The [parameter docs](https://build.fhir.org/ig/FHIR/fhir-tools-ig/branches/master/CodeSystem-ig-parameters.html) say:
> Relative path to a location in the IG repository from which to load resources. If the path ends in /*, then that means all sub-folders

On Windows, however, I think `path.sep` will be `\`, right? (Unless running in Bash shell, which I suspect is what @mint-thompson might be doing).

I think the parameter value is probably supposed to be cross-platform and not OS-dependent, since you don't know what OS will build the IG. So... I think we probably need to update our handling of this to either:
* Always expect `/` path separators in this value, regardless of OS; OR
* Be agnostic about separators regardless of OS (supporting `/` and `\`)

I think it is probably OK to normalize on `/` (assuming a regression comes back clean) since the documentation says `/*`, which would imply normalization on `/`. But... I'm open to agnosticism too if that's what the team thinks is best.

src/ig/IGExporter.ts
```diff
@@ -971,6 +971,19 @@ export class IGExporter {
...
+        .forEach(directoryPath => {
+          pathResourceDirectories.push(
+            ...readdirSync(path.join(this.inputPath, directoryPath.slice(0, -2)), {
+              withFileTypes: true,
+              recursive: true
+            })
+              .filter(file => file.isDirectory())
+              .map(dir => path.join(dir.path, dir.name))
+          );
+        });
```
cmoesel, line 986: Personal preference, but I think lines 971 - 986 above might be a little cleaner as a single foreach loop. E.g.,
```
const pathResourceDirectories: string[] = [];
pathResource.forEach(pr => {
  const fullPath = path.join(this.inputPath, pr);
  // if fullPath is a directory, add to pathResourceDirectories
  // else if fullPath ends with '/*' recursively add child directories to pathResourceDirectories
});
```

cmoesel (changes requested):
I've made a few comments below to consider.

In addition, I noticed that if you leave path-resource empty in the config file (for example, commenting out all the path-resource values in your example project), SUSHI crashes. This happens in line 751 of `importConfiguration.ts`. I know you didn't break it, but it might be nice to put in a null guard to avoid the crash.

src/ig/IGExporter.ts
mint-thompson: I can confirm that running the sample project on Windows when configured with `input/resources/nested/*` results in a warning for the nested resources. So, this will need to get updated as @cmoesel described.

src/ig/IGExporter.ts
jafeltra: Yeah you're right. I'm not sure why I was so worried about refactoring the previous code, but I'll clean it up.

jafeltra: Refactored in [6634c47](https://github.com/FHIR/sushi/pull/1427/commits/6634c47d1480dc56e2736b00349c6072307eaf9f).

src/ig/IGExporter.ts
```diff
@@ -971,6 +971,19 @@ export class IGExporter {
       pathResourceDirectories = pathResources
```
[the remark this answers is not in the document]
jafeltra: Updated in [6634c47](https://github.com/FHIR/sushi/pull/1427/commits/6634c47d1480dc56e2736b00349c6072307eaf9f). I feel like the wording isn't perfect, so let me know what you think.

src/ig/IGExporter.ts
jafeltra: I think since the documentation says `/*`, let's just normalize on it. I ran a regression and things came back clean, but I tweaked the logic around paths after that so I'll double check with another regression. I updated the tests to use the normalized paths in the `path-resource` parameters that they set, so they should be covering the OS-agnostic paths in the CI. @mint-thompson - the windows tests pass, but do you mind testing this again to be sure?

cmoesel (approved):
It's good!

PR #1428, opened 2024-02-19
Assign FSH Code or Reference to CodeableReference element

mint-thompson:
Completes task [CIMPL-1220](https://standardhealthrecord.atlassian.net/browse/CIMPL-1220).

When writing an assignment rule, if the path refers to a CodeableReference element, the value may be a FshCode or FshReference. The value will be applied to the element's concept or reference child element, respectively. It is possible to have two rules with the same CodeableReference path if one assigns a FshCode and the other assigns a FshReference.

I ran a full regression with these changes, and there were no differences in any of the outputs.


jafeltra (dismissed):
This makes sense to me! The code looks good, and the tests cover everything that I tested locally.

cmoesel (changes requested):
This is great. Your chosen approach for the challenge that we discussed came out very clean. I like it! I asked for one thing related to the tests...

jafeltra (approved):
I still approve.

cmoesel (approved):
Approved!

PR #1429, opened 2024-02-21
Do not allow aliases with unsupported characters

jafeltra:
The FSH 3.0 ballot specifies that Alias names can only contain a limited set of characters. Specifically, it says:

> Alias names SHOULD begin with a dollar sign (“$”) and MUST otherwise contain only ASCII letters, numbers, underscores (“_”), hyphens (“-“), and dots (“.”). 

This PR makes it so that SUSHI now follows these rules as well. Any alias that contains unsupported characters now ~logs an error and is not processed fully~ logs a warning and continues processing. ~This does make it so that aliases that use `$` to get the extra error checking do get extra error checking and log errors that an alias couldn't be resolved. Then, because the Alias isn't processed, another error is logged when processing entities that use the unprocessed alias. All that is to say, because SUSHI no longer allows these characters, there will be 2-3 errors logged when using an alias with unsupported characters (only 1 of which is new to this PR). However, the fix is pretty easy and the errors do show the places that the issue impacts, so I think this is a reasonable approach.~ After switching to the warning approach, this cascade of errors no longer happens, yay.

I did spend some time talking to @mint-thompson about the case where we have special handling for when you miss a space or two around the `=` sign when creating an Alias, but there wasn't a great way to know that that handling had happened and skip past this new error. So in the cases where you might have `Alias myalias= http://example.org` or `Alias myalias=http://example.org`, you'll get the error about missing a space and the ~error~ warning that there are unsupported characters. They aren't wrong and since it would be complicated to try to suppress the second one, @mint-thompson agreed with me that we'll just log both.

I ran a full regression ~and there are two repos that this causes issues for: [joofio/ufis-ig#main](https://github.com/joofio/ufis-ig/tree/main/) and [joofio/ufis-swe-ig#main](https://github.com/joofio/ufis-swe-ig/tree/main/). For these two repos, this PR causes ~300 new errors due to 1 alias with `~` in it and ~100 uses of that alias. Since it is just those two repos with issues, the issue is pretty easy to resolve, and the `~` is not allowed according to the FSH spec, I think this is okay, despite being a little (maybe a lot) annoying. I was thinking of putting up two PRs for the fixes for these repos.~ After switching to the warning approach, the only changes in these repos is one new warning, as expected.


test/import/FSHImporter.Alias.test.ts
```diff
@@ -355,7 +371,55 @@ describe('FSHImporter', () => {
...
+      // The B@dAlias alias does not resolve, since it is not created
+      expect(rule.value).toEqual(
+        new FshCode('foo', 'B@dAlias').withFile('Loinc.fsh').withLocation([6, 10, 6, 21])
+      );
+      expect(loggerSpy.getAllMessages('error')).toHaveLength(1);
+      expect(loggerSpy.getLastMessage('error')).toMatch(/B@dAlias includes unsupported characters/);
+    });
+
+    it('should log an error when an alias contains unsupported characters and starts with $', () => {
+      // The only supported characters in Aliases are ASCII letters, numbers, _, -, .
```
cmoesel, line 400: These two extra tests are nice. Could we get a test for an alias that has all the special allowed characters too? Like `Foo-foo_McFooferton_Jr._the_3rd`?

cmoesel (changes requested):
This looks good. The one thing I'm not sure of, however, is if we _need_ to make this an error. I know it is a MUST in the specification, but since we often _can_ process it correctly, and refusing to process it causes a cascade of other errors, I think we should consider logging it as a warning and processing it "best effort". We do best effort processing in other situations as well.

I was on the fence about this until I tried making a profile with an invalid name (`MyMedic@tion`) and it produced a _warning_ instead of an error, despite that also being a MUST in the spec. So it would be consistent with past behavior to do the same thing (and then we also don't have to feel so bad about crashing that other person's IG).  BTW -- invalid characters in an _id_ do cause an _error_, but I think that's because they also cause an error in IGP, whereas names in FHIR are less strict than FSH, so names with invalid characters are only a warning in IGP. So what do you think? I'm open to discussion on it.

I also asked for one more test to confirm that an alias with all the special characters does not get flagged.

test/import/FSHImporter.Alias.test.ts
jafeltra: Good idea! Added in [c97868b](https://github.com/FHIR/sushi/pull/1429/commits/c97868b063808f28e7764bd3c0126f8ac11d209b).

jafeltra:
I agree with you, especially since names already log a warning where these characters are not supported. If that is the case, then I think aliases can be even more flexible since they don't represent an element within FHIR. I switched to a warning and continue processing the alias when unsupported characters are encountered. Let me know what you think.

cmoesel (approved):
Looks good and passes the tests (automated and manual)!

mint-thompson (approved):

PR #1430, opened 2024-02-23
Quantity may omit numeric value

mint-thompson:
Completes task [INT-1241](https://standardhealthrecord.atlassian.net/browse/INT-1241) in alignment with [this FSH 3.0 spec change](https://jira.hl7.org/browse/FHIR-42337).

Update the grammar to allow a quantity to omit the numeric value component. Since a ratio contains two quantities, this also affects the allowed FSH for a ratio. A quantity without a numeric value is useful for setting the unit.


cmoesel (approved):
This works really well. I manually tested it against my own set of cases like this:
https://fshschool.org/FSHOnline/#/share/3V9pUor

It's nice we got support in ratios for free. I think that makes sense. Unfortunately, ratios are a hidden feature in FSH. As far as I can tell, they are not documented in the spec at all. So... we probably should put them into the next ballot.

test/import/FSHImporter.SDRules.test.ts
jafeltra, line 1: Is it worth testing somewhere in the importer that we can import a quantity with a non-UCUM system with the units/display and no value? Or is this covered by something else that I missed?

jafeltra (commented):
This looks really good. I didn't notice anything in the implementation at all, but I did ask one question about tests.

test/import/FSHImporter.SDRules.test.ts
mint-thompson: I think this scenario is handled by the existing implementation. A non-UCUM system with no value would look like `system#units "display"`, so the `code` parser rule would be used. The value on the rule is a `FshCode`. Assigning a `FshCode` to a Quantity-type element is handled safely in `ElementDefinition.assignFshCode`.

jafeltra: Ah, that makes sense. Thanks for the explanation!

jafeltra (approved):

PR #1432, opened 2024-03-04
Combine included concepts when exporting ValueSet

mint-thompson:
ValueSet concepts with the same system and version may appear as separate rules. This happens most often when RuleSets are used. Included concepts with the same system and version are added to the same concept list.

A full regression for this PR is in progress. Results will be added to this post when the regression is complete.


src/export/ValueSetExporter.ts
```diff
@@ -184,6 +184,28 @@ export class ValueSetExporter {
...
+  private addConceptComposeElement(
+    freshElement: ValueSetComposeIncludeOrExclude,
+    composeList: ValueSetComposeIncludeOrExclude[]
+  ): void {
+    if (freshElement.concept?.length > 0) {
+      const matchingSystemIndex = composeList.findIndex(compose => {
+        return (
+          compose.system === freshElement.system &&
+          compose.version === freshElement.version &&
+          compose.concept?.length > 0
```
cmoesel, line 196: I think we also need to check if the `compose.valueSet` matches the `freshElement.valueSet` -- because if valueSets are _also_ specified in the compose rule, then it selects the intersection of the concepts and valueSets. We don't want to add a concept that isn't supposed to be intersected to a compose rule that has an intersection (and vice-versa).

src/export/ValueSetExporter.ts
```diff
@@ -169,7 +169,7 @@ export class ValueSetExporter {
               }
             );
             if (composeElement.concept.length > 0) {
-              valueSet.compose.include.push(composeElement);
+              this.addConceptComposeElement(composeElement, valueSet.compose.include);
```
cmoesel, line 172: Wouldn't we also want this same behavior for `valueSet.compose.exclude`?

cmoesel (changes requested):
Thanks for finding the issue. I've found a couple of things that probably need to be addressed.

jafeltra (approved):
An excellent use of XOR. Looks good to me.

cmoesel (approved):
This looks good to me as well!

PR #1433, opened 2024-03-05
Replace type.aggregation on ElementDefinition when modified

### LINKED ISSUE #1411: SUSHI should replace ElementDefinition.type.aggregation array when modified in FSH
Some ElementDefinition array properties are additive (you should only add to them) while others should be fully replaced when edited (and perhaps there are other options too). I asked for more information on this when I filed [FHIR-31405](https://jira.hl7.org/browse/FHIR-31405) back in 2021.

The `ElementDefinition.type.aggregation` array is an array property that should be replaced in the snapshot by whatever is in the differential. To support this, however, SUSHI needs to perform special handling of how it generates the differential. More background can be seen in [this Zulip conversation](https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/restricting.20aggregation.20type/near/413120070).

Note that we will need to support things like this at a broader level, so any solution in FSH/SUSHI should be easy to apply to other element properties as well.

mint-thompson:
Fixes #1411 and completes task [CIMPL-1229](https://standardhealthrecord.atlassian.net/browse/CIMPL-1229).

If a rule modifies an element or child element of the type.aggregation array, any values on type.aggregation inherited from the parent are completely replaced. Properties of ElementDefinition with this behavior are defined in REPLACEMENT_PROPS.

No changes appeared in the regression output for this code change.


src/fhirtypes/ElementDefinition.ts
```diff
@@ -472,6 +473,56 @@ export class ElementDefinition {
     this._original = undefined;
   }
 
+  clearOriginalProperty(pathParts: PathPart[]): void {
+    let currentOriginalElement: any = this._original;
+    // eslint-disable-next-line
+    let currentActualElement: any = this;
```
jafeltra, line 479: I just want to double check my own understanding of why we need to keep track of both `this` and `this._original`. Is it just so that `this._original` has the same starting point for type.aggregation so when the diff is generated it is correct?

src/fhirtypes/ElementDefinition.ts
```diff
@@ -472,6 +473,56 @@ export class ElementDefinition {
...
+          }
+          if (Array.isArray(currentActualElement)) {
+            currentActualElement = currentActualElement[currentIndex];
+          }
+        }
+      } else {
+        if (
+          currentOriginalElement?.[pathPart.base] != null &&
+          currentActualElement?.[pathPart.base] != null
+        ) {
```
jafeltra, line 499: For this `if` statement, why do we only clear the actual element if the original and it both have non-null values? But we always clear the original element regardless after this block.

jafeltra (commented):
Wow, I didn't realize it was so tricky to clear `type.aggregation`! The tests were super helpful to see this working. I asked two small questions that were mostly for my own understanding of the code.

src/fhirtypes/ElementDefinition.ts
mint-thompson: Exactly. We need to clear out something on `this._original`, but we have to follow the `clearPath` to get there. Then, when the diff is generated, that part of `this._original` will be `undefined`, which means that the diff will be everything at that path on `this`.

src/fhirtypes/ElementDefinition.ts
mint-thompson: Okay, I'm not entirely sure why I chose that implementation, but in investigating this, I found a scenario that I missed: `clearPath` points to a primitive array, the parent has nothing at that path, but it _does_ have child elements of that path. I'll update this implementation to handle that scenario.

src/fhirtypes/ElementDefinition.ts
jafeltra: Thanks for confirming!

src/fhirtypes/ElementDefinition.ts
jafeltra: The updated logic makes more sense to me, and it looks like it covers your new test case well. Now I think I have one more question related to this part of the logic. Is there a case when the original element would differ from the actual element? As in, is there a time when `currentOriginalElement?.[pathPart.base]` would not be null but `currentActualElement?.[pathPart.base]` would be null? Maybe I'm just thinking they would never differ because of when we call this function, but that probably isn't something we want the function to depend on.

jafeltra (dismissed):
I like that we're able to avoid re-calling the code to clear elements now, and the process to clear looks about as straightforward as possible. Looks good to me.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -472,6 +475,52 @@ export class ElementDefinition {
...
+
+  private calculateClearPath(pathParts: PathPart[]): PathPart[] {
+    const replacementIndex = this._replacementProps.findIndex(
+      replacementPath =>
+        replacementPath.length <= pathParts.length &&
+        replacementPath.every((rp, index) => rp === pathParts[index].base)
+    );
+    if (replacementIndex >= 0) {
+      const clearPath = pathParts.slice(0, this._replacementProps[replacementIndex].length);
+      this._replacementProps.splice(replacementIndex, 1);
```
cmoesel, line 518: I had to think about this a little as I reviewed it. Might it be helpful to add a couple of comments? For example:
> // Build the clearPath from the actual pathParts since they might have indices or slicenames

and

> // Remove the replacement property from _replacementProps so we don't mistakenly clear it again later

src/fhirtypes/ElementDefinition.ts
```diff
@@ -3136,3 +3185,11 @@ const PROPS_AND_UNDERPROPS: string[] = PROPS.reduce((collect: string[], prop) =>
  * See http://hl7.org/fhir/elementdefinition.html#interpretation.
  */
 const ADDITIVE_PROPS = ['mapping', 'constraint'];
+
+/**
+ * These list properties are replaced in child profiles. If they are modified
+ * in a profile, the snapshot should contain only entries in that profile,
+ * and not in the parent profile. Each property is given as a list of
+ * path parts.
```
cmoesel, line 3193: Consider adding:
> For more context and a specific example, see this Zulip thread: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/restricting.20aggregation.20type/near/413120070

cmoesel (changes requested):
This looks good. I think the recent improvements really help a lot.

I tested it with this simple (and technically incorrect) case:
```
Profile: MyPatient1
Parent: Patient
* name ^type.aggregation[0] = #contained
* name ^type.aggregation[+] = #referenced

Profile: MyPatient2
Parent: MyPatient1
* name ^type.aggregation = #contained
```

I suggested a few comments to help future me (or future someone else) as they maintain the code. That said, if you think it's not necessary and the comments will just add clutter, then feel free to push back.

jafeltra (approved):
Still looks good to me.

cmoesel (approved):
Excellent. Thank you for working through all our feedback on this. It looks great!

PR #1436, opened 2024-03-07
Export new mappings only (or at top of list in snapshot mode)

### LINKED ISSUE #1422: Export StructureDefinition with only new mappings
The `StructureDefinition.mapping` element does not need to repeat all mappings from its parent. The IG Publisher will add in the parent mappings during processing. SUSHI should export StructureDefinitions with only the newly introduced mappings and then allow the IG Publisher to handle how they are integrated. Currently, this will result in new mappings being _first_ in the array, which then results in them being listed first in the rendered IG -- which is what authors generally want.

We should, however, consider an exception when SUSHI is run w/ snapshot mode on -- because in that case, we most likely want a complete StructureDefinition with all the mappings (since IG Publisher is less likely to be used in this case). If we do this, consider putting new mappings at the start of the array, like the publisher does.

We may be able to use an approach similar to the `ADDITIVE_PROPS` approach we use in `ElementDefinition`.

See: https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/mapping.20from.20IG.20listed.20first/near/416509917

jafeltra:
Fixes #1422.

This PR makes it so that only new mappings are included in the `StructureDefinition.mapping` list when running SUSHI normally (i.e. not with the `--snapshot` flag). When running SUSHI with the `--snapshot -s` flag, the new mappings will appear at the top of the `StructureDefinition.mapping` list.

SUSHI also supports using FSH `Mapping`s to add a new `comment` to existing mappings on the base definition. When that happens, those mappings are included in the list (or moved to the top with `--snapshot`).

I ran a regression for this in both regular (no `-s`) mode and `-s` mode. Without the `-s` (so a regular regression against `master`), there are a lot of changes because every `mapping` list changes. It is either removed entirely or the inherited mapping elements are removed. Running a regression comparing repos with `-s` on results in 35 repos with changes, which are all repos that added mappings, so the mapping array order has changed. There were a couple repos that directly added mappings with assignment rules. This previously overwrote inherited mapping entries, but those mappings are now re-added at the end of the list, so I think this is an improvement.

### DIFF
--- a/src/fhirtypes/StructureDefinition.ts
+++ b/src/fhirtypes/StructureDefinition.ts
@@ -415,6 +415,28 @@ export class StructureDefinition {
     this.originalMapping = cloneDeep(this.mapping) ?? [];
   }
 
+  buildMappingJSON(j: LooseStructDefJSON, snapshot: boolean) {
+    const newMappings: StructureDefinitionMapping[] = differenceWith(
+      this.mapping,
+      this.originalMapping,
+      isEqual
+    );
+    if (snapshot) {
+      const filteredOriginalMappings = this.originalMapping.filter(
+        m =>
+          !newMappings.some(
+            nm => nm.identity === m.identity && nm.name === m.name && nm.uri === m.uri
+          )
+      );
+      j.mapping = [...newMappings, ...filteredOriginalMappings];
+    } else {
+      j.mapping = newMappings;
+    }
+    if (isEmpty(j.mapping)) {
+      delete j.mapping;
+    }
+  }
+
   /**
    * Exports the StructureDefinition to a properly formatted FHIR JSON representation.
    * @returns {any} the FHIR JSON representation of the StructureDefinition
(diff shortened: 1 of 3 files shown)


mint-thompson (approved):

src/fhirtypes/StructureDefinition.ts
```diff
@@ -88,6 +91,7 @@ export class StructureDefinition {
    * A StructureDefinition instance of StructureDefinition itself.  Needed for supporting escape syntax.
    */
   private _sdStructureDefinition: StructureDefinition;
+  private originalMapping: StructureDefinitionMapping[] = [];
```
cmoesel, line 94: We're a little inconsistent with our use of `_` prefixed private properties in SUSHI -- but I wonder if we should use it here (e.g., `_originalMapping`) since the only other private property uses it -- and since the private `_original` in ElementDefinition.ts uses it as well.

cmoesel (commented):
This looks good. I left one comment for your consideration. @jafeltra - what do you think?

src/fhirtypes/StructureDefinition.ts
jafeltra: So this is a little bit of a silly reason for this, but I didn't use `_originalMapping` because StructureDefinitionExporter actually strips all properties that start with `_` in order to clear out things that do not apply to the new structure definition that is being built (lines [436-443](https://github.com/FHIR/sushi/blob/master/src/export/StructureDefinitionExporter.ts#L436-L443)). However, we do still want to keep track of the original mapping even on the new structure definition. I could add a separate check there to not clear out `_originalMapping` specifically, but I thought that seemed slightly less clean than just using `originalMapping` instead.

cmoesel: Oh, I see. That's a bit of an unintentional side-effect of FHIR also using `_` for primitive shadow properties. Got it. In that case, let's just leave it as is. Thanks!

cmoesel (approved):
👍

PR #1437, opened 2024-03-11
Update alias regex

jafeltra:
Small PR to update the Alias regular expression to require at least one other character besides `$`.

I ran a regression and there were three repos that actually did have Aliases that were just `$`, so they get a new warning now. The three repos are [joofio/ufis-ig#main](https://github.com/joofio/ufis-ig/tree/main/), [joofio/ufis-swe-ig#main](https://github.com/joofio/ufis-swe-ig/tree/main/), and [hl7-eu/gravitate-health#master](https://github.com/hl7-eu/gravitate-health/tree/master/), if you wanted to run a regression on them specifically. The first two also had aliases with unsupported characters, so it would already have been receiving at least one new warning before this PR.

### DIFF
--- a/test/import/FSHImporter.Alias.test.ts
+++ b/test/import/FSHImporter.Alias.test.ts
@@ -439,5 +439,28 @@ describe('FSHImporter', () => {
         /\$N\*tAll\*wed includes unsupported characters/
       );
     });
+
+    it('should resolve but log a warning when an alias contains ONLY a $', () => {
+      // Depending on how you read the FSH spec, this may technically be allowed FSH,
+      // but SUSHI will warn about it.
+      const input = leftAlign(`
+      Alias: $ = http://example.org
+
+      Profile: ObservationProfile
+      Parent: Observation
+      * code = $#foo
+      `);
+
+      const results = importText([new RawFSH(input, 'Alias.fsh')]);
+      expect(results.length).toBe(1);
+      const rule = results[0].profiles.get('ObservationProfile').rules[0] as AssignmentRule;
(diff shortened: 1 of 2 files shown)


test/import/FSHImporter.Alias.test.ts
```diff
@@ -439,5 +439,28 @@ describe('FSHImporter', () => {
...
+      const results = importText([new RawFSH(input, 'Alias.fsh')]);
+      expect(results.length).toBe(1);
+      const rule = results[0].profiles.get('ObservationProfile').rules[0] as AssignmentRule;
+      // The $ alias does still resolve because only a warning is logged
+      expect(rule.value).toEqual(
+        new FshCode('foo', 'http://example.org').withFile('Alias.fsh').withLocation([6, 10, 6, 14])
+      );
+      expect(loggerSpy.getAllMessages('error')).toHaveLength(0);
+      expect(loggerSpy.getAllMessages('warn')).toHaveLength(1);
+      expect(loggerSpy.getLastMessage('warn')).toMatch(/\$ includes unsupported characters/);
```
jafeltra, line 463: Note: This warning message might be a little confusing for the case when the alias is just `$`. Because this case comes up so infrequently and when it does the alias name is referenced and the line number is provided, it didn't seem like it made sense to add another warning message into the code to complicate the implementation for this case. However, if either of you think we should have a more specific error message, I can probably be convinced.

mint-thompson (approved):

cmoesel (approved):
Looks good!

Issue #1439, opened 2024-03-13
Allow setting version and releaseLabel from command line


LodewijkSioen:
This is a feature request for those of us who use a pipeline to publish our IG. In our company we use Git tags to trigger the publishing process. Right now we have a step in our pipeline that will update the sushi-config.yaml file to set the `version` and `releaseLabel` properties based in the tag name. Right now this is done with some dirty powershell scripting.

It would be nice if we could just override these parameters from the command line like this:

`sushi . --igVersion "1.3.5-alfa.1" --releaseLabel "qa-preview" `

cmoesel:
This is an interesting idea. Would you propose that this should actually change the `sushi-config.yaml` file on disk as well, or would you expect the file on disk to stay as-is even if you override its values on the command line?

LodewijkSioen:
I would not change the config file on disk as that would mean changing a file that is under version control. This could lead to problems when checking out the latest version on the next build.

LodewijkSioen:
So I thought to give it a try and this came out: #1442

jafeltra:
Completed in #1442

PR #1442, opened 2024-03-18
Override sushi-config options from command line

LodewijkSioen:
#1439 

Added command line option to override some elements in sushi-config.yaml:
```
 -c, --config <config>    override elements in sushi-config.yaml (supported: 'version', 'status', 'releaselabel') (eg: --config status:draft)
```

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -73,8 +73,12 @@ async function app() {
     )
     .option('-s, --snapshot', 'generate snapshot in Structure Definition output', false)
     .option(
-      '-c, --config <config...>',
-      "override elements in sushi-config.yaml (supported: 'version', 'status', 'releaselabel') (eg: --config 'status:draft')"
+      '-c, --config <config>',
+      "override elements in sushi-config.yaml (supported: 'version', 'status', 'releaselabel') (eg: --config 'status:draft')",
+      (value: string, previous = {}) => {
+        const [k,...v] = value.split(':');
+        return Object.assign(previous, {[k]: v.join(':')});
+      }
     )
     .action(async function (projectPath, options) {
       setLogLevel(options);
@@ -179,7 +183,7 @@ async function runBuild(input: string, program: OptionValues, helpText: string) 
     logger.info(`  --out ${path.resolve(program.out)}`);
   }
   if (program.config) {
-    logger.info(`  --config ${program.config}`);
+    Object.entries(program.config).forEach(([k, v]) => logger.info(`  --config ${k}:${v}`));
   }
   logger.info(`  ${path.resolve(input || '.')}`);
 
--- a/src/app.ts
+++ b/src/app.ts
@@ -72,7 +72,10 @@ async function app() {
       false
     )
     .option('-s, --snapshot', 'generate snapshot in Structure Definition output', false)
-    .option('-c, --config <config...>', 'override elements in sushi-config.yaml (supported: \'version\', \'status\', \'releaselabel\') (eg: --config \'status:draft\')')
+    .option(
+      '-c, --config <config...>',
+      "override elements in sushi-config.yaml (supported: 'version', 'status', 'releaselabel') (eg: --config 'status:draft')"
+    )
     .action(async function (projectPath, options) {
       setLogLevel(options);
       await runBuild(projectPath, options, program.helpInformation()).catch(logUnexpectedError);
--- a/src/app.ts
+++ b/src/app.ts
@@ -74,10 +74,10 @@ async function app() {
     .option('-s, --snapshot', 'generate snapshot in Structure Definition output', false)
     .option(
       '-c, --config <config>',
-      "override elements in sushi-config.yaml (supported: 'version', 'status', 'releaselabel') (eg: --config 'status:draft')",
+      "override elements in sushi-config.yaml (supported: 'version', 'status', 'releaselabel') (eg: --config status:draft)",
       (value: string, previous = {}) => {
-        const [k,...v] = value.split(':');
-        return Object.assign(previous, {[k]: v.join(':')});
+        const [k, ...v] = value.split(':');
+        return Object.assign(previous, { [k]: v.join(':') });
       }
     )
     .action(async function (projectPath, options) {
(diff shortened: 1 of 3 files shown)


cmoesel:
Thank you for the contribution, @LodewijkSioen! We're currently discussing this with the team. We're trying to decide if the way you've done it is best (which is very clear but adds _three_ new options to the CLI) or if it might be better to introduce a _single_ option for which the thing you're overriding is part of the value (e.g., `-ig version:1.2.3 -ig status:draft -ig releaselable:qa-draft`). The single option is a tiny bit harder to use, but allows for easily supporting other overrides without continuing to increase the number of different option flags. Do you have any thoughts on this?

LodewijkSioen:
@cmoesel I'm ok with that. I can imagine the option bloat. However, I don't see a way to do that using [commander.js](https://github.com/tj/commander.js?tab=readme-ov-file#options) except [custom option parsing](https://github.com/tj/commander.js?tab=readme-ov-file#options). This would introduce a bunch of parsing logic. Alternatively, the IG override options could be a json string.

One drawback is that the built-in `--help` documentation will be less useful.

cmoesel:
Hi @LodewijkSioen. I apologize for the delay in responding. We've discussed this some more as a team, and we do think that this feature would work better as a single option (perhaps `-c` / `--config` since you're overriding config data). We feel that this is actually more user-friendly because having too many command-line options can be overwhelming.

I don't think having the user express overrides as a JSON string would be very user-friendly. But I also don't think that you need to implement this using commander's custom option parsing. I'd suggest that you specify the option as a [variadic option](https://github.com/tj/commander.js?tab=readme-ov-file#variadic-option). This would allow users to pass the option multiple times (e.g., `-c version:1.2.3 -c status:draft`), which makes the resulting option value an array (e.g., `['version:1.2.3', 'status:draft']`). Then in [updateConfig](https://github.com/FHIR/sushi/blob/310cd4c7a9c0c8aa5555df8a4bb226ce613e45cc/src/utils/Processing.ts#L225-L243), you just iterate that array, split each item on `:` to get your key and value, and set it in the config (with perhaps a gating function to ensure only allowed keys are used).

How does that sound to you?

LodewijkSioen:
Sounds good to me. This seems more elegant and in line with how most command line options work. I can take a stab at it the week after next.

src/app.ts
```diff
@@ -71,6 +72,10 @@ async function app() {
       false
     )
     .option('-s, --snapshot', 'generate snapshot in Structure Definition output', false)
+    .option(
+      '-c, --config <config...>',
+      "override elements in sushi-config.yaml (supported: 'version', 'status', 'releaselabel') (eg: --config 'status:draft')"
+    )
```
cmoesel, line 78: I thought that variadic arguments worked via repeated uses of the option, but it turns out I was wrong. It just expects any number of values after the option. So when I tested this with `build --config status:draft /path/to/my/project` it mistakenly parsed `/path/to/my/project` as one of the config option values instead of as the final path argument.

So in the end, you were right. You need a custom option parser. That said, it's not that difficult. I tried this and it worked well, returning an array of values:

```ts
    .option(
      '-c, --config <config>',
      "override elements in sushi-config.yaml (supported: 'version', 'status', 'releaselabel') (eg: --config 'status:draft')",
      (value: string, previous: string[] = []) => {
        return previous.concat(value);
      }
    )
```

Now that you have a custom option parser already, you could go the next step and put the string parsing in there too and return an object instead:
```ts
    .option(
      '-c, --config <config>',
      "override elements in sushi-config.yaml (supported: 'version', 'status', 'releaselabel') (eg: --config 'status:draft')",
      (value: string, previous = {}) => {
        const [k, ...v] = value.split(':');
        return Object.assign(previous, { [k]: v.join(':') });
      }
    )
```

If you do that then don't forget to change the lines that log options:
```ts
  if (program.config) {
    Object.entries(program.config).forEach(([k, v]) => logger.info(`  --config ${k}:${v}`));
  }
```

And of course that means you can remove the `parseConfigOverrides` function in `Processing.ts` as well (since it would already be parsed into an object). But now you'll have to fix the tests too.

Anyway, it's up to you where you prefer to parse those strings. These are just a couple of options.

cmoesel (changes requested):
@LodewijkSioen - Yes, that is what I meant. But it turns out that I was mistaken regarding how variadic options work. Oops. See my inline comment for some other ideas. I'm sorry for the confusion and complication!

src/app.ts
LodewijkSioen: Yep, you're right. Variadic kind of forces you to use the clumsy `--` syntax.

Also: `const [k, ...v] = value.split(':');` *chef kiss*

src/app.ts
```diff
@@ -71,6 +72,14 @@ async function app() {
       false
     )
     .option('-s, --snapshot', 'generate snapshot in Structure Definition output', false)
+    .option(
+      '-c, --config <config>',
+      "override elements in sushi-config.yaml (supported: 'version', 'status', 'releaselabel') (eg: --config 'status:draft')",
```
cmoesel, line 77: I just noticed that you wrap the option value in single quotes (e.g. `--config 'status:draft'`). Those are not required, as `--config status:draft` works too. Did you do it that way because values with a space in them do require that?

I don't have a strong opinion, but I don't think that FHIR allows version, status, or releaselabel values to have spaces anyway (or at least as a best practice), so it may not be necessary to show the quotes.

LodewijkSioen: Fixed

LodewijkSioen:
Prettier should be fixed now.

cmoesel (approved):
This is fantastic. Thank you, @LodewijkSioen!

@jafeltra - Can you take a look at this so we have two reviews and approvals?

@LodewijkSioen - In our release notes, we link to the PR. Could you update the PR description to reflect how the feature works now? It currently describes your original approach.

LodewijkSioen:
Release notes updated

jafeltra (approved):
This looks great to me. This ended up being a very clean implementation, and the tests are very helpful.

Thanks for contributing this!

PR #1443, opened 2024-03-19
Improve performance when exporting Instances

### LINKED ISSUE #1426: Apply performance improvements when exporting large numbers of Instances
See [zulip thread](https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/SUSHI.203.2E3.2E0/near/412249552) for original discussion. A [sample repo](https://github.com/bkaney/sushi-perf) was provided to perform the analysis that led to these findings. Sample performance logs for the versions being compared are attached: [v2.10.2-performance.txt](https://github.com/FHIR/sushi/files/14224659/v2.10.2-performance.txt) [v3.3.0-performance.txt](https://github.com/FHIR/sushi/files/14224663/v3.3.0-performance.txt)

Based on an investigation of the differences between SUSHI v2.10.2 and v3.3.0, there are several contributors to longer runtime for the sample repo. 

One is the validation of path rules during Instance export. Path rules are validated with an assigned value of `null`. It may be the case that these rules do not need to be validated at all, and that they can be skipped. If there are cases where it is necessary to validate a path rule, it may be possible to do so with a function that is less expensive than `StructureDefinition.validateValueAtPath`.

Another is that when many instances of the same StructureDefinition are created, the StructureDefinition is fished up and created fresh for each instance. This means that operations such as unfolding elements will be repeated when rules are validated. In this case, `ElementDefinition.unfold` ends up using a large amount of process ticks. It may be possible to reuse a StructureDefinition for each instance being created. There may be limitations to this, though: it may only be possible to reuse a StructureDefinition of a core FHIR resource.

As another potential optimization, it may be possible to skip setImpliedPropertiesOnInstance when creating an instance of a core FHIR resource. This is because core FHIR resources do not have any fixed or pattern values on their elements.

mint-thompson:
Fixes #1426 and completes task [CIMPL-1242](https://standardhealthrecord.atlassian.net/browse/CIMPL-1242).

Instance export remains one of SUSHI's most time-consuming operations. This PR includes several changes that aim to reduce the amount of time needed to export Instances.

- When exporting an Instance, one of the first steps is to find the StructureDefinition that SUSHI is exporting an instance of. This StructureDefinition will usually have its elements unfolded during Instance export in order to validate the paths of assignment rules. These unfolding operations can be time-consuming. In order to avoid repeatedly unfolding the same elements, StructureDefinitions are reused by the InstanceExporter. This offers significant performance gains in cases where a project defines many Instances of the same StructureDefinition, and those Instance definitions include many assignment rules to the same paths.
- Before any of an Instance's rules are applied, its implied properties from its StructureDefinition are set. But, Instances of core FHIR resources will never have their own implied properties. The only case where there may be properties to set is when the Instance contains assignment rules on extension paths where the extension's URL is given as the slice name. Therefore, when the StructureDefinition is a core FHIR resource, other paths do not need to be checked for implied properties.
- When validating an assignment rule on an Instance, a frequent operation is finding an element's parent or children. This can be accomplished by searching the StructureDefinition's list of elements based on the starting element id. However, this can become expensive when the list of elements is long. Elements now maintain references to their parents and children in order to minimize the amount of times the full element list is searched or filtered.

As examples of the benefits of these changes, here are a few selected repos from the regression set, with run times for SUSHI 3.8.0 and for this change set. These times are wall-clock times from my personal system, which are not perfectly accurate measurements of performance. However, I feel that the magnitude of the time reductions are significant.

| Repo | Number of Instances | SUSHI 3.8.0 time (seconds) | PR time (seconds) |
|--------|--------|--------|--------|
| [HL7/fhir-saner#master](https://github.com/HL7/fhir-saner/tree/master/) | 204 | 164 | 114 |
| [HL7/v2-to-fhir#master](https://github.com/HL7/v2-to-fhir/tree/master/) | 247 | 114 | 68 |
| [joofio/test-epi-composition#main](https://github.com/joofio/test-epi-composition/tree/main/) | 2714 | 2362 | 290 |

In the regression set for this PR, some changes appear. However, these are due to existing bugs that were revealed by the changes in this PR. Two issues have been added to the SUSHI backlog for these bugs: https://standardhealthrecord.atlassian.net/browse/CIMPL-1248 and https://standardhealthrecord.atlassian.net/browse/CIMPL-1249

### DIFF
--- a/src/fhirtypes/ElementDefinition.ts
+++ b/src/fhirtypes/ElementDefinition.ts
@@ -282,6 +282,8 @@ export class ElementDefinition {
   binding: ElementDefinitionBinding;
   mapping: ElementDefinitionMapping[];
   structDef: StructureDefinition;
+  treeParent: ElementDefinition;
+  treeChildren: ElementDefinition[];
   private _original: ElementDefinition;
   private _edStructureDefinition: StructureDefinition;
 
@@ -2458,10 +2460,15 @@ export class ElementDefinition {
    * @returns {ElementDefinition|undefined} the parent element or undefined if this is the root element
    */
   parent(): ElementDefinition | undefined {
-    const parentId = this.id.slice(0, this.id.lastIndexOf('.'));
-    if (parentId !== '') {
-      return this.structDef.findElement(parentId);
+    if (!this.treeParent) {
+      const parentId = this.id.slice(0, this.id.lastIndexOf('.'));
+      if (parentId !== '' && this.structDef) {
+        this.treeParent = this.structDef.findElement(parentId);
+      } else if (!this.structDef) {
+        this.treeParent = undefined;
+      }
     }
+    return this.treeParent;
   }
 
   /**
@@ -2486,13 +2493,25 @@ export class ElementDefinition {
    * @returns {ElementDefinition[]} the child elements of this element
    */
   children(directOnly = false): ElementDefinition[] {
-    return this.structDef.elements.filter(e => {
-      return (
-        e !== this &&
-        e.id.startsWith(`${this.id}.`) &&
-        (!directOnly || e.path.split('.').length === this.path.split('.').length + 1)
-      );
-    });
+    if (!this.treeChildren) {
+      this.treeChildren = this.structDef.elements.filter(e => {
+        return (
+          e !== this &&
+          e.id.startsWith(`${this.id}.`) &&
+          e.path.split('.').length === this.path.split('.').length + 1
+        );
+      });
+    }
+    if (directOnly) {
+      return this.treeChildren;
+    } else {
+      const descendents: ElementDefinition[] = [];
+      for (const child of this.treeChildren) {
+        descendents.push(child);
+        descendents.push(...child.children());
+      }
+      return descendents;
+    }
   }
 
   /**
@@ -2924,9 +2943,17 @@ export class ElementDefinition {
     // We don't want to clone the reference to the StructureDefinition, so temporarily save it and remove it
     const savedStructDef = this.structDef;
     this.structDef = null;
+    // We don't want to clone the tree references either
+    const savedTreeParent = this.treeParent;
+    this.treeParent = undefined;
+    const savedTreeChildren = this.treeChildren;
+    this.treeChildren = undefined;
     const clone = cloneDeep(this);
-    // Set the reference to the StructureDefinition again
+    // Set the reference to the StructureDefinition and nearby ElementDefinitions again
+    // we want the clone to know about the StructureDefinition, but not have any tree relationships
     this.structDef = clone.structDef = savedStructDef;
+    this.treeParent = savedTreeParent;
+    this.treeChildren = savedTreeChildren;
     // Clear original if applicable
     if (clearOriginal) {
       clone.clearOriginal();
--- a/src/fhirtypes/StructureDefinition.ts
+++ b/src/fhirtypes/StructureDefinition.ts
@@ -273,7 +273,7 @@ export class StructureDefinition {
     let newMatchingElements: ElementDefinition[] = [];
     // Iterate over the path, filtering out elements that do not match
     for (const pathPart of parsedPath) {
-      // if we have all our elements can we do this with a tree instead
+      // TODO: if we have all our elements, can we do this with a tree instead?
       // Add the next part to the path, and see if we have matches on it
       fhirPathString += `.${pathPart.base}`;
       newMatchingElements = matchingElements.filter(
--- a/src/fhirtypes/StructureDefinition.ts
+++ b/src/fhirtypes/StructureDefinition.ts
@@ -1,6 +1,5 @@
 import upperFirst from 'lodash/upperFirst';
 import cloneDeep from 'lodash/cloneDeep';
-import escapeRegExp from 'lodash/escapeRegExp';
 import differenceWith from 'lodash/differenceWith';
 import isEqual from 'lodash/isEqual';
 import isEmpty from 'lodash/isEmpty';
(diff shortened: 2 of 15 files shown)


src/fhirtypes/StructureDefinition.ts
```diff
@@ -158,26 +157,75 @@ export class StructureDefinition {
...
+        // if this is the only child, splice in after parent
+        // if this is not the only child, splice in after your next-older sibling's last child
+        if (element.treeParent.treeChildren.length === 1) {
+          const parentIndex = this.elements.indexOf(element.treeParent);
+          this.elements.splice(parentIndex + 1, 0, element);
+        } else {
+          const olderSibling = element.treeParent.treeChildren.slice(
+            -2,
+            element.treeParent.treeChildren.length - 1
+          )[0];
```
jafeltra, line 202: Why is -2 the start index here?

src/fhirtypes/StructureDefinition.ts
```diff
@@ -225,6 +273,7 @@ export class StructureDefinition {
     let newMatchingElements: ElementDefinition[] = [];
     // Iterate over the path, filtering out elements that do not match
     for (const pathPart of parsedPath) {
+      // if we have all our elements can we do this with a tree instead
```
jafeltra, line 276: Is this something for a future improvement? Or is this outdated now?

jafeltra (commented):
The performance improvements with this branch really are great. I left a few comments/questions inline. One other thing I noticed is that the `ElementDefinition.assignFshCode` tests take a lot longer to run on this branch than on master. I don't really know why (but I also didn't look too much into it), and I'd rather have actual SUSHI run faster than have the tests run faster, but it seemed odd so I was wondering if anyone else noticed that and/or if there was a reason that might have happened on this branch that I missed.

src/fhirtypes/StructureDefinition.ts
mint-thompson: This slice gets us the second-to-last element of `element.treeParent.treeChildren`. Since `element` has already been added to the tree, it is the last element of `element.treeParent.treeChildren`. So, the second-to-last element is the sibling that needs to be checked for any child elements.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -2924,9 +2943,17 @@ export class ElementDefinition {
     // We don't want to clone the reference to the StructureDefinition, so temporarily save it and remove it
     const savedStructDef = this.structDef;
     this.structDef = null;
+    // We don't want to clone the tree references either
+    const savedTreeParent = this.treeParent;
+    this.treeParent = undefined;
+    const savedTreeChildren = this.treeChildren;
+    this.treeChildren = undefined;
```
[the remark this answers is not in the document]
mint-thompson: This is something of a concession to the tests that compare two `ElementDefinition` instances using `expect.toEqual`. A newly created `ElementDefinition` will have `undefined at these two properties, so in order for a clone to match, it must also have `undefined`, not `null`. The other option is for us to implement a custom jest matcher that treats `null` and `undefined` as equivalent. That might be the best idea here, since it does look a little strange.

mint-thompson:
> One other thing I noticed is that the `ElementDefinition.assignFshCode` tests take a lot longer to run on this branch than on master.

Okay, so the good news is, it's not just you: this is also happening for me. This makes me think it's worth investigating, so I'll... investigate.

src/fhirtypes/StructureDefinition.ts
mint-thompson: This is something for a future improvement. I'll label it TODO to make that more apparent.

src/fhirtypes/ElementDefinition.ts
```diff
@@ -2486,13 +2493,25 @@ export class ElementDefinition {
...
-        e.id.startsWith(`${this.id}.`) &&
-        (!directOnly || e.path.split('.').length === this.path.split('.').length + 1)
-      );
-    });
+    if (!this.treeChildren) {
+      this.treeChildren = this.structDef.elements.filter(e => {
+        return (
+          e !== this &&
+          e.id.startsWith(`${this.id}.`) &&
+          e.path.split('.').length === this.path.split('.').length + 1
```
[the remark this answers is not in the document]
mint-thompson: When `this.treeChildren` is falsy, it means we haven't set it up yet. This array should only contain direct children, so the condition is written with the idea that `directOnly` is true. This block only does set-up, so we always continue on to checking `directOnly` before returning anything.

src/fhirtypes/ElementDefinition.ts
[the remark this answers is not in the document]
mint-thompson: Update: I tried implementing a custom matcher, but it basically means reimplementing lodash equality, so I'm just going to leave these as undefined.

mint-thompson:
I've made some small updates to comments. There's also a [new issue](https://standardhealthrecord.atlassian.net/browse/CIMPL-1254) for yet another bug discovered when investigating the differences in output on this branch. Next is investigating `assignFshCode`.

src/fhirtypes/ElementDefinition.ts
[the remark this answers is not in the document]
jafeltra: Yeah I think that is the right choice. It is a little weird to have the code be written to make the tests happy, but not weird enough to say we should reimplement lodash equality 😳

src/fhirtypes/ElementDefinition.ts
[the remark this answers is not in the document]
jafeltra: Ahhh you're totally right. I read the `return` wrong here. Thanks for the explanation!

src/fhirtypes/StructureDefinition.ts
jafeltra: Ahhh okay, that makes sense. I think part of why I got confused was that the `slice` used `-2` and `element.treeParent.treeChildren.length - 1`. Is there a reason to no just use `-2` and `-1`? Or `element.treeParent.treeChildren.length - 2` and `element.treeParent.treeChildren.length - 1`? Or am I just messing up my math completely?

mint-thompson: Oh, it's because I forgot that the second argument can also be relative to the end of the array, so this can just be `slice(-2, -1)`. That's much simpler.

jafeltra (approved):
This looks great! I think these changes will be super helpful! I didn't notice anything else, so 💯

src/fhirtypes/StructureDefinition.ts
```diff
@@ -158,26 +157,72 @@ export class StructureDefinition {
...
+        for (; i < this.elements.length; i++) {
+          const currentId = this.elements[i].id;
+          if (new RegExp(`^${this.escapePath(currentId)}[.:/]`).test(element.id)) {
+            lastMatchId = currentId;
+          } else {
+            const lastMatchRegex = this.escapePath(lastMatchId);
+            if (
+              !new RegExp(`^${lastMatchRegex}[./:]`).test(currentId) ||
+              // If element is not a slice at this level, and the currentId is a slice, break to add children before slices
+              (new RegExp(`^${lastMatchRegex}[.]`).test(element.id) &&
```
cmoesel, line 184: This is very nitpicky, but since you're only matching on one character, I think this would suffice (and might be more efficient depending on RegEx implementation):
```suggestion
              (new RegExp(`^${lastMatchRegex}\\.`).test(element.id) &&
```

cmoesel (commented):
It looks like this was already merged, but I had a couple of very nitpicky things already commented on. So I'll leave them here in case we want to make the changes. But generally speaking, it all looks good -- and the implementation is smart!

PR #1444, opened 2024-03-20
Replace references when applying caret rules on CodeSystems

mint-thompson:
Completes task [CIMPL-1211](https://standardhealthrecord.atlassian.net/browse/CIMPL-1211).

The `replaceReferences` function is now used by `CodeSystemExporter` in a similar manner to how it is used by `StructureDefinitionExporter`. CodeSystems only need this for caret rules.

### DIFF
--- a/src/export/CodeSystemExporter.ts
+++ b/src/export/CodeSystemExporter.ts
@@ -7,7 +7,8 @@ import {
   determineKnownSlices,
   setImpliedPropertiesOnInstance,
   validateInstanceFromRawValue,
-  isExtension
+  isExtension,
+  replaceReferences
 } from '../fhirtypes/common';
 import { FshCodeSystem } from '../fshtypes';
 import { CaretValueRule, ConceptRule } from '../fshtypes/rules';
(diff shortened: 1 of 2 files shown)


jafeltra (commented):
I think the code changes here make sense. I just had one question about the issue reported on Zulip:

It looks like the test case shows the second issue from the Zulip thread where they try to se the `property.valueCoding` on a code. It looks like they were also trying to set the `property.uri` on the CodeSystem itself using `CodeSystemConceptProperties#country`. I'm kind of confused what the expected value is here. This branch logs no warnings and sets the `uri` to `country`, so I'm guessing that is what is expected, but I'm not sure exactly why.

mint-thompson:
The reason for this behavior comes from `ElementDefinition.assignFshCode`. When an element's type is code, uri, or string, the FshCode's code property is assigned. So, you are correct in observing that this isn't really helpful for the scenario in the zulip thread. Setting `CodeSystem.property.uri` to a specific code within a system appears to be the typical usage, in that [FHIR uses it](https://hl7.org/fhir/R5/codesystem-concept-properties.html) in the CodeSystem that defines some common concept properties. So, it would be useful for this to be easy to do in FSH. If uri elements were assigned with `system#code` instead of just `code`, that would let us do it. That seems reasonable, so I'll see if that causes any unfortunate regressions.

cmoesel:
Good observation, @jafeltra, and good investigation, @mint-thompson! I agree -- given that setting the URI to `{system}#{code}` is apparently common, I think it makes sense to look into this. It's a little outside of spec, because the spec says this format can be assigned to code, Coding, CodeableConcept, and CodeableReference (and does not mention uri) -- but we're already allowing the assignment in SUSHI anyway; so we might as well make it useful I guess. Of course this all depends on whether or not @mint-thompson's regression finds any reasonable use cases for doing it the other way.

mint-thompson:
I ran a regression that assigned `system#code` to uri-type elements, and it showed some undesirable regressions. So, I don't think we can make that change as part of this PR. But, it's probably still worth thinking about how to try to handle this situation in the future.

jafeltra (approved):
Thanks for looking into that other case from Zulip. This looks good to me!

cmoesel (approved):
Excellent.

PR #1446, opened 2024-03-22
Find value for choice slice when choice element has one type

mint-thompson:
Completes task [CIMPL-1248](https://standardhealthrecord.atlassian.net/browse/CIMPL-1248).

When a slice has a descendent that is a choice element, and that choice element has only one type, a choice slice is not created on the StructureDefinition to represent that choice slice. This is because the choice element itself has all the necessary information when it has only one type. When validating required elements on an instance, check for the presence of such an element. If it is found, use its type to determine the name of the property to look for on the instance being validated.

### DIFF
--- a/src/export/InstanceExporter.ts
+++ b/src/export/InstanceExporter.ts
@@ -434,6 +434,27 @@ export class InstanceExporter implements Fishable {
             break;
           }
         }
+        // if the choice only has one type, we may not have created the choice slice.
+        // if this happens, everything is fine.
+        if (instanceChild == null) {
+          const singleTypeChoiceElements = possibleChoiceSlices.filter(
+            c => c.path === child.path && c.type.length === 1
+          );
+          for (const choiceElement of singleTypeChoiceElements) {
+            const sliceName = childPathEnd.replace('[x]', upperFirst(choiceElement.type[0].code));
+            instanceChild = instance[`_${sliceName}`] ?? instance[sliceName];
+            const splitPath = splitOnPathPeriods(choiceElement.id);
+            if (
+              instanceChild != null &&
+              (instance._sliceName
+                ? splitPath[splitPath.length - 2].split(':')[1] === instance._sliceName
+                : true)
+            ) {
+              child = choiceElement;
+              break;
+            }
+          }
+        }
         // If we don't have instanceChild yet, it may be due to a rule that refers to a named slice using a numeric index somewhere in the path.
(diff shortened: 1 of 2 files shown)


src/export/InstanceExporter.ts
```diff
@@ -434,6 +434,27 @@ export class InstanceExporter implements Fishable {
...
+              instanceChild != null &&
+              (instance._sliceName
+                ? splitPath[splitPath.length - 2].split(':')[1] === instance._sliceName
+                : true)
+            ) {
+              child = choiceElement;
+              break;
+            }
+          }
+        }
```
jafeltra, line 457: This seems to work well, but it seems pretty similar to the block above it (on lines 414 - 436). The two main differences seem to be the condition in the choiceSlices filter and the slice name used to get the instanceChild. I briefly tried to update the [filter on line 419](https://github.com/FHIR/sushi/pull/1446/files#diff-7ccabd62381ce8ebf05457c24cc3884eeefb1ec67ff259b82ca7f2c14c4f51cfR419) to be something like this:

```
const choiceSlices = possibleChoiceSlices.filter(
  c => c.path === child.path && (c.sliceName || c.type.length === 1)
);
```
And then I made a variable to either use the `choiceSlice.sliceName` or replace the `[x]` portion like you added and used that variable when setting the `instanceChild` [on line 422](https://github.com/FHIR/sushi/pull/1446/files#diff-7ccabd62381ce8ebf05457c24cc3884eeefb1ec67ff259b82ca7f2c14c4f51cfR422), and things seemed to be okay.

This seems like it might reduce some duplicate code and reduce the chance that we might edit one of these without editing the other. But I'm curious what your thoughts are. I also didn't thoroughly test this, it was more just a quick attempt on my end, so there might be a reason that I missed that these two cases are different.

jafeltra (commented):
This seems to work well. I actually tried pulling this code into the performance improvements branch and ran it on the set of repos that had new warnings, and they all seemed resolved appropriately (except for the one repo where you already made a new Jira ticket). So yay!

I did just ask one question about a possible refactor, but besides that, the functionality seems to be great.

src/export/InstanceExporter.ts
cmoesel: Ohhh... smart. I like it!

jafeltra (approved):
This makes sense to me! Thanks for the mini-refactor!

PR #1448, opened 2024-03-25
Ignore all regresion/output* folders

jafeltra:
A small PR so I can continue to hold on to different regressions I still need without Git being sad. And may as well make ESLint and Prettier not sad.


cmoesel (approved):
YES

mint-thompson (approved):
👸 Good.

PR #1452, opened 2024-04-04
Support extensions on IG.definition in sushi-config.yaml

### LINKED ISSUE #1440: Add definition to sushi-config.yaml to support extensions on ImplementationGuide.definition
Some extensions, such as [ig-link-dependency](https://hl7.org/fhir/tools/0.1.0/StructureDefinition-ig-link-dependency.html) have context `ImplementationGuide.definition`. Since `sushi-config.yaml` abstracts away the `ImplementationGuide.definition` element, however, there is nowhere to put these extensions in `sushi-config.yaml`.

Consider allowing a top-level `definition` property on `sushi-config.yaml` so that there is somewhere to hang these extensions.

jafeltra:
Fixes #1440

This PR adds the option to include extensions directly on IG.definition. Right now, _only_ extensions are allowed on the definition property in the configuration file. All the other properties still need to use their top-level configuration properties.

This will also require an update to FSH School to reflect the new configuration property. Once the approach here is set, I can make a PR with the new information there.

### DIFF
--- a/src/fhirtypes/ImplementationGuide.ts
+++ b/src/fhirtypes/ImplementationGuide.ts
@@ -55,6 +55,7 @@ export type ImplementationGuideDefinition = {
   page?: ImplementationGuideDefinitionPage;
   parameter?: ImplementationGuideDefinitionParameter[];
   template?: ImplementationGuideDefinitionTemplate[];
+  extension?: Extension[];
 };
 
 export type ImplementationGuideDefinitionGrouping = {
--- a/src/fshtypes/Configuration.ts
+++ b/src/fshtypes/Configuration.ts
@@ -80,6 +80,8 @@ export type Configuration = {
   // the correct URL when generating the IG JSON.
   groups?: ConfigurationGroup[];
 
+  definition?: ConfigurationDefinition;
+
   // The resources property corresponds to IG.definition.resource. SUSHI can auto-generate all of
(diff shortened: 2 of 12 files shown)


cmoesel (commented):
Who knew that adding one little property would touch so many files!

This works well. The one thing I think we should consider is if we want a more helpful error message. Currently, if I try to put a `resource` property under `definition`, it says:
> Configuration property definition contains unexpected property: resource. Check that this property is spelled, capitalized, and indented correctly.

But maybe it would be more helpful if it said something like:
> Configuration property definition contains unexpected property: resource. Only the extension property is allowed under definition. All other definition properties are represented at the top-level of the configuration.

It looks like you could do this by overriding the recommendation text, similar to what we do for `pages` [here](https://github.com/FHIR/sushi/blob/299c285bef5d339ddb760a6b6006691ac161a6c7/src/import/importConfiguration.ts#L941-L945) (except replace the recommendation text instead of appending to it). What do you think?

jafeltra:
I had forgotten that we provided specific recommendations for pages, so I like this idea! I used your warning text in [8200756](https://github.com/FHIR/sushi/pull/1452/commits/8200756989cfb899ca73ac6698110e58799c40ea). Let me know what you think and if you notice anything else.

cmoesel (approved):
A little bit odd, but a lot a bit helpful. Thanks!

PR #1453, opened 2024-04-09
v3.9.0

jafeltra:
Version bump to prep for next release. No issues were reported by `npm audit`.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -7497,9 +7497,9 @@
       }
     },
     "node_modules/tar": {
-      "version": "6.2.0",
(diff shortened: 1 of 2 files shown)


cmoesel:
Between when you submitted this and now, `npm audit` found something to complain about. I pushed up the result of `npm audit fix`.

This all looks good, but since there were some substantial changes in here and PRs to address issues introduced by other PRs, etc, I'm running a full regression from 3.8.0 to this. I'll let you know how it goes later tonight. If you don't hear from me, then... I forgot to check it. :-/

cmoesel (approved):
Regression looks good! Two repos w/ changes because we now catch re-defined elements in logical models, and several repos with changes only in warning counts. This is what we expected, so I'm signing off on it!

PR #1455, opened 2024-04-15
Prevent logging an error when Parent doesn't have a mapping array

### LINKED ISSUE #1454: SUSHI logs an unexpected error when adding a Mapping to a Profile whose Parent does not have a mapping array
SUSHI recently made it so that StructureDefinitions only include new mappings when exported and the `mapping` array is removed entirely if there are no new mappings. However, this leads to an issue where SUSHI logs an unexpected error when trying to add mappings to a profile that is based off one of these StructureDefinitions without a `mapping`.

Based on Zulip thread: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Error.20cannot.20read.20properties.20of.20undefined.20.28reading.20'find'.29/near/432906929

Example from Zulip:

```
Profile: MyProcedure
Parent: https://www.medizininformatik-initiative.de/fhir/core/modul-prozedur/StructureDefinition/Procedure
* encounter MS

Mapping: MyMapping
Id: mapping-test
Title: "Mapping Test"
Source: MyProcedure
* encounter -> "Test"
```

with the following dependency in `sushi-config.yaml`:

```yaml
dependencies:
  de.medizininformatikinitiative.kerndatensatz.prozedur: 2024.0.0
```

Note: I had to use the Firely Terminal to build the snapshots.

jafeltra:
Fixes #1454 

Prevent logging an error when a parent that was fished using `fishForFHIR` does not have a `mapping` array. I don't think we should be logging this error regardless, so this PR adds a quick fix to prevent the error.

The error itself came up because a dependency that was used had an SD that was a profile's parent and did not have a `mapping` array on it, which is likely because SUSHI now only adds new or changed mappings, unless SUSHI uses `--snapshot` mode. I'm not sure if the fact that it has come up that at least one package is being published without mappings means we should consider other times to regenerate the full `mapping` list rather than just in `--snapshot` mode or relying on the IG Publisher to create it. I think this PR is relevant, regardless of what we decide about that, but we can decide whether another follow on task is needed.

### DIFF
--- a/src/export/MappingExporter.ts
+++ b/src/export/MappingExporter.ts
@@ -91,7 +91,7 @@ export class MappingExporter {
         Type.Extension,
         Type.Logical
       );
-      const matchingParentMapping = parent?.mapping.find(
+      const matchingParentMapping = parent?.mapping?.find(
         (m: StructureDefinitionMapping) => m.identity === fshDefinition.id
       );
       if (matchingParentMapping != null) {
--- a/test/export/MappingExporter.test.ts
+++ b/test/export/MappingExporter.test.ts
@@ -1,28 +1,44 @@
 import path from 'path';
 import { cloneDeep } from 'lodash';
 import { loadFromPath } from 'fhir-package-loader';
-import { MappingExporter, Package } from '../../src/export';
+import { MappingExporter, StructureDefinitionExporter, Package } from '../../src/export';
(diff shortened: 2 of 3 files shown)


cmoesel (approved):
10/10. Perfect.

I don't think we need to do anything else. The output still adds the top-level mapping for the mapping in the differential, so it's all good!

PR #1456, opened 2024-04-16
Set version based on sushi-config.yaml in FSHOnly projects

### LINKED ISSUE #1380: Support propagation of version, status, publisher to FSH-Only IGs
SUSHI 3.0 removed automatic propagation of version and status since the IG Publisher does this already (and in a more sophisticated manner). Authors using SUSHI in FSH-Only mode, however, don't use the IG Publisher, and therefore do not benefit from its built-in propagation of these properties.

We should consider adding back support for propagating these properties. In order to not introduce what might be deemed a breaking change, we should consider making this behavior opt-in via `sushi-config.yaml` configuration. We will need to determine the format of that config (multiple properties, single property w/ multiple values, etc) and whether it should apply only to FSH-Only IGs or to any IG that sets those config properties.

See: https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/Profiles.20Versioning/near/400510389

jafeltra:
Fixes #1380

This PR adds back support for using the version specified in `sushi-config.yaml` in Structure Definitions, Value Sets, and Code Systems as a default value in FSH Only projects. If the `version` is set with a rule, that will overwrite the version from the config file. This is only done in the case where `FSHOnly` is set to `true` in the config file because the IG Publisher will set the version when it is used.

Because `status` is already set using the value in `sushi-config.yaml` as a default in all projects, this should restore the functionality that was removed in SUSHI 3.x. As a reference, I was looking at PR #1143, which removed this functionality originally.

### DIFF
--- a/src/export/CodeSystemExporter.ts
+++ b/src/export/CodeSystemExporter.ts
@@ -38,7 +38,11 @@ export class CodeSystemExporter {
     }
     if (fshDefinition.title) codeSystem.title = fshDefinition.title;
     if (fshDefinition.description) codeSystem.description = fshDefinition.description;
-    delete codeSystem.version; // deleting to allow the IG Publisher default to take hold
+    if (this.tank.config.FSHOnly) {
+      codeSystem.version = this.tank.config.version;
+    } else {
+      delete codeSystem.version; // deleting to allow the IG Publisher default to take hold
+    }
     codeSystem.status = this.tank.config.status;
     codeSystem.url = `${this.tank.config.canonical}/CodeSystem/${codeSystem.id}`;
   }
--- a/src/export/StructureDefinitionExporter.ts
+++ b/src/export/StructureDefinitionExporter.ts
@@ -382,7 +382,11 @@ export class StructureDefinitionExporter implements Fishable {
       delete structDef.title;
     }
     structDef.status = this.tank.config.status;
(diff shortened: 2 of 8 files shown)


cmoesel (approved):
This looks good and works well. I wish we had thought to do this for FSHOnly back when we removed it for better alignment with the IG Publisher!

As I did this, I had one thought for us to consider. If someone defines an `Instance` with `#definition` usage, and the resource they are creating an instance of has `status` and/or `version`, should we also apply it there for FSHOnly projects? I don't think SUSHI 2.x did that, so this would be new. We auto set the `url` in these cases now, so it _might_ make sense to auto-set `status` and `version` -- but I can't decide for sure. What do you think?

This is approved either way, because I think if we decide to propagate those into instances, it should be a separate PR anyway.

jafeltra:
> We auto set the `url` in these cases now, so it _might_ make sense to auto-set `status` and `version` -- but I can't decide for sure. What do you think?

Oh I think I like that idea, especially since status is usually (always?) required. I was trying things out and happened to make an instance of a CapabilityStatement, which also has `fhirVersion` with cardinality `1..1`. I wonder if we should try to include that from the configuration if there's a `fhirVersion` element.

jafeltra:
I created #1457 to separately consider adding defaults for other values.

PR #1459, opened 2024-04-29
New options for sushi init command

### LINKED ISSUE #1445: sushi init with no prompts
@cmoesel 

I was initially thinking something like 

`sushi init -y`

would be handy, just to create the default example project with no prompts.  But I think the shortcut specifying the project name would be a good option, and if it was handy enough maybe something like

`sushi init -name MyIG -version 0.10`

which sort of leads on to maybe exposing more of the config options such as canonical, fhirVersion etc.  

I think being able to set a version could be possibly be useful for CI, as well as minimising required changes in config for the project after creation.  

The original reason I was looking at this was to create a package using the standard tooling for test automation, so being able to version and name the project so it ended up in the package json would be very useful in that context.  Although a little convoluted just to create a package, but I couldn't see more minimal approach that used the standard tool chain

jafeltra:
Fixes #1445 

This PR updates the `sushi init` command with the following new arguments and options:
- `name` argument: allows you to give the name of the project you are creating. If provided, `sushi init` will not prompt you for a project name.
- `-c, --config`: similar to the new option for `sushi build`, this option lets you specify the value for any of the prompts in `sushi init`. Specifying a value with this option will no longer prompt you for that question and instead will just log the value you set.
- `-d, --default`: if provided, any values not specified value using `-c, --config` will just use the default value. No prompts will be asked, and instead, the default values are logged.
- `-a, --auto-initialize`: If provided, the project will automatically be initialized once all prompts are answered and/or all values are set using `-c` or `-d`. The y/n prompt for initializing is skipped.

I also simplified the `--help` text for `build` and `init` by moving the defaults for the `path-to-fsh-project` and `out` onto the same line as the description of that argument/option and removing the `.on('--help', ...)` portion. I'm not entirely sure why we didn't do that from the beginning (maybe it wasn't possible when we first wrote that?), but if there was a reason I forgot about, I can revert that change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Options:
 
 Commands:
   build [options] [path-to-fsh-project]                build a SUSHI project
-  init [options]                                       initialize a SUSHI project
+  init [options] [name]                                initialize a SUSHI project
   update-dependencies [options] [path-to-fsh-project]  update FHIR packages in project configuration
   help [command]                                       display help for command
 ```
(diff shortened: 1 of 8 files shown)


cmoesel (changes requested):
I like it! It works quite well!

I did make one silly mistake that I think others could make, so I'd like to see if we can somehow avoid it. I did this:
```
sushi init -da -c version:1.2.3 status:active Foo
```
This resulted in creating a project named `status:active` because I forgot the second `-c`. And I think the `Foo` argument just got dropped. So... if there is a way to detect the extra argument (that didn't have an option), then I think it might be nice to issue an error in that case. Maybe something like:
> Too many arguments: status:active Foo. Did you forget an option flag?

jafeltra:
> So... if there is a way to detect the extra argument (that didn't have an option), then I think it might be nice to issue an error in that case.

I learned that commander lets you make this scenario an error using `.allowExcessArguments(false)`. Since you could run into this same problem with the `-c` option on the `build` command, I went ahead and added it there as well.

Let me know what you think since the error it logs is a little less specific than your suggestion.

Note that this doesn't quite solve the problem if you were to try to use the prompt for the project name but make the same mistake -- e.g. `sushi init -c version:1.2.3 status:active`.

cmoesel (approved):
Looks good! Thanks, Julia!

PR #1462, opened 2024-05-20
Remove releaseLabel coded options from `sushi-config.yaml` created by `sushi init`

### LINKED ISSUE #1461: Update releaseLabel section in SUSHI init's generated config
We have recently learned that despite the FHIR documentation on the `releaselabel` parameter, there is no set of suggested values (see [zulip](https://chat.fhir.org/#narrow/stream/196008-ig-publishing-requirements/topic/releaseLabel)). In fact, some of these values are _not_ preferred for real projects (e.g., `normative+trial-use` is better stated as `Mixed Normative-Trial Use`).

We need to update the config file generated as part of `sushi init` so it doesn't list these values anymore.

jafeltra:
This PR removes the comment that is included when you generate a `sushi-config.yaml` file using `sushi init`. Because `releaselabel` has recently been clarified to be any string value, we no longer need to include coded options for this property. 

Best practices will be documented on FSH School. We haven't fully explained other basic properties, like `copyrightyYear` or `fhirVersion`, within the config file itself, so I opted to not explain the best practices for `releaseLabel` directly in the config file either. However, if anyone thinks it would be useful to still include a comment with some options of values, let me know.

Fixes #1461

### DIFF
--- a/src/utils/Processing.ts
+++ b/src/utils/Processing.ts
@@ -721,11 +721,6 @@ export async function init(
         const node = YAML.createNode(userValue);
         node.comment = ' draft | active | retired | unknown';
         configDoc.set(field, node);
-      } else if (field === 'releaseLabel') {
-        const node = YAML.createNode(userValue);
-        node.comment =
-          ' ci-build | draft | qa-preview | ballot | trial-use | release | update | normative+trial-use';
-        configDoc.set(field, node);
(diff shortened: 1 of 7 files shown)


cmoesel (approved):
Looks like you scrubbed it all. Thanks!

PR #1464, opened 2024-05-28
Warn when config does not include menu and there is no menu.xml

### LINKED ISSUE #1161: Warn when config doesn't include menu and there is no menu.xml
The IG Publisher requires a menu.xml file -- and will either crash or hang if it does not exist.

There are two ways users can provide a menu when using SUSHI:
1. Via the `menu` property in `sushi-config.yaml`
2. Via `input/includes/menu.xml`

If SUSHI detects that there is no `menu` property in `sushi-config.yaml` and no `menu.xml` in `input/includes`, then it should issue a _warning_ that no menu was provided.

_NOTE: This applies only to FSH IG projects; not FSHOnly projects._

KaelynJefferson:
Fixes #1161

This PR includes a check if there is a config.menu or menu.xml, and if not issues a warning to the user. Tests affected by this addition of a warning message were modified to include a config.menu.

### DIFF
--- a/test/ig/IGExporter.IG.test.ts
+++ b/test/ig/IGExporter.IG.test.ts
@@ -1913,6 +1913,10 @@ describe('IGExporter', () => {
       loggerSpy.reset();
       tempOut = temp.mkdirSync('sushi-test');
       config = cloneDeep(minimalConfig);
+      config.menu = [
+        { name: 'Home', url: 'index.html' },
+        { name: 'Artifacts', url: 'artifacts.html' }
+      ];
       config.resources = [
         {
           reference: { reference: 'Patient/BazPatient' },
@@ -2928,6 +2932,10 @@ describe('IGExporter', () => {
       loggerSpy.reset();
       tempOut = temp.mkdirSync('sushi-test');
       config = cloneDeep(minimalConfig);
+      config.menu = [
+        { name: 'Home', url: 'index.html' },
+        { name: 'Artifacts', url: 'artifacts.html' }
+      ];
       config.parameters = [];
       config.parameters.push({
         code: 'path-resource',
@@ -3020,6 +3028,10 @@ describe('IGExporter', () => {
 
     it('should not warn on deeply nested resources that are included in the path-resource parameter with a directory and wildcard', () => {
       const config = cloneDeep(minimalConfig);
+      config.menu = [
+        { name: 'Home', url: 'index.html' },
+        { name: 'Artifacts', url: 'artifacts.html' }
+      ];
       config.parameters = [];
       config.parameters.push({
         code: 'path-resource',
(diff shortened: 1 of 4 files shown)


test/ig/IGExporter.IG.test.ts
```diff
@@ -1913,6 +1913,10 @@ describe('IGExporter', () => {
       loggerSpy.reset();
       tempOut = temp.mkdirSync('sushi-test');
       config = cloneDeep(minimalConfig);
+      config.menu = [
+        { name: 'Home', url: 'index.html' },
+        { name: 'Artifacts', url: 'artifacts.html' }
+      ];
```
cmoesel, line 1919: Since you do this six times, consider making a `minimalConfigWithMenu` instead -- then just use that instead of `minimalConfig` in these places. That way you only need to specify the config menu in one place. It's a small thing, but I feel like it's a little cleaner that way.

test/ig/IGExporter.IG.test.ts
```diff
@@ -3347,6 +3371,10 @@ describe('IGExporter', () => {
           }
         ]
       };
+      config.menu = [
+        { name: 'Home', url: 'index.html' },
+        { name: 'Artifacts', url: 'artifacts.html' }
+      ];
```
cmoesel, line 3377: This is super nit-picky, but given that you define the whole config JSON in the lines just above this, I think it makes more sense just to stick the menu in that JSON object in the first place.

test/ig/IGExporter.menu-xml.test.ts
```diff
@@ -30,14 +30,15 @@ describe('IGExporter', () => {
...
+    it('should log a warning when config.menu is undefined and none provided', () => {
       const pkg = new Package(minimalConfig);
       const igDataPath = path.resolve(__dirname, 'fixtures', 'simple-ig');
       const exporter = new IGExporter(pkg, null, igDataPath);
       exporter.addMenuXML(tempOut);
       const menuPath = path.join(tempOut, 'input', 'includes', 'menu.xml');
       expect(fs.existsSync(menuPath)).toBeFalsy();
-      expect(loggerSpy.getAllMessages()).toHaveLength(0);
+      expect(loggerSpy.getAllMessages('warn')).toHaveLength(1);
+      expect(loggerSpy.getLastMessage()).toMatch(/No "menu" property or file was found\./s);
```
cmoesel, line 41: Since you include some variables in the message (input folder and config file name), I think it would be good to make sure those are coming through correctly too. I think you can just add something like `.*input[/\]includes folder.* or .* in sushi-config\.yaml` to the regex.

cmoesel (changes requested):
Thank you for this first contribution, @KaelynJefferson! This looks great! I've reviewed and left a few small (nitpicky) comments and suggestions. There is nothing here that is _wrong_; just a few things that perhaps can be ever-so-slightly improved (and may just be related to my own preferences). Please take a look at my comments and either implement the suggestions or reply if you disagree with anything!

jafeltra (approved):
I think this all makes sense! I didn't notice anything else and your changes to address Chris's comments looked good to me.

Once Chris re-reviews and approves, we can merge this in!

cmoesel (approved):
This looks great and works well. Great first contribution, @KaelynJefferson! Thank you!

Issue #1128, opened 2022-08-10
Have sushiClient create definitions with snapshot


nico-ploner:
Hello sushi-team,

would it be possible for the sushiClient's api to create json resources including their snapshots?
The cli offers the switch --snapshot but the fshToFhir method's options are lacking this property and line 81 of FshToFhir.ts hard codes the snapshot value to 'false'.

Any help is highly appreciated.
Nico

cmoesel:
Hi @nico-ploner.  We can look into this, but it's worth noting that SUSHI's snapshot generation is likely not perfect and differs from the snapshots that the IG Publisher and/or Simplifier would create.  If you plan to publish these resources, it would probably be better to use one of those other tools to generate the snapshots.

nico-ploner:
Hey @cmoesel. Thank you for your response. I'll bear it in mind when creating profiles using SUSHI.
I'd be happy to see SUSHI's client provide this feature some day which allows us to directly use its ouput in code generation tools.

cmoesel:
Ah, yes, I see.  SUSHI's snapshots would probably work fine for code generation.  We'll add this feature to our backlog for consideration in future sprints.

joepaquette:
Ditto @nico-ploner... It would be very good to get rid of the IG Publisher in our workflow just to get `snapshot`  for our code generators!

jafeltra:
This was added in PR #1465. Thanks @joepaquette!

PR #1465, opened 2024-05-30
Issue #1128 - add snapshot option to FshToFhir

joepaquette:
Issue #1128 - Enable sushiClient (FshToFhir) to create StructureDefinitions with the snapshot data element

### DIFF
--- a/src/run/FshToFhir.ts
+++ b/src/run/FshToFhir.ts
@@ -48,6 +48,8 @@ export async function fshToFhir(
     logger.level = options.logLevel;
   }
 
+  const snapshot = options.snapshot ?? false;
+
   // set up a config so that sushi can run
   const config = {
     canonical: options.canonical ?? 'http://example.org',
@@ -87,7 +89,7 @@ export async function fshToFhir(
     ] as const
   ).forEach(artifactType => {
     outPackage[artifactType].forEach((artifact: { toJSON: (snapshot: boolean) => any }) => {
-      fhir.push(artifact.toJSON(false));
+      fhir.push(artifact.toJSON(snapshot));
     });
   });
 
@@ -104,6 +106,7 @@ type fshToFhirOptions = {
   fhirVersion?: string;
   dependencies?: ImplementationGuideDependsOn[];
   logLevel?: Level;
+  snapshot?: boolean;
 };
 
 // Winston levels: https://github.com/winstonjs/winston#logging-levels plus a silent option
--- a/src/run/FshToFhir.ts
+++ b/src/run/FshToFhir.ts
@@ -100,6 +100,11 @@ export async function fshToFhir(
   };
 }
 
+// *** WARNING ***
+// The 'snapshot' option, when set to true, triggers the generation of StructureDefinition.snapshot data elements.
+// Use of this option should be considered EXPERIMENTAL! The StructureDefinition.snapshot data elements generated
+// by SUSHI are likely not perfect and differ from the snapshots that the IG Publisher and/or Simplifier would create.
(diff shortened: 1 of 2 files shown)


test/run/FshToFhir.test.ts
```diff
@@ -163,6 +163,7 @@ describe('#FshToFhir', () => {
       expect(results.fhir[0].id).toBe('MyPatient');
       const name = results.fhir[0].differential.element.find((e: any) => e.id == 'Patient.name');
       expect(name.mustSupport).toBe(true);
+      expect(results.fhir[0].snapshot).not.toBeDefined();
```
cmoesel, line 166: This is nitpicky, and this code does not need to be changed, but for future consideration, Jest does support [expect(thing).toBeUndefined()](https://jestjs.io/docs/expect#tobeundefined).

src/run/FshToFhir.ts
```diff
@@ -104,6 +106,7 @@ type fshToFhirOptions = {
   fhirVersion?: string;
   dependencies?: ImplementationGuideDependsOn[];
   logLevel?: Level;
+  snapshot?: boolean;
```
cmoesel, line 109: Do you think it might be worth putting a comment here that warns that SUSHI snapshot generation is experimental (or some word like that) and that snapshots may differ from what the IG Publisher would generate? I just worry because snapshots are "good enough" for our own purposes within SUSHI (and likely for some other use cases), but they're not really a focus of the tool -- so I think maybe people should know that?

cmoesel (commented):
Thank you, Joe! I've left one question I'd like your opinion about.

src/run/FshToFhir.ts
cmoesel: I'm noticing now that we don't say anything about this in any of our documentation elsewhere (like the FSH School documentation) -- so maybe I'm being overly cautious. Feel free to push back on this if you'd like.

test/run/FshToFhir.test.ts
joepaquette: Somehow, I've missed that! I learn something new everyday.

src/run/FshToFhir.ts
joepaquette: Valid points! I'll add a comment to that affect. I'll also add a comment in the `snapshot` parameter in the FSH School doc update. After I resubmit bot PRs, let me know if you want me to do any additional word-smithing.

cmoesel (approved):
Thanks, Joe! This looks great. I did a manual test with the following code and confirmed it is working as expected.

```ts
import { sushiClient } from '.';

// Example usage with options
sushiClient
  .fshToFhir(
    `
Profile: MyPatient
Parent: Patient
* name 1..* MS

Profile: MyCondition
Parent: Condition
* subject only Reference(Patient)
`,
    {
      canonical: 'http://example.com',
      version: '1.2.3',
      fhirVersion: '4.0.1',
      snapshot: true
    }
  )
  .then(results => {
    console.log('FHIR', JSON.stringify(results.fhir, null, 2));
    console.log('WARNINGS', JSON.stringify(results.warnings, null, 2));
    console.log('ERRORS', JSON.stringify(results.errors, null, 2));
  })
  .catch(err => {
    console.error('Woah, Nellie!', err);
  });
```

As you know, we require two code reviews in order to merge, so I'll have someone from the team take a look next week. In the mean time, perhaps you can click the `Update Branch` button to sync this up with the master branch?

jafeltra (approved):
Hi @joepaquette! This PR looks good to me as well. I tested it out and looked through the code, and it all looks good to me!

Thanks for addressing this long-standing issue!

PR #1466, opened 2024-05-30
logical/resource root elements set default to prevent empty/missing definition

### LINKED ISSUE #1159: Logical/Resource root element cannot have empty/missing definition
For `Logical` and `Resource` entities only:

If the user does not provide a caret rule to set the root element's `short`/`definition`, the SUSHI will default the `short` to the specified `Title` and the `definition` to the specified `Description`.  In this case, if `Title` is not specified, then the root element will not have a `short`; and likewise, if `Description` is not specified, then the root element will not have a `definition`.

The problem is that [sdf-3](http://hl7.org/fhir/R4/structuredefinition.html#invs) states that _every_ element must have a `definition` -- so if SUSHI leaves it out, then the IG Publisher complains (rightfully so).  To avoid this, SUSHI should default the `short` and `definition` to the logical/resource _name_ when there otherwise would be no other value for them.

So...

Root element `short`:
* Use `^short` value if specified;
* Otherwise, use `Title:` value if specified;
* Otherwise, use name (specified after `Logical:` or `Resource:`)

Root element `definition`:
* Use `^definition` value if specified;
* Otherwise, use `Description:` value if specified;
* Otherwise, use the `short` value (following logic above, if necessary)

KaelynJefferson:
Fixes #1159 

This PR includes:
- Logical/Resource root element `short` and `definition` values were set by logic described in 1159 and prevents the case of empty/missing `definition`.
- Affected tests were modified.

### DIFF
--- a/src/export/StructureDefinitionExporter.ts
+++ b/src/export/StructureDefinitionExporter.ts
@@ -707,9 +707,13 @@ export class StructureDefinitionExporter implements Fishable {
     // Reset the root element's short and definition
     if (fshDefinition.title) {
       structDef.elements[0].short = fshDefinition.title;
+    } else {
+      structDef.elements[0].short = fshDefinition.name;
     }
     if (fshDefinition.description) {
       structDef.elements[0].definition = fshDefinition.description;
+    } else {
+      structDef.elements[0].definition = structDef.elements[0].short;
     }
   }
 
--- a/src/export/StructureDefinitionExporter.ts
+++ b/src/export/StructureDefinitionExporter.ts
@@ -705,16 +705,8 @@ export class StructureDefinitionExporter implements Fishable {
     // differential root element.
 
     // Reset the root element's short and definition
-    if (fshDefinition.title) {
-      structDef.elements[0].short = fshDefinition.title;
-    } else {
-      structDef.elements[0].short = fshDefinition.name;
-    }
-    if (fshDefinition.description) {
-      structDef.elements[0].definition = fshDefinition.description;
-    } else {
-      structDef.elements[0].definition = structDef.elements[0].short;
-    }
+    structDef.elements[0].short = fshDefinition.title ?? fshDefinition.name;
+    structDef.elements[0].definition = fshDefinition.description ?? structDef.elements[0].short;
   }
 
   /**
--- a/test/export/StructureDefinition.LogicalExporter.test.ts
+++ b/test/export/StructureDefinition.LogicalExporter.test.ts
@@ -861,6 +861,48 @@ describe('LogicalExporter', () => {
     expect(loggerSpy.getAllMessages('error')).toHaveLength(0);
   });
 
+  it('should create Logical root element with short equal to title if short not available AND definition equal to description if definition not available', () => {
+    const logical = new Logical('MyTestModel');
+    logical.id = 'MyModel';
+    logical.title = 'MyTestModel title is here';
+    logical.description = 'MyTestModel description is here';
+
+    doc.logicals.set(logical.name, logical);
+    exporter.exportStructDef(logical);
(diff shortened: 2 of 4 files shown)


src/export/StructureDefinitionExporter.ts
```diff
@@ -707,9 +707,13 @@ export class StructureDefinitionExporter implements Fishable {
...
     if (fshDefinition.title) {
       structDef.elements[0].short = fshDefinition.title;
+    } else {
+      structDef.elements[0].short = fshDefinition.name;
     }
     if (fshDefinition.description) {
       structDef.elements[0].definition = fshDefinition.description;
+    } else {
+      structDef.elements[0].definition = structDef.elements[0].short;
     }
```
cmoesel, line 717: This is perfectly good. But here is a neat TypeScript trick you could use instead, if you want:
```ts
structDef.elements[0].short = fshDefinition.title ?? fshDefinition.name;
structDef.elements[0].definition = fshDefinition.description ?? structDef.elements[0].short;
```
This also works in later versions of JavaScript. The `??` is called the [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing). Basically, if the thing to the left of the `??` is defined and not null, then that's the value returned; but if the thing to the left of the `??` is undefined or null, then it returns the thing to the right of the `??` instead. In other words, `A ?? B` is kind of like `if (A != null) return A else return B`.

test/export/StructureDefinition.LogicalExporter.test.ts
```diff
@@ -861,6 +861,48 @@ describe('LogicalExporter', () => {
...
+
+    doc.logicals.set(logical.name, logical);
+    exporter.exportStructDef(logical);
+    const exported = pkg.logicals[0];
+
+    expect(exported.elements).toHaveLength(1);
+    expect(exported.elements[0].short).toBe(logical.title);
+    expect(exported.elements[0].definition).toBe(exported.elements[0].short);
+  });
+
```
cmoesel, line 905: Are there any tests that ensure that if the user defined a rule to set the short and definition that the values from the rule take precedence? E.g., Given:
```
Logical: MyTestModel
Id: MyModel
Title: "MyTestModel title is here"
Description: "MyTestModel description is here"
* . ^short = "A short set by a rule"
* . ^definition = "A definition set by a rule"
```
Then the root element's short and definition should be the ones from the rules, not the title and description. If we already have a test for this, great! If we don't, can you add one?

test/export/StructureDefinition.ResourceExporter.test.ts
```diff
@@ -506,4 +506,43 @@ describe('ResourceExporter', () => {
...
+    resource.id = 'MyModel';
+    resource.title = 'MyTestModel title is here';
+
+    doc.resources.set(resource.name, resource);
+    exporter.exportStructDef(resource);
+    const exported = pkg.resources[0];
+
+    expect(exported.elements[0].short).toBe(resource.title);
+    expect(exported.elements[0].definition).toBe(exported.elements[0].short);
+  });
```
cmoesel, line 547: Same question as above re: setting the root element's short and definition using rules.

cmoesel (commented):
Good work! I left a few comments for your consideration. They may or may not require changes. I'm not sure!

cmoesel (approved):
Well done, Kaelyn. This looks great. 

I am approving -- but I do have one request: Can you edit this PR's title so it is short enough to fit? And edit the description too since it starts with `...`? If you need help w/ figuring out a good (but short) descriptive PR title, feel free to start a conversation w/ the team on Slack.

mint-thompson (approved):
Looks good! I think it'll be ready to merge once you update the title.

Issue #968, opened 2021-11-29
Sushi should check referential integrity of references


saulakravitz:
If a say Reference(x),  x should exist.
Currently, if it doesn't, sushi is happy, and the IG publisher throws an error.   
Annoying.

cmoesel:
We intentionally allow for unresolvable references so that authors can create simple examples without having to author a full web of instances.  For example, If you only want to show how a CarePlan works, why should you also have to create a completely valid Patient just to satisfy the reference?  Some resources are even worse and would require 3 or 4 additional instances just to have "valid" references.

I don't think that IG Publisher always threw an error for unresolvable references (although I may be remembering wrong).   If it's doing that now, perhaps we should revisit this.

saulakravitz:
The IG publisher error is for a broken link.
At the risk of suggesting more and more options, a mode where unresolved references are errors would be helpful.

jafeltra:
#1468 was recently added and would log a warning in the example above. See #1155 for other similar, but slightly different, reference cases.

PR #1468, opened 2024-06-05
Warn when reference values do not resolve and have incorrect format

KaelynJefferson:
This PR includes:
- Issues a warning when the user provides reference values to entities that do not resolve and have incorrect FHIR Reference format
- Tests affected were modified


mint-thompson (changes requested):
I think there may need to be some extra logic to test if the resulting value conforms to the allowed format for a FHIR Reference. Values that look like `Type/Id`, such as `Patient/MyPatient`, are allowed. URLs are also allowed. So, these values should not result in a message that says that the reference does not conform to the FHIR Reference format. There are more details in the [documentation for Reference](https://hl7.org/fhir/R5/references.html#Reference).

src/fhirtypes/common.ts
```diff
@@ -980,6 +980,34 @@ export function replaceReferences<T extends AssignmentRule | CaretValueRule>(
...
+        if (
+          value.reference.includes('/') &&
+          value.reference.indexOf('/') !== value.reference.lastIndexOf('/')
+        ) {
+          // ensure it is not an absolute url
+          if (
+            value.reference.indexOf('http://') == -1 &&
+            value.reference.indexOf('https://') == -1 &&
+            value.reference.indexOf('www.') == -1
+          ) {
```
mint-thompson, line 1004: For checking if the value of the reference is an absolute url, you should be able to use the `valid-url` package. It is used in a few places in SUSHI for this purpose.

src/fhirtypes/common.ts
```diff
@@ -980,6 +980,34 @@ export function replaceReferences<T extends AssignmentRule | CaretValueRule>(
...
+            `Cannot find the entity referenced at ${value.reference}. The provided reference value will be used, however, this reference does not conform to the FHIR Reference() format.`
+          );
+        }
+
+        // if reference is a relative reference with too many parts
+        // example: PatientExample/Patient/Patient-oops
+        if (
+          value.reference.includes('/') &&
+          value.reference.indexOf('/') !== value.reference.lastIndexOf('/')
+        ) {
```
mint-thompson, line 998: Based on an example in the FHIR documentation, I think that a value with more than one `/` might actually be acceptable? The last example in the [Literal References](https://hl7.org/fhir/R5/references.html#literal) section shows `Observation/1x2/_history/2` as a version-specific reference value. So maybe, only a value without _any_ `/` would be one that we warn about? I'd like to get opinions from the rest of the team on this.

mint-thompson (changes requested):
Thank you for the updates. I have a few more changes to request: turns out the Reference type can get a little complicated!

mint-thompson (approved):
This looks fine to me! Since so many formats are allowed, we don't even need to check if something is a URL (since those will contain a `/`, and therefore not emit a warning). Thank you for making all the requested updates!

jafeltra (changes requested):
This is looking good. I found one case where we want to avoid logging this warning message and left a few other nitpicky comments. Let me know if you disagree with any of the nitpicky ones.

src/fhirtypes/common.ts
```diff
@@ -979,6 +979,16 @@ export function replaceReferences<T extends AssignmentRule | CaretValueRule>(
         clone = cloneDeep(rule);
         (clone.value as FshReference).reference = value.reference.slice(0, firstPipe);
         clone = replaceReferences(clone, tank, fisher);
+      } else if (type == null && id == null && value.reference) {
+        // if the reference to an entity is provided but is unable to be resolved
+        if (!value.reference.startsWith('urn:')) {
+          if (!value.reference.includes('/')) {
```
jafeltra, line 985: The way this logic is written with the nested if statements makes me think there's something else we want to do if the value starts with `urn:` but does contain a `/`. I think it would be simpler to just have one if statement with both conditions combined with "and" (i.e. `!value.reference.startsWith('urn:') && !value.reference.includes('/')`). While that doesn't change the logic as it currently is, I think it makes the intent a little clearer.

jafeltra (changes requested):
This looks good! Thanks for adding that new case and adding a test! I just had one small comment inline.

There are still 24 repos that get new warnings, but from looking through the warning messages, it looks like they're all relevant warnings, so I think that's okay. In case anyone else is curious, this is the list of repos that I see with warnings:

```
davidhay25/models#main
hl7-be/referral#master
ansforge/IG-fhir-telesurveillance#main
BIH-CEI/interagent#main
cander2/epi-test#master
CEOsys/cpg-on-ebm-on-fhir#master
hl7-be/lab#master
hl7-eu/gravitate-health#master
HL7/codex-radiation-therapy#master
HL7/cqf-recommendations#master
HL7/ReportIIITraveler-ig#master
HL7NZ/nes-ig#master
IHE/PCC.mAPS#master
IHE/PCC.mAPS.us#master
InstitutNationalduCancer/ImplementationGuide_OsirisFHIR#master
Interop-Sante/hl7.fhir.fr.core#main
joofio/ufis-ig#main
joofio/ufis-swe-ig#main
medizininformatik-initiative/kerndatensatz-bildgebung#main
medizininformatik-initiative/kerndatensatzmodul-medikation#master
medizininformatik-initiative/kerndatensatzmodul-onkologie#dev
servicewell/wof-portal-fhir-ig#master
swantrace/ASU-research-consent-on-FHIR#main
tewhatuora/fhir-rheumatic-fever#master
```

src/fhirtypes/common.ts
cmoesel: Or there is, of course, the equivalent inverse expression: `!(value.reference.startsWith('urn:') || value.reference.includes('/'))`. Take your pick!

cmoesel: Oh right. That's his name!

jafeltra:
I also noticed we still have #968 open. I think this PR addresses that issue, and we should be able to close it when this gets merged.

jafeltra (approved):
Looks great! Thanks for making those updates!

Issue #1492, opened 2024-07-16
Error when adding a caret rule to a versioned code in a ValueSet


sebg-mio42:
Adding a caret rule like a designation to a versioned code in a ValueSet currently causes an error.

For example this FSH ValueSet with a designation for a specific version of SNOMED:

```
ValueSet: Test
Id: test
Title: "Test"
Description: "test"
* http://snomed.info/sct#182869005 "Drug not taken - patient lost tablets (situation)"
  * ^designation[0].language = #de
  * ^designation[0].value = "Medikament nicht eingenommen wegen Verlust des Arzneimittels"
* http://snomed.info/sct|http://snomed.info/sct/11000274103/version/20240515#182879007 "Medication decreased (finding)"
  * ^designation[0].language = #de
  * ^designation[0].value = "Dosisreduktion" 
```

This causes the following errors:

```
error Could not find concept http://snomed.info/sct|http://snomed.info/sct/11000274103/version/20240515#182879007, skipping rule.
  File: ...\input\fsh\vs_test.fsh
  Line: 9
error Could not find concept http://snomed.info/sct|http://snomed.info/sct/11000274103/version/20240515#182879007, skipping rule.
  File: ...\input\fsh\vs_test.fsh
  Line: 10
```
and the output JSON does not include the designation:

```
{
  "resourceType": "ValueSet",
  "status": "draft",
  "name": "Test",
  "id": "test",
  "title": "Test",
  "description": "test",
  "url": "http://example.org/ValueSet/test",
  "compose": {
    "include": [
      {
        "system": "http://snomed.info/sct",
        "concept": [
          {
            "code": "182869005",
            "display": "Drug not taken - patient lost tablets (situation)",
            "designation": [
              {
                "language": "de",
                "value": "Medikament nicht eingenommen wegen Verlust des Arzneimittels"
              }
            ]
          }
        ]
      },
      {
        "system": "http://snomed.info/sct",
        "version": "http://snomed.info/sct/11000274103/version/20240515",
        "concept": [
          {
            "code": "182879007",
            "display": "Medication decreased (finding)"
          }
        ]
      }
    ]
  }
}
```
Expected output would be no errors and the following JSON:

```
{
  "resourceType": "ValueSet",
  "status": "draft",
  "name": "Test",
  "id": "test",
  "title": "Test",
  "description": "test",
  "url": "http://example.org/ValueSet/test",
  "compose": {
    "include": [
      {
        "system": "http://snomed.info/sct",
[...]

PR #1493, opened 2024-07-16
Valueset versioned caret rules

sebg-mio42:
**Description:**
added the version to the check in ValueSetExporter and added a new test case

**Related Issue:**
Fixes https://github.com/FHIR/sushi/issues/1492


cmoesel (approved):
Thank you for this excellent contribution, @sebg-mio42. I've reviewed the code and it looks great to me. I've also run a regression test on the 20 most recently updated FSH projects and that came out clean (as expected).

We typically look for _two_ reviews and approvals before merging, so I will wait for one more review before merging this code. Thanks again!

mint-thompson (approved):
Looks good, thank you!

PR #1496, opened 2024-07-22
Upgrade ESLint and Prettier

cmoesel:
**Description:** Upgrade ESLint and Prettier to their latest versions. ESLint 9.x is a major update which requires a new configuration file format as well as using an 8.0.0 alpha version of @typescript-eslint. This upgrade resulted in new linting and prettier violations being reported, so this PR also fixes those new violations (several of which were real bugs).

**Testing Instructions:** Run `npm run check` to ensure that lint, prettier, and tests all pass.

**Related Issue:** N/A


mint-thompson (approved):
Everything works fine on my system. Glad that this caught some existing problems!

jafeltra (approved):
Whoa, `FlatCompat` is cool! And very cool (and mildly concerning) there were so many bugs. But this looks good to me now!

mint-thompson (approved):

cmoesel:
> Whoa, `FlatCompat` is cool! And very cool (and mildly concerning) there were so many bugs. But this looks good to me now!

Yeah. I got that for free. For future reference, if you need to upgrade to ESLint 9, this is the easiest path: `npx @eslint/migrate-config .eslintrc`!

Issue #1215, opened 2023-02-17
SUSHI parses round decimals as integers


florianschoffke:
When I create instances on profiles that contain decimals then those do not get converted accordingly when they are rounded.

For instance, if I specify an extension to have a .valueDecimal and then assign an instance
.extension.valueDecimal = 19.00 
then the .json output is: 
{ "url": "http://url.com/test",
 "valueDecimal": 19 },

However the HAPI Validator complains that the field is not a decimal. Is there a way to state that it needs to be exactly 19.00?

If I assign 19.01 then SUSHI converts correctly to "valueDecimal": 19.01

cmoesel:
See [Precision is lost using Sushi and or GoFSH](https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Precision.20is.20lost.20using.20Sushi.20and.20or.20GoFSH)](https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Precision.20is.20lost.20using.20Sushi.20and.20or.20GoFSH/near/363052371) on Zulip for more discussion.

PR #1497, opened 2024-07-26
Maintaining trailing zeros for decimal precision in json output

florianschoffke:
**Description:**
This PR attempts to address the issue of trailing zeros for decimal numbers which are omitted when using Typescript and JSON  tooling.
The main idea for solving this issue are twofold:
1. In case of decimals the rawValue as string is being used for assignment
2. When outputting the JSON content a regex replaces Numbers with trailing zeros that are strings to exclude the strings

There was no 'elegant' way that I could see to solve this issue as numbers in Typescript with trailing zeros are represented as plain numbers and fs.outputJSONsync transforms the number 19.00 to 19 as well.

Would be interested if we could work with that solution. If I missed other side effects please let me know.

**Testing Instructions:**
Run Sushi with this FSH File:
```
Instance: ObservationExample
InstanceOf: Observation
Usage: #example
* id = "example-xxx"
* extension[+]
  * url = "decimal-test-round"
  * valueDecimal = 19.00
* extension[+]
  * url = "decimal-test-off"
  * valueDecimal = 19.01
* extension[+]
  * url = "money-test-round"
  * valueMoney.value = 19.00
* extension[+]
  * url = "money-test-off"
  * valueMoney.value = 19.01
* status = #final
* code = http://loinc.org#11557-6
```

Before the commit the output would be ` "valueDecimal": 19`
Afterwards it is ` "valueDecimal": 19.00`

**Related Issue:**
#1215

### DIFF
--- a/src/export/InstanceExporter.ts
+++ b/src/export/InstanceExporter.ts
@@ -178,13 +178,15 @@ export class InstanceExporter implements Fishable {
             value = `#${matchingContainedReferenceId}`;
           }
         }
+        let rawValue = (rule instanceof AssignmentRule) ? rule.rawValue : null
         const validatedRule = instanceOfStructureDefinition.validateValueAtPath(
           rule.path,
           value,
           this.fisher,
           inlineResourceTypes,
           rule.sourceInfo,
-          manualSliceOrdering
+          manualSliceOrdering,
+          rawValue
         );
         // Record each valid rule in a map
         // Choice elements on an instance must use a specific type, so if the path still has an unchosen choice element,
--- a/src/fhirtypes/ElementDefinition.ts
+++ b/src/fhirtypes/ElementDefinition.ts
@@ -1830,7 +1830,7 @@ export class ElementDefinition {
    * @throws {ValueAlreadyAssignedError} when the value is already assigned to a different value
    * @throws {MismatchedTypeError} when the value does not match the type of the ElementDefinition
    */
-  assignValue(value: AssignmentValueType, exactly = false, fisher?: Fishable): void {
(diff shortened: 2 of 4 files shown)


mint-thompson:
Unfortunately, I think this implementation has some undesirable side effects. Any string which can be turned into a number will end up as a number in the output JSON, regardless of the FHIR field's type. If you can check the FHIR field's type before making this conversion, and only change it when it is an appropriate type, then it would be good. I'm not sure how feasible it is to make that check during `writeFHIRResources`, though, since that is very late in the pipeline.

This FSH demonstrates a few cases where a string looks enough like a number that it gets turned into a number. The output JSON will have numeric values for all three `valueString` elements.
```
Instance: ObservationExample
InstanceOf: Observation
Usage: #example
* id = "example-xxx"
* extension[+]
  * url = "leading-zero-string"
  * valueString = "003413"
* extension[+]
  * url = "exponential-string"
  * valueString = "12E9"
* extension[+]
  * url = "hexadecimal-string"
  * valueString = "0X3A"
* status = #final
* code = http://loinc.org#11557-6
```

cmoesel:
Hi @florianschoffke. Thanks again for the contribution! We really appreciate it! Preserving trailing 0s is definitely a pain point that JavaScript/JSON does not make easy. It would be great to have a solution for this. As Mint said, however, we need to be sure we don't break anything else in the process. Have you had a chance to think about the issues that Mint has raised? Do you plan to attempt to address them?

florianschoffke:
First off thanks for considering!
As of now I can't be sure that I will be able to solve it in the near future. But will have a look at the issues raised.

cmoesel:
I am closing this PR as it is nearly 2 years old. Feel free to re-open it if you have updates.

PR #1498, opened 2024-07-31
Minor Code Cleanup

cmoesel:
**Description:** This PR contains minor code cleanup based on things I noticed while preparing the "Developing SUSHI" knowledge sharing session. These include:
* Simplified the path setting logic now that we know we'll never need choice type renaming
* Removed an outdated code comment about choice type renaming
* Removed comment about copying IG scripts (we download them now)
* Updated the `package.json` file list to remove files that no longer exist
* Updated the ANTLR README to refer to _all_ of the grammar files

**Testing Instructions:** Basic tests to ensure nothing breaks. This passed a regression with a full-year lookback.

### DIFF
--- a/antlr/README.md
+++ b/antlr/README.md
@@ -2,9 +2,12 @@
 
 The FSH Grammar is defined using [ANTLR4](https://www.antlr.org/).
 
-To edit the grammar, edit the file:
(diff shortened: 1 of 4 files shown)


jafeltra (approved):
Makes sense to me!

PR #1500, opened 2024-08-11
Add optional argument for making clone of predefined resources

mint-thompson:
This PR applies to task [CIMPL-1250](https://standardhealthrecord.atlassian.net/browse/CIMPL-1250).

The FHIRDefinitions class provides a public method to return all predefined resources. Hypothetically, the caller of this method could alter the predefined resources, which is why it could be useful to return a clone of these resources. However, in all places this method is called by SUSHI, the resources are used for a search, not for any modifications. By not cloning the resources for the search, significant time savings are possible. Just in case someone else is using SUSHI as a dependency and calling this method and expecting to receive a clone, the default behavior is to return a clone.

When searching for slices, the slice of an element is always a sibling of that element. Therefore, it is preferable to search the smaller list containing only the element's parent's children, when possible. This takes advantage of the underlying ElementDefinition tree. Because the root element's parent is undefined, use the full element list when working with the root element. This shouldn't be possible in most cases, as revealed by the test coverage, but the implementation is defensive just in case something unusual has happened to the root element.

A regression test of public FSH projects updated within the last year indicated no changes. The initial test run indicated changes in two repos, but both of those repos produced identical results in a second run without any code changes. The changes to `allPredefinedResources` are most visible in projects with many StructureDefinitions. The changes to `ElementDefinition` are most visible in projects with many Instance definitions. The impact of the `ElementDefinition` changes is relatively small, but the `allPredefinedResources` changes are significant for some FSH projects.

The regression test results summary is attached as html. Three projects showed increased run times, but all of those projects are relatively small, with the longest run time being 61 seconds. Some larger projects showed meaningful decreases in run time. As a few examples:
* [fhir-fi/finnish-base-profiles#master](https://github.com/fhir-fi/finnish-base-profiles/tree/master/) decreased from 224 seconds to 15 seconds
* [hl7-it/lab-report#master](https://github.com/hl7-it/lab-report/tree/master/) decreased from 865 seconds to 31 seconds
* [HL7/US-Core#master](https://github.com/HL7/US-Core/tree/master/) decreased from 187 seconds to 99 seconds

While these are just a few examples I've picked out, and not every project has changes that are so dramatic, I think the overall effect is meaningfully positive without adding an odious amount of complexity to the implementation.

[Regression summary HTML](https://github.com/user-attachments/files/16576371/index.html.txt)

### DIFF
--- a/src/fhirdefs/FHIRDefinitions.ts
+++ b/src/fhirdefs/FHIRDefinitions.ts
@@ -35,8 +35,12 @@ export class FHIRDefinitions extends BaseFHIRDefinitions implements Fishable {
     return flatten(Array.from(this.supplementalFHIRDefinitions.keys()));
   }
 
-  allPredefinedResources(): any[] {
-    return Array.from(this.predefinedResources.values()).map(v => cloneDeep(v));
+  allPredefinedResources(makeClone = true): any[] {
+    if (makeClone) {
+      return Array.from(this.predefinedResources.values()).map(v => cloneDeep(v));
+    } else {
+      return Array.from(this.predefinedResources.values());
+    }
   }
 
   add(definition: any): void {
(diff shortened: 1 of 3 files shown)


jafeltra (approved):
These changes all make sense to me! I looked through the regression from the last year, and nothing looks changed and the changes in time are honestly so impressive.

Good finds! Hopefully this helps people who were still struggling with performance!

cmoesel (approved):
Wow. What a find! Well done!

PR #1503, opened 2024-08-27
ValueSet self reference

### LINKED ISSUE #1487: Log error for ValueSet self-reference
Transferred from internal JIRA issue originally created 2024-03-06.

A component rule on a ValueSet should not reference itself. If this happens, log an error and skip the rule.

```
Alias: SCT = http://snomed.info/sct

Profile: MyServiceReqeust
Parent: ServiceRequest
Title: "Radiotherapy Plan Prescription"
Description: "A Radiotherapy Plan Prescription is a request for radiotherapy treatment with a single treatment plan."
* category from MaxVS (extensible)

ValueSet: MaxVS
Title: "Category Code Max Value Set"
Description: "Category value set required to support backward compatibility by including an inactive SNOMED code"
* ^experimental = false
* include codes from valueset MaxVS
* SCT#108290001 "Radiation oncology AND/OR radiotherapy (procedure)" // this code is inactive  
```
In the above FSH, the rule `* include codes from valueset MaxVS` should cause an error and be skipped.

KaelynJefferson:
**Description:** This PR addresses when a component rule on a ValueSet attempts to self-reference by id, ValueSet name, or url. For this case, it logs an error and removes the self reference.

**Testing Instructions:** Verify with unit tests.

**Related Issue:** Fixes #1487


src/export/ValueSetExporter.ts
```diff
@@ -86,6 +86,14 @@ export class ValueSetExporter {
...
             return this.fisher.fishForMetadata(vs, Type.ValueSet)?.url ?? vs;
           });
+          composeElement.valueSet = composeElement.valueSet.filter(vs => {
+            if (vs == valueSet.url) {
+              logger.error(
+                `Value set with id ${valueSet.id} has component rule with self referencing value set (by id, value set name, or url). Skipping rule.`
+              );
+            }
+            return vs != valueSet.url;
+          });
```
jafeltra, line 96: This is a question of what we want the desired output to be. What you have here might be totally correct, but I just want to double check.

Now that we're filtering the `composeElement.valueSet`, there's a chance that we're left with no items in `composeElement.valueSet`. Then on line 187, we push the `composeElement` object onto `valueset.compose.include` no matter what the value of `composeElement` is. Since we could have filtered out all the value sets, we might be pushing on a compose element that doesn't have any value sets, which then gets converted to a `null` value within an array when we create the FHIR resources. Since I might have explained that poorly, here's an example of what I'm trying to say:

If you have the following value set (and another value set defined for `OtherVS`)

```
ValueSet: MaxVS
Id: max-vs
* include codes from valueset MaxVS
* include codes from valueset OtherVS
```
the first rule will filter out the `MaxVS` value set because it is self-reference. That then results in  the following `compose` element to be created:

```
"compose": {
  "include": [
    null,
    {
      "valueSet": [
        "http://example.org/ValueSet/other-vs"
      ]
    }
  ]
}
```
Since we're skipping the rule and telling authors about it, I wasn't expecting a `null` entry where the filtered value set was removed. Is that what we were expecting?

src/export/ValueSetExporter.ts
```diff
@@ -86,6 +86,14 @@ export class ValueSetExporter {
           composeElement.valueSet = component.from.valueSets.map(vs => {
             return this.fisher.fishForMetadata(vs, Type.ValueSet)?.url ?? vs;
           });
+          composeElement.valueSet = composeElement.valueSet.filter(vs => {
+            if (vs == valueSet.url) {
+              logger.error(
+                `Value set with id ${valueSet.id} has component rule with self referencing value set (by id, value set name, or url). Skipping rule.`
+              );
```
jafeltra, line 93: I'm not sure if this error message is completely accurate. If you have a rule on a ValueSet like this:

```
ValueSet: MaxVS
Id: max-vs
* include codes from valueset MaxVS and OtherVS
```
the filter will filter out the `MaxVS` reference, but _keep_ the OtherVS reference. So the rule itself isn't skipped, we're only skipping _part_ of the rule.

That said, I actually am not sure what we want to do in this scenario. If someone did use this rule, I'm not sure if we want to just filter out `MaxVS` because if they asked for the intersection of codes between MaxVS and OtherVS and we end up giving them just all the codes in OtherVS, is that the behavior we want? If it is, we should update the error message to just say "skipping this value set" rather than "skipping rule". If it isn't and we decide that we're making it so SUSHI is giving back something unexpected and not what an author is trying to do, we probably want to fully skip the rule if _any_ of the value sets are self referencing.

src/export/ValueSetExporter.ts
```diff
@@ -86,6 +86,14 @@ export class ValueSetExporter {
           composeElement.valueSet = component.from.valueSets.map(vs => {
             return this.fisher.fishForMetadata(vs, Type.ValueSet)?.url ?? vs;
           });
+          composeElement.valueSet = composeElement.valueSet.filter(vs => {
+            if (vs == valueSet.url) {
+              logger.error(
+                `Value set with id ${valueSet.id} has component rule with self referencing value set (by id, value set name, or url). Skipping rule.`
```
jafeltra, line 92: This is very small, but I think `self-referencing value` should have the hyphen in it since `self-referencing` is the adjective.

jafeltra (commented):
Even though the Jira issue didn't link to it, I do think this resolves #1487. If you add "Fixes #1487" to the PR description, it should automatically close the issue once this PR is merged. Then again, maybe me typing this in the comment will do that too. I guess we'll see!

I also left some comments inline with questions about how the behavior here works. They probably are worth discussing with the team (unless these questions have been answered last week and I just missed it!).

src/export/ValueSetExporter.ts
jafeltra: After talking through this in person, we decided that the expected outcome of that FSH snippet would be

```
"compose": {
  "include": [
    {
      "valueSet": [
        "http://example.org/ValueSet/other-vs"
      ]
    }
  ]
}

```

jafeltra: If it is "a real pain in the neck" to get rid of the `null`, don't worry too much because we logged an error.

src/export/ValueSetExporter.ts
jafeltra: After talking through this in person, we decided that the behavior is correct (skip over MaxVS and use any remaining value sets) but we should update the error message.

Perhaps something along the lines of "removing self-reference" (or some better wording you think of).

src/export/ValueSetExporter.ts
KaelynJefferson: Looking into this further with the debugger, I do not see null values added to the compose.include array. The test I added does not show resulting null values in its array either (If it did, there would be 3 null values since we show the 3 different self references in that example). 

I see compose.include = null when given only self-referencing values in the value set (but null is not added to the compose.include array) - in this case there is no array.

jafeltra: Hm I think the case I was describing is a little different from what you described. I agree that the test shows we skip over the self-referenced value set and no `null` ends up in the `compose.include.valueset` array it constructs. I think the other example you described where there is no array at all is happens when the only rule on the value set has only a reference to itself (I might be wrong though, so let me know if I misunderstood this case).

The case that I was trying to ask about is if you have two include rules and one has a valid reference to a value set and the other has a self-referenced value set. In that case, because the `compose.include` array is being created, I'm seeing a `null` value entered into that array for the self-referenced value set. If you use the example FSH I included and run it on this branch and look at the generated ValueSet for MaxVS, I think you should see the `null` value. Then again, maybe I'm doing something wrong, so let me know if you can't reproduce that.

Also, we did decide that if "it is a 'real pain in the neck' to get rid of the `null`, then we just would leave it. So if this is reaching pain in the neck status, let me know.

KaelynJefferson: Updated with the recent changes. Let me know if this still does not cover this case.

src/export/ValueSetExporter.ts
```diff
@@ -176,7 +184,9 @@ export class ValueSetExporter {
               this.addConceptComposeElement(composeElement, valueSet.compose.include);
             }
           } else {
-            valueSet.compose.include.push(composeElement);
+            if (composeElement.valueSet?.length != 0) {
+              valueSet.compose.include.push(composeElement);
+            }
```
mint-thompson, line 189: I think this condition might be a little too restrictive. Consider the following FSH:
```
ValueSet: SomeVS
* include codes from system http://example.org/SomeCS and valueset SomeVS
```
In this case, the `composeElement.valueSet` will be an empty array, so `composeElement` will not be added to the value set. But, `composeElement.system` has a value, so I think that `composeElement` should be added.

mint-thompson (changes requested):
Thank you for making these updates. I have a comment about the conditional statement you added in the exporter.

src/export/ValueSetExporter.ts
```diff
@@ -176,7 +184,9 @@ export class ValueSetExporter {
               this.addConceptComposeElement(composeElement, valueSet.compose.include);
             }
           } else {
-            valueSet.compose.include.push(composeElement);
+            if (composeElement.valueSet?.length != 0) {
```
jafeltra, line 187: This is nitpicky, but I think it is preferred to use `!==` here because I think we do want to be checking strict equality (or inequality) here.

```suggestion
            if (composeElement.valueSet?.length !== 0) {
```

jafeltra (commented):
This looks good and covers the case I had mentioned! I left one little comment, but if you disagree with it, just let me know.

jafeltra (approved):
Looks good to me! Thanks for making all those updates! This seems like it will be very helpful to catch these sneaky mistakes.

mint-thompson (approved):
Thank you for the updates! Approved!

PR #1511, opened 2024-09-16
Assign date or dateTime that contains only year

### LINKED ISSUE #1483: Can't assign DateTime with only year component
Transferred from internal JIRA issue originally created 2023-09-20.

A four-digit number is a valid DateTime, but it matches the NUMBER token, so it doesn’t get assigned correctly:
```
* valueDateTime = 2023
```
When trying to assign a number to a DateTime, check if the raw value of the assignment is a four-digit number. If so, allow it.

mint-thompson:
**Description:**
A FHIR date or dateTime may be 4-digit year. The FSH parser will assign this value a NUMBER token type. When assigning to a date or dateTime element, if a type mismatch occurs, and the original value is a number, try to assign the raw value. If it is a valid 4-digit year, it will be assigned successfully. Otherwise, the original error will be thrown and logged.

I don't really like my implementation here: it's very similar code in all four exporters. I didn't manage to come up with a good way to isolate the "retry with another value" code, but I think it would be better if it looked like that.

**Testing Instructions:**
Tests are added to exporter classes to demonstrate assignment of these values.

**Related Issue:**
Fixes #1483

### DIFF
--- a/src/export/CodeSystemExporter.ts
+++ b/src/export/CodeSystemExporter.ts
@@ -16,7 +16,7 @@ import { logger } from '../utils/FSHLogger';
 import { MasterFisher, assembleFSHPath, resolveSoftIndexing } from '../utils';
 import { InstanceExporter, Package } from '.';
 import { CannotResolvePathError, MismatchedTypeError } from '../errors';
-import { isEqual } from 'lodash';
+import { cloneDeep, isEqual } from 'lodash';
 
 export class CodeSystemExporter {
   constructor(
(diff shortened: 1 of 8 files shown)


jafeltra (approved):
Looks good to me! It looks like we handled a similar situation already in all the different exporters, and I trust past us, so I think this is probably the right approach to handle the error.

cmoesel:
Just to capture some conversations we've had offline:
* The spec doesn't explicitly say how dates and datetimes should be assigned.
* The grammar does include a token format for dates and times without string delimiters
* Most examples in the spec assign using quoted date strings (e.g., `valueDateTime = "2012-01-01"`).
* This issue (year-only dates parsed as numbers) is _avoided_ if you just assign it using a string date instead.

While this PR does make it so you can assign year-only dates without using quoted strings, it introduces a fair amount of complexity. So the question we need to answer is if the complexity is worth it or if it's better to leave the complexity out of the code and just say you need to use a string date to make it unambiguous.

mint-thompson:
Closing without merge, as this adds a lot of complexity for something with a fairly simple workaround of "quote the date".

Issue #1485, opened 2024-06-28
Support using FHIR R6 in SUSHI


mint-thompson:
Transferred from internal JIRA issue originally created 2024-01-18.

SUSHI should be able to build an IG that uses R6 (currently `6.0.0-ballot1`). We’ll need to first be able to load R6 as a FHIR version. Once we can load R6, this may also require looking at any changes to StructureDefinition, ElementDefinition, ValueSet, CodeSystem, etc. and making sure there aren’t changes to things that SUSHI relies on.

SUSHI seems to be the one limiting the FHIR versions that can be used (here) and builds the package name and version to pass to FPL to load. However, there may still be changes in FPL that are required to properly load the R6 package.

PR #1517, opened 2024-09-23
Allow 6.x versions of FHIR

mint-thompson:
**Description:**
FHIR version 6 is currently in balloting stages. It does not appear that any code changes are needed yet for allowing an R6 IG. The usual warning about prerelease versions will be shown when one of the ballot versions is used.

**Testing Instructions:**
The main question here is whether supporting R6 is actually as simple as I think it is. Having looked over [information about the R6 ballot changes](https://hl7.org/fhir/6.0.0-ballot2/ballot-intro.html), I don't think there's anything that will need special handling. 

The new datatype [RelativeTime](https://hl7.org/fhir/6.0.0-ballot2/datatypes.html#RelativeTime) is complex enough that it doesn't make sense to try to support assigning it in a single rule. But, it may be worth considering if we want to support checking [its invariants](https://hl7.org/fhir/6.0.0-ballot2/datatypes-definitions.html#RelativeTime). rlt-1 is a very simple check, and rlt-2 isn't too much more complicated. We do something like this for Extensions for their [ext-1 invariant](https://hl7.org/fhir/extensibility-definitions.html#Extension).

**Related Issue:**
#1485

### DIFF
--- a/test/utils/Processing.test.ts
+++ b/test/utils/Processing.test.ts
@@ -426,6 +426,12 @@ describe('Processing', () => {
       expect(config.fhirVersion).toEqual(['4.5.0']);
     });
 
+    it('should allow FHIR R6', () => {
+      const input = path.join(__dirname, 'fixtures', 'fhir-r6');
+      const config = readConfig(input);
+      expect(config.fhirVersion).toEqual(['6.0.0-ballot2']);
+    });
+
     it('should allow FHIR current', () => {
       const input = path.join(__dirname, 'fixtures', 'fhir-current');
       const config = readConfig(input);
--- a/test/utils/Processing.test.ts
+++ b/test/utils/Processing.test.ts
@@ -426,10 +426,16 @@ describe('Processing', () => {
       expect(config.fhirVersion).toEqual(['4.5.0']);
     });
 
-    it('should allow FHIR R6', () => {
-      const input = path.join(__dirname, 'fixtures', 'fhir-r6');
+    it('should allow FHIR R6 prerelease', () => {
+      const input = path.join(__dirname, 'fixtures', 'fhir-r6-prerelease');
       const config = readConfig(input);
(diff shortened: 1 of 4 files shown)


cmoesel (dismissed):
Well, that was easier than expected! I tested it out by creating an R6 IG and profiling an element that doesn't exist in R5. It worked! I was even able to run the IG Publisher after.

One of the reasons this is so simple now is that past us thought ahead to this moment and added this: https://github.com/FHIR/sushi/blob/04b9d7beb12ad99489e03d0feeb142cf79c09762/src/utils/FHIRVersionUtils.ts#L61-L75

Anyway, nice job!

test/utils/Processing.test.ts
```diff
@@ -426,6 +426,12 @@ describe('Processing', () => {
       expect(config.fhirVersion).toEqual(['4.5.0']);
     });
 
+    it('should allow FHIR R6', () => {
+      const input = path.join(__dirname, 'fixtures', 'fhir-r6');
+      const config = readConfig(input);
+      expect(config.fhirVersion).toEqual(['6.0.0-ballot2']);
+    });
```
jafeltra, line 433: I don't know what the difference is, but I was also able to use just `6.0.0` as a `fhirVersion` already. Is it worth adding a test for that? SUSHI recognizes it as a non-preRelease version, thanks to the snippet of code Chris linked.

jafeltra (commented):
I just left one question about a test, but I might be getting ahead of myself, so let me know what you think.

test/utils/Processing.test.ts
mint-thompson: Getting a package downloaded for `6.0.0` right now is not so good, but either way, I can add a test for when that version is specified.

jafeltra (approved):
Looks good to me!

cmoesel (approved):
Re-approving!

Issue #1404, opened 2023-12-13
Support fully defining contained resources using caret rules


cmoesel:
It should be possible to fully define a contained resource, element-by-element, using caret rules in a non-Instance FSH definition. E.g.:
```
ValueSet: ExampleValueset
Id: example-valueset
* ^contained.resourceType = "CodeSystem"
* ^contained.id = "example-codesystem"
* ^contained.url = "http://example.org/codesystem"
* ^contained.version = "1.0.0"
* ^contained.content = #complete
* ^contained.concept[0].code = #example-code-1
* ^contained.concept[0].display = "Example Code 1"
* ^url = "http://example.org/valueset"
* ^version = "1.0.0"
* ^status = #active
```
Currently this results in an error because SUSHI does not recognize the path `^contained.url` since it is not in the `Resource` datatype. Ideally, SUSHI should look at the assigned `resourceType` and use that type when resolving paths in the contained resource FSH. Note that this _does_ work correctly when assigning contained resources in an `Instance` definition.

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Contained.20code.20system.20in.20the.20value.20set/near/407685349

Issue #1403, opened 2023-12-13
Allow fragments where reference URLs or Canonicals would go (in ValueSet syntax, but also elsewhere)


cmoesel:
Since fragments can be used as references or canonicals pointing to contained resources, anywhere the syntax allows a reference/canonical, it should also allow a fragment. For example, the compose rule in the following Value Set should work (but doesn't):
```
ValueSet: ExampleValueset
Id: example-valueset
* ^contained = example-codesystem
* ^url = "http://example.org/valueset"
* ^version = "1.0.0"
* ^status = #active
* include codes from system #example-codesystem

Instance: example-codesystem
InstanceOf: CodeSystem
Usage: #inline
* url = "http://example.org/codesystem"
* version = "1.0.0"
* status = #active
* content = #complete
// more rules
```

The above is an example. There may be other places this also needs to be supported.

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Contained.20code.20system.20in.20the.20value.20set/near/407685349

mint-thompson:
I've made some progress on this task in the `fragment-references-and-canonicals` branch. Note that the branch is not yet complete, so I do not recommend opening a pull request with that branch. Additionally, given the complexity of the implementation and change to the grammar, this may not be a feature that should be included.

PR #1520, opened 2024-09-24
Allow ValueSet to reference contained inline CodeSystem

### LINKED ISSUE #1402: Cannot refer to inline CodeSystems in ValueSet compose rules
The following code does not work, but should work (resulting in `compose > include > system` value `#example-codesystem`):
```
ValueSet: ExampleValueset
Id: example-valueset
* ^contained = example-codesystem
* ^url = "http://example.org/valueset"
* ^version = "1.0.0"
* ^status = #active
* include codes from system example-codesystem

Instance: example-codesystem
InstanceOf: CodeSystem
Usage: #inline
* url = "http://example.org/codesystem"
* version = "1.0.0"
* status = #active
* content = #complete
// more rules
```
Similarly, it should work correctly if `example-codesystem` is defined using `CodeSystem` syntax (and still use the fragment instead of the full URL).

We should also check to see if you can refer to inline value sets from compose rules (and fix it if you can't).

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Contained.20code.20system.20in.20the.20value.20set/near/407685349

mint-thompson:
**Description:**
If a ValueSetComponentRule specifies a system, check if that system is present in the list of contained resources. If so, add the [valueset-system extension](https://hl7.org/fhir/extensions/StructureDefinition-valueset-system.html) to the system, using the relative reference as the extension's value. If the system is not present in the list of contained resources, and the system is an inline Instance, log an error and do not add the component to the ValueSet.

Fishing in the tank for a CodeSystem will now return inline instances of CodeSystem. This matches the operation of fishing in the tank for a ValueSet.

Add `_system.extension` to the type definition for elements of a ValueSet's include and exclude lists. Use the Extension fhirtype as part of this definition. Replace the Extension fshtype with the Extension fhirtype in other parts of the ValueSet definition.

**Testing Instructions:**
Tests are added to ValueSetExporter and FHIRExporter as part of this change set.

As part of this implementation, I chose to make the failure case log an error and skip the component, but not throw anything. This means that processing of the ValueSet will continue. Please let me know if you think that this should instead throw an error and halt processing the ValueSet.

**Related Issue:**
Fixes #1402. Note that this does _not_ resolve the somewhat related issues #1403 (for allowing fragments as references to contained ValueSets) and #1404 (which would allow an inline CodeSystem to be constructed using caret rules on the ValueSet).

Additional information that led to this issue available in [this zulip thread](https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Contained.20code.20system.20in.20the.20value.20set/near/424938537). Additionally, I checked the list of FHIR core extensions for a similar extension to be used when a contained ValueSet is referenced within another ValueSet, and I did not find any such extension.

### DIFF
--- a/src/export/ValueSetExporter.ts
+++ b/src/export/ValueSetExporter.ts
@@ -76,8 +76,10 @@ export class ValueSetExporter {
             .replace(/^([^|]+)/, csMetadata?.url ?? '$1')
             .split('|');
           composeElement.system = foundSystem[0];
-          // if the code system is also a contained resource, add the special extension
-          // if it's not a contained resource, and the system we found is an inline instance, that's a problem
+          // if the code system is also a contained resource, add the valueset-system extension
+          // this zulip thread contains a discussion of the issue and an example using this extension:
+          // https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Contained.20code.20system.20in.20the.20value.20set/near/424938537
+          // additionally, if it's not a contained resource, and the system we found is an inline instance, that's a problem
           const containedSystem = valueSet.contained?.find((resource: any) => {
             return resource?.id === csMetadata.id && resource.resourceType === 'CodeSystem';
           });
--- a/src/export/ValueSetExporter.ts
+++ b/src/export/ValueSetExporter.ts
@@ -81,7 +81,7 @@ export class ValueSetExporter {
           // https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Contained.20code.20system.20in.20the.20value.20set/near/424938537
           // additionally, if it's not a contained resource, and the system we found is an inline instance, that's a problem
           const containedSystem = valueSet.contained?.find((resource: any) => {
-            return resource?.id === csMetadata.id && resource.resourceType === 'CodeSystem';
+            return resource?.id === csMetadata?.id && resource.resourceType === 'CodeSystem';
           });
           if (containedSystem != null) {
             composeElement._system = {
--- a/src/export/ValueSetExporter.ts
+++ b/src/export/ValueSetExporter.ts
@@ -76,6 +76,28 @@ export class ValueSetExporter {
             .replace(/^([^|]+)/, csMetadata?.url ?? '$1')
             .split('|');
           composeElement.system = foundSystem[0];
+          // if the code system is also a contained resource, add the special extension
+          // if it's not a contained resource, and the system we found is an inline instance, that's a problem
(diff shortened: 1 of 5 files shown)


src/fhirtypes/ValueSet.ts
```diff
@@ -83,6 +82,9 @@ export type ValueSetCompose = {
 
 export type ValueSetComposeIncludeOrExclude = {
   system?: string;
+  _system?: {
+    extension?: Extension[];
+  };
```
cmoesel, line 87: I think `_system` is better represented as an [Element](https://github.com/FHIR/sushi/blob/contained-codesystem-in-valueset/src/fhirtypes/dataTypes.ts#L62-L65) since it technically could have an `id` or `extension` array.

cmoesel (changes requested):
This works as advertised. I left of could of small comments to address. Thanks!

src/export/ValueSetExporter.ts
```diff
@@ -76,6 +76,30 @@ export class ValueSetExporter {
             .replace(/^([^|]+)/, csMetadata?.url ?? '$1')
             .split('|');
           composeElement.system = foundSystem[0];
+          // if the code system is also a contained resource, add the valueset-system extension
+          // this zulip thread contains a discussion of the issue and an example using this extension:
+          // https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Contained.20code.20system.20in.20the.20value.20set/near/424938537
+          // additionally, if it's not a contained resource, and the system we found is an inline instance, that's a problem
+          const containedSystem = valueSet.contained?.find((resource: any) => {
+            return resource?.id === csMetadata.id && resource.resourceType === 'CodeSystem';
```
jafeltra, line 84: Part of #1402 says

> We should also check to see if you can refer to inline value sets from compose rules (and fix it if you can't).

I went to go look to see if this was fixed with this PR, so I tried to add an inline valueset to the contained list and reference it in a compose rule. Maybe it is worth discussing if we want that fix as part of this or if we want to track it as part of a separate issue. Either way, we probably should fix this line because I get an error that says "Cannot read properties of undefined (reading 'id')". So at the very least, I think we want to avoid that. I think updating the line like this should do it:
```suggestion
            return resource?.id === csMetadata?.id && resource.resourceType === 'CodeSystem';
```

jafeltra (commented):
Looks good. I just noticed one issue, that also asks a slightly larger question.

jafeltra (approved):
Makes sense. Thanks for adding the skipped test!

PR #1523, opened 2024-10-02
Fix FSH Finder URL for regression script

cmoesel:
**Description:** The FSH Finder URL changed when FSH School was transferred to HL7. As a result, the regression script failed to download repo information from FSH Finder. It also displayed a big ugly error message that was difficult to interpret.

This PR adds better handling for when the FSH Finder download fails and also updates the FSH Finder URL to the new location.

**Testing Instructions:** Run a regression. You don't need to do a full regression. You could just do: `npm run regression -- run -a gh:master -b local --count 2`. You can test the new error handling by changing the `FSH_FINDER_URL` in the regression script back to its old value and attempt a regression again.

**Related Issue:** N/A


mint-thompson (approved):
New URL makes the regression work, and the new error message is much more useful. Thank you!

jafeltra (approved):
Looks good. The error message is much nicer to read now, but also doesn't happen!

PR #1524, opened 2024-10-07
Only return FshType id from rule value if the value is a string

### LINKED ISSUE #1519: Instance with unquoted numeric id causes call stack overflow
Assignment rules on Instance `id` should be quoted strings. But, if the value is unquoted and numeric, the initial assignment during export will fail due to being the wrong type. SUSHI sees that the value is numeric, and makes the guess that maybe it's supposed to be assigning an instance identified by that numeric value. It will fish for such an Instance, find the one already being exported. This is because the Instance's `id` accessor method casts the rule's value to a string. This leads to the Instance being exported again, eventually leading to call stack overflow. It would probably make more sense if a non-string value on a rule on `id` did _not_ get returned by the accessor.

Sample FSH:
```
Instance: requireddataforantimicro-response
InstanceOf: Questionnaire
Usage: #example
* id = 3193360
```

In this case, the expected behavior would be an error on the `id` rule indicating that the numeric value is the wrong type.

[Zulip thread](https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Sushi.3A.20.20Maximum.20call.20stack.20size.20exceeded.20.3F/near/472525892)

mint-thompson:
**Description:**
When getting the id for a FshType, rules are checked to see if there is one that will set the id and does not refer to an instance. A value that is not a string will not be valid and not be applied. Therefore, only return the value if it is a string.

Rename test file for FshCodeSystem to match name of source file.

**Testing Instructions:**
Confirm that using a non-string value in a rule that sets an entity's `id` does not allow for other rules to fish up that entity using that id. The sample FSH from the issue is an example of this on an Instance:
```
Instance: requireddataforantimicro-response
InstanceOf: Questionnaire
Usage: #example
* id = 3193360
```
In this case, an error will be logged for the `id` rule since it is the wrong type, and there should not be a call stack overflow as described in the linked issue. Behavior should be the same for Profile, Extension, Logical, Resource, CodeSystem, and ValueSet, although these entities will set `id` using a caret rule, not an assignment rule.

**Related Issue:**
Fixes #1519.

### DIFF
--- a/src/fshtypes/FshCodeSystem.ts
+++ b/src/fshtypes/FshCodeSystem.ts
@@ -25,8 +25,8 @@ export class FshCodeSystem extends FshEntity {
 
   get id() {
     const assignedId = getNonInstanceValueFromRules(this, 'id', '', 'id');
-    if (assignedId) {
-      return assignedId.toString();
+    if (typeof assignedId === 'string') {
+      return assignedId;
     }
     return this._id;
   }
--- a/src/fshtypes/FshStructure.ts
+++ b/src/fshtypes/FshStructure.ts
@@ -22,8 +22,8 @@ export abstract class FshStructure extends FshEntity {
 
   get id() {
     const assignedId = getNonInstanceValueFromRules(this, 'id', '', 'id');
-    if (assignedId) {
-      return assignedId.toString();
+    if (typeof assignedId === 'string') {
+      return assignedId;
     }
     return this._id;
   }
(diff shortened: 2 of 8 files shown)


cmoesel (approved):
I agree that this fixes the problem and remains true to the intent of the current code.

That said, I do wonder why we intentionally only get non-Instance values. In theory, something like this should be valid:
```
Instance: Id3193360
InstanceOf: string
Usage: #inline
* value = "3193360"

Instance: MyQuestionnaire
InstanceOf: Questionnaire
Usage: #example
* id = Id3193360
* status = #final
```
and should detect the id as having the value `"3193360"`. But... that is outside the scope of this PR and probably covered by #1526.

jafeltra (approved):
Confirmed this resolves the issue. I agree with Chris's comments, but that seems like something to address in the new issue.

PR #1528, opened 2024-10-11
entities with same name

### LINKED ISSUE #1469: `Canonical()` behavior when a CodeSystem and a ValueSet have exact same Ids
If we define a CodeSystem and a ValueSet with the exact Ids:

```
CodeSystem: FooBar
Id:         foo-bar
Title:     "Foo Bar CodeSystem"

* #v1
    "value1"
* #v2
    "value2"
```

```
ValueSet: FooBar
Id: foo-bar
Title: "Foo Bar ValueSet"

* include codes from system FooBar
```

`Canonical(FooBar)` always return the canonical url of the ValueSet. 

* Is there a way we could choose which resource's url we'd like to receive?
* If not, it would be great if SUSHI could through an error to highlight there is some ambiguity here.

KaelynJefferson:
**Description:** This PR adds conditionals to check for entities that exist with the same name, and logs an error when entities are defined with a duplicate name.

**Testing Instructions:** jest unit tests

**Related Issue:** Fixes #1469


src/export/CodeSystemExporter.ts
```diff
@@ -273,7 +273,16 @@ export class CodeSystemExporter {
...
-    if (this.pkg.codeSystems.some(cs => cs.name === fshDefinition.name)) {
+    if (
+      this.pkg.codeSystems.some(cs => cs.name === fshDefinition.name) ||
+      this.pkg.instances.some(instance => fshDefinition.name === instance._instanceMeta.name) ||
+      this.pkg.profiles.some(prof => fshDefinition.name === prof.name) ||
+      this.pkg.extensions.some(extn => fshDefinition.name === extn.name) ||
+      this.pkg.logicals.some(logical => fshDefinition.name === logical.name) ||
+      this.pkg.resources.some(resource => fshDefinition.name === resource.name) ||
+      this.pkg.valueSets.some(valueSet => fshDefinition.name === valueSet.name)
+    ) {
```
mint-thompson, line 284: I think it would be possible to write this condition by checking `this.pkg.fshMap` rather than each of the package's lists of FSH entities individually.

test/export/InstanceExporter.test.ts
```diff
@@ -10311,7 +10342,10 @@ describe('InstanceExporter', () => {
       afterEach(() => {
         // None of the test expect warnings or errors. All should be clean.
         expect(loggerSpy.getAllLogs('warn')).toHaveLength(0);
-        expect(loggerSpy.getAllLogs('error')).toHaveLength(0);
+        expect(loggerSpy.getAllLogs('error')).toHaveLength(1);
+        expect(loggerSpy.getMessageAtIndex(0, 'error')).toMatch(
+          /Multiple FSH entities created with name/
+        );
```
mint-thompson, line 10348: I'm a little confused by this. Why is this error appearing for these tests? There don't appear to be any entities with the same names defined.

test/export/InstanceExporter.test.ts
```diff
@@ -2217,7 +2248,7 @@ describe('InstanceExporter', () => {
       thisIsName.isInstance = true;
       thisIsSeacow.rules.push(thisIsName);
       const exported = exportInstance(thisIsSeacow);
-      expect(loggerSpy.getAllMessages('error')).toHaveLength(0);
+      // expect(loggerSpy.getAllMessages('error')).toHaveLength(0);
```
mint-thompson, line 2251: I think this should still be producing 0 errors.

test/export/InstanceExporter.test.ts
```diff
@@ -6876,7 +6907,7 @@ describe('InstanceExporter', () => {
 
       const exported = exportInstance(carePlanInstance);
       expect(exported.instantiatesCanonical).toEqual(undefined); // instantiatesCanonical is not set with invalid type
-      expect(loggerSpy.getMessageAtIndex(1, 'error')).toMatch(
+      expect(loggerSpy.getMessageAtIndex(7, 'error')).toMatch(
```
mint-thompson, line 6910: What's causing the six additional errors here?

mint-thompson (changes requested):
Thank you for working on this! I have some feedback about some of your tests, and a suggestion for the message being logged.

test/export/InstanceExporter.test.ts
KaelynJefferson: I believe the extra error messages were due to how this test was calling exportInstance multiple times (exportInstance is defined as a function in the beforeEach that when called, calls export for cs, vs, and sd and the instance given) so on the second time, it logged the errors. I adjusted this test (and the one above) to just call export within the function instead.

test/export/InstanceExporter.test.ts
KaelynJefferson: This error is occurring for a different reason than below. 

Stepping through this test, I see that when export is called in this group of tests, it adds the profile (CarProfile) and the logical (Car) at the same time, so when calling export for the Logical, its already in the this.pkg.FshMap when it checks for it, and throws the error.

mint-thompson (dismissed):
Thank you!

cmoesel:
UPDATE: I ran the regression and this caused new errors in almost 60 projects of out the 510 projects updated in the last year (so... > 10% of projects are affected). That being the case, I don't think we want to introduce this as an error. And if we make it a warning, we might want to consider some additional guidance on how you can make the entity name unique while retaining the same name in the exported definition (as it seems that people must be doing this intentionally).

cmoesel:
@KaelynJefferson: The rest of the team discussed this in a standup and agreed that we don't want this to be an error. We also thought that logging 50 warnings (as would be the case for a few IGs) might be too much. So we suggest that you collect all of the duplicate entity names and issue _one_ **warning** message. Perhaps something like:
> Detected FSH entity definitions with duplicate names. While FSH allows for duplicate names across entity types, they can lead to ambiguous results when referring to these entities by name elsewhere (e.g., in references). Consider using unique names in FSH declarations and assigning duplicated names using caret assignment rules instead. Detected duplicate names: Foo, Bar, Baz.

src/import/FSHTank.ts
```diff
@@ -168,15 +169,19 @@ export class FSHTank implements Fishable {
...
       }
     });
+
+    if (duplicateEntities.size > 0) {
+      logger.warn(
+        'Detected FSH entity definitions with duplicate names. While FSH allows for duplicate ' +
+          'names across entity types, they can lead to ambiguous results when referring to these ' +
+          'entities by name elsewhere (e.g., in references). Consider using unique names in FSH ' +
+          'declarations and assigning duplicated names using caret assignment rules instead. ' +
+          `Detected duplicate names: ${Array.from(duplicateEntities)}.`
```
cmoesel, line 182: When there are multiple affected entities, this will string them all together without any spaces (e.g., `foo,bar,baz`). This would likely cause some usability issues for long lists. I'd suggest: `${Array.from(duplicateEntities).join(', ')}`

test/run/FshToFhir.test.ts
```diff
@@ -206,16 +206,17 @@ describe('#FshToFhir', () => {
...
-        /Duplicate entity name: multiple entity types with name MyPatient1 exist/
-      );
-      expect(results.errors[1].message).toMatch(
-        /Duplicate entity name: multiple entity types with name MyPatient1 exist/
+      expect(results.warnings[0].message).toMatch(
+        'Detected FSH entity definitions with duplicate names. While FSH allows for duplicate ' +
+          'names across entity types, they can lead to ambiguous results when referring to these ' +
+          'entities by name elsewhere (e.g., in references). Consider using unique names in FSH ' +
+          'declarations and assigning duplicated names using caret assignment rules instead. ' +
+          'Detected duplicate names: MyPatient1.'
```
cmoesel, line 219: Ideally, we should probably test this for the case with multiple duplicates.

cmoesel (changes requested):
I've made a small request for a change. In addition, I think this needs to be re-synched with the master branch.

cmoesel (approved):
This looks great. Thank you, @KaelynJefferson!

mint-thompson (approved):

PR #1530, opened 2024-10-25
Support fully defining resources using caret rules

### LINKED ISSUE #1404: Support fully defining contained resources using caret rules

mint-thompson:
**Description:**

A contained resource can be defined within another resource one rule at a time using caret rules. By setting the resourceType of the contained resource, SUSHI will be able to correctly validate other rules that build up the contained resource. Most of the logic for this already existed, but some additional code was needed in order to collect information about inline resource types. Additionally, if an instance is assigned with a caret rule, further caret rules may be used to modify the contained resource. This gets a little complicated when exporting a StructureDefinition.

A ValueSet may refer to a CodeSystem that is defined with caret rules on the ValueSet. The resulting compose element will look the same as if the inline CodeSystem were defined separately.

Log a warning when an example instance is assigned within a conformance resource.

Update object definitions for inline instance tests so that the Instances are, in fact, inline.

**Testing Instructions:**
Please check for completeness of test cases. This change involves a lot of small variations in possible input scenarios, and while the code coverage tool reports that my new code is covered by the new tests, there may be unique combinations that merit their own tests.

The new warning message for example instances within conformance instances might need some revision. It's maybe a little too casual right now.

The changes to the exporters are similar, but not always the same. But, there may be some opportunities to pull out similar blocks into shared functions.

Additionally, I left a question in my implementation in `StructureDefinitionExporter` in regards to re-cleaning contained resources. Please consider whether a different strategy for re-cleaning would be necessary or useful.

An initial regression of the 100 most recently updated projects showed no changes. This isn't too surprising, as this change is mostly to allow for new things that currently do not work.

**Related Issue:**
Fixes #1404.

### DIFF
--- a/src/export/CodeSystemExporter.ts
+++ b/src/export/CodeSystemExporter.ts
@@ -8,7 +8,8 @@ import {
   setImpliedPropertiesOnInstance,
   validateInstanceFromRawValue,
   isExtension,
-  replaceReferences
+  replaceReferences,
+  splitOnPathPeriods
 } from '../fhirtypes/common';
 import { FshCodeSystem } from '../fshtypes';
 import { CaretValueRule, ConceptRule } from '../fshtypes/rules';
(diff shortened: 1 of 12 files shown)


jafeltra (approved):
I think this all makes sense to me. The code was a little confusing to me because it really needed to touch so many different areas, and it is a bummer that there is similar but not exactly the same code in the different exporters, but it seems like that's just how it needs to be (or at least that's the approach we've taken in these sections in the past). The tests looked like they covered the cases that this is aiming to address, so it looks good to me!

cmoesel (approved):
This looks good. It's complex, but I think it needs to be -- so here we are! Good work!

PR #1534, opened 2024-11-01
Load automatic dependencies for FHIR R6 versions

### LINKED ISSUE #1533: SUSHI does not auto-load extensions / terminology packages for FHIR R6
SUSHI does not automatically load the extensions package or terminology package for projects that specify their fhirVersion as 6.0.0-ballot2. I think this is because our auto-package loading mechanism looks up those packages based on FHIR version, and SUSHI is missing a mapping for version 6.

We should update that code to account for R6 and/or to fall back to the latest known FHIR-specific versions (E.g., currently the R5 versions) when we don't  have a mapping for the requested FHIR version.

Currently, authors can work around this by adding the latest R5 versions to their dependencies:
```yaml
dependencies:
  hl7.fhir.uv.extensions.r5: latest
  hl7.terminology.r5: latest
```

mint-thompson:
**Description:**
R6 prerelease versions of FHIR are available, but R6 versions of the terminology and extension packages are not yet available. Load the R5 versions of this packages when an R6 version of FHIR is used.

**Testing Instructions:**
Confirm that tests pass. Run SUSHI on a project that uses R6 prerelease FHIR along with official extensions and confirm that no errors are produced.

**Related Issue:**
Fixes #1533.

### DIFF
--- a/src/utils/Processing.ts
+++ b/src/utils/Processing.ts
@@ -70,7 +70,7 @@ export const AUTOMATIC_DEPENDENCIES: AutomaticDependency[] = [
   {
     packageId: 'hl7.terminology.r5',
     version: 'latest',
-    fhirVersions: ['R5']
+    fhirVersions: ['R5', 'R6']
   },
   {
     packageId: 'hl7.fhir.uv.extensions.r4',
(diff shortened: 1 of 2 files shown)


cmoesel (approved):
Looks good to me. I tried it on an R6 project and confirmed that the automatic dependencies are now properly loaded. Thank you, @mint-thompson!

jafeltra (approved):
Looks good to me! Worked as expected and the tests look good!

PR #1536, opened 2024-11-07
v3.12.1

mint-thompson:
**Description:**
Bump version to v3.12.1. This is a patch version, since we're fixing the automatic extension issue for R6 that was introduced in v3.12.0.

**Testing Instructions:**
Run SUSHI to build a project that uses an R6 ballot version, has no configured dependencies, and makes use of extensions in the [core registry](https://hl7.org/fhir/extensions/extension-registry.html). These extensions should be automatically loaded, so there should be no errors.

**Related Issue:**
N/A

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,12 +1,12 @@
 {
   "name": "fsh-sushi",
-  "version": "3.12.0",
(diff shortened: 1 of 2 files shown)


cmoesel (approved):
Looks good!

PR #1537, opened 2024-11-18
Check exported resources for presence of multiple choice values

### LINKED ISSUE #1137: SUSHI should warn when multiple value[x] choices are used
SUSHI should log a warning if multiple value[x] choice elements are applied to an Instance. For example, both `valueString` and `valueCodeableConcept` cannot both be assigned on an Observation instance. In this case, the FHIR validator reports an error, but SUSHI does not.

Example FSH: https://fshschool.org/FSHOnline/#/share/3pKpG5X
Zulip thread: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Multiple.20Observation.20values

mint-thompson:
**Description:**
When a choice element is defined in FHIR, it allows for values of different types with names based on those types. However, any choice element that is present on resource should have only one named value. For example, Observation.value[x] allows for many types, but an Observation resource should not have both valueString and valueInteger. After exporting a resource, check the resource for the presence of multiple values for the same choice element, and log an error when they are detected.

As assigned values are validated in exporters, the relevant StructureDefinitions will have their elements unfolded. Keep the same instance of that StructureDefinition that was used for validation so that the unfolded elements are present when performing the multiple choice validation check.

**Testing Instructions:**
Confirm that all tests pass. Confirm that tests provide sufficient coverage for the scenario described. I put most of the tests in `InstanceExporter.test.ts` since those were the first ones I wrote, but since all exporters use the same functions, I felt that it was not necessary to have as many tests in the other exporters.

**Related Issue:**
Fixes #1137.


jafeltra (approved):
This all looks good to me! I think just having the full coverage in the InstanceExporter and keeping the simpler tests in each other exporter test suite makes sense.

test/export/InstanceExporter.test.ts
```diff
@@ -5321,6 +5321,109 @@ describe('InstanceExporter', () => {
...
+        valueString: "patient's gender is unknowable",
+        valueCode: 'unknowable'
+      });
+      expect(loggerSpy.getAllMessages('error')).toHaveLength(1);
+      expect(loggerSpy.getLastMessage('error')).toMatch(
+        /Bar contains multiple choice value assignments for choice element Patient\.gender\.extension\.value\[x\]\..*File: PatientInstance\.fsh.*Line: 10 - 20\D*/s
+      );
+      expect(loggerSpy.getAllMessages('warn')).toHaveLength(0);
+    });
+
```
cmoesel, line 5426: I think we need some negative tests too (i.e., cases where the error should not be logged). For example, we want to ensure that repeated elements that have child elements w/ choices allow for different choice types in each repetition. For example, you could write a test with multiple `Observation.component`s, each with a different `Observation.component.value[x]` type.

src/fhirtypes/common.ts
```diff
@@ -1633,3 +1633,64 @@ export function getMatchingContainedReferenceId(
...
+  fshDef: Profile | Extension | Logical | Resource | FshCodeSystem | FshValueSet | Instance,
+  instance: { [key: string]: any },
+  element: ElementDefinition
+) {
+  const children = element.children(true);
+  children.forEach(child => {
+    // does this child represent a choice element, such as value[x]?
+    // if so, check for choices
+    if (child.id.endsWith('[x]')) {
+      const availableChoices = child.getSlices();
```
cmoesel, line 1655: I'm trying to understand how this works. Does `child.getSlices()` only work if the SD used at the top-level is the same instance of the SD that was used to assign the values (e.g., it relies on choice slices having been materialized during instance assignment)? If so, that seems to me like it might be fragile (as in, it would stop working if we changed how we do assignments or if someone invoked this function w/ a fresh instance of the SD). Or am I missing something? As far as I can tell, `ElementDefinition.getSlices()` doesn't automatically expand all choice type for you -- it only works for the ones that were already expanded.

I expected you would use `element.type` to build up the list of potential value properties.

cmoesel (changes requested):
Thank you, Mint! I've asked for some additional tests and asked a question about the approach. See my individual comments.

cmoesel (approved):
Thanks, @mint-thompson. This looks good to me!

jafeltra (approved):

Issue #1391, opened 2023-11-21
Add support for R5 #position discriminator in slicing


cmoesel:
FHIR R5 introduces a `#position` discriminator type with the following description:
> The slices are differentiated by their index. This is only possible if all but the last slice have min=max cardinality, and the (optional) last slice contains other undifferentiated elements.

It further notes that position is
> Used to match slices based on their index. This is only possible if all but the last slice have a fixed cardinality where min > 0 and min = max. The last slice MAY have min != max. Typical example: slice on Practitioner.name to require that the first name be the usual name for the practitioner, and it must be present.

If you try to set a discriminator type to `#position` now, SUSHI issues an error, even in R5 projects:
> ^slicing.discriminator[0].type: Invalid value: #position. Value must be selected from one of the following: #value, #exists, #pattern, #type, #profile

We need to allow for this discriminator. Furthermore, our instance exporter needs to be smart enough to put positional slices in the right order, regardless of the order they are assigned in the instance.

cmoesel:
For someone looking to do a `good first issue`, it would be helpful to have SUSHI allow the `#position` discriminator to be used in slicing rules. The idea about having SUSHI export the slices in the right order could be done at a later time (since authors can already manually author their instances to use correct slice order).

PR #1538, opened 2024-11-22
Support "position" slicing discriminator

mint-thompson:
**Description:**
FHIR R5 added "position" as an allowed slice discriminator, where slices are differentiated by their order within a list element. Because a consistent order is necessary for this slicing to be meaningful, emit an error if "ordered" is not set to true for the slicing. Previous versions of FHIR may not use this discriminator.

Minor changes to npm-shrinkwrap.json due to issues detected by `npm audit`.

**Testing Instructions:**
Confirm that new tests adequately cover this new feature. Try building projects with R4 and R5 resources using SUSHI, observing that the "position" discriminator can't be used with R4, but can be used with R5. There is also a new error message, so please let me know if you have suggestions for its wording.

**Related Issue:**
Partially addresses #1391, but does _not_ validate Instances to confirm that slices are provided in the required order when position-based slicing is used.

### DIFF
--- a/src/fhirtypes/ElementDefinition.ts
+++ b/src/fhirtypes/ElementDefinition.ts
@@ -386,13 +386,27 @@ export class ElementDefinition {
     validationErrors.push(
       this.validateIncludes(slicing.rules, ALLOWED_SLICING_RULES, 'slicing.rules')
     );
-
+    const isR4 = this.structDef.fhirVersion.startsWith('4.');
     slicing.discriminator?.forEach((d, i) => {
       const discriminatorPath = `slicing.discriminator[${i}]`;
       validationErrors.push(this.validateRequired(d.type, `${discriminatorPath}.type`));
-      validationErrors.push(
-        this.validateIncludes(d.type, ALLOWED_DISCRIMINATOR_TYPES, `${discriminatorPath}.type`)
-      );
+      if (isR4) {
+        validationErrors.push(
+          this.validateIncludes(d.type, ALLOWED_DISCRIMINATOR_TYPES, `${discriminatorPath}.type`)
+        );
+      } else {
+        validationErrors.push(
+          this.validateIncludes(d.type, ALLOWED_DISCRIMINATOR_TYPES_R5, `${discriminatorPath}.type`)
+        );
+        if (d.type === 'position' && slicing.ordered !== true) {
+          validationErrors.push(
+            new ValidationError(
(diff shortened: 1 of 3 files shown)


jafeltra (approved):
This makes sense! I tested out the new slicing discriminator in R4 and R5, and it worked as expected!

src/fhirtypes/ElementDefinition.ts
```diff
@@ -386,13 +386,27 @@ export class ElementDefinition {
...
+          this.validateIncludes(d.type, ALLOWED_DISCRIMINATOR_TYPES_R5, `${discriminatorPath}.type`)
+        );
+        if (d.type === 'position' && slicing.ordered !== true) {
+          validationErrors.push(
+            new ValidationError(
+              'Slicing ordering must be true when a position discriminator is used.',
+              'slicing.ordered'
+            )
+          );
+        }
```
cmoesel, line 408: Here's something to think about: If the user uses a `position` discriminator type, should we just set `ordered` to true for them? I.e., we know what it should be, so... just do it? Putting the "short" in "shorthand"?

I'm torn on it. I feel like if they don't set the `ordered` property themself using a caret rule, then it makes sense to set it for them. But if they had set it themself to the _wrong_ value (false), then I'm not sure if silently setting it to true is the right thing to do...

src/fhirtypes/ElementDefinition.ts
```diff
@@ -3076,6 +3090,15 @@ export type ElementDefinitionSlicingDiscriminator = {
...
 const ALLOWED_DISCRIMINATOR_TYPES = ['value', 'exists', 'pattern', 'type', 'profile'];
+// R5 introduced the 'position' discriminator
+const ALLOWED_DISCRIMINATOR_TYPES_R5 = [
+  'value',
+  'exists',
+  'pattern',
+  'type',
+  'profile',
+  'position'
+];
```
cmoesel, line 3101: I don't have a strong opinion on this, but... it could be:
```suggestion
const ALLOWED_DISCRIMINATOR_TYPES_R5 = [...ALLOWED_DISCRIMINATOR_TYPES, 'position'];
```

cmoesel (approved):
I'm approving this, but I did make a minor suggestion that you can take or leave. And I also started a discussion on an alternate approach to `ordered` -- but I'm comfortable with it as-is if that's what other think is best.

PR #1539, opened 2024-11-26
Allow assigning an inline instance of a primitive

### LINKED ISSUE #1526: Improve handling of complex id, and assignment of primitive instances generally
If an entity’s id is an object and not just a string, it can produce errors and undesirable results. As long as that object has a “value” attribute, we should be able to use that attribute for the actual string. This only happens when assigning an inline instance of string to “id”.

The right thing to do is to make it so that the value ends up on the actual “id” and the other parts end up on “_id”.

While we discovered this with id, this is applicable when assigning any inline instance of a primitive.

jafeltra:
**Description:**

This PR makes it so you can assign an inline instance to an element that is a primitive data type. The `value` property of the inline instance is assigned directly to the primitive element. If the instance defines any other properties, they get added at the `_` property of the primitive instance.

**Testing Instructions:**

I added two tests to test out two slightly different scenarios for this task, but if you wanted to try to FSH for these, this is what I've been using to test:

The following FSH should create a file called Questionnaire-Id3193360.json and assign the id as `Id3193360`:

```
Instance: Id3193360
InstanceOf: id
Usage: #inline
* value = "3193360"

Instance: MyQuestionnaire
InstanceOf: Questionnaire
Usage: #example
* id = Id3193360
* status = #final
```

The following FSH should add "Important Observation" to the `valueString` and add the extension to `_valueString`:

```
Instance: ValueString
InstanceOf: string
Usage: #inline
* value = "Important Observation"
* extension.url = "http://example.com/fake"
* extension.valueBoolean = true

Instance: MyObservation
InstanceOf: Observation
* valueString = ValueString
* status = #final
* code = #123
```

I ran a regression of all the repos changed in the last year, and no changes were reported.

**Related Issue:** Fixes #1526

### DIFF
--- a/test/export/InstanceExporter.test.ts
+++ b/test/export/InstanceExporter.test.ts
@@ -266,6 +266,13 @@ describe('InstanceExporter', () => {
       carePlanInstance = new Instance('C');
       carePlanInstance.instanceOf = 'CarePlan';
       doc.instances.set(carePlanInstance.name, carePlanInstance);
+      const carePlanStatus = new AssignmentRule('status');
+      carePlanStatus.value = new FshCode('draft');
+      const carePlanIntent = new AssignmentRule('intent');
+      carePlanIntent.value = new FshCode('plan');
+      const carePlanSubject = new AssignmentRule('subject');
+      carePlanSubject.value = new FshReference('Patient/fake');
+      carePlanInstance.rules.push(carePlanStatus, carePlanIntent, carePlanSubject);
       lipidInstance = new Instance('Bam')
         .withFile('LipidInstance.fsh')
         .withLocation([10, 1, 20, 30]);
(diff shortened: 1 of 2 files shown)


test/export/InstanceExporter.test.ts
```diff
@@ -11354,6 +11371,63 @@ describe('InstanceExporter', () => {
           ]
         });
       });
+
+      it('should assign an inline instance of a primitive to a primitive element', () => {
+        const primitiveInstannce = new Instance('MyFancyId')
```
mint-thompson, line 11376: minor typo here and in following tests: variable name should be `primitiveInstance`

jafeltra: Whoops, good call! Fixed in [5a78ac5](https://github.com/FHIR/sushi/pull/1539/commits/5a78ac52ec0a7a3fb931e8759d6c43a2e7a00bde).

jafeltra:
After discussing at stand up, we decided that the validation portion of this is out of scope for this PR. I fixed the typo that @mint-thompson pointed out, so I think this is ready for re-review.

cmoesel (approved):
Approved (again)!

PR #1540, opened 2024-12-02
Improve error message for ValueSet compose component without any concept

### LINKED ISSUE #1518: Improve error message for code caret rule on ValueSet compose component without any concept
When writing rules for a ValueSet, it is invalid to make a code caret rule on a concept that is not specifically present within the ValueSet.. However, when there are _no_ specific concepts present, a slightly strange error message is produced.

Sample FSH:
```
ValueSet: SomeVS
* include codes from system http://example.org/CS
* http://example.org/CS#"some-code" ^designation.value = "some value"
```

Error message received:
```
error The element or path you referenced does not exist: compose.include[0].concept[undefined].designation.value
  Line: 3
```
The rule path in the error message is the path constructed within the ValueSetExporter for the rule on line 3. But, it might be more useful to directly indicate that the rule is invalid because there is no concept array in the compose element with `http://example.org/CS` as the system.

KaelynJefferson:
**Description:** This PR adds an improved error message for when there is a code caret rule on ValueSet where the compose component does not have a concept.

**Testing Instructions:** Run unit tests.

**Related Issue:** This fixes #1518 .


jafeltra (commented):
This looks good! I ran a regression and it doesn't show anyone is hitting this error right now, but it will definitely be helpful for someone writing FSH and running into this.

I just left one small suggestion. Let me know if you agree.

cmoesel (changes requested):
The logic looks good! I've suggested a friendly message to help the user know how to resolve this.

cmoesel (approved):
Thank you, @KaelynJefferson!

jafeltra (approved):
Looks good to me! Thanks for the updates!

Issue #1389, opened 2023-11-17
SUSHI does not handle using the Canonical() keyword with SearchParameters


jafeltra:
When using the `Canonical` keyword with a SearchParameter that is part of FHIR R5, SUSHI cannot find the resource and logs the following error:

> error Cannot use canonical URL of individual-gender because it does not exist. Be sure that individual-gender exists and it has a URL.

FSH Online example that demonstrates this problem: https://fshschool.org/FSHOnline/#/share/3R14kQe
Based on Zulip question: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Checking.20Canoncials/near/402723450

cmoesel:
Addressed by #1548.

Issue #1043, opened 2022-03-09
Support for importing an example bundle by reference to another IG, or directly


saulakravitz:
Sushi current supports neither of these:
  1) importing an example bundle from an IG that is listed as a dependency.   I wanted to import the example [death certificate bundle](http://build.fhir.org/ig/HL7/vrdr/branches/master/Bundle-DeathCertificateDocument-Example1.html) defined in VRDR into the Vital Records FHIR Messaging IG.
  2) importing an example JSON bundle by including it in the input/resources folder and referencing it by ID.
  
  In the end I needed to include the FSH for the complete example in the [Vital Records FHIR Messaging IG](http://build.fhir.org/ig/nightingaleproject/vital_records_fhir_messaging_ig/branches/main/Bundle-DeathCertificateDocument-Example1.html)....

cmoesel:
The primary issue here is that when SUSHI loads dependencies, it only holds onto structure definitions, code systems, value sets, and implementation guides.  Everything else (including Bundles) gets dropped.  The same behavior applies to predefined resources in `input/resources` (or elsewhere).

SUSHI should probably load and keep the entire package, but we'll need to determine how the other resource types are stored, how they should be fished for, and if this causes any issues (using extra memory, fishing up things it didn't used to fish up, etc).

cmoesel:
#1548 addresses point 2 above. It does not address point 1 above, but the IG Publisher doesn't load example resources from dependencies, so I don't think you can reference an example across IGs anyway.

cmoesel:
Closing this as I believe SUSHI now supports all it can/should support based on current IG Publisher behavior (see comment above).

Issue #988, opened 2021-12-17
SUSHI doesn't load all resources in dependency packages


cmoesel:
Currently, SUSHI only loads `StructureDefinition`, `ValueSet`, `CodeSystem`, and `ImplementationGuide` resources from dependency packages.  BUT... it probably _should_ be loading other types such as `ConceptMap`, `SearchParameter`, `PlanDefinition`, etc.  In a [recent Zulip conversation](https://chat.fhir.org/#narrow/stream/179166-implementers/topic/Reference.20instances.20between.20IGs), we even saw a use case for loading examples.

Once the new separate package loading module is developed, we should update it to support these features.

cmoesel:
Addressed by #1548.

PR #1548, opened 2024-12-23
Support fishing all resource types from dependency packages and input/*

cmoesel:
**Description:** In the past, SUSHI only loaded specific resource types from dependency packages and predefined resources in input/* (i.e., StructureDefinitions, ValueSets, and CodeSystems). The recent integration of FPL 2.0.0 now loads all resource types, but SUSHI still did not return them in its results since it only fished (i.e. searched) for the previously supported types. This PR adds support for fishing those types by interpreting a fishing call for the `Type.Instance` type as a wildcard type (since a FSH `Instance` can be literally any type).

This also adds a new `fishForMetadatas` function that will return _all_ the matching `Metadata`s instead of just the first one. This allows for smarter resolution of `Canonical(...)` keywords to ensure that a matched type is resolved when one is available. This change, however, required that I add support for `fishForMetadatas` in all the `Fishable` implementations (not just FHIRDefinitions, but also Tank, Package, some exporters, etc.).

Backstory for `fishForMetadatas` change: Regression failed when I added support for fishing on all resource types because it introduced new types that _could_ now come up in search results _before_ the types it returned in previous versions. In one specific case, an IG was assigning [CarePlan.instantiatesCanonical](https://hl7.org/fhir/R4/careplan-definitions.html#CarePlan.instantiatesCanonical) to `Canonical(SomeGuideline)`. In previous versions of SUSHI, `Canonical(SomeGuideline)` resolved to a `PlanDefinition` defined in `input/fsh` -- but with support for finding additional resource types in packages and predefined inputs, it was now resolving to a `Library` in `input/resources`. This was correct behavior because SUSHI prioritizes predefined resource above FSH definitions (to allow for "patched" resources to override FSH resources) -- but it caused errors because `CarePlan.instantiatesCanonical` allowed for `canonical<PlanDefinition>` but not `canonical<Library>`, so it threw a mismatched type exception (_Canonical(Library) is not one of the allowed types..._). This seemed like an unfortunate error since there _was_ a valid type in scope (the `PlanDefinition` defined in FSH), so I updated SUSHI to look through all the matches if the first match does not satisfy the canonical requirements. Now the regression passes with no changes.

**Testing Instructions:** I've created a simple project that tests both aspects of this: (1) That additional resource types are now loaded (in this case, from predefined resource), and (2) that SUSHI intelligently chooses the right resource based on required Canonical type. You can try running the project using SUSHI 3.13.1, and then this PR branch. The project: [AllTypesTest.zip](https://github.com/user-attachments/files/18230610/AllTypesTest.zip)

**Related Issues:**
* #988 
* #1389 
* #1043


test/fhirtypes/ElementDefinition.assignString.test.ts
```diff
@@ -339,6 +339,7 @@ describe('ElementDefinition', () => {
       );
     });
 
+    // MAYBE?
```
mint-thompson, line 342: That's... a good question. Should assigning a string be allowed if it corresponds to a Canonical URL of the appropriate type? Or more likely, from the other direction: should it be allowed as long as it's a valid URI, and if we find something with that url, it has the appropriate type? It's extra steps to check, but it would nice to allow it, particularly for cases when it's the canonical of something that is not in SUSHI's scope.

mint-thompson (approved):
These changes look good, and the sample project you provided produces the expected output. I left one comment that contains a question, but that's something that would be a future task. Thank you!

jafeltra (approved):
This all makes sense to me!

test/fhirtypes/ElementDefinition.assignString.test.ts
cmoesel: TBH, I don't remember why I wrote this comment -- and I suspect I left it in here by mistake!

I think I'm fine w/ keeping behavior as it is. We have traditionally allowed string-based assignment to references as a back-door way to force a specific reference URL without checking it. I think canonical should probably be the same. If you want extra type-checking, using `Canonical(...)`, otherwise use a string to force whatever URL value you want.

cmoesel: But I am going to delete this comment.

mint-thompson (approved):

Issue #1544, opened 2024-12-13
Cross-version extensions for choice types should escape [ and ] in URL


cmoesel:
Currently, SUSHI only recognizes cross-version extensions for choice types if they use unescaped `[` and `]` in their URL. E.g.:
```
http://hl7.org/fhir/5.0/StructureDefinition/extension-Questionnaire.versionAlgorithm[x]
```

[RFC 2732](http://www.ietf.org/rfc/rfc2732.txt), however, reserves `[` and `]` for IPv6 host syntax, which means that they should be escaped in URL paths. Based on that, the above extension should actually be:
```
http://hl7.org/fhir/5.0/StructureDefinition/extension-Questionnaire.versionAlgorithm%5Bx%5D
```

SUSHI should be liberal about what it accepts and conservative about what it produces. Therefore:
* SUSHI should accept both `[x]` and `%5Bx%5D` in cross-version extension URLs used in FSH
* SUSHI should always use `%5Bx%5D` when generating JSON definitions with these URLs

This came up on the [Issue with polymorphic inter-version extensions](https://chat.fhir.org/#narrow/channel/215610-shorthand/topic/Issue.20with.20polymorphic.20inter-version.20extensions/near/488766521) Zulip thread.

cmoesel:
Addressed by #1549

PR #1549, opened 2024-12-23
URL Encode Cross-Version Extension URLs

cmoesel:
**Description:** Cross-version extensions can represent choice elements, which means that `[x]` might be part of the extension URL -- in which case, it should be URL-encoded (per specifications) to `%5Bx%5D`. This also means that the corresponding `id` should also be URL-encoded to match the tail-end of the URL. Similarly, relative URLs in sub-extensions of complex extensions should also be URL-encoded and the slice names should match the URL-encoded relative URLs. See this Zulip discussion for reference: https://chat.fhir.org/#narrow/channel/215610-shorthand/topic/Issue.20with.20polymorphic.20inter-version.20extensions/near/488766521

This PR does all those things. It follows the approach of being liberal in what it accepts and conservative in what it produces -- which is to say that authors can use unencoded `[x]` in the URLs and slice names and SUSHI will correctly handle them, but SUSHI will always produce outputs with encoded `%5Bx%5D`. This PR also fixes the `parseFSHPath` function  because the old implementation did not deal well with slice names containing `[x]` -- and, as it turns out, the old approach allowed for syntactically incorrect paths like `component[1}.code` or `component[systolic}.code`.

**Testing Instructions:** See the tests for different variations of examples. One thing you can try:
```
Profile: MyQuestionnaire
Parent: Questionnaire
* extension contains http://hl7.org/fhir/5.0/StructureDefinition/extension-Questionnaire.versionAlgorithm%5Bx%5D named r5VersionAlgorithm 0..1
* extension[r5VersionAlgorithm].value[x] only string

Instance: MyQuestionnaireInstance
InstanceOf: MyQuestionnaire
* status = #active
* extension[r5VersionAlgorithm].valueString = "My Algorithm"
```
In the above case, the master branch will not recognize this cross-version extension, but this PR branch should.

You can also use the same example above, but replace the `contains` rule with this one:
```
* extension contains http://hl7.org/fhir/5.0/StructureDefinition/extension-Questionnaire.versionAlgorithm[x] named r5VersionAlgorithm 0..1
```
In this case, the master branch will find the extension, but its output will _not_ encode the URL. On this PR branch, it will find the extension but also output the URL in its encoded form.

**Related Issue:** #1544


cmoesel:
NOTE: Running regression does detect changes in some repos. For most, it is just the difference of `[x]` vs `%5Bx%5D` in the output. In one case, however, it also detects an invalid path (that previously went by undetected) and issues an error (as well as different output since it cannot process the path correctly).

Affected repos (if you want to see for yourself):
```
HL7/fhir-us-mcc#master
HL7/sdc#master
HL7/davinci-crd#master
hl7-eu/Oncology#master
hl7-eu/laboratory#master
hl7ch/ch-lab-report#master
HL7/uv-lab-rep-ig#main
```

test/utils/PathUtils.test.ts
```diff
@@ -362,6 +370,63 @@ describe('PathUtils', () => {
...
+        slices: ['foo[x]']
+      });
+      expect(pathParts[1]).toEqual({
+        base: 'item2',
+        brackets: ['11', 'bar[x]', 'baz[x]'],
+        slices: ['foo[x]', 'bar[x]', 'baz[x]']
+      });
+      expect(pathParts[2]).toEqual({
+        base: 'value[x]'
+      });
```
mint-thompson, line 392: I had forgotten that `parseFSHPath` doesn't include anything about previous slices for path parts that don't have any slices of their own. It's what we want, but I got scared for a minute.

src/utils/PathUtils.ts
```diff
@@ -2,48 +2,50 @@ import { flatten } from 'lodash';
...
-        brackets = brackets.slice(1);
+    const parsedPart: { base: string; brackets?: string[]; slices?: string[] } = {
+      base: pathPart.match(/^([^\[]+(\[x\])?)/)?.[0] ?? ''
+    };
+    if (pathPart.length > parsedPart.base.length) {
+      // Get the content from the outermost bracket pairs. (?![^\[]*\]) ensures we don't
+      // match nested closing brackets (thank you, claude.ai)
+      parsedPart.brackets = Array.from(
+        pathPart.slice(parsedPart.base.length).matchAll(/\[([^\[\]]|\[(?:[^\[\]]*)\])*\]/g)
+      ).map(match => match[0].slice(1, -1));
```
mint-thompson, line 27: It looks like claude had the right regular expression in the code, but not the comment. Typical claude!

mint-thompson (changes requested):
This looks good! I only have one minor change to request. It feels a little weird that we're breaking some of the FHIR spec around `id`, but since this is a temporary measure to maintain tool compatibility until the trailing `[x]`s get removed from extensions, I think it's fine.

test/utils/PathUtils.test.ts
cmoesel: Yeah, me too. I originally rewrote this to include the previous slices even in this case -- and it broken a bunch of tests, so I reverted back to the original approach.

src/utils/PathUtils.ts
cmoesel: Typical claude!

cmoesel: Fixed in b6caaee.

mint-thompson (approved):
hooray for cross-version extensions!

jafeltra (approved):
This looks good! The repos with regressions all make sense to me.

PR #1551, opened 2024-12-31
Use latest hl7.fhir.uv.tools release-specific package

cmoesel:
**Description:** The IG Publisher now uses the "latest" release-specific version of `hl7.fhir.uv.tools` (e.g., `hl7.fhir.uv.tools.r4#0.3.0`) instead of a non-release-specific current build. This PR updates SUSHI to do the same. As a result, automatic dependency loading no longer uses any `current` builds, so it is more compatible with FSH Online.

**Testing Instructions:** Run this branch against an R4 IG and confirm it loads the most recent R4 tools package (currently 0.3.0). Also run against an R5 IG and confirm it loads the most recent R5 tools package (also 0.3.0).

I ran a full regression against master and there were no changes, so this change does not impact any current projects.

**Related Issue:** N/A

### DIFF
--- a/src/utils/Processing.ts
+++ b/src/utils/Processing.ts
@@ -51,8 +51,14 @@ type FshFhirMapping = {
 // For some context on implicit packages, see: https://chat.fhir.org/#narrow/stream/179239-tooling/topic/New.20Implicit.20Package/near/325318949
 export const AUTOMATIC_DEPENDENCIES: AutomaticDependency[] = [
   {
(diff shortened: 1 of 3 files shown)


jafeltra (approved):
Looks good! I tested that the correct versions are used based on the FHIR version and that the package is only loaded once if it is also provided in the dependencies.

mint-thompson (approved):

PR #1572, opened 2025-06-11
Fix sdf-8a validation errors by always including root element in differential

### LINKED ISSUE #1571: StructureDefintions with differentials fail on constraint sdf-8a
Medplum's fhir validator https://www.medplum.com/docs/sdk/core.evalfhirpathtyped uses the fhirpath constraint definitions to validate resources.

fsh scripts that generate differentials in the output StructureDefinitions are failing on the following contraint

> sdf-8a from https://build.fhir.org/structuredefinition-definitions.html
In any differential, all the elements must start with the StructureDefinition's specified type for non-logical models, or with the same type name for logical models

and the fhirpath expression
```
(%resource.kind = 'logical' or element.first().path.startsWith(%resource.type)) and (element.tail().empty() or element.tail().all(path.startsWith(%resource.differential.element.first().path.replaceMatches('\\..*','')&'.')))
```

Consider the following fsh script

```
Profile: Medplum_Tenant
Parent: Organization
Id: MedplumTenant
Title: "Medplum Tenant"
Description: "Medplum Tenant"

* name 1..1
* address 1..1
```

In the differential section of the structure definition, this will produce the following

```
"differential": {
  "element": [
    {
      "id": "Organization.name",
      "path": "Organization.name",
      "min": 1
    },
    {
      "id": "Organization.address",
      "path": "Organization.address",
      "min": 1
    }
  ]
}
```

This fails validation against constraint sdf-8a.

Instead the differential section should look like the following:

```
"differential": {
  "element": [
    {
      "id": "Organization",
      "path": "Organization"
    },
    {
      "id": "Organization.name",
      "path": "Organization.name",
      "min": 1
    },
    {
      "id": "Organization.address",
      "path": "Organization.address",
      "min": 1
     }
  ]
}
```

ianplunkett:
### Problem
FHIR StructureDefinitions generated from FSH scripts were failing validation against constraint **sdf-8a** when processed by Medplum's FHIR validator. The constraint requires that "In any differential, all the elements must start with the StructureDefinition's specified type for non-logical models, or with the same type name for logical models."

**Root Cause**: The differential section was missing the root element (e.g., `Organization`) and only included child elements that had changes (e.g., `Organization.name`). This violated the sdf-8a constraint which expects the first element in the differential to be the root type.

**Example Failure Case**:
```fsh
Profile: Medplum_Tenant
Parent: Organization
* name 1..1
```

This generated a differential missing the root element:
```json
"differential": {
  "element": [
    {
      "id": "Organization.name",
      "path": "Organization.name", 
      "min": 1
    }
  ]
}
```

### Solution
Modified the `toJSON()` method in `StructureDefinition.ts` to **always include the root element** in the differential section for both logical models and resources (derivation = 'specialization').

**Key Changes**:
1. **Always add root element first**: Extract the root element and add its diff to the differential
2. **Process remaining elements**: Add diffs for elements that have changes (slice(1) to skip root)
3. **Cleaner logic**: Simplified the conditional logic by handling root element separately

**Fixed Output**:
```json
"differential": {
  "element": [
    {
      "id": "Organization",
      "path": "Organization"
    },
    {
      "id": "Organization.name",
      "path": "Organization.name",
      "min": 1
    }
  ]
}
```

### Testing
Updated all affected test cases to expect the additional root element in differential sections:
- ✅ Profile generation tests now expect `length + 1` elements
- ✅ All test assertions updated to include root element validation
- ✅ Complex scenarios with slicing and nested elements properly tested

### Impact
- ✅ **Fixes sdf-8a validation failures** with Medplum and other FHIR validators
- ✅ **Maintains backward compatibility** - only adds the required root element
- ✅ **Consistent behavior** across all StructureDefinition types
- ✅ **No breaking changes** to existing functionality

**Files Changed**:
- `src/fhirtypes/StructureDefinition.ts` - Core logic fix
- `test/export/StructureDefinitionExporter.test.ts` - Updated test expectations  
- `test/fhirtypes/StructureDefinition.test.ts` - Updated test expectations

Closes #1571

### DIFF
--- a/src/fhirtypes/StructureDefinition.ts
+++ b/src/fhirtypes/StructureDefinition.ts
@@ -514,10 +514,14 @@ export class StructureDefinition {
 
     // Populate the differential
     j.differential = { element: [] };
-    this.elements.forEach((e, idx) => {
-      // If this is a logical model or a resource (derivation = 'specialization'),
-      // we need to make sure the root element is included in the differential.
-      if (e.hasDiff() || (this.derivation === 'specialization' && idx === 0)) {
+    // Always include the root element for both logical models and resources
+    const rootElement = this.elements[0];
+    const rootDiff = rootElement.calculateDiff().toJSON();
+    j.differential.element.push(rootDiff);
+
+    // Add other elements that have differences
+    this.elements.slice(1).forEach(e => {
+      if (e.hasDiff()) {
         const diff = e.calculateDiff().toJSON();
         j.differential.element.push(diff);
       }
--- a/test/export/StructureDefinitionExporter.test.ts
+++ b/test/export/StructureDefinitionExporter.test.ts
@@ -10524,8 +10524,9 @@ describe('StructureDefinitionExporter R4', () => {
       const sd = pkg.profiles[0];
       const json = sd.toJSON();
 
-      expect(json.differential.element).toHaveLength(1);
+      expect(json.differential.element).toHaveLength(2);
       expect(json.differential.element).toEqual([
+        { id: 'Observation', path: 'Observation' },
         { id: 'Observation.subject', path: 'Observation.subject', min: 1 }
       ]);
     });
@@ -10552,8 +10553,12 @@ describe('StructureDefinitionExporter R4', () => {
       const sd = pkg.profiles[0];
       const json = sd.toJSON();
 
-      expect(json.differential.element).toHaveLength(2);
+      expect(json.differential.element).toHaveLength(3);
       expect(json.differential.element).toEqual([
+        {
+          id: 'Observation',
+          path: 'Observation'
+        },
         {
           id: 'Observation.code.coding',
           path: 'Observation.code.coding',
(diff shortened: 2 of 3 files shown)


cmoesel:
Thanks for this PR. I'm having trouble understanding how sdf-8a requires a root element in the differential? What about that constraint says that?

In fact, the FHIR specification explicitly says that the root element is _not_ required in the section marked [Differential vs. Snapshot](https://hl7.org/fhir/R5/profiling.html#snapshot) (**emphasis** mine):
> Differential statements describe only the differences that they make relative to the structure definition they constrain (which is most often the base FHIR resource or datatype). For example, a profile may make a single element mandatory (cardinality 1..1). In the example of a differential structure, it will contain a single element with the path of the element being made mandatory, and a cardinality statement. Nothing else is stated - all the rest of the structural information is implied (note that this means that **a differential profile can be sparse and only mention the elements that are changed, without having to list the full structure. This rule includes the root element - it is not needed in a sparse differential**).

ianplunkett:
closing see #1571

PR #1586, opened 2025-09-13
Improve error message for unsupported versions of FHIR

grahamegrieve:
Description: Improved error message from Sushi when version isn't suitable


cmoesel (approved):
I like this change. Thanks @grahamegrieve!

Issue #1564, opened 2025-04-11
SUSHI constrains too much when type constraining to a choice of resource and a profile of that resource


cmoesel:
Given:
```
* identifier only Identifier or MyIdentifier
```
SUSHI produces the type:
```json
"type": [
  {
    "code": "Identifier",
    "profile": [
      "http://example.org/StructureDefinition/MyIdentifier"
    ]
  }
],
```
[FSH Online Link](https://fshonline.fshschool.org/#/share/eJwLKMpPy8xJtVLwrQxJLM7mCkgsSs0rsVIAc7QUMlOAvMy0zNQiBd9gVH5+Xk6lgicSH6ikEsHn4gpAGI0kDLMASUhLobiyuCQ1V8FQT88QAGrEMjM=)

This is incorrect because any `Identifier` should be allowed, but the constraint SUSHI outputs requires instances to conform to `MyIdentifier` profile.

The correct behavior should be either a no-op (because the adding the profile to the choice is meaningless since all Identifiers are allowed) or should list the canonical for the resource in the profile array (which more closely aligns with what the user asked for):
```
"type": [
  {
    "code": "Identifier",
    "profile": [
      "http://hl7.org/fhir/StructureDefinition/Identifier",
      "http://example.org/StructureDefinition/MyIdentifier"
    ]
  }
],
```

See: [#shorthand > Issue specifying two types, one not rendered @ 💬](https://chat.fhir.org/#narrow/channel/215610-shorthand/topic/Issue.20specifying.20two.20types.2C.20one.20not.20rendered/near/511061454)

Teruk15:
Hello, @cmoesel  may I work on this?

cmoesel:
Hi @Teruk15. Yes, absolutely! See the [SUSHI README Developer Instructions](https://github.com/fhir/sushi?tab=readme-ov-file#sushi-developer-instructions) for core information about SUSHI development. We also have files on contributing, best practices, and code of conduct in the root of the repository.

Thanks for offering to help. I look forward to you PR. Let me know if you run into any issues. I marked this as a "good first issue" based on my recollection of the code -- but it's always possible that it might turn out more complicated than expected.

JayMurdoch:
Thanks for raising this @cmoesel - very much appreciated.

cmoesel:
@Teruk15 - Are you still interested in taking on this task?

@JayMurdoch - I should have thought to mention this before, but there is a workaround for this issue. Instead of saying:
```
[...]

PR #1589, opened 2025-09-18
Allow base types as profiles in type constraint rules

benjaminflessner:
**Description:**
Provides the ability to specify both a profile and the base type as a given only-rule. Previously, when calculating the intersection between types and profiles, if the profile === type, it was just dropped. Now it will be included as an additional profile, but only if it was not the only instance of that type  (e.g. setting `* resource only Observation` will still only set the type to be only "Observation", but `* resource only CustomProfile or Observation` will set the type as Observation and the profile as a 2-element array containing CustomProfile and Observation.

Found a failing test re: duplicate types after attempting my monkey-patch. I'm not sure if there was more that the test was expecting, but throwing a warning for duplicate types seems to be a side-effect of the code area I changed. To more specifically prevent this, I added a `uniqWith` on the initial line of `constrainType` so truly duplicate entries will be filtered (and then later the warning will trigger because the counts are different).

**Testing Instructions:**
Tests should continue to pass. Can create a simple IG with a field that contains both a profile and its type in the only list and verify that the type appears as profile alongside the specified profile.

**Related Issue:**
https://github.com/FHIR/sushi/issues/1564

### DIFF
--- a/src/fhirtypes/ElementDefinition.ts
+++ b/src/fhirtypes/ElementDefinition.ts
@@ -1048,7 +1048,7 @@ export class ElementDefinition {
    * @throws {SliceTypeRemovalError} when a rule would eliminate all types on a slice
    */
   constrainType(rule: OnlyRule | AddElementRule, fisher: Fishable, target?: string): void {
-    const types = rule.types;
+    const types = uniqWith(rule.types, isEqual);
     // Establish the target types (if applicable)
     const targetType = this.getTargetType(target, fisher);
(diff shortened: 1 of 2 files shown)


cmoesel (approved):
Nice work, @benjaminflessner. I've run a full regression and the only changes are ones that we would expect: profiles that constrained types to resources and profiles and previously had the resource erroneously filtered out. I also saw a few changes where authors had mistakenly repeated types in their `only` rule (and now the duplicates are removed). So it's all good!

PR #1599, opened 2025-11-21
Update init message to mention modern sushi syntax

vadi2:


vadi2:
Mm, I think `sushi build` is great for scripts (removes ambiguity as to what is desired) but for humans, `sushi` is simpler and easier.

cmoesel (approved):

PR #1620, opened 2026-04-13
Use Official Cross-Version (xver) Packages

cmoesel:
**Description:** SUSHI implemented cross-version extensions [in 2021](https://github.com/FHIR/sushi/pull/805) by allowing authors to put special cross-version extension packages in their dependencies. The package names and versions followed the conventions described in the spec [here](https://hl7.org/fhir/R4/versions.html#extensions) and we interpreted it as `hl7.fhir.extensions.r5:4.0.1` indicating the R5 extensions in the R4 format. We processed cross-version extensions by using pattern-matching to find the corresponding targeted element in the target version of FHIR (which we loaded as "supplemental" FHIR definitions).

Now that official cross-version packages are available (see [FAQs](https://confluence.hl7.org/spaces/FHIRI/pages/413256623/FAQs) and [#IG creation > issue with link to xver extension @ 💬](https://chat.fhir.org/#narrow/channel/179252-IG-creation/topic/issue.20with.20link.20to.20xver.20extension/near/582031656)), SUSHI's old implementation is no longer necessary. This PR makes the following changes:
* Removes all code for the old implementation of cross-version extensions (often called "implied" extensions in the code) and "supplemental" FHIR definitions (which are no longer needed).
* Replaces old cross-version package dependencies (e.g., `hl7.fhir.extensions.r5:4.0.1`) with the corresponding new cross-version package dependencies (e.g., `hl7.fhir.uv.xver-r5.r4`) and logs a warning telling the user to fix it in their sushi-config.yaml.
* Detects use of cross-version choice extensions ending in `[x]` or URL-encoded `%5Bx%5D` and strips the `[x]` suffix since the new cross-version packages do not expect it. In this case, SUSHI also warns the user to fix it in their source code.
* Detects use of cross-version extensions that are not valid according to the new cross-version package (i.e., the cross-version package is loaded but the extension can't be found) and issues an error directing the user to the cross-version package IG documentation.
* Detects use of cross-version extensions with no cross-version extension package loaded and issues an error directing the user to add the appropriate cross-version extension package to their dependencies.

Since SUSHI implemented cross-version extensions before all of the rules were worked out, there are differences between SUSHI's implementation and the final implementation. As a result, some IGs will see new errors for cross-version extensions that differ between the two implementations.

I ran a regression over 1000+ existing SUSHI projects and found there were 61 projects that used the old SUSHI cross-version extension packages. Of those 61, here are how the regression results turned out:
* 13 projects with new errors and changes in output
* 7 projects with no new errors and with incompatible output changes due to use of extensions ending in `[x]`
* 9 projects with no new errors and with compatible output changes
* 32 projects with no new errors and with the only output change being the new xver dependency in the IG JSON file

I've attached a zip of the regression results. You may review them if you'd like by unzipping it and opening the `index.html` file in your browser: [xver-regression.zip](https://github.com/user-attachments/files/26677284/xver-regression.zip)

I'm also including a screenshot of the affect projects below:

CleanShot 2026-04-13 at 09 49 03@2x

**Testing Instructions:**

Review the regression results above to ensure they are as expected.

Create a project with cross-version extensions using SUSHI's old cross-version extension packages or clone one of the affected projects from the regression. Run this version of SUSHI against it to determine it works as expected (`ts-node src/app.ts build /path/to/project`).

**Related Issue:** N/A


jduteau (approved):
argh, this affects one of my guides!  But a good PR in any case.

PR #1629, opened 2026-06-24
Fix processing of slices on choice elements

cmoesel:
**Description:** When manually creating slices on choice elements (e.g., `Observation.value[x]`), SUSHI runs into a few issues:

1. If you constrain a slice by type (using `only`) and then constrain another by type (also using `only`), in some cases SUSHI emits an error like this: _Type constraint on Observation.value[x] makes rules in ObservationWithValueAttachment obsolete for choices: valueString_.
2. If you attempt to traverse into a sub-element of a slice, in some cases SUSHI emits an error like this: _No element found at path value[x][valueAttachment].extension for ContainsRule in ObservationWithValueAttachment, skipping rule_. (NOTE: error depends on type of rule applied to the path).

These issues were discovered and discussed on Zulip: [#shorthand > Slicing and extending with a xver extension @ 💬](https://chat.fhir.org/#narrow/channel/215610-shorthand/topic/Slicing.20and.20extending.20with.20a.20xver.20extension/near/604137635).

This PR resolves both issues by improving handling for the special case of slices on choice elements.

**Testing Instructions:** The unit tests show the issue, but you may also test manually using this [XVerObsValueAttachment.zip](https://github.com/user-attachments/files/29293911/XVerObsValueAttachment.zip) project based on the Zulip thread use case. First run the release version of SUSHI to confirm the errors are being logged. Then run this PR branch (using `ts-node`) to confirm the errors go away and the resulting StructureDefinition looks correct.

**Related Issue:** See https://chat.fhir.org/#narrow/channel/215610-shorthand/topic/Slicing.20and.20extending.20with.20a.20xver.20extension/near/604137635.

### DIFF
--- a/src/fhirtypes/StructureDefinition.ts
+++ b/src/fhirtypes/StructureDefinition.ts
@@ -301,7 +301,12 @@ export class StructureDefinition {
             unfoldedElements = matchingElements[0].unfoldChoiceElementTypes(fisher);
             newMatchingElements = unfoldedElements.filter(e => e.path.startsWith(fhirPathString));
           }
-        } else if (matchingElements[0].id.endsWith('[x]')) {
+        } else if (
+          matchingElements[0].id.endsWith('[x]') ||
+          (matchingElements[0].path.endsWith('[x]') && matchingElements[0].sliceName != null)
+        ) {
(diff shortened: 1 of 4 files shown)


jduteau (approved):
Looks good to me

PR #1630, opened 2026-07-10
Fix versioned extension checks

GinoCanessa:
# Fix false validation errors for versioned extension slices and colliding slice names

## Summary

This branch fixes two related, high-impact false-positive validation errors that
cause SUSHI to exit non-zero ("shipwreck") on IGs that use cross-version
extension packages (`hl7.fhir.uv.xver-*`), even though the generated FHIR JSON is
correct. Both are surfaced by the **FHIR Subscriptions R5 Backport IG**
(`hl7.fhir.uv.subscriptions-backport`), which today reports 31 + 12 = **43 false
errors**; with this change it validates cleanly.

Neither fix alters the generated FHIR JSON - they only remove false-positive
diagnostics from the required-element and extension-usage validators.

## Motivation

When a profile is supplied by a cross-version extension package, its extension
slices pin a **version** on `type.profile`
(e.g. `…/extension-SubscriptionStatus.subscription|0.1.0`), while instances
reference the extension by its **unversioned** canonical URL. SUSHI compared these
with exact string equality in a few places, and separately identified a slice's
extension by fishing the **bare slice name**. Both assumptions break for
cross-version IGs:

1. **Versioned slice never matches (31 false errors).** In
   `StructureDefinition.findMatchingSlice`, a versioned `type.profile`
   (`…|0.1.0`) never `===` the unversioned resolved extension URL, so the assigned
   extension is never bound to its named slice. `validateRequiredChildElements`
   then counts 0 occurrences and emits
   `Element … has minimum cardinality 1 but occurs 0 time(s).` even though the
   element is present in the output.

2. **Slice name collides with an unrelated extension (12 false errors).**
   InstanceExporter's "were extensions used correctly along the path?" check
   fished the bare slice name as an extension. A slice named `type` resolves to
   R4/R4B core `familymemberhistory-type` (whose `name` is literally `"type"`, and
   which is a non-modifier), so a genuine modifier extension assigned via
   `modifierExtension[type]` is falsely flagged
   `Non-modifier extension type used on modifierExtension element.` (This becomes
   observable once fix #1 correctly rewrites the path to the profile slice name.)

Both patterns are FHIR-version-agnostic (R4 and R4B are equally affected) and the
IG author cannot fix them from the FSH side, because the offending version/slice
name comes from a published dependency package.

### Version-insensitive canonical comparison

- **`src/fhirtypes/common.ts`** - new `canonicalsAreEqualIgnoringVersion(a, b)`
  helper. It compares the two values **exactly as-presented first** (so a URL that
  legitimately contains a literal `|` still matches when both sides carry the same
  literal), and only falls back to stripping the suspected trailing `|version`
  (everything after the **last** `|`) when the exact compare fails. This is
  deliberately pipe-safe and intentionally differs from `FishingUtils`, which
  splits on the *first* `|` to peel off a fishing key; that path is left unchanged.

### Fix 1 - versioned extension-slice matching

- **`src/fhirtypes/StructureDefinition.ts`** (`findMatchingSlice`) - match an
  extension slice by comparing `type.profile` against the resolved extension URL
  with the new helper instead of `===`. When a match is achieved **only** by
  ignoring a pinned version that differs from the resolved extension's actual
  version, emit a **debug-level** note (no warn/error - the author can change
  neither the unversioned instance URL nor the version pinned in a published
  package). Stays silent when versions agree or either side is unversioned.
- **`src/export/StructureDefinitionExporter.ts`** - when a duplicate extension
  `ContainsRule` slice is version-pinned, compare with the helper so it is treated
  as a harmless duplicate (**warning**, not error), matching the pre-existing
  behavior for unversioned duplicates.

### Fix 2 - slice-name / extension-name collision

- **`src/export/InstanceExporter.ts`** - new `resolveSliceExtension(...)` resolves
  the Extension bound to a sliced `(modifier)extension` path part via the matched
  slice's **`type.profile`** (fished version-insensitively) rather than by fishing
  the bare slice name. Applied to **both** mirror-image checks (a non-modifier
  flagged on a `modifierExtension` slice, and a modifier flagged on an `extension`
  slice), since they share the same root cause. The reliable URL-based first check
  is left untouched.

## Testing

- **`test/fhirtypes/canonicalsAreEqualIgnoringVersion.test.ts`** - 10 unit tests
  for the helper: unversioned/versioned/mixed equality, differing versions,
  differing base URLs, the literal-`|`-in-URL exact-match path and its symmetric
  strip limitation, and null/undefined handling (no throw).
- **`test/fhirtypes/StructureDefinition.test.ts`** - a version-pinned
  `type.profile` slice matched via the unversioned URL, plus the debug drift note
  firing on version mismatch and staying silent when versions agree.
- **`test/export/StructureDefinitionExporter.test.ts`** - a version-pinned
  duplicate `ContainsRule` slice is downgraded from error to warning.
- **`test/export/InstanceExporter.test.ts`** - no false
  `Non-modifier extension …` / `Modifier extension …` errors when a slice name
  collides with an unrelated extension's `name` (both directions), and a required
  version-pinned extension slice referenced by unversioned URL is assigned without
  a false `minimum cardinality … occurs 0 time(s)` error.

All affected suites pass locally (`915` passing across the four suites; `10/10`
for the helper).

### How to reproduce / verify against a real IG

```
npm ci && npm run build
node dist/app.js <path-to-fhir-subscription-backport-ig branch:tickets>
```

Before: `31 Errors` (min-cardinality), then `12 Errors` (non-modifier) once the
first fix lands. After: **0 errors** from these two causes, with byte-for-byte
identical generated JSON.

## Compatibility & risk

- No change to generated FHIR JSON - only false-positive validation errors are
  removed. New diagnostics are **debug-level** only.
- Version-insensitive comparison is pipe-safe (exact match first), so URLs that
  contain a literal `|` are unaffected.
- The InstanceExporter lookup reuses the already-matched slice's element via
  `findElementByPath`, so it does not introduce new external fetches.

## Related

N/A (have not filed issues for this)

## Pre-merge checklist

- [x] Targeted Jest suites for the changed files pass (`npm test -- <files>`)
- [x] `npm run check` (full `test` + `lint` + `prettier`) is clean
  - Test coverage for all additions in this PR - does not address existing gaps.
- [x] Branch is rebased/merged on the latest `master`
- [x] `npm audit` shows no new vulnerabilities (no dependency changes in this PR)


src/export/InstanceExporter.ts
```diff
@@ -409,6 +414,50 @@ export class InstanceExporter implements Fishable {
...
+      return undefined;
+    }
+    let extension = this.fisher.fishForFHIR(profileUrl, Type.Extension);
+    if (extension == null && profileUrl.includes('|')) {
+      extension = this.fisher.fishForFHIR(
+        profileUrl.slice(0, profileUrl.lastIndexOf('|')),
+        Type.Extension
+      );
+    }
+    return extension;
```
cmoesel: This is pretty much what `fishForFHIRBestVersion` does, so you can replace this whole thing with `return fishForFHIRBestVersion(this.fisher, profileUrl, undefined, Type.Extension)`.

src/export/InstanceExporter.ts
```diff
@@ -409,6 +414,50 @@ export class InstanceExporter implements Fishable {
...
+    // part's trailing numeric occurrence index, since findElementByPath treats brackets as slice
+    // identifiers rather than array positions.
+    const path = pathParts
+      .slice(0, index + 1)
+      .map(pp => {
+        const brackets =
+          getArrayIndex(pp) == null ? (pp.brackets ?? []) : (pp.brackets ?? []).slice(0, -1);
+        return pp.base + brackets.map(b => `[${b}]`).join('');
+      })
+      .join('.');
```
cmoesel: I think you can simplify this a little using the existing `assembleFSHPath` function:

```ts
const path = assembleFSHPath(
  pathParts
    .slice(0, index + 1)
    .map(pp => (getArrayIndex(pp) == null ? pp : { ...pp, brackets: pp.brackets.slice(0, -1) }))
);
```

src/export/StructureDefinitionExporter.ts
```diff
@@ -1260,7 +1261,11 @@ export class StructureDefinitionExporter implements Fishable {
           // then it is most likely a harmless no-op.  In this case, treat it as a warning.
           if (e instanceof DuplicateSliceError) {
             const slice = element.getSlices().find(el => el.sliceName === item.name);
-            if (slice?.type[0]?.profile?.some(p => p === extension.url)) {
+            if (
+              slice?.type[0]?.profile?.some(p =>
+                canonicalsAreEqualIgnoringVersion(p, extension.url)
+              )
+            ) {
```
cmoesel: I think you can actually just check `p` against `profileUrl` from line 1253 instead of using the `canonicalsAreEqualIgnoringVersion` function. I.e.,

```ts
if (slice?.type[0]?.profile?.some(p => p === profileUrl)) {
```

If you do that, I think we more appropriately distinguish true error conditions from warnings:

1. If `profileUrl` (representing the new slice rule) has a version and the existing slice has the same version, then `profileUrl` === `p`, resulting in only a warning.
2. If `profileUrl` has a version and the existing slice has a different version, then `profileUrl` !== `p`, resulting in an error (which seems appropriate).
3. If `profileUrl` has a version and the existing slice does not have a version, then `profileUrl` !== `p`, resulting in an error (which seems appropriate because technically I _think_ you should re-slice in that case?).
4. If `profileUrl` has no version and the existing slice also has no version, then `profileUrl` === `p`, resulting in only a warning.
5. If `profileUrl` has no version and the existing slice has a version, then `profileUrl` !== `p`, resulting in an error (which seems appropriate since the new rule attempts to widen the slice definition, which I _think_ is not allowed?).

If we do change this to compare against `profileUrl`, then the `profileUrl` assignment will have to happen _before_ the `try` block instead of inside of it. 

My only doubts are that I'm not 100% sure about numbers 3 and 5 above, so if I'm wrong about those, let me know. Either way, if we leave it as-is in this PR, then all five cases above would receive only a warning, which definitely seems wrong for number 2 above.

src/fhirtypes/common.ts
```diff
@@ -1504,6 +1504,32 @@ export function isModifierExtension(extension: any): boolean {
...
+  if (a === b) {
+    return true;
+  }
+  if (a == null || b == null) {
+    return false;
+  }
+  const stripVersion = (s: string): string =>
+    s.includes('|') ? s.slice(0, s.lastIndexOf('|')) : s;
+  return stripVersion(a) === stripVersion(b);
+}
```
cmoesel, line 1524: Consider switching to split on the _first_ `|` instead of the last `|`. Your comment calls out this discrepancy, but I didn't understand the argument for diverging from the existing approach. I'm not aware of a definitive answer regarding which is right (first `|` or last `|`), but...
* [CanonicalResource.url](https://hl7.org/fhir/R5/canonicalresource-definitions.html#CanonicalResource.url) `cnl-1` says `URL should not contain | or #` (constraint severity: warning).
* [CanonicalResource.version](https://hl7.org/fhir/R5/canonicalresource-definitions.html#CanonicalResource.version) says `The version SHOULD NOT contain a '#'` (but there is no formal constraint).

Since the `url` explicitly says you should not use `|` (and has a formal constraint to back it up), while the `version` only says that about `#`, then `"url": "http://my.canonicalwith|bar.com"` would produce a warning whereas `"version": "12|3"` would not. That being the case, I think a bar in `url` is discouraged more, so is therefore less likely to occur -- which is why we should assume that when a canonical has more than one `|`, the literal `|` is more likely to be in the `version`, so we should split on the _first_ bar.

But even if that's not convincing, please still consider doing it anyway just to be consistent with the rest of the SUSHI codebase. ;-)

test/export/StructureDefinitionExporter.test.ts
```diff
@@ -7719,6 +7719,102 @@ describe('StructureDefinitionExporter R4', () => {
...
+        loggerSpy
+          .getAllMessages('error')
+          .some(m => /Slice named precondition already exists/.test(m))
+      ).toBe(false);
+    });
+
+    it('should downgrade a different-version duplicate of the same extension slice from error to warning', () => {
+      // Deliberate side effect of version-insensitive duplicate detection (Phase 3): declaring the
+      // same slice name twice with different versions of the same extension is a warning (SUSHI
+      // keeps only the first profile per slice) rather than a "conflicting duplicate" error.
```
cmoesel: I've already touched on this in a prior comment, but why _shouldn't_ this be an error? It seems to me that it's something the author should fix in their FSH, especially if we're basically tossing the second slice definition anyway (which means the FSH code doesn't reflect the actual final representation of the SD).

src/export/InstanceExporter.ts
```diff
@@ -409,6 +414,50 @@ export class InstanceExporter implements Fishable {
...
   }
 
+  /**
+   * Resolves the Extension definition bound to a sliced (modifier)extension path part via the
+   * matched slice's type.profile in the instance's StructureDefinition, rather than by fishing the
+   * bare slice name. Slice names are arbitrary profile-local identifiers that routinely collide with
+   * the name of unrelated published extensions (e.g. a slice named "type" vs. R4 core
+   * familymemberhistory-type, whose name is literally "type"), so fishing the slice name can resolve
+   * the wrong extension. The type.profile canonical is the authoritative extension URL and is fished
+   * version-insensitively (retrying with the trailing |version stripped if the exact fish misses).
```
cmoesel: Not a super big deal, but... 

a) I'm not sure all of the discussion about why this is better than fishing by the slice name is necessary. It's helpful in the context of this PR, but once the fish-by-slicename logic is gone, then I don't think we really need to convince the reader of this approach.

b) The comment does not reflect the full flexibility of the `extension[sliceName]` syntax. On an extension path, the token inside the brackets _could_ refer to the actual slice name in the profile, but the "sliceName" also could refer to the extension by its `id`, `name`, or `url`. In that case `sliceName` is not referring to an "arbitrary profile-local identifier" but to a defined extension in the project's scope (including dependencies). We allow this so that you can still easily set an extension in an instance even if the profile doesn't declare it.

Anyway, you don't _have_ to remove or edit this comment (I'll accept the PR with the comment as-is), but I think you should consider it.

test/fhirtypes/canonicalsAreEqualIgnoringVersion.test.ts
```diff
@@ -0,0 +1,85 @@
...
+    expect(
+      canonicalsAreEqualIgnoringVersion('http://example.org/Foo|bar', 'http://example.org/Foo|bar')
+    ).toBe(true);
+  });
+
+  it('should NOT match a literal-pipe URL against its versioned form (symmetric strip limitation)', () => {
+    // The helper strips the suspected trailing version from BOTH sides symmetrically, so it
+    // cannot distinguish a literal '|' that is part of the URL from a '|version' suffix on the
+    // shorter side: 'Foo|bar' strips to 'Foo' while 'Foo|bar|0.1.0' strips to 'Foo|bar'.
+    // Per bugreport §1 the literal-'|' case is only guaranteed safe via the exact-match-first
```
cmoesel: bugreport §1?

src/fhirtypes/StructureDefinition.ts
```diff
@@ -908,8 +909,33 @@ export class StructureDefinition {
...
+              pinnedVersion !== sliceDefinition.version
+            ) {
+              logger.debug(
+                `Matched extension slice '${matchingSlice.sliceName}' by canonical URL ` +
+                  `${sliceDefinition.url}, ignoring the version pinned on its type.profile ` +
+                  `(${pinnedVersion}); the resolved extension is version ${sliceDefinition.version}.`
+              );
+            }
+          }
+        }
```
cmoesel, line 930: The PR description describes the use case for an instance setting an extension using it's version-free canonical when the profile has it pinned. I understand why we want to support that use case and let it go.

I can also understand why we might let an instance set an extension using a version-pinned canonical when a profile specifies the extension canonical w/ no version (although that's not the primary use case of this PR).

I don't understand why we'd want to silently ignore a case where the profile pins to one version and the instance sets the extension using a different version. E.g., if the profile says use version `123` then shouldn't we bark if an instance uses version `456`?

cmoesel (changes requested):
Thanks for submitting this PR. I think it's definitely headed in the right direction and it works pretty well. I've also run a regression and confirmed it doesn't break any existing projects that have been modified in the last 6 months or so.

I've left a few questions, comments, and suggestions. Let me know what you think!

src/fhirtypes/common.ts
GinoCanessa: I will switch over to first pipe for consistency within FSH (missed that you had that the other way elsewhere).

I default to last because at a WGM (Phoenix or Dallas?) we had a discussion where someone volunteered that they included vertical pipes in their IGs at the time (they would stop this moving forward, but historical is historical), and there are instances of that in FHIR R4 (e.g., `http://terminology.hl7.org/CodeSystem/v2-0006|2.1` : [v2-006|2.1](https://hl7.org/fhir/R4/v2/0006/2.1/index.html)).

I think that consistency is better in this PR and can review at some point for _all_ checks if it is an issue.

src/fhirtypes/StructureDefinition.ts
GinoCanessa: I believe much of this is cleaned up by switching to `fishForFHIRBestVersion`.  I was leaning towards ignoring the version because _theoretically_ it is ok to do with extensions, but a warning feels better.

I will switch to a warning here and can upgrade to error if preferred.

test/fhirtypes/canonicalsAreEqualIgnoringVersion.test.ts
GinoCanessa: Leakage from internal process. Fixing.

cmoesel (approved):
Thanks again for the contribution. The updates look good! I've run another regression and confirmed this does not cause any issues for existing projects.

PR #1633, opened 2026-08-18
Cache type lookups in MasterFisher

octoshikari:
### Problem

`MasterFisher.fishForFHIR` is called with only `Type.Type` from some very hot paths. The biggest one is `ElementDefinition.isPrimitive`, which runs for every child element of every instance being validated and fishes the type's StructureDefinition just to read its `kind`. The same handful of type codes (`string`, `code`, `uri`, ...) end up being looked up thousands of times per run, and each lookup is a query against the package database plus a possible read of the resource JSON from disk.

In a CPU profile of a build of [HL7/fhir-mCODE-ig](https://github.com/HL7/fhir-mCODE-ig), `isPrimitive` accounted for 36% of the total run time and package database lookups for 47%.

### Change

Cache `fishForFHIR` results when the only requested type is `Type.Type`. That case can only ever be answered by the FHIR definitions — `Package.internalFish` has an explicit "Package doesn't currently support types" branch and `FSHTank.internalFish` skips the type as well — and the definitions don't change while fishing, so the answer is stable for the life of the fisher. Everything else is fished exactly as before, so results that can change as the package fills up are not cached.

### Effect

Best of five runs with a warm package cache, node 24:

| IG | before | after |
| --- | --- | --- |
| HL7/fhir-mCODE-ig | 17.4s | 11.3s |
| HL7/fhir-sdoh-clinicalcare | 19.6s | 14.6s |

For both IGs the generated `fsh-generated` tree is byte-identical to the one produced without this change.

### Tests

Two tests in `MasterFisher.test.ts`: repeated `Type.Type` fishing reaches the FHIR definitions only once, and fishing for other types is still not cached.

### DIFF
--- a/test/utils/MasterFisher.test.ts
+++ b/test/utils/MasterFisher.test.ts
@@ -8,7 +8,7 @@ import { minimalConfig } from './minimalConfig';
 import { cloneDeep } from 'lodash';
 import { getTestFHIRDefinitions, testDefsPath, TestFHIRDefinitions } from '../testhelpers';
 import { PREDEFINED_PACKAGE_NAME, PREDEFINED_PACKAGE_VERSION } from '../../src/ig';
-import { Metadata } from '../../src/utils';
+import { Metadata, Type } from '../../src/utils';
 
 describe('MasterFisher', () => {
   let fisher: MasterFisher;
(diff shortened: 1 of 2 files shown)


test/utils/MasterFisher.test.ts
```diff
@@ -410,6 +410,26 @@ describe('MasterFisher', () => {
...
 
+  it('should only look up a type once when fishing for it repeatedly with only Type.Type', () => {
+    const typeFisher = new MasterFisher(undefined, defs, undefined);
+    const fhirSpy = jest.spyOn(defs, 'fishForFHIR');
+    const first = typeFisher.fishForFHIR('string', Type.Type);
+    const second = typeFisher.fishForFHIR('string', Type.Type);
+    expect(first.id).toBe('string');
+    expect(second).toBe(first);
+    expect(fhirSpy).toHaveBeenCalledTimes(1);
+    fhirSpy.mockRestore();
```
cmoesel: If the test fails, we'll never reach this line and the `fhirSpy` mock won't get restored. I'd suggest either wrapping this in a `try`/`finally` and restoring the mock in the `finally` block, e.g.

```ts
it('should only look up a type once when fishing for it repeatedly with only Type.Type', () => {
  const typeFisher = new MasterFisher(undefined, defs, undefined);
  const fhirSpy = jest.spyOn(defs, 'fishForFHIR');
  try {
    const first = typeFisher.fishForFHIR('string', Type.Type);
    const second = typeFisher.fishForFHIR('string', Type.Type);
    expect(first.id).toBe('string');
    expect(second).toBe(first);
    expect(fhirSpy).toHaveBeenCalledTimes(1);
  } finally {
    fhirSpy.mockRestore();
  }
});
```

... or creating a small `describe` block w/ appropriate setup/teardown methods, e.g.

```ts
describe('type caching', () => {
  let fhirSpy: jest.SpyInstance;
  beforeAll(() => {
    fhirSpy = jest.spyOn(defs, 'fishForFHIR');
  });

  beforeEach(() => {
    fhirSpy.mockClear();
  });

  afterAll(() => {
    fhirSpy.mockRestore();
  });

  // it('...') tests go here
}); 
```

test/utils/MasterFisher.test.ts
```diff
@@ -410,6 +410,26 @@ describe('MasterFisher', () => {
...
+  });
+
+  it('should not cache results when fishing for types other than Type.Type', () => {
+    const typeFisher = new MasterFisher(undefined, defs, undefined);
+    const fhirSpy = jest.spyOn(defs, 'fishForFHIR');
+    typeFisher.fishForFHIR('Patient', Type.Resource);
+    typeFisher.fishForFHIR('Patient', Type.Resource);
+    expect(fhirSpy).toHaveBeenCalledTimes(2);
+    fhirSpy.mockRestore();
```
cmoesel: Same comment as above regarding the `fhirSpy`.

cmoesel (changes requested):
This looks good to me. I've made a small comment on the tests. It also needs to be rebased or have merged from main. I'm going to run a regression on it (which takes some time), but I don't expect to see any issues.

cmoesel (approved):
Thanks for the updates. This looks great and I'm sure users will appreciate the speed boost!

PR #1634, opened 2026-08-18
Don't block the build on the SUSHI version check

octoshikari:
### Problem

`checkSushiVersion()` is awaited before any FSH is read. It shells out to `npm view fsh-sushi version` (with an HTTP request to the npm registry as a fallback), which takes about half a second here and can take up to the full 5 second timeout on a slow or offline network. Nothing in the build needs the result until the results box is printed at the end.

### Change

Start the version check without awaiting it, then await it where it is used. `--require-latest` still awaits it up front so that it can fail before doing any work; everything else overlaps the check with the build.

No change to the output or to any of the messages.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -192,19 +192,20 @@ async function runBuild(input: string, program: OptionValues, helpText: string) 
   }
   logger.info(`  ${path.resolve(input || '.')}`);
 
-  const sushiVersions = await checkSushiVersion();
-  if (
-    program.requireLatest &&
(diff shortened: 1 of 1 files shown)


cmoesel (approved):
Thanks for contributing this. It's a smart approach and also helps with another problem: some companies block npm when there is a major npm worm or other security incident, which was causing SUSHI to hang for a long time before starting the build. Now it will still hang, but at least it will do so _after_ it's built everything (which is less annoying since you can quit and still use the build results).

PR #1635, opened 2026-08-23
Fix multiline strings ending in quotes

### LINKED ISSUE #1569: Multi-Line Delimiter fails to parse content when last character is a "
If the `"""` string delimiter is used, the parser fails if the character immediately preceding the closing `"""` is `"`. E.g., the following FSH:
```
Instance: TestPatientBroken
InstanceOf: Patient
* name.text = """Porous Robert "SpongeBob""""
```
results in the following error:
> error extraneous input '"' expecting {<EOF>, KW_ALIAS, KW_PROFILE, KW_EXTENSION, KW_INSTANCE, KW_INVARIANT, KW_VALUESET, KW_CODESYSTEM, KW_RULESET, KW_MAPPING, KW_LOGICAL, KW_RESOURCE}

If you put the text content on its own line, however, you get the expected results. E.g., the following works:
```
Instance: TestPatientWorks
InstanceOf: Patient
* name.text = 
  """
  Porous Robert "SpongeBob"
  """
```

[FSH Online Link](https://fshonline.fshschool.org/#/share/eJzzzCsuScxLTrVSCEktLglILMlMzSsJzy/KLubyhEr5p1kpQCW4tBTyEnNT9UpSK0oUbBW4FBSUlJSAZEB+UX5psUJQflJqUYmCUnBBfl56qlN+khJcBdw0FIucivKzU/OIsAloCE47gAAAYY5ATA==)

Originally reported on Zulip here: [#shorthand > ✔ Escape Parameterized Ruleset @ 💬](https://chat.fhir.org/#narrow/channel/215610-shorthand/topic/.E2.9C.94.20Escape.20Parameterized.20Ruleset/near/520644881).

shin4141:
**Description:**

### Problem

A multiline FSH string whose content ends in a quote creates four consecutive quote characters: one content quote plus the three-character closing delimiter. The current non-greedy lexer rule consumes the first three as the delimiter and leaves the final quote outside the multiline token.

### Fix

Allow one or two quote characters immediately before the final triple-quote delimiter so that the final three quotes remain the closing delimiter while preceding quotes remain string content.

### Tests

Add regression coverage for strings ending in quotes and preserve existing multiline-string behavior.

AI-assisted development was used for investigation and implementation support. I reviewed the final grammar change, generated artifacts, regression coverage, and scope of the submitted fix.

**Testing Instructions:**

Run `npm run check`.

**Related Issue:**

Fixes #1569.


cmoesel (changes requested):
Thanks for fixing this! This looks good and the approach seems simple enough not to cause any trouble. It turns out we have a similar rule in the [MiniFSHLexer.gr](https://github.com/shin4141/sushi/blob/fix/1569-multiline-ending-quotes/antlr/src/main/antlr/MiniFSHLexer.g4#L9). Would you mind updating that as well? That lexer is used when parsing parameterized rulesets. Thanks again!

shin4141:
Thanks for catching that. I updated MiniFSHLexer with the same multiline-string rule and added regression coverage through the parameterized ruleset path. I also regenerated the affected grammar artifacts and ran the full checks. Thanks again for pointing me to the second lexer.

cmoesel (changes requested):
Thanks @shin4141. This looks great and passes regression. It is now out of synch with master. Could you rebase or merge from master? Once it's up-to-date I am ready to approve and merge.

shin4141:
Thanks! I merged the latest `master` into the PR branch, so it is now up to date and has no merge conflicts. The new “Lint and Test” workflow run is currently awaiting approval and has not started any jobs yet.

cmoesel (approved):
Thank you for the contribution! This was an annoying bug, so it's great to have a fix!

PR #1637, opened 2026-09-02
Improve performance with large CodeSystems

vadi2:
**Description:**

SUSHI slows down sharply as a CodeSystem grows. Profiling a 9,347-concept CodeSystem (about 56,000 code caret rules for designations and properties) showed two costs that grow with the number of rules, so this PR fixes both:

1. **Rebuilding the StructureDefinition for every caret rule.** `setPropertyOnDefinitionInstance` called `instance.getOwnStructureDefinition(fisher)` (a package lookup plus `StructureDefinition.fromJSON`) on each call. For a CodeSystem that is once per code caret rule, and it was 83% of the build. The function now takes the StructureDefinition as an optional final parameter (default unchanged), and `CodeSystemExporter` and `ValueSetExporter` pass in the one they already built. Sharing one StructureDefinition across an entity's rules has precedent: the validation pass in these exporters already did it, and `InstanceExporter` reuses cached StructureDefinitions across all of an instance's rules. In the same code path, `findConceptPath` and `setConcepts` index concepts by code instead of scanning the concept list for every rule, and `setPropertyOnInstance` fills a large array from its current end instead of from index 0 (the old loop re-read the array length each iteration, and each iteration pushed one element, so starting at the current length adds exactly the same elements).

2. **Scanning every rule to find a definition's id, url, name, or version.** `findAssignmentByPath` did a `findLast` over all rules, and `FSHTank.fish` does that for every entity it checks on every fish. In an IG with several large CodeSystems, every lookup of a Reference or code system walked tens of thousands of rules, which was 63% of what remained after fix 1. The found rule is now cached per rules array. Rules arrays are only appended to or replaced wholesale (as `applyInsertRules` does), so the cache is invalidated when the array is replaced, when its length changes, or when its last rule changes. Because the rule rather than its value is cached, changes to a rule's value are still seen, and a cached rule is re-checked against the lookup on each hit because `CodeSystemExporter` rewrites a code caret rule's `path` in place once it resolves the concept.

Measurements (Node 22, `sushi` on the same machine):

| Build | master | fix 1 | fix 1 + 2 |
|---|---|---|---|
| One CodeSystem with 9,347 concepts, FSHOnly | 147 s, 750 MB | 15 s, 594 MB | 8 s, 598 MB |
| IG with 32 large CodeSystems written in FSH | not measured (many minutes) | 142 s | 52 s |

For comparison, the same IG with those 32 CodeSystems pre-rendered to JSON in `input/` (the workaround the project uses today) builds in 31 s. Most of the remaining gap is ANTLR parsing of the large FSH files.

Output is unchanged: the generated JSON for the 9,347-concept CodeSystem is byte-identical to master, and every other resource in the IG is identical to the pre-rendered build.

Developed with AI assistance (Claude Code). The diff, tests, and measurements were reviewed before submitting.

**Testing Instructions:**

Run `npm run check`. New tests in `test/fshtypes/FshCodeSystem.test.ts`, `test/fshtypes/Instance.test.ts` and `test/import/FSHTank.test.ts` cover the cache being invalidated when rules are added, removed, replaced, their values change, or a code caret rule's path is resolved. New tests in `test/export/CodeSystemExporter.test.ts` and `test/export/ValueSetExporter.test.ts` pin the output when several caret rules target the same extension slice and a numerically indexed extension, since the shared StructureDefinition now carries the slices added by earlier rules.

To see the performance difference, run `sushi` on a project containing a CodeSystem with a few thousand concepts that each have a designation or property.

**Related Issue:**

None.

https://claude.ai/code/session_01AKNbxASp6vABedSLJMGW97

### DIFF
--- a/src/export/CodeSystemExporter.ts
+++ b/src/export/CodeSystemExporter.ts
@@ -293,13 +293,17 @@ export class CodeSystemExporter {
   }
 
   /**
-   * Finds the FSH path to the concept identified by codePath, e.g. ['#a', '#b'] becomes concept[2].concept[0].
+   * Finds the FSH path to the concept identified by codePath. For example, if #a is the third top-level
+   * concept and #b is its first child, ['#a', '#b'] becomes concept[2].concept[0]. An empty codePath
+   * (a caret rule that is not on a concept) returns an empty path.
    * @param {CodeSystem} codeSystem - The CodeSystem containing the concepts
    * @param {string[]} codePath - The codes (with a leading #) leading to the concept
-   * @param {Map<CodeSystemConcept[], Map<string, number>>} conceptIndices - Cache of the index of each code in
-   *   each concept list. Every code caret rule needs one of these lookups, so the concept lists are indexed once
-   *   rather than scanned for each rule.
+   * @param {Map<CodeSystemConcept[], Map<string, number>>} conceptIndices - Cache of the index of the first
+   *   concept with each code in each concept list. Every code caret rule needs one of these lookups, so the
+   *   concept lists are indexed once rather than scanned for each rule. The cache is only valid while the
+   *   concept lists are not modified, so callers should use a new Map for each set of rules they resolve.
    * @returns {string} the path to the concept
+   * @throws {CannotResolvePathError} when a code in codePath is not found
    */
   private findConceptPath(
     codeSystem: CodeSystem,
--- a/src/export/CodeSystemExporter.ts
+++ b/src/export/CodeSystemExporter.ts
@@ -98,11 +98,13 @@ export class CodeSystemExporter {
               return;
             }
           }
-          conceptContainer.push(newConcept);
-          if (!conceptsByCode.has(conceptContainer)) {
-            conceptsByCode.set(conceptContainer, new Map());
+          let siblingsByCode = conceptsByCode.get(conceptContainer);
+          if (siblingsByCode == null) {
+            siblingsByCode = new Map();
+            conceptsByCode.set(conceptContainer, siblingsByCode);
           }
-          conceptsByCode.get(conceptContainer).set(newConcept.code, newConcept);
+          conceptContainer.push(newConcept);
+          siblingsByCode.set(newConcept.code, newConcept);
           existingConcepts.set(concept.code, concept);
         }
       });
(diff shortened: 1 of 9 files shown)


src/export/CodeSystemExporter.ts
```diff
@@ -116,9 +122,10 @@ export class CodeSystemExporter {
     // Because this.findConceptPath can potentially throw an error,
     // build a list of successful rules that will actually be applied.
     const successfulRules: CaretValueRule[] = [];
+    const conceptIndexCache = new Map<CodeSystemConcept[], Map<string, number>>();
```
cmoesel: The `setConcepts` function already iterates all rules and builds a cache for its purposes. Since `setConcepts` is called before `setCaretPathRules`, might it be possible for `setConcepts` to build a cache once (in a way that satisfies both use cases) and return it so that it can be passed into the `setCaretPathRules` function and re-used there? Would that simplify the code and/or be even more performant?

cmoesel (changes requested):
Thanks for the contribution, @vadi2. Caching adds more complication, but in this case it seems there is a very good payoff. We'll take it!

I've left one comment for you to consider. Let me know what you think.

This also needs to be rebased or merged from master.

Thanks again!

vadi2:
Thanks @cmoesel! Merged master, and `setConcepts` now builds the index once and `findConceptPath` reuses it - simpler, same speed (~4.3 s vs 82 s on master). One behaviour difference from master to flag: mixing a numeric index with two aliases of the same extension now gives two well-formed extensions instead of master's malformed three, because the shared StructureDefinition keeps the slice from validation.

```
Extension: MyExt
Id: my-ext
* value[x] only string

ValueSet: MixedVS
* ^extension[0].url = "http://example.org/numeric"
* ^extension[0].valueString = "n"
* ^extension[MyExt][0].valueString = "one"
* ^extension[my-ext][1].valueString = "two"
```

master:
```json
[{"url": "http://example.org/numeric", "valueString": "one"},
 {"url": "http://example.org/StructureDefinition/my-ext"},
 {"valueString": "two"}]
```

this branch:
```json
[{"url": "http://example.org/numeric", "valueString": "one"},
 {"url": "http://example.org/StructureDefinition/my-ext", "valueString": "two"}]
```

src/fshtypes/common.ts
```diff
@@ -52,6 +52,35 @@ export function fshifyString(input: string): string {
...
+// Two changes are not detected: replacing or reordering rules within the same array without changing its
+// length or last rule, and changing a rule in place so that it newly matches a lookup. Code that does either
+// must call clearRuleLookupCache afterwards.
+type RuleLookupCache = {
+  length: number;
+  lastRule: Rule;
+  // a lookup that found no rule is cached with an undefined value
+  rulesByLookup: Map<string, AssignmentRule | CaretValueRule | undefined>;
+};
+const ruleLookupCache = new WeakMap<Rule[], RuleLookupCache>();
```
cmoesel: Should we be concerned about the potential size of the lookup cache for an IG that contains many large codesystems?

PR #1638, opened 2026-09-10
Fix missing null-padded value arrays on repeating primitives

### LINKED ISSUE #1631: Repeating primitive with extensions but no values omits the null-padded value array
When every entry of a repeating primitive element carries only an `extension` (or `id`) and no value, SUSHI emits the `_element` array **without** the parallel value array. Per the FHIR JSON spec the value array should still be present, null-filled, so the two arrays align.

SUSHI already gets this right as soon as *one* entry has a value — it emits `"given": ["Xavier", null]` alongside `"_given": [null, {...}]`. Only the all-values-absent case drops the value array, so this looks like a gap in an otherwise-correct code path rather than a design decision.

**Version:** SUSHI v3.20.0 (implements FHIR Shorthand specification v3.0.0), `fhirVersion: 4.0.1`.

## Steps to reproduce

`sushi-config.yaml`:

```yaml
canonical: http://example.org/repro
fhirVersion: 4.0.1
FSHOnly: true
```

`input/fsh/repro.fsh`:

```
Alias: $display = http://hl7.org/fhir/StructureDefinition/display

// 1. Repeating primitive, extension but NO value anywhere in the array.
Instance: NoValues
InstanceOf: Patient
Usage: #example
* name[0].given[0].extension[0].url = $display
* name[0].given[0].extension[0].valueString = "ext on given[0]"

// 2. Same, but one sibling entry has a value.
Instance: SomeValues
InstanceOf: Patient
Usage: #example
* name[0].given[0] = "Xavier"
* name[0].given[1].extension[0].url = $display
* name[0].given[1].extension[0].valueString = "ext on given[1]"

// 3. Non-repeating primitive, extension but no value.
Instance: NonRepeating
InstanceOf: Patient
Usage: #example
* name[0].family.extension[0].url = $display
* name[0].family.extension[0].valueString = "ext on family"

// 4. Repeating primitive with an id (not an extension) and no value.
Instance: NoValuesId
InstanceOf: Patient
Usage: #example
* name[0].given[0].id = "given-0"
```

Then `sushi .` and inspect `name[0]` in each generated instance.

## Actual

```jsonc
// 1. NoValues — value array missing entirely
{"_given":[{"extension":[{"url":"…/display","valueString":"ext on given[0]"}]}]}

// 2. SomeValues — correct, arrays aligned and null-padded
{"given":["Xavier",null],"_given":[null,{"extension":[{"url":"…/display","valueString":"ext on given[1]"}]}]}

// 3. NonRepeating — correct, no array involved
{"_family":{"extension":[{"url":"…/display","valueString":"ext on family"}]}}

// 4. NoValuesId — same gap, with id rather than extension
{"_given":[{"id":"given-0"}]}
```

## Expected

Cases 1 and 4 should null-pad the value array the same way case 2 does:

```jsonc
// 1.
{"given":[null],"_given":[{"extension":[{"url":"…/display","valueString":"ext on given[0]"}]}]}

// 4.
{"given":[null],"_given":[{"id":"given-0"}]}
```

Cases 2 and 3 are already correct and are included only as controls.

## Spec reference

[FHIR R4 §JSON Representation of Primitive Elements](https://hl7.org/fhir/R4/json.html):

> In the case where the primitive element may repeat, it is represented in two arrays. JSON null values are used to fill out both arrays so that the id and/or extension are aligned with the matching value in the first array

> Note: when one of the repeating elements has no value, it is represented in the first array using a null. When an element has a value but no extension/id, the second array will have a null at the position of that element.

The second sentence covers this case directly: the repeating element has no value, so it should be represented in the first array using a `null`. The spec's worked example happens to show a mixed array, which may be why the all-absent case was missed.

## Why it matters

We author SDC template-based `$extract` templates ([`sdc-questionnaire-templateExtractValue`](http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire-templateExtractValue)), where a directive is an extension on an element that deliberately has **no** value — the value is produced at extraction time. On repeating primitives such as `HumanName.given` and `Address.line` that is exactly case 1.

Uploading such a questionnaire to HAPI FHIR (tested on 8.10.0 and on `hapi.fhir.org` 8.11.16-SNAPSHOT) returns `201`/`200` with no `OperationOutcome`, but the stored resource has silently dropped those extensions.

cpruijsen:
**Description:**

Keep a null-padded value array next to `_element` when every entry of a repeating primitive has only an `id` or `extension` and no value. FHIR JSON requires the value array to stay, null-filled, so the two arrays align (https://hl7.org/fhir/json.html#primitive). Mixed arrays (`"given": ["Xavier", null]` next to `"_given"`) already did this; the all-values-absent case did not.

`setPropertyOnInstance` already builds both arrays. `cleanResource` → `replaceField` then turned empty `{}` placeholders into `null` and deleted any array that was all null. That deletion now leaves the value array in place when the companion `_` array still has id or extension data. All-null `_` arrays are still dropped. Non-repeating primitives (`family` + `_family` only) are unchanged.

The same cleanup runs for StructureDefinitions, so a caret rule that puts only extensions on a repeating primitive such as `type.aggregation` now keeps `"aggregation": [null, null]` next to `_aggregation`. An existing test had encoded the old omission and was updated.

Keep an all-null value array only when `_${name}` has a non-empty element. Empty `{}` on the companion is not treated as data, because those objects become `null` in the same pass. The alternative (keep whenever the sibling is `!= null`) would leave `searchInclude: [null]` on PathRule-only primitives and break the existing CapabilityStatement instance test. A two-pass clean (recurse first, then drop all-null arrays) would also work.

Fixes #1631.

**Testing Instructions:**

- `npm test -- test/export/InstanceExporter.test.ts test/export/StructureDefinitionExporter.test.ts`
- `npm run check`
- Or in FSH Online / a local FSHOnly project, the four instances from #1631: `NoValues` and `NoValuesId` should now emit `"given": [null]` next to `"_given"`; `SomeValues` and `NonRepeating` should stay as they are.

**Related Issue:**

https://github.com/FHIR/sushi/issues/1631

### DIFF
--- a/src/fhirtypes/common.ts
+++ b/src/fhirtypes/common.ts
@@ -1170,9 +1170,19 @@ export function replaceField(
     } else if (typeof object[prop] === 'object' && !skipFn(prop)) {
       replaceField(object[prop], matchFn, replaceFn, skipFn);
 
-      // If the object[prop] was an array and all items were replaced by null using the replaceFn, get rid of the whole array
+      // If the object[prop] was an array and all items were replaced by null using the replaceFn, get rid of the whole array.
+      // Repeating FHIR primitives are two parallel arrays (value and _value). Keep a null-padded value array when the
+      // companion _value array still has id or extension data so the arrays stay aligned.
+      // See https://hl7.org/fhir/json.html#primitive and https://github.com/FHIR/sushi/issues/1631.
+      // Empty objects are not companion data; they become null in this same pass.
       if (Array.isArray(object[prop]) && object[prop].every((v: any) => v == null)) {
-        delete object[prop];
+        const companion = prop.startsWith('_') ? undefined : object[`_${prop}`];
+        const companionHasData =
+          Array.isArray(companion) &&
+          companion.some((v: any) => v != null && (typeof v !== 'object' || !isEmpty(v)));
+        if (!companionHasData) {
+          delete object[prop];
+        }
       }
 
       // Since an array could have been deleted, the whole object[prop] could have ended up as empty.
--- a/test/export/InstanceExporter.test.ts
+++ b/test/export/InstanceExporter.test.ts
@@ -5777,6 +5777,112 @@ describe('InstanceExporter', () => {
       ]);
     });
 
+    // Issue #1631: repeating primitives with only id/extension must keep a null-padded value array
+    describe('Issue #1631', () => {
+      const displayUrl = 'http://hl7.org/fhir/StructureDefinition/display';
+
+      it('should keep a null-padded value array when a repeating primitive has only an extension', () => {
+        // Instance: NoValues
+        // InstanceOf: Patient
+        // * name[0].given[0].extension[0].url = http://hl7.org/fhir/StructureDefinition/display
+        // * name[0].given[0].extension[0].valueString = "ext on given[0]"
+        const instance = new Instance('NoValues');
+        instance.instanceOf = 'Patient';
+        const urlRule = new AssignmentRule('name[0].given[0].extension[0].url');
+        urlRule.value = displayUrl;
+        const valueRule = new AssignmentRule('name[0].given[0].extension[0].valueString');
+        valueRule.value = 'ext on given[0]';
+        instance.rules.push(urlRule, valueRule);
+        const exported = exportInstance(instance);
+        expect(exported.name[0]).toEqual({
+          given: [null],
+          _given: [
+            {
+              extension: [
+                {
+                  url: displayUrl,
+                  valueString: 'ext on given[0]'
+                }
+              ]
+            }
+          ]
+        });
+      });
+
+      it('should keep aligned value and underscore arrays when a repeating primitive has mixed values and extensions', () => {
+        // Instance: SomeValues
+        // InstanceOf: Patient
+        // * name[0].given[0] = "Xavier"
+        // * name[0].given[1].extension[0].url = http://hl7.org/fhir/StructureDefinition/display
+        // * name[0].given[1].extension[0].valueString = "ext on given[1]"
+        const instance = new Instance('SomeValues');
+        instance.instanceOf = 'Patient';
+        const givenRule = new AssignmentRule('name[0].given[0]');
+        givenRule.value = 'Xavier';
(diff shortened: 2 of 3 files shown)


src/fhirtypes/common.ts
```diff
@@ -1170,9 +1170,19 @@ export function replaceField(
...
+      // Repeating FHIR primitives are two parallel arrays (value and _value). Keep a null-padded value array when the
+      // companion _value array still has id or extension data so the arrays stay aligned.
+      // See https://hl7.org/fhir/json.html#primitive and https://github.com/FHIR/sushi/issues/1631.
+      // Empty objects are not companion data; they become null in this same pass.
       if (Array.isArray(object[prop]) && object[prop].every((v: any) => v == null)) {
-        delete object[prop];
+        const companion = prop.startsWith('_') ? undefined : object[`_${prop}`];
+        const companionHasData =
+          Array.isArray(companion) &&
+          companion.some((v: any) => v != null && (typeof v !== 'object' || !isEmpty(v)));
```
cmoesel: Regarding `typeof v !== 'object'`, aren't items in the `_${prop}` array _always_ objects? They're either `null` (which technically is type `'object'`) or they're an object containing an `id` and/or `extensions` (e.g., `{ "id": "foo" }`). Are there cases where the `_${prop}` array might contain something that isn't an object? Do we need to check `typeof v !== 'object'` at all?

cmoesel (commented):
Thanks for the contribution, @cpruijsen. This looks good. I've left one comment asking a question about your `companion` check. Please also rebase the branch or merge master into it so it's up-to-date with master.

src/fhirtypes/common.ts
cpruijsen: You're right, it can't fire. Entries in a `_<prop>` array are created by
`setPropertyOnInstance` as `{}`, `{_sliceName: ...}` or `null` and then merged
with id and extension data, so a non-null one is always an object. Dropped the
check, `isEmpty` covers it. Full suite passes.

Merged master in as well.

cmoesel (approved):
Thanks! This looks great!

PR #1639, opened 2026-09-10
Don't strip invalid characters from ImplementationGuide name (#1271)

### LINKED ISSUE #1271: Don't try to clean up IG name
FHIR says IG names should follow this constraint: `name.matches('[A-Z]([A-Za-z0-9_]){0,254}')`. If a user supplies an IG name with invalid characters, it will _quietly_ remove them, but it doesn't go the next step and capitalize the first letter -- so the name may still violate the constraint.

It would probably be better just to leave the name alone and warn if it violates the constraint. The current behavior is confusing to users.

See: https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/IG.20Name/near/357017209

cpruijsen:
**Description:**

SUSHI was quietly removing non-alphanumeric characters from `ImplementationGuide.name` (except `_`) and not capitalizing the first letter, so the result could still violate FHIR `name.matches('[A-Z]([A-Za-z0-9_]){0,254}')`.

`initIG` now assigns `config.name` as authored. If the name is present and does not match that constraint, SUSHI logs a warning that quotes the constraint and points at the `"name"` property in `sushi-config.yaml`. It does not auto-capitalize and does not keep stripping.

This matches how Profile / ValueSet / CodeSystem names are already handled (`HasName.setName`): leave the value, warn, let the author fix it.

**What I chose and the alternative:** leave the name as authored and warn. The alternative is to keep stripping (and maybe capitalize the first letter) and also warn. I chose leave-and-warn because that is what #1271 asks for, and because silent cleanup was the confusing part, including cases like `Minimal IG` that used to become a FHIR-valid `MinimalIG` with no log. Happy to switch if maintainers would rather share `nameRegex` from `mixins.ts`, keep a sanitized export plus a warn, or treat this as an error instead of a warning.

**Testing Instructions:**

- `npm test -- test/ig/IGExporter.test.ts`
- In a project whose `sushi-config.yaml` has `name: Minimal IG`, confirm the generated `ImplementationGuide-*.json` keeps `name: "Minimal IG"` and the build log warns with `[A-Z]([A-Za-z0-9_]){0,254}`.
- With `name: MinimalIG`, confirm the name is unchanged and that constraint warning is not emitted.

Note: authors who previously relied on SUSHI stripping spaces or hyphens will now export the invalid name and see a warning (and IG Publisher may reject it) until they fix `sushi-config.yaml`. Use `title` for the human label; `name` is the machine identifier.

**Related Issue:**

Fixes #1271

### DIFF
--- a/src/ig/IGExporter.ts
+++ b/src/ig/IGExporter.ts
@@ -51,6 +51,8 @@ const IG_RESOURCE_FORMAT_EXTENSIONS = [
   DEPRECATED_RESOURCE_FORMAT_EXTENSION,
   CURRENT_RESOURCE_FORMAT_EXTENSION
 ];
+// FHIR ImplementationGuide.name invariant: name.matches('[A-Z]([A-Za-z0-9_]){0,254}')
+const IG_NAME_REGEX = /^[A-Z]([A-Za-z0-9_]){0,254}$/;
 
 function isR4(fhirVersion: string[]) {
   return fhirVersion.some(v => /^R4B?$/.test(getFHIRVersionInfo(v).name));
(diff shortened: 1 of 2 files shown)


cmoesel (approved):
Thanks for the contribution, @cpruijsen. This looks great. The warning message is helpful too. Approved!

Thanks again!

PR #1640, opened 2026-09-13
Don't clean up the IG name

### LINKED ISSUE #1271: Don't try to clean up IG name

AltangerelG:
Fixes #1271.

## Problem

`IGExporter.initIG()` stripped every character outside `[A-Za-z0-9_]` from the
configured IG name:

```js
name: this.config.name.replace(/[^A-Za-z0-9_]/g, ''),
```

The strip is silent, and it stops half way: it never capitalizes the first
letter, so `my ig` became `myig`, which still violates the constraint the strip
was meant to satisfy. The author loses the name they configured and gains
nothing.

## Change

The name is used as configured, and a warning is logged when it does not match
the name invariant.

`nameRegex` moves from `fhirtypes/mixins.ts` to `fhirtypes/primitiveTypes.ts`
alongside `idRegex`, so the IG name and the names checked by
`HasName.setName()` are held to one regex and report one message. That was
option 1 in my comment on the issue; happy to keep the regex local to
`IGExporter` instead if you would rather not move it.

## Three decisions worth your attention

**The message says "The IG name", not "The name".** I said in the issue I would
reuse `HasName`'s wording verbatim, and the constraint sentence is verbatim. But
`HasName` passes `sourceInfo`, so its warning carries a file and line; this one
has no source info, and without the word "IG" there is nothing to tell the
reader which name is being complained about. Happy to drop it if you prefer
strict consistency.

**An absent name is skipped rather than warned.** `Configuration.name` is
optional — `required()` only enforces it for non-FSH-only configs — so
`this.config.name` may be undefined, in which case the old `.replace()` would
have thrown. The check is guarded with `!= null`, leaving `ig.name` undefined
like any other unset optional. This is slightly beyond what the issue asked for,
so call it out if you would rather see it handled separately.

**This changes behaviour for existing IGs.** A project configured with
`name: My IG` was silently getting `MyIG` and will now get `My IG` plus a
warning — a name the IG Publisher will reject. That is what the issue asks for,
and I think it is right that the author fixes their own name rather than having
it quietly rewritten, but it is a behaviour change rather than a pure bug fix
and existing projects will see new warnings.

## Related

`Package.ts` already uses `config.name` unmodified for the package metadata, so
before this change the ImplementationGuide resource and the package manifest
could disagree about the name. They now agree.

## Testing

Three tests added in `test/ig/IGExporter.IG.test.ts` under `#ig-name`: a
conforming name passes through with no warning; a name with invalid characters
is preserved and warned about rather than stripped; and a name that starts with
a lower-case letter is preserved and warned about, which is the case the old
strip could not have fixed.

`npm run check` passes locally: 3538 tests in 104 suites, no lint errors, no
Prettier differences.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


AltangerelG:
Closing this — #1639 already covers it and got there first, three days before I
started. I only checked the issue thread, which had no comments, and missed the
open PR. Apologies for the duplicate review load.

@cpruijsen's version reaches the same conclusion for the same reason, so there
is nothing here worth splitting attention over. #1639 is the one to review.

Issue #1480, opened 2024-06-27
Add tests for FshToFhir API for more thorough coverage


mint-thompson:
Transferred from internal JIRA issue originally created 2023-09-18.

The tests for the FshToFhir API should be built out more to test more of what it does. There may be other tests to add, but possible things to add include:

* testing what types (profiles/extensions/logicals/etc) are converted
* testing the full end to end function works (taking some FSH and ensuring the appropriate FHIR is exported)

drlalithapranathi:
I'd like to take this one. I'm planning to build out `test/run/FshToFhir.test.ts` with end-to-end coverage: every entity type that gets converted (profiles, extensions, logicals, resources, instances, value sets, code systems), the canonical/version/snapshot options, invariants and mappings, aliases and rule sets shared across multiple inputs, and error handling across inputs. Happy to adjust the scope if you'd rather it went a different way.

drlalithapranathi:
Opened #1642 for this.

PR #1642, opened 2026-09-17
Add more tests for the fshToFhir API

drlalithapranathi:
**Description:**

Picking up #1480. The tests for `fshToFhir` were mostly about log levels and a couple of profile conversions, so I added tests for what the API actually hands back when you feed it FSH.

What's covered now:

- each entity type it converts: profiles, extensions, logicals, resources, instances, value sets, code systems
- artifacts come back grouped by type, no matter what order they were written in
- the canonical, version and snapshot options show up in the output
- invariants and mappings land on the profile instead of being exported on their own
- an alias or ruleset defined in one input works in another
- one bad input doesn't stop the good ones, and errors don't leak into the next call

Tests only, no source changes.

**Testing Instructions:**

`npm run check` should pass. The new tests are in the `#EndToEnd` block at the bottom of test/run/FshToFhir.test.ts.

**Related Issue:**

#1480


drlalithapranathi:
First time contributing here, so happy to adjust anything. Two things I wasn't sure about: I put the new tests in their own `#EndToEnd` block instead of extending `#Conversion`, since they need the full R4 definitions loaded. And a few assertions check exact error and warning counts. Easy to change either one.

